PR #9, opened 2016-05-24
GUACAMOLE-25: Only stream audio when requested.

mike-jumper:


guacamole-common-js/src/main/webapp/modules/AudioRecorder.js
```diff
@@ -213,6 +240,14 @@ Guacamole.RawAudioRecorder = function RawAudioRecorder(stream, mimetype) {
     var writtenSamples = 0;
 
     /**
+     * The audio stream provided by the browse, if allowed. If no stream has
```
jmuehlner: browser?

mike-jumper: :cold_sweat:

PR #10, opened 2016-05-28
GUACAMOLE-32: extended variable support for linked databases

vchrisb:
- adding support for leveraging `MYSQL_ENV_MYSQL_USER`, `MYSQL_ENV_MYSQL_PASSWORD`, `MYSQL_ENV_MYSQL_DATABASE` and `MYSQL_ENV_MYSQL_ROOT_PASSWORD` for linked
- removed deprecated environmental variables `MYSQL_NAME`, `MYSQL_PORT_3306_TCP_ADDR` and `MYSQL_PORT_3306_TCP_PORT`
- adding support for leveraging `POSTGRES_ENV_POSTGRES_USER`, `$POSTGRES_ENV_POSTGRES_PASSWORD` and `POSTGRES_ENV_POSTGRES_DB` for linked
- removed deprecated environmental variables `POSTGRES_NAME`, `POSTGRES_PORT_5432_TCP_ADDR` and `POSTGRES_PORT_5432_TCP_PORT`

### DIFF
--- a/guacamole-docker/bin/start.sh
+++ b/guacamole-docker/bin/start.sh
@@ -90,15 +90,21 @@ set_optional_property() {
 ## instead of a linked container.
 ##
 associate_mysql() {
-
-    # Use linked container if specified
-    if [ -n "$MYSQL_NAME" ]; then
-        MYSQL_HOSTNAME="$MYSQL_PORT_3306_TCP_ADDR"
-        MYSQL_PORT="$MYSQL_PORT_3306_TCP_PORT"
-    fi
-
-    # Use default port if none specified
-    MYSQL_PORT="${MYSQL_PORT-3306}"
+	: ${MYSQL_HOSTNAME:=mysql}
+	# overwrite "MYSQL_PORT" with default port if using linked container
+	if [ -n "$MYSQL_ENV_MYSQL_DATABASE" ]; then
+		MYSQL_PORT=3306
+	fi
+	# Use default port if none specified
+    	: ${MYSQL_PORT:=3306}
+    	
+	# if we're linked to MySQL and thus have credentials already, let's use them
+	: ${MYSQL_USER:=${MYSQL_ENV_MYSQL_USER:-root}}
+	if [ "$MYSQL_USER" = 'root' ]; then
+		: ${MYSQL_PASSWORD:=$MYSQL_ENV_MYSQL_ROOT_PASSWORD}
+	fi
+	: ${MYSQL_PASSWORD:=$MYSQL_ENV_MYSQL_PASSWORD}
+	: ${MYSQL_DATABASE:=${MYSQL_ENV_MYSQL_DATABASE:-guacamole}}
 
     # Verify required connection information is present
     if [ -z "$MYSQL_HOSTNAME" -o -z "$MYSQL_PORT" ]; then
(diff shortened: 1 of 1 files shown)


guacamole-docker/bin/start.sh
```diff
@@ -90,15 +90,21 @@ set_optional_property() {
...
-
-    # Use linked container if specified
-    if [ -n "$MYSQL_NAME" ]; then
-        MYSQL_HOSTNAME="$MYSQL_PORT_3306_TCP_ADDR"
-        MYSQL_PORT="$MYSQL_PORT_3306_TCP_PORT"
-    fi
-
-    # Use default port if none specified
-    MYSQL_PORT="${MYSQL_PORT-3306}"
+	: ${MYSQL_HOSTNAME:=mysql}
```
mike-jumper: Why default the hostname to "mysql"?

guacamole-docker/bin/start.sh
```diff
@@ -90,15 +90,21 @@ set_optional_property() {
...
+	# overwrite "MYSQL_PORT" with default port if using linked container
+	if [ -n "$MYSQL_ENV_MYSQL_DATABASE" ]; then
+		MYSQL_PORT=3306
+	fi
+	# Use default port if none specified
+    	: ${MYSQL_PORT:=3306}
+    	
+	# if we're linked to MySQL and thus have credentials already, let's use them
+	: ${MYSQL_USER:=${MYSQL_ENV_MYSQL_USER:-root}}
```
mike-jumper: Defaulting the user (and password) to the root credentials seems awfully scary. The documentation covers creating a restricted-privilege user with only the necessary permissions to work with the Guacamole database.

I'd think it'd be far better to refuse to start the image and kick back an error noting that required environment variables are missing.

mike-jumper:
1. Will *removing support* for the old form cause things to break?
2. Why the variable initialization starting with colons? I'm not familiar with this syntax.
3. Please follow the code style of the rest of the script - 4 space indent, no tabs.
4. Why all the new defaults? If the goal is to leverage additional variables, changing the behavior of default values seems out of scope.

guacamole-docker/bin/start.sh
vchrisb: Using `MYSQL_PORT_3306_TCP_ADDR` and `MYSQL_PORT_3306_TCP_PORT` is deprecated as of [Link environment variables reference](https://docs.docker.com/compose/link-env-deprecated/).
Linked services are reachable by the service name or its alias. In this case the alias of the database service is `mysql`. Please see [Links](https://docs.docker.com/compose/compose-file/#links).
Docker is either adding a `Hosts`entry for the linked docker image, newer docker versions do have a build in DNS server.

guacamole-docker/bin/start.sh
vchrisb: The mysql docker image by default does not create a `MYSQL_USER` and this parameter is optional. See [Environmental Variables](https://hub.docker.com/_/mysql/)
Setting the `MYSQL_USER` to `root`, if none is specified, is only happening for linked docker images.
Not providing this fallback to `root`, for linked services, is also ok for me.

vchrisb:
1. I'm not aware of any drawbacks. It should be 100% backward compatible. That's the reason why I haven't renamed `MYSQL_PORT`, though it is preset with something like `tcp://172.17.0.5:5432`. See `name_PORT` in [Link environment variables reference](https://docs.docker.com/compose/link-env-deprecated/)
2. Please have a look at `Example 10-7` in [Parameter Substitution](http://www.tldp.org/LDP/abs/html/parameter-substitution.html)
3. I can rewrite the variable assignment using if statements.
4. new defaults are for `MYSQL_HOSTNAME` and `MYSQL_PORT` because the old are deprecated. Defaults for `MYSQL_USER` is as comment above discussable

necouchman:
Pinging on this one, trying to clean up some old PRs.  @mike-jumper: any other comments for Chris, here, after his last post?

PR #14, opened 2016-06-02
GUACAMOLE-44: Slice uploaded files into chunks (DO NOT slurp entire file into memory)

mike-jumper:
The old file upload code was inherently flawed as it required reading the entire file into memory. Besides being a poor user experience (long delay before upload begins), this could even crash the browser if the file was large enough.

This change adds a new writer-style object, `Guacamole.FileWriter`, whose `sendFile()` efficiently streams a provided `File` over an underlying `Guacamole.OutputStream`. The implementation streams the `File` properly via `slice()`; it is not read in its entirety into memory.

As this required wrapping a `Guacamole.ArrayBufferWriter` and being aware of the Guacamole protocol's instruction size restrictions, this change replaces the old magic 6048 with a well-documented constant, `Guacamole.ArrayBufferWriter.DEFAULT_BLOB_LENGTH`, and exposes the blob length used by any particular `Guacamole.ArrayBufferWriter` instance with a mutable `blobLength` property.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/ArrayBufferWriter.js
+++ b/guacamole-common-js/src/main/webapp/modules/ArrayBufferWriter.js
@@ -62,6 +62,20 @@ Guacamole.ArrayBufferWriter = function(stream) {
     }
 
     /**
+     * The maximum length of any blob sent by this Guacamole.ArrayBufferWriter,
+     * in bytes. Data sent via
+     * {@link Guacamole.ArrayBufferWriter#sendData|sendData()} which exceeds
+     * this length will be split into multiple blobs. As the Guacamole protocol
+     * limits the maximum size of any instruction or instruction element to
+     * 8192 bytes, and the contents of blobs will be base64-encoded, this value
+     * should only be increased with extreme caution.
+     *
+     * @type {Number}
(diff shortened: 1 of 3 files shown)


guacamole-common-js/src/main/webapp/modules/ArrayBufferWriter.js
```diff
@@ -62,6 +62,20 @@ Guacamole.ArrayBufferWriter = function(stream) {
     }
 
     /**
+     * The maximum length of any blob sent by this Guacamole.ArrayBufferWriter,
+     * in bytes. Data sent via
+     * {@link Guacamole.ArrayBufferWriter#sendData|sendData()} which exceeds
```
jmuehlner: What's this syntax? "sendData|sendData()"

mike-jumper: It's the way you provide text for JSDoc3's `@link`: http://usejsdoc.org/tags-inline-link.html

Re-reading the above ... it looks like there's a more readable, more markdown-y syntax available. Perhaps I should switch over to that before precedent is set?

I believe this is our first use of `@link` in the docs.

jmuehlner: If there's a more readable syntax available, we should probably use that. :)

PR #16, opened 2016-06-05
GUACAMOLE-44: Implement uploads within tunnel REST endpoint

mike-jumper:
Similar to apache/incubator-guacamole-client#13, this change implements the upload half of the tunnel REST endpoint. While it is possible to upload using JavaScript alone via the new `Guacamole.FileWriter` added via apache/incubator-guacamole-client#14, doing so instead via native HTTP is **much faster**.

To facilitate translation of Guacamole protocol status codes (received via "ack" instructions along the stream) to HTTP errors and `APIError` entities, this change also:

1. Adds a new `fromGuacamoleStatusCode()` to the `GuacamoleStatus` enum, allowing the integer status codes to be directly translated to typesafe `GuacamoleStatus` values.
2. Adds a new `statusCode` property to `APIError` (and it's JavaScript sibling `Error`), along with a new error type, `STREAM_ERROR`, for the sake of representing errors received along an intercepted Guacamole stream.

The logic surrounding routing a Guacamole stream to an `OutputStream` has been extracted and generalized such that the corresponding `InputStream` implementation can leverage much of the same logic. This is done through the following new classes:

1. `InterceptedStream` - a pairing of `OutputStream` or `InputStream` with the stream index used by the Guacamole protocol, here represented by a `String` despite its integer nature due to the way such things are represented internally. It's only parsed out as an integer when actually needed as such.
2. `InterceptedStreamMap` - a mapping of stream index (again, in `String` form) to the corresponding `InterceptedStream`. This class also automatically handles notification of stream closure, and provides its own `waitFor()` function that blocks until a given stream has closed. This logic used to be part of `StreamInterceptingTunnel` when it was `OutputStream`-specific.
3. `StreamInterceptingFilter` - a filter which intercepts stream-related instructions for streams of an arbitrary type, automatically handling mapping of stream indices and blocking. It is up to the implementation to filter the instructions it is actually interested in, and route the I/O associated with those instructions accordingly.
4. `OutputStreamInterceptingFilter` - an `OutputStream`-specific implementation of `StreamInterceptingFilter`. This contains the remaining logic that used to be part of `StreamInterceptingTunnel`.
5. `InputStreamInterceptingFilter` - an `InputStream`-specific implementation of `StreamInterceptingFilter`. This is the new logic, implementing support for routing uploaded file data along an established Guacamole stream.

As receipt of the error via "ack" happens asynchronously and outside the blocking call to `interceptStream()`, errors are instead reported via a call to the `setStreamError()` function of the newly-added `InterceptedStream` object.

### DIFF
--- a/guacamole/src/main/java/org/apache/guacamole/rest/tunnel/TunnelRESTService.java
+++ b/guacamole/src/main/java/org/apache/guacamole/rest/tunnel/TunnelRESTService.java
@@ -136,7 +136,12 @@ public class TunnelRESTService {
 
             @Override
             public void write(OutputStream output) throws IOException {
-                tunnel.interceptStream(streamIndex, output);
+                try {
+                    tunnel.interceptStream(streamIndex, output);
+                }
+                catch (GuacamoleException e) {
+                    throw new IOException(e);
+                }
             }
 
         };
@@ -167,8 +172,9 @@ public class TunnelRESTService {
      *     stream.
      *
      * @throws GuacamoleException
-     *     If the session associated with the given auth token cannot be
-     *     retrieved, or if no such tunnel exists.
+     *     If the session associated with the given auth
+     *     token cannot be retrieved, if no such tunnel exists, or if the
+     *     intercepted stream itself closes with an error.
      */
     @POST
     @Consumes(MediaType.WILDCARD)
--- a/guacamole/src/main/java/org/apache/guacamole/rest/tunnel/TunnelRESTService.java
+++ b/guacamole/src/main/java/org/apache/guacamole/rest/tunnel/TunnelRESTService.java
@@ -21,12 +21,14 @@ package org.apache.guacamole.rest.tunnel;
 
 import com.google.inject.Inject;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.Map;
 import java.util.Set;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.DefaultValue;
 import javax.ws.rs.GET;
+import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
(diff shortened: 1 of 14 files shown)


guacamole/src/main/java/org/apache/guacamole/rest/tunnel/TunnelRESTService.java
```diff
@@ -143,4 +150,53 @@ public void write(OutputStream output) throws IOException {
...
+     *
+     * @param filename
+     *     The filename to use for the sake of identifying the data being sent.
+     *
+     * @param data
+     *     An InputStream containing the data to be sent over the intercepted
+     *     stream.
+     *
+     * @throws GuacamoleException
+     *     If the session associated with the given auth
```
jmuehlner: Not a big deal, but this is a strange place to wrap this comment.

guacamole/src/main/java/org/apache/guacamole/tunnel/GuacamoleStreamException.java
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * Creates a new GuacamoleStreamException representing an error returned by
+     * an intercepted stream.
+     *
+     * @param status
+     *     The status code of the error condition reported by the intercepted
+     *     Guacamole stream.
+     *
+     * @param message
+     *     The human readable description of the error that occurred, as
```
jmuehlner: "as as"

guacamole/src/main/java/org/apache/guacamole/tunnel/InterceptedStream.java
```diff
@@ -0,0 +1,161 @@
...
+     */
+    public void setStreamError(GuacamoleException streamError) {
+        this.streamError = streamError;
+    }
+
+    /**
+     * Reports that this InterceptedStream did not complete successfully due to
+     * an error described by the given status code and human-readable message.
+     * The error reported by this call can later be retrieved as a
+     * GuacamoleStreamException by calling getStreamError().
```
jmuehlner: Isn't it just a GuacamoleException here? Or was the type meant to be "GuacamoleStreamException"?

guacamole/src/main/java/org/apache/guacamole/rest/tunnel/TunnelRESTService.java
mike-jumper: True. Will fix.

guacamole/src/main/java/org/apache/guacamole/tunnel/InterceptedStream.java
mike-jumper: It's just a `GuacamoleException`, but the subclass of `GuacamoleException` that will be set due to an invocation of this function will be `GuacamoleStreamException`.

mike-jumper: Since `GuacamoleStreamException` represents a received "ack" instruction, it can really only apply to streams where we are acting as the source of data and the server is sending acknowledgements ("ack" is only sent in response to "blob" or the instructions which start a stream). In practical terms of this change, if we changed `GuacamoleException` to `GuacamoleStreamException` for `setStreamError()` in general, that would mean that `setStreamError()` could only be used in context of `InputStreamInterceptingFilter`. If anything else ever wants to report an error, it will be impossible.

It is true that this is already the case - only `InputStreamInterceptingFilter` uses this function, but narrowing the type to that extent would guarantee that it can never be used anywhere else, and I felt that would be bad.

jmuehlner: Ok

PR #18, opened 2016-06-09
GUACAMOLE-1: Interface should say "Apache Guacamole", not "Guacamole".

mike-jumper:
This change updates the application name from "Guacamole" to "Apache Guacamole" via the translation strings. As the version number is now quite long, I've also separated the old `APP.NAME` into `APP.NAME` and `APP.VERSION`, and reorganized the login screen to keep things clean:

> apache-guac-login


jmuehlner:
Looks good.

PR #22, opened 2016-06-15
GUACAMOLE-53: Implement session affinity for balancing groups

mike-jumper:
This change adds a new connection group attribute which, if enabled, causes balancing groups to consistently route the same users to the same connections until those users log out.


extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connectiongroup/ModeledConnectionGroup.java
```diff
@@ -168,6 +177,10 @@ public Type getType() {
         // Set per-user connection limit attribute
         attributes.put(MAX_CONNECTIONS_PER_USER_NAME, NumericField.format(getModel().getMaxConnectionsPerUser()));
 
+        // Set per-user connection limit attribute
```
jmuehlner: Wrong label

PR #24, opened 2016-06-28
GUACAMOLE-55: Add interface-level support for image clipboard contents

mike-jumper:
This change implements a new directive, `guacClipboard`, which provides support for image clipboard contents in addition to text. All clipboard contents intended for (or coming from) this directive are communicated via the `data` property on the scope as `ClipboardData` objects. These objects are also consumed and produced by the `clipboardService` which has been refactored to support them. All of these objects, services, etc. have been collected together into their own "clipboard" module.


guacamole/src/main/webapp/app/clipboard/directives/guacClipboard.js
```diff
@@ -0,0 +1,312 @@
...
+        var element = $element[0];
+
+        // Intercept paste events, handling image data specifically
+        element.addEventListener('paste', function dataPasted(e) {
+
+            // Always clear the current clipboard contents upon paste
+            element.innerHTML = '';
+
+            // If we can't read the clipboard contents at all, abort
+            var clipboardData = e.clipboardData;
```
jmuehlner: Should we keep the existing content if a paste fails? I'm not sure...

guacamole/src/main/webapp/app/clipboard/services/clipboardService.js
```diff
@@ -0,0 +1,409 @@
...
+            else
+                deferred.reject();
+
+            // Unfocus the clipboard DOM event to avoid mobile keyboard opening,
+            // restoring whichever element was originally focused
+            clipboardContent.blur();
+            originalElement.focus();
+            popSelection();
+
+        }, 100);
```
jmuehlner: Maybe constantify this timeout value 100?

guacamole/src/main/webapp/app/clipboard/directives/guacClipboard.js
mike-jumper: No, you're right - this is when an actual paste occurs, not the magic clipboard-reading fake paste used by `clipboardService`.

We could certainly get rid of this behavior, and allow the clipboard contents to remain. This is the way things used to work. However, it would have the downside that nothing will happen if you attempt to paste an image into the clipboard unless you first delete everything already present (or select everything first and then paste).

The code which keeps the contents in sync with the internal clipboard data would need to decide whether to represent the clipboard contents as a single image or as text. If text already exists in the clipboard, or there are other things present besides an image, it will assume text is correct and strip out anything else, including the newly-pasted image.

I'm not sure which behavior would be better.

guacamole/src/main/webapp/app/clipboard/services/clipboardService.js
mike-jumper: OK.

guacamole/src/main/webapp/app/clipboard/directives/guacClipboard.js
jmuehlner: If there's a way to decide whether it's an image before clearing the text, I think that would be good to do. It's unexpected and undesirable in my opinion for all the text currently in the clipboard to disappear without warning if you paste more text. I think it will make people unhappy.

mike-jumper: I agree. Unfortunately, it is impossible to tell if the clipboard data is an image until after the paste occurs, and then only from harvesting the actual pasted data:

https://bugzilla.mozilla.org/show_bug.cgi?id=891247

I suppose we could check if the clipboard *isn't* empty ... and only clear the clipboard field if the contents cannot be checked ...

I'll give this a shot...

mike-jumper: OK - I've got this mostly working by handling some other properties of the native `ClipboardData` object (not the one we define here, but the one defined by the W3C Clipboard API), but I'm still seeing some issues with Firefox. We'll have to rehash those in tonight's dev session... so I guess this PR will just have to wait.

Other than that, the behavior is nice. Paste works as expected. In the case of Firefox, where data is simply unavailable, pasting an image into text will not work. This is fine IMHO - the problem I'm encountering is that pasting an image *never* works after the first attempt succeeds.

Anyway, more on that later.

mike-jumper:
Closing this until we can resolve the latest ridiculous quantity of browser-specific issues.

PR #26, opened 2016-07-12
GUACAMOLE-5: Refactor /api/data/... REST services into resources

mike-jumper:
The REST API changes required for the interface side of screen sharing require exposing additional resources beneath the existing `TunnelRESTService`, but this is inconvenient due to the structure of the REST services within Guacamole in general. Each such REST service re-implements the same lookup logic for every single subresource, and there is little potential for reuse among services that might need to return resources of a common type.

This enormous change refactors the main REST services under `/api/data` as resources under `/api/session/data`, with the root `/api/data` being the only remaining REST service: `SessionDataRESTService`.

Beneath that, each `UserContext` applying to the user's session is exposed via a `UserContextResource`, which contains resources for each directory within the `UserContext`, which contain resources for each object within the directory, etc.

Common base classes have been used where possible, so the universal add/update/remove operations which apply to `Directory` need not be re-implemented across all resources which expose a `Directory`, though there are still cases where such operations are overridden to enforce additional semantics (such as denying access to editing your own user account).

Each new resource is split across a dedicated module which loads the required injections, an optional `DirectoryResource` implementation which exposes the related `Directory` (if the base class is insufficient), a `DirectoryObjectResource` implementation which exposes the objects themselves (always required as assisted injection cannot handle the additional generically-typed parameter for the corresponding factory `create()` method), and a `DirectoryObjectTranslator` implementation which translates between the internal and external versions of an object. The various types of resources are all created with factories, the implementations of which are generated dynamically by Guice's assisted injection extension.


guacamole/src/main/java/org/apache/guacamole/rest/ObjectRetrievalService.java
```diff
@@ -23,19 +23,11 @@
...
 import org.apache.guacamole.net.auth.UserContext;
 import org.apache.guacamole.GuacamoleSession;
-import org.apache.guacamole.rest.connectiongroup.APIConnectionGroup;
 
 /**
- * Provides easy access and automatic error handling for retrieval of objects,
- * such as users, connections, or connection groups. REST API semantics, such
- * as the special root connection group identifier, are also handled
- * automatically.
+ * Provides easy access and automatic error handling for retrieval of objects.
```
jmuehlner: Should this even be called ObjectRetrievalService anymore? All it does is get userContexts out of the session at this point.

mike-jumper: I think I can get rid of this entirely, actually ...

guacamole/src/main/java/org/apache/guacamole/rest/activeconnection/ActiveConnectionResource.java
```diff
@@ -0,0 +1,70 @@
...
+     * @param connection
+     *     The ActiveConnection that this ActiveConnectionResource should
+     *     represent.
+     *
+     * @param translator
+     *     A DirectoryObjectTranslator implementation which handles
+     *     ActiveConnections.
+     */
+    @AssistedInject
+    public ActiveConnectionResource(@Assisted UserContext userContext,
```
jmuehlner: Where is the userContext parameter used?

guacamole/src/main/java/org/apache/guacamole/rest/directory/DirectoryResource.java
```diff
@@ -0,0 +1,265 @@
...
+        this.resourceFactory = resourceFactory;
+    }
+
+    /**
+     * Returns a map of all objects available within this DirectoryResource,
+     * filtering the returned map by the given permission, if specified.
+     *
+     * @param permissions
+     *     The set of permissions to filter with. A user must have one or more
+     *     of these permissions for a user to appear in the result.
```
jmuehlner: "for a user to appear"?

guacamole/src/main/java/org/apache/guacamole/rest/directory/DirectoryResource.java
```diff
@@ -0,0 +1,265 @@
...
+        boolean isAdmin = systemPermissions.hasPermission(SystemPermission.Type.ADMINISTER);
+
+        // Filter objects, if requested
+        Collection<String> identifiers = directory.getIdentifiers();
+        if (!isAdmin && permissions != null && !permissions.isEmpty()) {
+            ObjectPermissionSet objectPermissions = self.getUserPermissions();
+            identifiers = objectPermissions.getAccessibleObjects(permissions, identifiers);
+        }
+
+        Map<String, ExternalType> apiObjects = new HashMap<String, ExternalType>();
```
jmuehlner: For consistency, should maybe comment this section too?

guacamole/src/main/java/org/apache/guacamole/rest/user/UserResource.java
```diff
@@ -0,0 +1,195 @@
...
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+public class UserResource
+        extends DirectoryObjectResource<User, APIUser> {
+
+    /**
+     * The prefix of any path within an operation of a JSON patch which
+     * modifies the permissions of a user regarding a specific connection.
+     */
+    private static final String CONNECTION_PERMISSION_PATCH_PATH_PREFIX = "/connectionPermissions/";
```
jmuehlner: What are all these constants for?

guacamole/src/main/java/org/apache/guacamole/rest/activeconnection/ActiveConnectionResource.java
mike-jumper: It's required by Guice's assisted injection. If the parameter is missing, it can't figure out how to map the factory's `create()` function.

PR #28, opened 2016-07-14
GUACAMOLE-5: Fix regression introduced by recent refactor of the REST API

mike-jumper:
The recent refactor caused issues where actual type-specific implementations were not provided for the REST resources. All REST resources **MUST** be non-generic, thus `DirectoryResource` and `DirectoryObjectResource` **MUST NOT** be used directly. Type parameters of generic classes will otherwise be handled by Jersey as `Object` due to type erasure, resulting in class cast exceptions due to received JSON being deserialized as a `Map` instead of the intended type.

This change also updates the logging and error handling within `RESTExceptionWrapper` such that this sort of thing will be easier to notice going forward. Internal errors were not being properly logged, especially those which were not subclasses of `GuacamoleException`.


mike-jumper:
The refactor in question was #26.

PR #29, opened 2016-07-16
GUACAMOLE-5: Move common equals() / hashCode() of Identifiable implementations to base class

mike-jumper:
To make developers lives easier, including my own, this change moves the common `equals()` and `hashCode()` of the `Identifiable` implementations used throughout the Guacamole API to a common base class: `AbstractIdentifiable`.

The logic is identical to all past version of Guacamole. The code in question is simply no longer duplicated.


guacamole-ext/src/main/java/org/apache/guacamole/net/auth/AbstractIdentifiable.java
```diff
@@ -0,0 +1,75 @@
...
+        if (identifier == null)
+            return 0;
+
+        return identifier.hashCode();
+    }
+
+    @Override
+    public boolean equals(Object other) {
+
+        // Not equal if null or not a User
```
jmuehlner: Bad comment copy/paste

PR #30, opened 2016-07-16
GUACAMOLE-5: Add concept of sharing profiles to guacamole-ext and REST API.

mike-jumper:
Background
-------------------

Independent of the web application and surrounding APIs, screen sharing has already been implemented within guacamole-server. When a user wants to join an existing connection, differing levels of access for the joining user are implemented through providing different connection parameters.

To provide this functionality through the web application, the extension API needs to be modified to support those semantics, and the REST API needs to be modified to expose them.

The goal of [GUACAMOLE-5](https://issues.apache.org/jira/browse/GUACAMOLE-5) in general is to implement the necessary API and interface changes to provide a means for users to generate temporary "share links" for their connections, where the ability to generate such links (and the level of access provided to such links) is dictated by the administrator.

Sharing profiles
-------------------------

The API changes proposed here involve a new object: `SharingProfile`. This object is the abstraction of the semantics described above. Like a `Connection`, defines a set of parameter name/value pairs which dictate the level of access provided by that connection. Like all other objects within the extension subsystem, it's stored within a `Directory<SharingProfile>` in the `UserContext`, and it has its own `CREATE_SHARING_PROFILE` system permission and set of object-level permissions.

*Unlike* a `Connection`, however, it cannot be connected to directly, and does not implement `Connectable`. To use a `SharingProfile` to access a connection, the following needs to happen:

1. A user with `READ` access to both the `SharingProfile` and its associated `Connection` connects to the `Connection`. The backend creates a new `GuacamoleTunnel` for this newly-established connection, as well as an `ActiveConnection` to track its existence.
2. The user that connected looks up the available sharing profiles for their active connection via the REST API, and makes another request to retrieve temporary credentials (an instance of `UserCredentials`). The client-side JavaScript parses these credentials to determine the format of the link, and provides the link to the user.
3. The user sends the link to the person intending to join. Upon visiting the link, that person is temporarily authorized by the extension that generated the temporary credentials. The single `Connection` available to that user will actually join the existing connection according to the restrictions of the `SharingProfile`.

Summary of changes
----------------------------------

1. New `SharingProfile` object. The connection that can be joined by a sharing profile is referred to as the "primary connection", with the sharing profile being subordinate.
2. Corresponding `Directory<SharingProfile>`, `CREATE_SHARING_PROFILE` system-level permission, and object-level `getSharingProfilePermissions()` on `User`.
3. New `Shareable` interface, which defines the function to be used to generate temporary credentials for a shared object. In the case of these changes, `ActiveConnection` implements `Shareable<SharingProfile>`.
4. New `UserCredentials` object, allowing extensions to describe the actual credentials of a user in a way that can be parsed by an automated system.
5. New `QUERY_PARAMETER` field type, allowing `UserCredentials` to specify a token, etc. that must be included in the URL.
6. REST API changes supporting retrieval of the `ActiveConnection` from a tunnel, temporary sharing `UserCredentials` from an `ActiveConnection`, etc.
7. All necessary base changes to keep the existing extensions buildable, despite support for `SharingProfile` not yet being implemented therein.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
@@ -27,6 +27,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import org.apache.guacamole.auth.jdbc.tunnel.GuacamoleTunnelService;
 import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.auth.jdbc.JDBCEnvironment;
@@ -154,6 +155,12 @@ public class ModeledConnection extends ModeledGroupedDirectoryObject<ConnectionM
         // Update model
         getModel().setProtocol(config.getProtocol());
         
+    }
+
+    @Override
+    public Set<String> getSharingProfileIdentifiers()
+            throws GuacamoleException {
+        return Collections.<String>emptySet();
     }
 
     @Override
--- a/guacamole/src/main/java/org/apache/guacamole/rest/permission/PermissionSetResource.java
+++ b/guacamole/src/main/java/org/apache/guacamole/rest/permission/PermissionSetResource.java
@@ -212,7 +212,7 @@ public class PermissionSetResource {
 
             }
 
-            // Create sharing profile permission if path has sharing profileprefix
+            // Create sharing profile permission if path has sharing profile prefix
             else if (path.startsWith(SHARING_PROFILE_PERMISSION_PATCH_PATH_PREFIX)) {
 
                 // Get identifier and type from patch operation
--- a/guacamole/src/main/java/org/apache/guacamole/rest/permission/PermissionSetResource.java
+++ b/guacamole/src/main/java/org/apache/guacamole/rest/permission/PermissionSetResource.java
@@ -57,6 +57,12 @@ public class PermissionSetResource {
 
     /**
      * The prefix of any path within an operation of a JSON patch which
+     * modifies the permissions of a user regarding a specific sharing profile.
+     */
+    private static final String SHARING_PROFILE_PERMISSION_PATCH_PATH_PREFIX = "/sharingProfilePermissions/";
+
+    /**
+     * The prefix of any path within an operation of a JSON patch which
      * modifies the permissions of a user regarding a specific active connection.
      */
     private static final String ACTIVE_CONNECTION_PERMISSION_PATCH_PATH_PREFIX = "/activeConnectionPermissions/";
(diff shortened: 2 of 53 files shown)


guacamole/src/main/java/org/apache/guacamole/rest/permission/PermissionSetResource.java
```diff
@@ -205,6 +212,19 @@ else if (path.startsWith(CONNECTION_GROUP_PERMISSION_PATCH_PATH_PREFIX)) {
 
             }
 
+            // Create sharing profile permission if path has sharing profileprefix
```
jmuehlner: Probably want a space - "profile prefix"

guacamole/src/main/java/org/apache/guacamole/rest/sharingprofile/APISharingProfile.java
```diff
@@ -0,0 +1,207 @@
...
+    /**
+     * Returns a map of all associated connection parameter values which apply
+     * when the sharing profile is used, indexed by parameter name.
+     *
+     * @return
+     *     A map of all associated connection parameter values which apply when
+     *     the sharing profile is used, indexed by parameter name.
+     */
+    public Map<String, String> getParameters() {
+        return parameters;
```
jmuehlner: So this is just null until set through the setter?

guacamole/src/main/java/org/apache/guacamole/tunnel/UserTunnel.java
```diff
@@ -0,0 +1,120 @@
...
+    public ActiveConnection getActiveConnection() throws GuacamoleException {
+
+        // Pull the UUID of the current tunnel
+        UUID uuid = getUUID();
+
+        // Get the directory of active connections
+        Directory<ActiveConnection> activeConnectionDirectory = userContext.getActiveConnectionDirectory();
+        Collection<String> activeConnectionIdentifiers = activeConnectionDirectory.getIdentifiers();
+
+        // Search all  connections for a tunnel which matches this tunnel
```
jmuehlner: Extra space here - "all  connections"

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
```diff
@@ -157,6 +158,12 @@ public void setConfiguration(GuacamoleConfiguration config) {
     }
 
     @Override
+    public Set<String> getSharingProfileIdentifiers()
```
jmuehlner: So I don't see any implementation "getSharingProfileIdentifiers" that actually returns anything - am I missing something about how this works?

guacamole/src/main/java/org/apache/guacamole/rest/permission/PermissionSetResource.java
mike-jumper: Oops. Good catch.

guacamole/src/main/java/org/apache/guacamole/rest/sharingprofile/APISharingProfile.java
mike-jumper: Yup. Don't want to expose parameters unless explicitly needed, as they usually contain sensitive information. Like with `APIConnection`, the parameters of `APISharingProfile` aren't ever really populated in the outbound direction; they're used for update based on received data from the client. If the parameters are needed by the client, that information is queried directly from the `.../parameters` resource, and permission to do *that* is explicitly denied unless `UPDATE` permission is available (as it's an information leak otherwise).

guacamole/src/main/java/org/apache/guacamole/tunnel/UserTunnel.java
mike-jumper: Heh ... I'll move this space to that spot where the space is missing ...

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
mike-jumper: No, it's just not implemented. These are the API changes only, with the actual implementation of that API coming later.

I figured that if I held off on the review of the API until the underlying implementation was complete, that would distract from the review of the API design itself, and make the PR way too large.

jmuehlner: Ah ok, make sense.

PR #32, opened 2016-07-22
GUACAMOLE-5: Implement connection sharing within database auth

mike-jumper:
This change adds support for generation of temporary user credentials for shared connections.

The implementation produces sets of credentials which define only a single query parameter (`key`), which is accepted by the `AuthenticationProvider` and tied to a new `SharedConnectionUserContext`. The user context provides access only to a single connection which transparently joins the underlying existing connection, with the whole tree of data residing only in memory.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/sharing/HashSharedConnectionMap.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/sharing/HashSharedConnectionMap.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.guacamole.auth.jdbc.sharing;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+/**
+ * A HashMap-based implementation of the SharedConnectionMap.
+ *
+ * @author Michael Jumper
+ */
+public class HashSharedConnectionMap implements SharedConnectionMap {
+
+    /**
+     * Keeps track of the share key to SharedConnectionDefinition mapping.
+     */
+    private final ConcurrentMap<String, SharedConnectionDefinition> connectionMap =
+            new ConcurrentHashMap<String, SharedConnectionDefinition>();
+
+    @Override
+    public SharedConnectionDefinition get(String key) {
+        
+        // There are no null share keys
+        if (key == null)
+            return null;
+
+        // Update the last access time and return the SharedConnectionDefinition
+        return connectionMap.get(key);
+
+    }
+
+    @Override
+    public void put(String key, SharedConnectionDefinition definition) {
+        connectionMap.put(key, definition);
+    }
+
+    @Override
+    public SharedConnectionDefinition remove(String key) {
+
+        // There are no null share keys
+        if (key == null)
+            return null;
+
+        // Attempt to retrieve only if non-null
+        return connectionMap.remove(key);
+
+    }
+
+}
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/sharing/SecureRandomShareKeyGenerator.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/sharing/SecureRandomShareKeyGenerator.java
@@ -1,0 +1,45 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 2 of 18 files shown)


extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/sharing/HashSharedConnectionMap.java
```diff
@@ -0,0 +1,67 @@
...
+            return null;
+
+        // Update the last access time and return the SharedConnectionDefinition
+        return connectionMap.get(key);
+
+    }
+
+    @Override
+    public void put(String key, SharedConnectionDefinition definition) {
+        connectionMap.put(key, definition);
```
jmuehlner: No null check here?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/sharing/SecureRandomShareKeyGenerator.java
```diff
@@ -0,0 +1,45 @@
...
+public class SecureRandomShareKeyGenerator implements ShareKeyGenerator {
+
+    /**
+     * Instance of SecureRandom for generating sharing keys.
+     */
+    private final SecureRandom secureRandom = new SecureRandom();
+
+    @Override
+    public String getShareKey() {
+        byte[] bytes = new byte[33];
```
jmuehlner: I'm curious... why 33 bytes?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/sharing/SharedConnection.java
```diff
@@ -0,0 +1,171 @@
...
+
+    @Override
+    public void setAttributes(Map<String, String> attributes) {
+        // Do nothing - no attributes supported
+    }
+
+    @Override
+    public List<? extends ConnectionRecord> getHistory()
+            throws GuacamoleException {
+        return Collections.<ConnectionRecord>emptyList();
```
jmuehlner: We do support tracking of sharing connection history, right? Just don't expose it here?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/sharing/HashSharedConnectionMap.java
mike-jumper: Well, the other two methods do a `null` check so they can safely return that the key cannot possibly be present, as otherwise even attempting to do a lookup with a `null` key on a `ConcurrentHashMap` will result in a `NullPointerException`.

What would be the safe behavior here for such a `null` check? Silently don't store anything? Sounds like an NPE would be preferable (which is what will currently happen thanks to `ConcurrentHashMap`).

jmuehlner: Ok, that seems fair.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/sharing/SecureRandomShareKeyGenerator.java
mike-jumper: It's 33 because the key is output as base64, and thus needs to be a multiple of 3 for each byte of the output to have meaning. If it were 32, there would be one or more useless padding bytes at the end of the base64-encoded result.

If the non-power-of-two thing makes you cringe, we could switch to 32 and hex.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/sharing/SharedConnection.java
mike-jumper: Correct. Access is limited from the perspective of the user that authenticated with a share key, but that info is still stored and exposed elsewhere to a user with sufficient privileges.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/sharing/SecureRandomShareKeyGenerator.java
jmuehlner: No, that's fine. Just curious.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/GuacamoleTunnelService.java
```diff
@@ -145,4 +148,39 @@ GuacamoleTunnel getGuacamoleTunnel(AuthenticatedUser user,
      */
     public Collection<ActiveConnectionRecord> getActiveConnections(ConnectionGroup connectionGroup);
 
+    /**
+     * Creates a socket for the given user which joins the given active
+     * connection. The given client information will be passed to guacd when
+     * the connection is established. This function will apply any concurrent
+     * usage rules in effect, but will NOT test object- or system-level
```
jmuehlner: "object-" is a bit funny looking. Maybe just "object or system-level"? Or "object-level or system-level"?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/sharing/SharedConnection.java
jmuehlner: Alright

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/GuacamoleTunnelService.java
mike-jumper: It's standard. "object or system-level" would be like saying (object) or (system-level), wheras with the repeated hypen it's clear that the "-level" is meant to be common to both.

Take a look at https://en.wikipedia.org/wiki/High-_and_low-level, for example.

jmuehlner: I stand(ard) corrected!

jmuehlner:
Everything looks good to me.

PR #33, opened 2016-07-25
GUACAMOLE-5: Invalidate share keys and close associated tunnels when the shared connection is closed

mike-jumper:
This change alters the behavior of connections shared via the JDBC auth backend such that:

1. The share keys generated for an active connection are invalidated once that connection closes.
2. Share keys created for a closed connection are immediately invalidated.
3. The tunnels created by users of a share key are closed once the share key is invalidated.
4. Tunnels created by users of a share key are immediately closed if the share key is already invalid.

From the user's perspective, the effect of all this is that shared connections automatically close when the original connection is closed, and cannot be reestablished. Reconnecting after a disconnect is only possible for a shared connection if the original connection is still running.

It is particularly important that this is guaranteed regardless of the timing of concurrent share/join attempts, as only the original connection is affected by concurrency limits. If shared connections can somehow outlive the original connection due to race conditions, then concurrency limits may be violated. The `SharedObjectManager` class was written to provide such strict guarantees.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
@@ -109,6 +109,12 @@ public abstract class AbstractGuacamoleTunnelService implements GuacamoleTunnelS
     private ConnectionRecordMapper connectionRecordMapper;
 
     /**
+     * Provider for creating active connection records.
+     */
+    @Inject
+    private Provider<ActiveConnectionRecord> activeConnectionRecordProvider;
+
+    /**
      * The hostname to use when connecting to guacd if no hostname is provided
      * within guacamole.properties.
      */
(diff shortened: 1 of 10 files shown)


extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
```diff
@@ -681,13 +696,22 @@ public GuacamoleTunnel getGuacamoleTunnel(AuthenticatedUser user,
...
     public GuacamoleTunnel getGuacamoleTunnel(SharedConnectionUser user,
-            TrackedActiveConnection activeConnection,
-            ModeledSharingProfile sharingProfile,
+            SharedConnectionDefinition definition,
             GuacamoleClientInformation info)
             throws GuacamoleException {
 
-        // Connect to shared connection
-        return assignGuacamoleTunnel(new ActiveConnectionRecord(user, activeConnection, sharingProfile), info);
+        // Create a connection record which described the shared connection
```
jmuehlner: "described" - past tense?

mike-jumper: OK - fixed.

PR #39, opened 2016-07-29
GUACAMOLE-5: Separate sharing semantics from database-backed AuthenticationProvider

mike-jumper:
The database authentication backend currently implements support for unique "share keys", which can be used as an alternative and temporary method of authentication for accessing shared connections. Prior to this change, it was the duty of the `AuthenticationProvider` to handle both username/password authentication *and* share key authentication.

This change splits the PostgreSQL/MySQL `AuthenticationProvider` implementations into two: one which is specific to username/password auth against the database, and the other which is specific to share key authentication against in-memory shared connections. Both share a common Guice `Injector` and, thus, a common set of core services. Collaborative management of shared connections between the two is achieved through a common service which handles registration and retrieval of shared connections and their corresponding keys.

This change is architectural, as it is a necessary step to cleanly implement support for use of multiple share keys within the same session, even if that session was due to a username/password login. The existing functionality is unaffected.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/PostgreSQLInjectorProvider.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/PostgreSQLInjectorProvider.java
@@ -19,29 +19,33 @@
 
 package org.apache.guacamole.auth.postgresql;
 
-import org.apache.guacamole.auth.jdbc.JDBCAuthenticationProvider;
+import com.google.inject.Guice;
+import com.google.inject.Injector;
+import org.apache.guacamole.GuacamoleException;
+import org.apache.guacamole.auth.jdbc.JDBCAuthenticationProviderModule;
+import org.apache.guacamole.auth.jdbc.JDBCInjectorProvider;
 
 /**
- * Provides a PostgreSQL-based implementation of the AuthenticationProvider
- * functionality.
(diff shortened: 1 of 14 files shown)


mike-jumper:
There may be a better way ... holding off for now.

mike-jumper:
Never mind - this definitely still seems to be the best way. It establishes proper separation of concerns and allows for logging of usernames (if known).

We will need to figure out how to provide the expected user experience for the case that a user is given multiple share links, as the interface will not automatically direct the user to the connection in that case, but that solution should really be independent of this.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/PostgreSQLInjectorProvider.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+public class PostgreSQLInjectorProvider extends JDBCInjectorProvider {
+
+    @Override
+    protected Injector create() throws GuacamoleException {
+
+        // Get local environment
+        PostgreSQLEnvironment environment = new PostgreSQLEnvironment();
+
+        // Set up Guice injector.
```
jmuehlner: In another of my Minor Comment Quibbles™, I'll point out that this is a strange place for a period in a comment.

mike-jumper: No problem. I'll rebase such that it never happened.

PR #40, opened 2016-07-29
port the auth tutorial from glyptodon to apache

OllieJones:
Hello, I worked through the Chapter 18 authorization tutorial. It needed some tweaking to get it to work with the apache incubating version of the system rather than the older glyptodon version.

I have put it into the guacamole client source tree, with a view to saving the next guy the work of sorting it out.


mike-jumper:
Hi @OllieJones,

Fixing the tutorial is well and good, but the source for that is actually part of the incubator-guacamole-manual repository:

https://github.com/apache/incubator-guacamole-manual/tree/master/tutorials/guacamole-auth-tutorial

That would be the place to make such changes. Since the idea of guacamole-auth-tutorial is that it follows the tutorial in the manual, if these changes cause the source to diverge from the manual contents, the manual will need to be updated as well.

Please also keep in mind our [contribution guidelines](https://github.com/apache/incubator-guacamole-manual/blob/master/CONTRIBUTING), specifically:

> 
>     All changes to Guacamole must have corresponding issues in JIRA so the
>     change can be properly tracked:
> 
>         https://issues.apache.org/jira/browse/GUACAMOLE/
> 
>     If you do not already have an account on the Apache Software Foundation's
>     JIRA, you will need to create one before creating your new issue.
> 

There needs to be a corresponding issue in JIRA, and that issue needs to be tagged in the commit message in the same manner as the other messages.

Thanks!

OllieJones:
On Fri, Jul 29, 2016 at 9:55 PM, Mike Jumper <notifications@github.com>
wrote:

> There needs to be a corresponding issue in JIRA,


OK. Thanks.

necouchman:
Ping.

@OllieJones: Don't think a JIRA issue was ever opened?  Can you do this and update the PR so we can push this along and close it out?

necouchman:
@OllieJones Ping...

necouchman:
@OllieJones: Another friendly ping, here, to open a JIRA issue and update the PRs and commits...

PR #41, opened 2016-07-30
GUACAMOLE-5: Distinguish the authenticating AuthenticationProvider from the data source

mike-jumper:
The changes from #36 were partly wrong, as `getAuthenticationProvider()` of `AuthenticatedUser` will not always return the same `AuthenticationProvider` that owns the data within the `UserContext`. As written, the value returned by `getAuthenticationProvider()` of `UserContext` will be wrong for any case where the `AuthenticatedUser` was authenticated by an `AuthenticationProvider` that isn't the source of that same `UserContext`.

This change establishes a clear distinction, within the context of the database extension, between the `AuthenticationProvider` which authenticated the user and the database-specific `AuthenticationProvider` implementation which serves as the data source.

Much of this is a rename of the database-specific `AuthenticatedUser` implementation to `ModeledAuthenticatedUser`, and `UserContext` to `ModeledUserContext`, both for the sake of future clarity, but there are also substantive changes which fix the actual bug introduced by #36, particularly in the use of the new `getModelAuthenticationProvider()` instead of `getAuthenticationProvider()`.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/AuthenticationProviderService.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/AuthenticationProviderService.java
@@ -63,6 +63,10 @@ public interface AuthenticationProviderService  {
      * user. A new placeholder account will be created for any user that does
      * not already exist within the database.
      *
+     * @param authenticationProvider
+     *     The AuthenticationProvider on behalf of which the UserContext is
+     *     being produced.
+     *
      * @param authenticatedUser
      *     The credentials to use to produce the UserContext.
      *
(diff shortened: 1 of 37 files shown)


jmuehlner:
In which case would you have an AuthenticatedUser who was authenticated by a different AuthenticationProvider than the one who made the UserContext for that user?

mike-jumper:
Yes. This is how multiple extensions can work together, like the common LDAP + database configuration. The authentication process is as follows:

1. Each `AuthenticationProvider` is polled, in order, with the `Credentials` supplied in the authentication request.
2. If successful, the `AuthenticationProvider` returns an `AuthenticatedUser` object representing that user's identity as well as the `Credentials` supplied when that user authenticated.
3. Each `AuthenticationProvider` is polled again, in order, with the `AuthenticatedUser` object from the extension which successfully identified the user.
4. If any `AuthenticationProvider` wishes to provide data for the authenticated user, they return a new `UserContext` which exposes that data.
5. Each `UserContext` is exposed via the REST API separately, and the availability of each is exposed via the JSON authentication response to the client. The user interface unifies the view of all this data so this is all transparent.

jmuehlner:
Ok, that makes sense.

PR #42, opened 2016-07-30
GUACAMOLE-5: Provide most recent Credentials to updateUserContext()

mike-jumper:
The `updateUserContext()` function of `AuthenticationProvider` has existed for a while, but it was written under the assumption that `AuthenticatedUser` will always receive an updated set of credentials, thus the possibly-updated `UserContext` will be able to access this information.

This is false. The `AuthenticatedUser` will only be updated if `updateAuthenticatedUser()` does so, and relying on that is bad. There's not much for `updateUserContext()` to update with if it can't be guaranteed access to fresh information.

Updating `UserContext` may require peeking at the most recent HTTP request / credentials, so the most recent `Credentials` object should be passed to `updateUserContext()` as well.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/InjectedAuthenticationProvider.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/InjectedAuthenticationProvider.java
@@ -95,7 +95,8 @@ public abstract class InjectedAuthenticationProvider implements AuthenticationPr
 
     @Override
     public UserContext updateUserContext(UserContext context,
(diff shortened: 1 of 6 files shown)


jmuehlner:
So none of the auth providers are actually using the new Credentials as of yet?

mike-jumper:
Correct, though I will have to do so when adding support for multiple share keys within the same session.

In the case of screen sharing via the database auth backend, there will be at least two or three auth providers involved:

1. The auth provider which authenticated the user (not necessarily part of the database backend, but likely so)
2. The auth provider which stores the connection data and tracks the shared connections (either the PostgreSQL or MySQL auth providers).
3. The secondary auth provider used by the database backend to track strictly shared connections.

The secondary auth provider will only be the one that authenticates the user if they authenticated using a share key in the first place. If they authenticated via the database, or even via LDAP, then the `AuthenticatedUser` comes from elsewhere and cannot be guaranteed to have the new `Credentials` (ie: any new share keys).

PR #44, opened 2016-07-30
GUACAMOLE-5: Fix identifier validity regression.

mike-jumper:
Pull request #38 introduced a regression which made it impossible to retrieve specific users. This is because the assumption driving #38 (that all identifiers within the JDBC auth are numeric strings) is incorrect. `User` objects within the JDBC auth have pure string identifiers which are not derived from internal numeric IDs and need not be numeric.

This change moves the validity checking from the base `ObjectModel` class into the object-specific `ModeledDirectoryObjectService` class, such that the `Directory` implementation itself can override the default numeric-only behavior if the objects truly do not require numeric identifiers.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
@@ -23,7 +23,6 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Set;
-import javax.xml.stream.events.Characters;
 import org.apache.guacamole.auth.jdbc.user.ModeledAuthenticatedUser;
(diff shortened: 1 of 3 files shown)


jmuehlner:
So how does the ObjectModel objectID field work for users then? The comments specify that it should be null only in the case that the object was not retrieved from the DB. I'm not sure if this makes sense in this context.

mike-jumper:
In the case of users, `ObjectModel` is the same - there is a numeric user ID. It's just the relationship between user ID and user identifier is different than that for connections and connection groups.

This actually happens naturally due to the identifier and object ID being separate. The difference in handling happens for the other objects, where the direct tie between object ID and identifier is enforced. For example, from `ConnectionModel`:

```java

    @Override
    public String getIdentifier() {

        // If no associated ID, then no associated identifier
        Integer id = getObjectID();
        if (id == null)
            return null;

        // Otherwise, the identifier is the ID as a string
        return id.toString();

    }

    @Override
    public void setIdentifier(String identifier) {
        throw new UnsupportedOperationException("Connection identifiers are derived from IDs. They cannot be set.");
    }
```

That type of override is not present in `UserModel`, and the specific SQL queries surrounding users take that into account.

mike-jumper:
Rebased to resolve conflicts with other recent merges.

PR #45, opened 2016-07-31
GUACAMOLE-5: Add support for multiple share keys.

mike-jumper:
This change (finally) leverages the changes from the other recent PRs to implement support for multiple share keys. Rather than create a `UserContext` which is restricted to a single connection (the connection associated with the share key), the initial share key only serves as the means of authentication if no other means has been used. Any further share keys received via new `Credentials` add further connections to the existing `UserContext`.

This has the effect that:

1. Legitimate users of the Guacamole system can use share keys without logging out first.
2. Accessing a shared connection while logged in as a normal Guacamole user causes the correct username to be recorded in the history.
3. Clicking a share key while already in a Guacamole session results in that connection being added to the existing session, rather than having no effect at all.

This change also leverages the HTML patching features of the extension subsystem to add a nifty "shared by" tag next to the shared connection in the home screen, now that such a thing can exist:

shared-by-tag


extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/sharing/connectiongroup/SharedRootConnectionGroup.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    @Override
+    public void setAttributes(Map<String, String> attributes) {
+        // Do nothing - no attributes supported
+    }
+
+    @Override
+    public int getActiveConnections() {
+        return 0;
```
jmuehlner: Why is this hardcoded to 0?

mike-jumper: Simply because the number of active connections shared connections by a particular user is not tracked. From the definition of `getActiveConnections()` within `Connectable`:

> Implementations may simply return 0 if this value is not tracked.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/sharing/connectiongroup/SharedRootConnectionGroup.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    @Override
+    public void setType(Type type) {
+        throw new UnsupportedOperationException("The root group is immutable.");
+    }
+
+    @Override
+    public Type getType() {
+        return Type.BALANCING;
```
mike-jumper: This is bonkers, however. It should be `ORGANIZATIONAL`.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/sharing/connectiongroup/SharedRootConnectionGroup.java
jmuehlner: Ok

PR #46, opened 2016-08-02
GUACAMOLE-5: Implement user-facing connection sharing interface.

mike-jumper:
This change adds a new "Share" menu to the Guacamole menu, right next to the existing user menu, which appears only if the current user has permission to share the current connection. The menu lists all sharing profiles available to the user. When a sharing profile is clicked, a share link is generated and displayed within the Guacamole menu:

connection-sharing

These links persist within the Guacamole menu of the connection in question. If the user navigates to a different part of Guacamole while a connection is running in the background, the links associated with that connection will still be visible in the Guacamole menu when they return.


guacamole/src/main/webapp/app/navigation/templates/guacMenu.html
```diff
@@ -0,0 +1,7 @@
+<div class="menu-dropdown" ng-class="{open: menuShown}" ng-click="toggleMenu()">
+    <div class="menu-title">{{menuTitle}}</div>
+    <div class="menu-indicator"></div>
+
+    <!-- Menu options -->
+    <div class="options" ng-transclude></div>
```
jmuehlner: Why is this called options? It's basically just the content of whatever you want to throw in there, right?

mike-jumper: It's the container for the menu options which are displayed when the menu is open, which are provided by transclusion in the case of this directive.

mike-jumper: The naming itself comes from its use within the `guacUserMenu` directive, which is where the meat of the `guacMenu` directive comes from.

Now that the menu is all abstract, it might make more semantic sense for this to be `menu-contents`.

jmuehlner:
Ok, looks better now.

PR #47, opened 2016-08-05
GUACAMOLE-5: Add support for managing sharing profile permissions.

mike-jumper:
Though previous changes did add support within the REST API for management of sharing profile permissions, such support remained absent in the admin interface.

This change adds support for managing sharing profile permissions on a per-user basis, as well as granting/revoking the system-level permission for creating sharing profiles. Management of sharing profile permissions is handled within the same hierarchical tree previously used only for connections and connection groups, handling the sharing profiles associated with a connection as if they were the connections within a group:

sharing-profile-permission-management

To make the tree more readable, I altered the rendering of the tree lines such that they have the traditional right-angle bends and terminate at the rough midpoint of each list item. This style is inherited by the other parts of the interface which use the `guacGroupList` directive, including the home screen and the location chooser (the dropdown used within the management interface for selecting a parent connection group).

These changes also required altering the connection group tree REST resource such that it includes sharing profiles within the tree.

### DIFF
--- a/guacamole/src/main/webapp/app/manage/controllers/manageUserController.js
+++ b/guacamole/src/main/webapp/app/manage/controllers/manageUserController.js
@@ -709,10 +709,10 @@ angular.module('manage').controller('manageUserController', ['$scope', '$injecto
     $scope.systemPermissionChanged = function systemPermissionChanged(type) {
 
         // Determine current permission setting
-        var value = $scope.permissionFlags.systemPermissions[type];
+        var granted = $scope.permissionFlags.systemPermissions[type];
 
         // Add/remove permission depending on flag state
-        if (value)
+        if (granted)
             addSystemPermission(type);
         else
             removeSystemPermission(type);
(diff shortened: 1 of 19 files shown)


guacamole/src/main/webapp/app/manage/controllers/manageUserController.js
```diff
@@ -918,6 +961,28 @@ angular.module('manage').controller('manageUserController', ['$scope', '$injecto
...
+         * applies to READ permissions.
+         *
+         * @param {String} identifier
+         *     The identifier of the sharing profile affected by the changed
+         *     permission.
+         */
+        sharingProfilePermissionChanged : function sharingProfilePermissionChanged(identifier) {
+
+            // Determine current permission setting
+            var value = $scope.permissionFlags.sharingProfilePermissions.READ[identifier];
```
jmuehlner: Small quibble, but "value" is not a very meaningful variable name here (to me). Are the expected values true/false?

mike-jumper: Yes, mostly. It will be true if the permission is granted, and false or undefined otherwise. The `PermissionFlagSet` definition states (copied from the diff of this PR):

>
> ```
 +         * The granted state of each permission for each sharing profile, as a
 +         * map of object permission type string to permission map. The
 +         * permission map is, in turn, a map of sharing profile identifier to
 +         * boolean value. A particular permission is granted if its
 +         * corresponding boolean value is set to true. Valid permission type
 +         * strings are defined within PermissionSet.ObjectPermissionType.
 +         * Permissions which are not granted may be set to false, but this is
 +         * not required.
> ```
>

How would you feel about `granted` instead?

jmuehlner: That seems much better to me.

mike-jumper: OK - change made via f3a201a. I went ahead and updated the other four cases where this naming was used within the same controller.

PR #56, opened 2016-08-13
GUACAMOLE-78: Handle anonymous users distinctly.

mike-jumper:
There has historically been no way to properly represent an anonymous user within the Guacamole extension API. All users are required to have unique identifiers, and the user interface renders a user-specific menu with logout button at all times.

This change defines a reserved identifier (`AuthenticatedUser.ANONYMOUS_IDENTIFIER` in Java, `AuthenticationResult.ANONYMOUS_USERNAME` in JavaScript, in both cases simply the empty string) which implicitly declares the user as anonymous, but must not be used for any other purpose. The JDBC authentication has been modified to return this username when a user authenticated via a share key (and thus has no associated user account).

Unlike a traditionally-authenticated user, an anonymous user is not presented the "user menu" which  typically contains the "Home", "Settings", and "Logout" options. When rendered within the admin interface, anonymous users are listed as "Anonymous" with contrasting styling:

anonymous-username-style

Anonymous users' authentication results are also not persisted via a cookie like normal users, thus their session is transient. Closing the tab will effectively log the user out.

Part of this involved actually defining a type within JavaScript for the JSON returned from authentication attempts (now `AuthenticationResult`, in the `auth` module), which up until now has never been documented client-side.

### DIFF
--- a/guacamole/src/main/webapp/app/index/config/indexRouteConfig.js
+++ b/guacamole/src/main/webapp/app/index/config/indexRouteConfig.js
@@ -88,7 +88,7 @@ angular.module('index').config(['$routeProvider', '$locationProvider',
 
                 // Otherwise, reject and reroute
                 else {
-                    $location.url(homePage.url);
+                    $location.path(homePage.url);
                     route.reject();
                 }
 
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/sharing/user/SharedAuthenticatedUser.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/sharing/user/SharedAuthenticatedUser.java
@@ -19,7 +19,6 @@
 
 package org.apache.guacamole.auth.jdbc.sharing.user;
 
-import java.util.UUID;
 import org.apache.guacamole.auth.jdbc.user.RemoteAuthenticatedUser;
 import org.apache.guacamole.net.auth.AuthenticatedUser;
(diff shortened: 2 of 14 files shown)


jmuehlner:
As far as changing the username displayed in the active sessions list, isn't this change removing information, and making it harder to determine which anonymous user is which?

I guess the gist of the question is, did the old method where we displayed the anonymous user's unique ID instead of the string "Anonymous" allow for easier identification, or differentiation between users? If not, I'm fine with this behavior change.

guacamole/src/main/webapp/app/index/config/indexRouteConfig.js
```diff
@@ -88,7 +88,7 @@ angular.module('index').config(['$routeProvider', '$locationProvider',
 
                 // Otherwise, reject and reroute
                 else {
-                    $location.url(homePage.url);
+                    $location.path(homePage.url);
```
jmuehlner: Is this change in order to preserve potential url parameters?

mike-jumper: Yep. Lacking this change, URL parameters are wiped out when the user is routed to their only connection.

Without the share key continuing to be present in the URL, reloading the tab fails results in a login page rather than the connection they were just using (a pretty severe violation of the [Principle of Least Surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)). Navigating back to the connection after visiting a different page would fail for the same reason.

mike-jumper:
> EDIT: Rereading the code informs me that this is a silly question - the identifier is randomly generated and adds no useful information. So nevermind...

Yeah - the random identifier was pretty useless to begin with. The most useful information in the case of an anonymous user would be the remote host, and that's still present.

PR #57, opened 2016-08-14
GUACAMOLE-79: Adding an option to allow the default LDAP result size limit (1000) to be overridden

hguehl:
This is a humble contribution for large Ldap directory (over 1000 accounts).

### DIFF
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
@@ -206,4 +206,23 @@ public class ConfigurationService {
         );
     }
 
+    /**
+     * Returns maximum number of results a LDAP query can return,
+     * as configured with guacamole.properties.
+     * By default, this will be 1000.
+     *
+     * @return
+     *     The maximum number of results a LDAP query can return,
+     *     as configured with guacamole.properties.
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed.
+     */
+    public int getMaxResults() throws GuacamoleException {
+        return environment.getProperty(
+            LDAPGuacamoleProperties.LDAP_MAX_SEARCH_RESULTS,
+            1000 
+        );
+    }
+
 }
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
@@ -145,4 +145,14 @@ public class LDAPGuacamoleProperties {
 
     };
 
+    /**
+     * The maximum number of results a LDAP query can return.
+     */
+    public static final IntegerGuacamoleProperty LDAP_MAX_SEARCH_RESULTS = new IntegerGuacamoleProperty() {
+
+        @Override
+        public String getName() { return "ldap-max-search-results"; }
+
+    };
+
 }
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
@@ -25,6 +25,7 @@ import com.novell.ldap.LDAPConnection;
 import com.novell.ldap.LDAPEntry;
 import com.novell.ldap.LDAPException;
 import com.novell.ldap.LDAPSearchResults;
+import com.novell.ldap.LDAPSearchConstraints;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -86,6 +87,9 @@ public class UserService {
             String usernameAttribute) throws GuacamoleException {
 
         try {
+            // Set search limits
+            LDAPSearchConstraints constraints = new LDAPSearchConstraints();
+            constraints.setMaxResults(confService.getMaxResults());
 
             // Find all Guacamole users underneath base DN
             LDAPSearchResults results = ldapConnection.search(
@@ -93,7 +97,8 @@ public class UserService {
                 LDAPConnection.SCOPE_SUB,
                 "(&(objectClass=*)(" + escapingService.escapeLDAPSearchFilter(usernameAttribute) + "=*))",
                 null,
-                false
+                false,
+                constraints
             );
 
             // Read all visible users


extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
```diff
@@ -206,4 +206,22 @@ public EncryptionMethod getEncryptionMethod() throws GuacamoleException {
         );
     }
 
+    /**
+     * Returns Max results a ldap query can return
```
jmuehlner: It looks like maybe a few words were left out here? Did you mean to say "return, as specified with guacamole.properties"?

jmuehlner: Also, when used in a comment, the acronym LDAP should be in all uppercase.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
```diff
@@ -145,4 +145,17 @@ private LDAPGuacamoleProperties() {}
 
     };
 
+    /**
+     * The maxResult a ldap search can return
```
jmuehlner: The formatting of this comment is a bit off (asterisks don't line up, capitalization, no period at the end of line).

Also, "The max results" would read better than "the maxResult".

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
```diff
@@ -86,14 +87,18 @@ private void putAllUsers(Map<String, User> users, LDAPConnection ldapConnection,
             String usernameAttribute) throws GuacamoleException {
 
         try {
+	    //change search limits
```
jmuehlner: For single line comments, we like to start with a capital letter, and leave a space after the slashes:

// This is a friendly comment

jmuehlner:
@mike-jumper You're more familiar with LDAP than I am. Can you comment on this approach as a whole?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
mike-jumper: And in the same vein, "max" should similarly *not* be capitalized.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
```diff
@@ -145,4 +145,17 @@ private LDAPGuacamoleProperties() {}
...
+     *
```
mike-jumper: Why is this `*` misaligned from the rest?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
mike-jumper: And "The maximum number of results" would be better still. Just "max results" seems rather wonky to me.

mike-jumper:
> @mike-jumper You're more familiar with LDAP than I am. Can you comment on this approach as a whole?

The only issues I have with the code are with the surrounding documentation. It needs cleanup for readability and consistency.

In terms of what the code actually *does*, everything looks fine to me. From the [documentation for `setMaxResults()`](https://www.novell.com/documentation/developer/jldap/jldapenu/api/com/novell/ldap/LDAPSearchConstraints.html#setMaxResults(int)):

> Sets the maximum number of search results to be returned from a search operation. The value 0 means no limit. The default is 1000. The search operation will be terminated with an LDAPException.SIZE_LIMIT_EXCEEDED if the number of results exceed the maximum.

The problem addressed by [GUACAMOLE-79](https://issues.apache.org/jira/browse/GUACAMOLE-79) looks like an internally-enforced limitation of the LDAP library, and it looks like the code is doing the only what is necessary to override that default limit if desired. Overall, this is definitely an improvement.

Assuming the readability/consistency issues get addressed, we still probably shouldn't merge the changes (once ready) until after 0.9.10-incubating is released, to avoid unnecessarily increasing the scope and testing surface of the release.

mike-jumper:
I guess my only other nitpick would be that the commit message, "LDAPConnection has a size limit of 1000", doesn't really describe the high level nature of the change. It describes the *need* for the change, but that's already covered by the "GUACAMOLE-79" issue tag.

It would be much better if the history recorded what is actually happening. In this case, we're adding an option to allow the default LDAP result size limit to be overridden.

hguehl:
Hi Guys,
updated the comments. Hope this will fit your requirements.
Hervé.
Ps: sorry for the number of commits, had to learn to deal with git

jmuehlner:
Hey Hervé, looks much better to me. The only thing remaining before we can merge this would be to clean up the git history a bit. We like to preserve a clean history, with logical commits. The last commit message you made looks like it describes the whole change pretty well.

What do you think of replacing the 10 commits you currently have with a single commit, having that message? Luckily this is pretty easy to do in git.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
```diff
@@ -145,4 +145,16 @@ private LDAPGuacamoleProperties() {}
 
     };
 
+    /**
+     * The maximum number of results a LDAP query can return.
+     */
+    public static final IntegerGuacamoleProperty LDAP_MAX_SEARCH_RESULTS = new IntegerGuacamoleProperty() {
+
+        @Override
+        public String getName() { return "ldap-maxresults"; }
```
mike-jumper: Just noticed:

`LDAP_MAX_SEARCH_RESULTS` is very nice and readable, so why is the property just `ldap-maxresults`? (Nice dash following the "ldap", but not between "max" and "results"). For consistency with other properties used by Guacamole, this should at least be `ldap-max-results`, though given the nice `LDAP_MAX_SEARCH_RESULTS` name, I'd suggest `ldap-max-search-results` instead.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
```diff
@@ -86,14 +87,18 @@ private void putAllUsers(Map<String, User> users, LDAPConnection ldapConnection,
             String usernameAttribute) throws GuacamoleException {
 
         try {
+	    // Set search limits
+	    LDAPSearchConstraints constraints = new LDAPSearchConstraints();
+	    constraints.setMaxResults(confService.getMaxResults());
```
mike-jumper: Didn't notice this before, but please use 4-space indents (no tabs). I saw that these three lines weren't quite aligned with the surrounding code, and decided to download the patch to take a closer look. The misalignment is due to the tabs (see yellow highlight in screenshot below):

ick-tabs

You might have to change your editor settings to get it to stop adding tabs instead of spaces.

Other than that, I see no further issues. With the exception of the tabs, all looks good to me, even the commit history. ;)

hguehl:
Tab Corrected

jmuehlner:
Looks good; we'll merge this after the release of 0.9.9-incubating.

mike-jumper:
> Looks good; we'll merge this after the release of 0.9.9-incubating.

0.9.10-incubating *

wnovak:
Any status when this pull request will be merged into master ?

mike-jumper:
@wnovak - I've started a new thread on the dev mailing list to hopefully address that:

http://mail-archives.apache.org/mod_mbox/incubator-guacamole-dev/201611.mbox/%3CCALKeL-NLti4gYHc22wYXQDJhcVza0RYfsOpeLawFsmE8iwCovg%40mail.gmail.com%3E

PR #58, opened 2016-08-17
GUACAMOLE-73: Clear permission cache when saving or updating to force permission reload.

jmuehlner:
The permissions live in the user cache, so clearing the user cache upon save/update of connections or connection groups solves this issue.


mike-jumper:
What about when you add a user or a sharing profile?

jmuehlner:
Good catch!

PR #60, opened 2016-08-20
GUACAMOLE-76: Fix N+1 problem when querying connection group tree.

mike-jumper:
Speed of the connection group tree REST resource fell dramatically with the addition of sharing profiles. This is because the `getSharingProfileIdentifiers()` function of `Connection` issues a new query for every call, and is not affected by retrieval of connections in batches via `getAll()` on the `Directory<Connection>`.

This change addresses this by retrieving child object identifiers when the parents are retrieved. If multiple parents are retrieved, the identifiers of each of their children is retrieved with *one query* (rather than *N queries*). This is accomplished through issuing multiple `SELECT` statements within the same logical query defined in the MyBatis mapping. Beware that while PostgreSQL will handle multiple statements within a single query by default, the MySQL JDBC driver requires a special `allowMultiQueries` property for this.

We should look into making the connection tree retrieval process more efficient, as there are definitely ways it could be sped up (use breadth-first only if attempts to retrieve the entire directory contents at once fail?), but this should fix things with respect to release-blocking regressions. Query performance should be back to where it was.


jmuehlner:
Looks good to me.

PR #61, opened 2016-08-20
GUACAMOLE-77: Ensure non-existent user accounts are not displayed.

mike-jumper:
This change establishes the distinction within the REST API between the user object *representing* the current user (`.../self`) and the user object representing an *existing* user (`.../users/{username}`).

This distinction has existed within `UserContext` for some time, but was never exposed within the REST API. It's an important distinction, as just because a backend agrees on a user's identity does not mean the user actually exists within the backend.

By implementing the above change (b23dcf8), and by not listing pseudo-accounts within the user directory of the sharing auth provider (36dc375), the non-relevant pseudo-accounts are no longer displayed as options in the admin interface. In case they ever *do* get displayed somewhere unexpectedly, they will at least be readable, as I've also added actual translation strings (87b517f).


jmuehlner:
Looks good to me.

PR #62, opened 2016-08-21
GUACAMOLE-81: Verify primary connection permissions when modifying sharing profiles.

mike-jumper:
This change generalizes the connection group / connection permission logic such that it can be applied to any parent/child relationship, and reapplies that logic against sharing profiles and their primary (parent) connections.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ChildObjectModel.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ChildObjectModel.java
@@ -20,12 +20,13 @@
 package org.apache.guacamole.auth.jdbc.base;
 
 /**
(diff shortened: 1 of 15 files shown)


extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ChildObjectModel.java
```diff
@@ -20,12 +20,12 @@
 package org.apache.guacamole.auth.jdbc.base;
 
 /**
- * Object representation of a Guacamole object, such as a user or connection,
- * as represented in the database.
+ * Object representation of a Guacamole object which can be the child of another
+ * object, such as a user or connection, as represented in the database.
```
jmuehlner: Can users be the children of other objects?

mike-jumper: They can be the children of other *humans* ...

I'll fix that.

PR #65, opened 2016-08-26
GUACAMOLE-95: Minor change to postgresql schema. Fixes saving sessions to history. 

ljruiten:
The colum sharing_profile_name in guacamole_connection_history was set to NOT NULL. Regular sessions do not have a sharing_profile_name so this caused an exception when guacamole tried to save them. Changed it to DEFAULT NULL (like in the mysql schema) to solve to problem.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/schema/001-create-schema.sql
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/schema/001-create-schema.sql
@@ -359,7 +359,7 @@ CREATE TABLE guacamole_connection_history (
   connection_id        integer      DEFAULT NULL,
   connection_name      varchar(128) NOT NULL,
   sharing_profile_id   integer      DEFAULT NULL,
-  sharing_profile_name varchar(128) NOT NULL,
+  sharing_profile_name varchar(128) DEFAULT NULL,
   start_date           timestamptz  NOT NULL,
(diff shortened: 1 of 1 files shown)


mike-jumper:
Hey @ljruiten,

This all looks great, but please also create an issue in JIRA (and tag that issue in the commit message). We require this for all changes for the sake of proper tracking, and it's exceptionally important for this contribution, as it fixes what would otherwise be a release-blocking defect.

From the [contribution guidelines](https://github.com/apache/incubator-guacamole-client/blob/master/CONTRIBUTING):

> 
> All changes to Guacamole must have corresponding issues in JIRA so the change can be properly tracked:
> 
> https://issues.apache.org/jira/browse/GUACAMOLE/
> 
> ... Commits should be made in logical units and must reference the JIRA issue number:
>
>

Thanks!

ljruiten:
Yes, I expected that, but I thought for the one word that I changed.... Does it follow the guidelines now?

mike-jumper:
> Does it follow the guidelines now?

Yup. Much better. Thanks again.

> Yes, I expected that, but I thought for the one word that I changed....

Well, the complexity of the change isn't the issue. For a maintainable project, changes require at least:

1. Making the code change itself.
2. Documenting that change.

The ease of making the change doesn't remove the need for documentation. For very simple changes, the documentation may end up being more work than the change itself, but both are still very necessary.

In the case of Guacamole, the specific changes are documented via the git history, while the effort surrounding those changes, their relation to releases and other changes, etc. is documented via JIRA.

PR #68, opened 2016-09-03
GUACAMOLE-101: add additional search filter function for extentions/guacamole-auth-ldap module

annieweng:
update extentions/guacamole-auth-ldap module.
Provide a way to do additional filter of ldap authentication.  the "ldap-additional-search-filter" can be added to guacamole.properties. it is default to empty string.  it's also flexible enough to accept any ldap search filter parameters.
  For example,  to ensure user login is also part of user group "exampleGroup"  
following filter indicate user have to be memeberof exampleGroup:

```
ldap-additional-search-filter: (memberOf=cn=exampleGroup,ou=groups,dc=example,dc=net) 
```

### DIFF
--- a/extensions/guacamole-auth-ldap/README
+++ b/extensions/guacamole-auth-ldap/README
@@ -91,6 +91,11 @@ guacamole.properties such that the authentication provider is available.
 
     # The base DN within which all role based groups can be found.
     ldap-group-base-dn:      ou=groups,dc=example,dc=net
+    
+    # The additional search filter for the user to be authenticated. 
+    # for example , following filter indicate user have to be memeberof exampleGroup. otherwise, authentication will failed.
+    #the filter is also used to filter out all users visible to guacadmin
+    ldap-additional-search-filter: (memberOf=cn=exampleGroup,ou=groups,dc=example,dc=net) 
 
 ------------------------------------------------------------
  Reporting problems
--- a/extensions/guacamole-auth-ldap/README
+++ b/extensions/guacamole-auth-ldap/README
@@ -92,10 +92,10 @@ guacamole.properties such that the authentication provider is available.
     # The base DN within which all role based groups can be found.
     ldap-group-base-dn:      ou=groups,dc=example,dc=net
     
-    # The additional search filter for the user to be authenticated. 
-    # for example , following filter indicate user have to be memeberof exampleGroup. otherwise, authentication will failed.
-    #the filter is also used to filter out all users visible to guacadmin
-    ldap-additional-search-filter: (memberOf=cn=exampleGroup,ou=groups,dc=example,dc=net) 
+    
+    # The ldap filter which appended to the base DN and user identifier attribute,
+    # produces the full DN of the user being authenticated.
+    ldap-user-filter: (memberOf=cn=exampleGroup,ou=groups,dc=example,dc=net) 
 
 ------------------------------------------------------------
  Reporting problems
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
@@ -234,52 +234,51 @@ public class AuthenticationProviderService {
 
         boolean authenticated=true;
         // Check if login in user also meet additional search filter.
-        if(confService.getUserSearchFilter() !=null)
-        {
+        if (confService.getUserSearchFilter() != null) {
+
             authenticated=false;
-    	    for (String usernameAttribute : confService.getUsernameAttributes()) {
+            for (String usernameAttribute : confService.getUsernameAttributes()) {
                 try {
-    			     String ldapSearchFilter= "(&(objectClass=*)(" + escapingService.escapeLDAPSearchFilter(usernameAttribute) + "="+credentials.getUsername()+")"
-    						+confService.getUserSearchFilter().trim()+")";
+                    String ldapSearchFilter= "(&(objectClass=*)(" + escapingService.escapeLDAPSearchFilter(usernameAttribute) + "="+credentials.getUsername()+")"
+                            +confService.getUserSearchFilter().trim()+")";
 
-    				 logger.debug("ldap search filter is :"+ldapSearchFilter);
-    				 // Find all Guacamole users underneath base DN
-    				 LDAPSearchResults results = ldapConnection.search(
-    						confService.getUserBaseDN(),
-    						LDAPConnection.SCOPE_SUB,
-    						ldapSearchFilter,
-    						null,
-    						false
-    						);
+                    logger.debug("ldap search filter is :"+ldapSearchFilter);
+                    // Find all Guacamole users underneath base DN
+                    LDAPSearchResults results = ldapConnection.search(
+                            confService.getUserBaseDN(),
+                            LDAPConnection.SCOPE_SUB,
+                            ldapSearchFilter,
+                            null,
+                            false
+                            );
 
-    				 // Read all visible users
-    				 if (results.hasMore()) {	
-    					 authenticated=true;
-    					 logger.debug("LDAP search find at least one match with filter: "+ldapSearchFilter);
-    					
-    				 }
-    			}
-    			catch (LDAPException e) {
-    				logger.warn("LDAP search failed with user filter: {} ", e.getMessage());
-    				throw new GuacamoleServerException("Error while querying users with additional filter", e);
-    			}
-    		}
-    	}
+                    // Read all visible users
+                    if (results.hasMore()) {	
+                        authenticated=true;
+                        logger.debug("LDAP search find at least one match with filter: "+ldapSearchFilter);
+
+                    }
+                }
+                catch (LDAPException e) {
+                    logger.warn("LDAP search failed with user filter: {} ", e.getMessage());
+                    throw new GuacamoleServerException("Error while querying users with additional filter", e);
+                }
+            }
+        }
 
 
         try {
-             if (authenticated)
-    	     {
-    		    // Return AuthenticatedUser if bind succeeds
-    		     AuthenticatedUser authenticatedUser = authenticatedUserProvider.get();
-    		     authenticatedUser.init(credentials);
-    		     return authenticatedUser;
-    	     }
-    	     else
-    	     {
-    		     throw new GuacamoleInvalidCredentialsException("Permission denied. doesn't meet additional filter"+confService.getUserSearchFilter(),
-    				CredentialsInfo.USERNAME_PASSWORD);
-    	     }
+            if (authenticated) {
+                // Return AuthenticatedUser if bind succeeds
+                AuthenticatedUser authenticatedUser = authenticatedUserProvider.get();
+                authenticatedUser.init(credentials);
+                return authenticatedUser;
+            }
+            else
+            {
+                throw new GuacamoleInvalidCredentialsException("Permission denied. doesn't meet additional filter"+confService.getUserSearchFilter(),
+                        CredentialsInfo.USERNAME_PASSWORD);
+            }
         }
 
         // Always disconnect
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
@@ -80,8 +80,8 @@ public class AuthenticationProviderService {
     private Provider<UserContext> userContextProvider;
 
     /**
-     * Service for escaping parts of LDAP queries.
-     */
+    * Service for escaping parts of LDAP queries.
+    */
     @Inject
     private EscapingService escapingService;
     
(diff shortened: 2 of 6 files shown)


jmuehlner:
Hey Annie, thanks for the contribution! There's a few things that would need to be done before we can accept any change, however.

First, any pull requests need an associated JIRA issue in the Apache JIRA. Have a look at our [contribution guidelines](https://guacamole.incubator.apache.org/open-source/) for more information.

In addition, there seem to be some formatting issues in this code. We like to have all of our code follow the same coding style - I have pointed out a few places in the code where these issues exist, but in general just doing things the same way as seen in other parts of the code is a good bet.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -205,38 +214,80 @@ private LDAPConnection bindAs(Credentials credentials)
...
-
-        // Always disconnect
-        finally {
-            ldapService.disconnect(ldapConnection);
-        }
-
+    		throws GuacamoleException {
+
+    	// Attempt bind
+    	LDAPConnection ldapConnection;
```
jmuehlner: This should be indented with spaces, not tabs. It should be 4 spaces per indentation level.a

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -205,38 +214,80 @@ private LDAPConnection bindAs(Credentials credentials)
...
+    	// If bind fails, permission to login is denied
+    	if (ldapConnection == null)
+    		throw new GuacamoleInvalidCredentialsException("Permission denied.", CredentialsInfo.USERNAME_PASSWORD);
+
+    	boolean authenticated=true;
+    	// check if login in user also meet additional search filter.
+    	if(confService.getAdditionalSearchFilter().length()>0)
+    	{
+    		authenticated=false;
+    		for (String usernameAttribute : confService.getUsernameAttributes()) {
```
jmuehlner: More tab issues.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -205,38 +214,80 @@ private LDAPConnection bindAs(Credentials credentials)
...
+    			catch (LDAPException e) {
+    				logger.warn("ldap search failed with additional filter"+confService.getAdditionalSearchFilter());
+    				throw new GuacamoleServerException("Error while querying users with additional filter", e);
+    			}
+    		}
+    	}
+
+
+    try{
+    	if(authenticated)
```
jmuehlner: There should be a space between "try" and the following curly bracket => "try {",
and a space between "if" and the following parenthesis => "if (".

In general, only function calls should appear with no space, .e.g. "results.hasMore()".

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -205,38 +214,80 @@ private LDAPConnection bindAs(Credentials credentials)
...
+    				LDAPSearchResults results = ldapConnection.search(
+    						confService.getUserBaseDN(),
+    						LDAPConnection.SCOPE_SUB,
+    						ldapSearchFilter,
+    						null,
+    						false
+    						);
+
+    				// Read all visible users
+    				while (results.hasMore()) {
```
jmuehlner: Isn't this an infinite loop if you never call next()?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
```diff
@@ -84,14 +84,17 @@
      */
     private void putAllUsers(Map<String, User> users, LDAPConnection ldapConnection,
             String usernameAttribute) throws GuacamoleException {
-
+
```
jmuehlner: We should avoid any leading spaces like this.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -205,38 +214,80 @@ private LDAPConnection bindAs(Credentials credentials)
...
+    		ldapConnection = null;
+    	}
+
+    	// If bind fails, permission to login is denied
+    	if (ldapConnection == null)
+    		throw new GuacamoleInvalidCredentialsException("Permission denied.", CredentialsInfo.USERNAME_PASSWORD);
+
+    	boolean authenticated=true;
+    	// check if login in user also meet additional search filter.
+    	if(confService.getAdditionalSearchFilter().length()>0)
```
jmuehlner: There should be spaces between these operators and operands => ".length() > 0"

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -205,38 +214,80 @@ private LDAPConnection bindAs(Credentials credentials)
...
+    		logger.debug("Error binding with LDAP server.", e);
+    		ldapConnection = null;
+    	}
+
+    	// If bind fails, permission to login is denied
+    	if (ldapConnection == null)
+    		throw new GuacamoleInvalidCredentialsException("Permission denied.", CredentialsInfo.USERNAME_PASSWORD);
+
+    	boolean authenticated=true;
+    	// check if login in user also meet additional search filter.
```
jmuehlner: Inline comments should start with a capital letter => "// Check if"

extensions/guacamole-auth-ldap/README
```diff
@@ -91,6 +91,11 @@ guacamole.properties such that the authentication provider is available.
 
     # The base DN within which all role based groups can be found.
     ldap-group-base-dn:      ou=groups,dc=example,dc=net
+    
+    # The additional search filter for the user to be authenticated. 
+    # for example , following filter indicate user have to be memeberof exampleGroup. otherwise, authentication will failed.
+    #the filter is also used to filter out all users visible to guacadmin
+    ldap-additional-search-filter: (memberOf=cn=exampleGroup,ou=groups,dc=example,dc=net)
```
mike-jumper: "ldap-additional-search-filter" and "The additional search filter for the user to be authenticated" don't really provide enough information to be a useful property name and description. Can you think of something more descriptive? Something which inherently implies, at a high level, what this property is used for?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -205,38 +214,80 @@ private LDAPConnection bindAs(Credentials credentials)
...
+    		logger.error("Cannot bind with LDAP server: {}", e.getMessage());
+    		logger.debug("Error binding with LDAP server.", e);
+    		ldapConnection = null;
+    	}
+
+    	// If bind fails, permission to login is denied
+    	if (ldapConnection == null)
+    		throw new GuacamoleInvalidCredentialsException("Permission denied.", CredentialsInfo.USERNAME_PASSWORD);
+
+    	boolean authenticated=true;
```
mike-jumper: Have you looked into logic which would not need such a flag?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
mike-jumper: I'd go one further and say that using the empty string for default is bad practice. If the property is switched over to use `null` (ie: no default provided), then this check becomes much clearer:

`if (confService.getAdditionalSearchFilter() != null)`

There are other issues here beyond just that check, though. Most IDEs would be warning you at this point to use `isEmpty()`, not to just manually compare the length against zero.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
```diff
@@ -205,5 +205,12 @@ public EncryptionMethod getEncryptionMethod() throws GuacamoleException {
             EncryptionMethod.NONE
         );
     }
+    
+    public String getAdditionalSearchFilter() throws GuacamoleException{
+    	 return environment.getProperty(
+    	        LDAPGuacamoleProperties.ADDITIONAL_LDAP_SEARCH_FILTERS,
+    	        ""
```
mike-jumper: Why not just:

```
return environment.getProperty(LDAPGuacamoleProperties.ADDITIONAL_LDAP_SEARCH_FILTERS);
```

and rely on `null`?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPConnectionService.java
```diff
@@ -154,6 +154,7 @@ public LDAPConnection bindAs(String userDN, String password)
 
             // Bind as user
             ldapConnection.bind(LDAPConnection.LDAP_V3, userDN, passwordBytes);
+
```
mike-jumper: Please do not arbitrarily add blank lines to otherwise-untouched code.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
mike-jumper: More generally, we should avoid purely-whitespace changes. There's no reason for this line to be modified, let alone just for the sake of adding an unused indent.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -205,38 +214,80 @@ private LDAPConnection bindAs(Credentials credentials)
...
+
+    				// Read all visible users
+    				while (results.hasMore()) {	
+    					authenticated=true;
+    					logger.debug("ldap search find at least one match with filter: "+ldapSearchFilter);
+    					break;
+    				}
+    			}
+    			catch (LDAPException e) {
+    				logger.warn("ldap search failed with additional filter"+confService.getAdditionalSearchFilter());
```
mike-jumper: Did you notice how the logger functions are invoked elsewhere in the code? Consider:

```
logger.error("Cannot bind with LDAP server: {}", e.getMessage());
```

There is no need to manually concatenate strings. The logger will substitute the `{}` within the message with the provided parameters, in order.

That said, this message would be very difficult to read anyway. Besides not matching the formatting of other messages, there is no separation between the message itself and the filter you intend to log. If the filter is "someFilter", then your message will look like:

```
ldap search failed with additional filtersomeFilter
```

rather than:

```
LDAP search failed with additional filter: someFilter
```

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
```diff
@@ -144,5 +144,8 @@ private LDAPGuacamoleProperties() {}
         public String getName() { return "ldap-encryption-method"; }
 
     };
-
+    public static final StringGuacamoleProperty ADDITIONAL_LDAP_SEARCH_FILTERS = new StringGuacamoleProperty(){
+    	  @Override
+          public String getName() { return "ldap-additional-search-filter"; }
```
mike-jumper: Again, please try to follow the style of the rest of the code. Consider this existing property from earlier in this file:

```
    /**
     * The base DN to search for Guacamole configurations.
     */
    public static final StringGuacamoleProperty LDAP_CONFIG_BASE_DN = new StringGuacamoleProperty() {

        @Override
        public String getName() { return "ldap-config-base-dn"; }

    };
```

There is a descriptive property name which is consistent between the Java code (`LDAP_CONFIG_BASE_DN`) and the property that would be used in guacamole.properties (`ldap-config-base-dn`), and the Java property is properly documented with a JavaDoc comment.

Compare to:

```
    public static final StringGuacamoleProperty ADDITIONAL_LDAP_SEARCH_FILTERS = new StringGuacamoleProperty(){
    	  @Override
          public String getName() { return "ldap-additional-search-filter"; }
    };
```

Note:

1. The JavaDoc comment is missing.
2. The Java naming (`ADDITIONAL_LDAP_SEARCH_FILTERS`) does not match that of the property (`ldap-additional-search-filter`).
3. The spacing surrounding the `getName()` implementation with respect to blank lines is inconsistent with the rest of the code in this file.

extensions/guacamole-auth-ldap/README
annieweng: how about ldap-user-base-dn-filter? 
 # The ldap filter which appended to the base DN and user identifier attribute,
 # produces the full DN of the user being authenticated.
 ldap-user-base-dn-filter: (memberOf=cn=exampleGroup,ou=groups,dc=example,dc=net)

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
annieweng: I have. 
only way will be if(confService.getAdditionalSearchFilter().length()>0), else .
however that will mean we have to repeat following code
                    // Return AuthenticatedUser if bind succeeds
    		    authenticatedUser.init(credentials);
in both if and else block. 

    		    return authenticatedUser;

mike-jumper: > only way will be ...

I don't think that *has* to be the only other possibility. There really must be a better way.

> however that will mean we have to repeat following code ...

If only there were a programming construct which allowed us to reuse pieces of code ... ;)

Anyway, it's clear the logic here needs some cleaning, but it's hard to read through given the other comments already made.

Perhaps we should revisit this after the stylistic issues are addressed?

extensions/guacamole-auth-ldap/README
mike-jumper: Doesn't this also apply to the general user search, though?

I'd say `ldap-user-filter` sounds about right given its purpose.

annieweng:
Hi guys
   thanks for the constructive suggestions. I believe I have update the baseline to match what you guys suggested, please let me know if there are any other action are required. 

thanks
Annie

jmuehlner:
Hey Annie, it looks like there are still some indentation issues - I still see a lot of tab-indented lines, lines that don't line up quite right, and lines that are indented overmuch. Can you make sure that all lines are indented by exactly 4 spaces per indentation level?

mike-jumper:
@annieweng, you might have to alter the settings in your IDE to get things to actually be indented in the same way as the rest of the code. I also recommend just looking at the diff within this PR, compared against the existing code. Look at the location of braces, the distance of each level of indent, etc. compared to what is already present. I'm pretty sure you'll agree that the style of the new code is starkly different.

The rule of thumb is to match the style of the code surrounding you. If you're having trouble determining that style, we do have a rough writeup of the style we try to follow:

http://guacamole.incubator.apache.org/guac-style/

This may not be *your* style, or even *our* style as individuals, but that's beside the point. Google's style guide sums this up pretty well (emphasis added):

> BE CONSISTENT.
> 
> If you’re editing code, take a few minutes to look at the code around you and determine its style. If they use spaces around all their arithmetic operators, you should too. If their comments have little boxes of hash marks around them, make your comments have little boxes of hash marks around them too.
> 
> **The point of having style guidelines is to have a common vocabulary of coding so people can concentrate on what you’re saying rather than on how you’re saying it.** We present global style rules here so people know the vocabulary, but local style is also important. **If code you add to a file looks drastically different from the existing code around it, it throws readers out of their rhythm when they go to read it. Avoid this.**

annieweng:
@mike-jumper I have update my editor to use 4 space and avoid tab, hopefully I got all the indentation and format inconsistency correctly this time.
thanks again for taking time review the code.

necouchman:
Ping.

@mike-jumper: Latest on this was Annie had updated formatting.  Anything else (aside from handling conflicts)?

@annieweng: Unfortunately this has lapsed long enough that the updates made to those files during this time require some conflict resolution.  Can you do this and update this PR?

annieweng:
Will do. I will try to do the update by the end of week.

thanks
Annie
On Jun 13, 2017, at 10:36 AM, necouchman <notifications@github.com<mailto:notifications@github.com>> wrote:


Ping.

@mike-jumper<https://github.com/mike-jumper>: Latest on this was Annie had updated formatting. Anything else (aside from handling conflicts)?

@annieweng<https://github.com/annieweng>: Unfortunately this has lapsed long enough that the updates made to those files during this time require some conflict resolution. Can you do this and update this PR?

annieweng:
looks like someone already merged the change, although the format is slightly different.  I will go ahead close the pull request. @necouchman @mike-jumper

PR #69, opened 2016-09-05
 GUACAMOLE-102 - Loadbalancing based on resource

wnovak:
Now there is ROUNDROBIN which is the old one (v0.9.10-master) and BALANCING which sorts the Connections by a connection parameter called "balancing". This parameter should be set by i.e. a remote service running on the target mashine and collecting CPU, Memory, I/O, Loggedin Users calculation.


mike-jumper:
I'll comment on the style as-is, as it's worth calling attention to a few things, but there are a few issues with this approach as a whole:

1. It's an abuse of connection parameters, which really should be *only* connection parameters (as defined by the protocol of the connection). For arbitrary connection-specific values, there are connection attributes.
2. If the intent is that connections defined within the database can be dynamically weighted for the sake of balancing, there is no need to define a whole new connection group type just to represent this. It can be built into the database auth and kept behind-the-scenes.

I recommend taking a look over how arbitrary data is stored for other non-parameter properties of connections within the database auth, and then rethinking this. For example, the [`max_connections_per_user`](https://github.com/apache/incubator-guacamole-client/blob/30639f2805f656784ff82400ec86b1abadf90e12/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/schema/001-create-schema.sql#L98) column is exposed internally via [`getMaxConnectionsPerUser()`](https://github.com/apache/incubator-guacamole-client/blob/30639f2805f656784ff82400ec86b1abadf90e12/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java#L238-L261) and externally via the [`max-connections-per-user`](https://github.com/apache/incubator-guacamole-client/blob/30639f2805f656784ff82400ec86b1abadf90e12/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java#L189-L190) attribute (with data defining the interface for this exposed via [`getConnectionAttributes()`](https://github.com/apache/incubator-guacamole-client/blob/30639f2805f656784ff82400ec86b1abadf90e12/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledUserContext.java#L177-L180) of the `UserContext`).

If we *really* need a dynamic weighting for connections within a balancing group, I suggest:

1. We shouldn't hack it into place, let alone as a connection parameter. It should be a legitimate thing (ie: has a column).
2. We should attempt to minimize the impact to the rest of the API, and avoid defining an entirely new connection group type, especially when this really still is "balancing".

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
```diff
@@ -714,4 +714,15 @@ public GuacamoleTunnel getGuacamoleTunnel(RemoteAuthenticatedUser user,
 
     }
 
+
+    public Integer getIdleConnections(Connection connection) {
```
mike-jumper: All functions must be documented with a proper JavaDoc comment. See the surrounding code for how this should be done.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
```diff
@@ -714,4 +714,15 @@ public GuacamoleTunnel getGuacamoleTunnel(RemoteAuthenticatedUser user,
...
+    	Integer idle = 999999;
```
mike-jumper: I expect this will not be done when the approach is revisited. That said, please beware:

1. As written, this is an unnecessary use of `Integer`. Why not `int`? Same for the return value of the function as a whole. The way things are being used here, `Integer` does nothing.
2. `999999` is a magic number - avoid these. It would make more sense for such a function to return `null` if the value is unset, rather than an arbitrarily large number. Perhaps that's why you initially went with `Integer`? Either way, it really should be done differently.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
```diff
@@ -714,4 +714,15 @@ public GuacamoleTunnel getGuacamoleTunnel(RemoteAuthenticatedUser user,
...
+        // Set parameters from associated data
+        Collection<ConnectionParameterModel> parameters = connectionParameterMapper.select(connection.getIdentifier());
+        for (ConnectionParameterModel parameter : parameters)
+        	if (parameter.getName().equalsIgnoreCase("balancing"))
```
mike-jumper: I mentioned this earlier, but I'll repeat here for the sake of context: this really is an abuse of connection parameters, which have a specific meaning.

It's also a little odd (or paranoid?) to use `equalsIgnoreCase()` here. If we *were* to define a magic parameter, better that it be strict. As written, this would be the only case where a parameter was interpreted in a case-insensitive way. Throughout the rest of the Guacamole stack, configuration options, parameters, attributes, properties, etc. are all case-sensitive and strictly interpreted.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/RestrictedGuacamoleTunnelService.java
```diff
@@ -183,6 +185,19 @@ protected ModeledConnection acquire(RemoteAuthenticatedUser user,
             @Override
             public int compare(ModeledConnection a, ModeledConnection b) {
 
+                Directory<ConnectionGroup> directory;
+                ConnectionGroup group;
+				try {
```
mike-jumper: Please do not indent with tabs - our code uses a 4-space indent absolutely everywhere. Deviating from this causes horrible readability issues like:

please-no-tabs-thanks

You may need to change your editor's settings. Even after that, it's worth double-checking the diff to make sure something crazy like the above hasn't happened.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/RestrictedGuacamoleTunnelService.java
```diff
@@ -183,6 +185,19 @@ protected ModeledConnection acquire(RemoteAuthenticatedUser user,
...
+	            	// Get connection group directory
+					directory = a.getCurrentUser().getAuthenticationProvider().getUserContext(a.getCurrentUser()).getConnectionGroupDirectory();
+	                // Get authorized connection group
+	                group = directory.get(a.getParentIdentifier());
+                if (group.getType() == ConnectionGroup.Type.BALANCING)
+                	return getIdleConnections(a)
+                			- getIdleConnections(b);
+				} catch (GuacamoleException e) {
+					e.printStackTrace();
```
mike-jumper: Production code should not contain `e.printStackTrace()`. If there is an error condition here, it should be logged in a human-readable way (and without a stack trace) at the appropriate level, and logged again at the debug level with the underlying exception (such that the logging framework will include the stack trace, but only at debug level).

If you take a peek at the other parts of Guacamole, you'll see that's what's been done elsewhere.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/RestrictedGuacamoleTunnelService.java
```diff
@@ -183,6 +185,19 @@ protected ModeledConnection acquire(RemoteAuthenticatedUser user,
...
 
+                Directory<ConnectionGroup> directory;
+                ConnectionGroup group;
+				try {
+	            	// Get connection group directory
+					directory = a.getCurrentUser().getAuthenticationProvider().getUserContext(a.getCurrentUser()).getConnectionGroupDirectory();
+	                // Get authorized connection group
+	                group = directory.get(a.getParentIdentifier());
+                if (group.getType() == ConnectionGroup.Type.BALANCING)
+                	return getIdleConnections(a)
```
mike-jumper: Magic changes in expected behavior really need to be documented. The comparator here is earlier defined as "Sort connections in ascending order of usage", which is being (intentionally) violated here. To avoid confusion, and for the sake of maintainability, that behavior needs to be clearly defined just as it previously was.

guacamole/src/main/webapp/app/rest/types/ConnectionGroup.js
```diff
@@ -129,6 +129,13 @@ angular.module('rest').factory('ConnectionGroup', [function defineConnectionGrou
         BALANCING : "BALANCING",
 
         /**
+         * The type string associated with balancing connection groups.
```
mike-jumper: This is the exact same message as used for the `BALANCING` type. This is probably a moot point, as I think adding a new `ROUNDROBIN` type should not be the solution here, but if it *were* this documentation would need to be updated to match what this type actually represents.

guacamole/src/main/webapp/translations/de.json
```diff
@@ -223,6 +223,7 @@
         "FIELD_HEADER_TYPE"     : "Typ:",
 
         "NAME_TYPE_BALANCING"       : "Balancing",
+        "NAME_TYPE_ROUNDROBIN"      : "RoundRobin",
```
mike-jumper: FYI - you do not need to define translation strings for languages other than English if the string does not differ from English. The translation system will pull the English strings as a fallback.

If you speak the language in question, and it just happens to be the same as English in this case, then that's a different matter, but otherwise - it's better to leave it blank. Leaving it blank allows others to see that the string has not been defined for that language yet, and to then provide a real translation.

mike-jumper:
Please also take a look at our [contribution guidelines](https://github.com/apache/incubator-guacamole-client/blob/master/CONTRIBUTING), specifically:

> 
> All changes to Guacamole must have corresponding issues in JIRA so the change can be properly tracked:
> 
> https://issues.apache.org/jira/browse/GUACAMOLE/
>
> If you do not already have an account on the Apache Software Foundation's JIRA, you will need to create one before creating your new issue.
>

and:

> Commits should be made in logical units and must reference the JIRA issue number:
>

For these changes (once ready) to be merged, there needs to be an associated issue in JIRA and related commit messages need to be tagged with that issue number. For examples of this, take a quick look through our git history.

In general, it's best that the JIRA issue be created before work even starts, such that the project and surrounding community can know that you're working on something and provide feedback (which is the primary reason for this), but better late than never. Tagging commits with the JIRA issue provides important context when reviewing changes historically, writing release notes, etc.

Just as you shouldn't have undocumented code, the overall changes themselves need documentation, too.

wnovak:
Hi, 
I took a thought about you suggestions and came up with the following results. 
1. Remove additional balancing group
2. Add 2 new connectiongroup definition parameter
2a. Checkbox for resource balancing ON/OFF
2b. Formula input for the resource weight calculation with parameters for CPU, MEM, IO and Loggedin Users, maybe more needed
3. If Checkbox is ON the compare sorts the groupmembers regarding the weight from a new column in the guacamole_connection table, if it'S OFF than the current balancing (guacamole active connections) is used.
4. Last but not least, create a servlet listener or remote java app service, running on the target servers which does the weight calculation with the formula parameter an put it into the guacamole_connection table as a the new column.
5. the max connections and max user connections should also stay in place with resource balancing to give it a overall limit.

PR #70, opened 2016-09-05
GUACAMOLE-58: fix for URLs with parameters

heinzbeinz:
See Ticket: https://issues.apache.org/jira/browse/GUACAMOLE-58

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Tunnel.js
+++ b/guacamole-common-js/src/main/webapp/modules/Tunnel.js
@@ -165,6 +165,12 @@ Guacamole.Tunnel.State = {
  *     will be made to the same domain.
  */
 Guacamole.HTTPTunnel = function(tunnelURL, crossDomain) {
+    function appendPart(url, part) {
+        if (!part) {
+            return url;
+        }
+        return url.indexOf('?') >= 0 ? url + "&" + part : url + "?" + part;
+    }
 
     /**
      * Reference to this HTTP tunnel.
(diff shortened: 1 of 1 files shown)


necouchman:
Ping.

This looks relatively straight-forward to me - no obvious issues here.

@heinzbeinz: Have you tested on multiple browsers to make sure this works as expected?  In particular IE...the availability of URL handling methods in IE lags behind the others.

@mike-jumper @jmuehlner: Any concerns here?  This is a pretty simply PR, and looks okay to me - only thing I wonder is if this is impacted at all by the changes Mike made to support the CAS authentication module - there was some work done on the URL rewriting stuff, and I'm not sure if that's related at all?

mike-jumper:
Just took a quick look over these changes - they look generally sane, but I have a couple concerns:

1. The `appendPart()` function is duplicated verbatim within the same file. It would be better if this function were defined only once. It could be defined privately as `Guacamole.Tunnel._appendPart()` namespace, for example, and would be excluded from the guacamole-common-js docs as long as it's properly annotated as `@private`.
2. Regardless of the repeating, all functions need to be documented. This new `appendPart()` function will need proper JSDoc.
3. The `HTTPTunnelServlet` class actually relies on the `read:`, `write:`, etc. strings being at the very beginning of the query string. It doesn't parse those values as parameters. The HTTP tunnel will not function with these changes unless the tunnel class itself is also modified to be compatible.

mike-jumper:
Regarding the incompatibility with the `GuacamoleHTTPTunnelServlet` implementation, the relevant pars are:

1. The handling of the "read:" and "write:" prefixes:

   https://github.com/apache/incubator-guacamole-client/blob/728d9b937c80bbf61ac79dd563dc1775203b34e6/guacamole-common/src/main/java/org/apache/guacamole/servlet/GuacamoleHTTPTunnelServlet.java#L238-L250

2. The handling of the "connect" string:

   https://github.com/apache/incubator-guacamole-client/blob/728d9b937c80bbf61ac79dd563dc1775203b34e6/guacamole-common/src/main/java/org/apache/guacamole/servlet/GuacamoleHTTPTunnelServlet.java#L209-L211

All of the above are currently handled through direct comparisons against `request.getQueryString()`. Since "read:" and "write:" are tested using `startsWith()`, any URL having parameters prior to those prefixes will fail there. The "connect" string is tested through a `equals()` comparison, and so will fail if any query parameters are present at all.

heinzbeinz:
As I recently switched companies, maybe my old colleague @sebastian-ohmsen would like to look at this. We only used the guacamole client without the java backend, so we didn't experience any errors on that side.

necouchman:
@heinzbeinz In that case, it might be best if you could close this PR.  The way the Apache-mirrored repos work, the only two ways to close a PR are for the original person who opened it (you) to close it manually, or for it to be merged into the ASF repo and the asfgit bot to close it.  If you close this, we can pick back up the SR and do a new PR for it - either your colleague or someone else.

PR #71, opened 2016-09-08
GUACAMOLE-104: Sanitize filenames before upload or download.

jmuehlner:
To make tomcat happy, we can just replace slashes with something else - in this case underscores.


mike-jumper:
_ :+1:

PR #76, opened 2016-11-11
GUACAMOLE-99: Support linking Docker container to guacd via GUACD_HOSTNAME and GUACD_PORT environment variables.

mike-jumper:


guacamole-docker/bin/start.sh
```diff
@@ -352,20 +352,42 @@ mkdir -p "$GUACAMOLE_LIB"
 # Point to associated guacd
 #
 
-# Verify required link is present
-if [ -z "$GUACD_PORT_4822_TCP_ADDR" -o -z "$GUACD_PORT_4822_TCP_PORT" ]; then
+# Use linked container for guacd if specified
+if [ -n "$GUACD_NAME" ]; then
```
jmuehlner: What is GUACD_NAME used for?

mike-jumper: Only for determining whether a Docker link is being used: https://docs.docker.com/compose/link-env-deprecated/

jmuehlner: Ah, ok.

PR #78, opened 2016-11-11
GUACAMOLE-70: Allow access to be restricted to strictly the users in the database.

mike-jumper:


extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/JDBCAuthenticationProviderService.java
```diff
@@ -73,8 +80,22 @@ public ModeledUserContext getUserContext(AuthenticationProvider authenticationPr
 
         // Retrieve user account for already-authenticated user
         ModeledUser user = userService.retrieveUser(authenticationProvider, authenticatedUser);
-        if (user == null)
-            return null;
+        if (user == null) {
```
jmuehlner: That's a different method, isn't it? org.apache.guacamole.auth.jdbc.AuthenticationProviderService makes no mention of this that I can see.

mike-jumper: It is true that the documentation of the JDBC auth's `AuthenticationProviderService` does not mention this. I can quickly update that.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/AuthenticationProviderService.java
```diff
@@ -72,7 +71,7 @@ public AuthenticatedUser authenticateUser(AuthenticationProvider authenticationP
      *
      * @return
      *     A new UserContext instance for the user identified by the given
-     *     credentials.
+     *     credentials, or null if no such user exists within the database.
```
jmuehlner: Should we maybe mention what the meaning of returning null is? The method itself says "allow access" and then returns null, but this comment doesn't say anything about returning null implying allowing access.

mike-jumper: If that part is confusing, I'd say we need to improve the wording of the comment which reads "allow access".

`null` does not mean "allow access"; it means the user that just authenticated has no context within scope of this `AuthenticationProvider`. We're only allowing access in the sense that we're not invalidating the auth result. Access to data within the scope of the `AuthenticationProvider` is denied, but no stance is taken as to the validity of the authentication attempt in the first place.

This is in contrast to the branch where `getUserContext()` throws an exception, in which case not only is access to the data denied, but we're revoking the positive auth result that would otherwise be valid.

jmuehlner: Ok, maybe updating the wording of the comment that says it allows access would be good then.

mike-jumper: OK - I reworded things to hopefully make more sense.

PR #85, opened 2016-11-29
GUACAMOLE-131: Migrate to Webjars for JavaScript library dependencies.

mike-jumper:
This change migrates to Webjars (http://www.webjars.org/) for the JavaScript library dependencies, and upgrades messageformat.js to 1.0.2, which is now MIT-licensed. Serving the static files provided by each Webjar requires using the `WebjarServlet`, in our case mapped by Guice. Lacking this, the static files are unreachable.

I needed to exclude two dependencies:

1. `angular`, which was being pulled in at version 1.3.0-beta9 due to `angular-module-shim`.
2. `messageformat`, which was being pulled in at version 0.1.7 due to `angular-translate-interpolation-messageformat`.


jmuehlner:
What do you mean by exclude dependencies? I see the dependencies declared in the POM.

mike-jumper:
There are `<exclusions>...</exclusions>` sections for two of the declared dependencies, preventing older versions of transitive dependencies from overriding the explicitly-declared versions we want.

jmuehlner:
Ah yes, there it is. I totally missed that. 😞

PR #89, opened 2016-12-03
GUACAMOLE-133: Document all dependencies of Guacamole within LICENSE.

mike-jumper:
Note that there are **two versions of `LICENSE`**:

1. `LICENSE` (in the top-level directory) which covers the source, and thus describes only the dependencies bundled with the source.
2. `guacamole/src/licenses/LICENSE` which is added to the Guacamole .war file during the build, and describes all dependencies which are bundled with the .war ... of which there are a fair few.

Within each of the above `LICENSE` files, *bundled* dependencies and their licenses are listed is in alphabetical order using the following rough format:

    Some Dependency (http://url.for.project/if/known/)
    --------------------------------------------------
    
        Version: A.B.C
        From: 'Carl P. Wrightholder' (http://url.for.author/if/known)
        License(s):
            LICENSE1 (guacamole/src/dependencies/some-dependency-A.B.C/License1.txt)
            LICENSE2 (guacamole/src/dependencies/some-dependency-A.B.C/License2.txt)

    Full text of license(s), if not too large to include.

I've kept the license filenames identical to the upstream source (if they included the license text at all). If the project instead included a URL to the license instead, I downloaded the license and used the filename from the URL. Some projects have multiple licenses. Depending on how the project represents their licenses, multiple licenses may be within in the same `LICENSE` / `LICENSE.txt` / `License.txt` / `COPYING` / `MITBSDCDDPLv1.3.27.text` file.

See also: http://www.apache.org/dev/licensing-howto.html

### DIFF
--- a/guacamole/src/licenses/dependencies/jsr250-api-1.0/CDDLv1.0.txt
+++ b/guacamole/src/licenses/dependencies/jsr250-api-1.0/CDDLv1.0.txt
@@ -1,0 +1,119 @@
+COMMON DEVELOPMENT AND DISTRIBUTION LICENSE (CDDL) Version 1.0
+
+1. Definitions.
+
+1.1. Contributor means each individual or entity that creates or contributes to the creation of Modifications.
+
+1.2. Contributor Version means the combination of the Original Software, prior Modifications used by a Contributor (if any), and the Modifications made by that particular Contributor.
+
+1.3. Covered Software means (a) the Original Software, or (b) Modifications, or (c) the combination of files containing Original Software with files containing Modifications, in each case including portions thereof.
+
+1.4. Executable means the Covered Software in any form other than Source Code.
+
+1.5. Initial Developer means the individual or entity that first makes Original Software available under this License.
+
+1.6. Larger Work means a work which combines Covered Software or portions thereof with code not governed by the terms of this License.
+
+1.7. License means this document.
+
+1.8. Licensable means having the right to grant, to the maximum extent possible, whether at the time of the initial grant or subsequently acquired, any and all of the rights conveyed herein.
+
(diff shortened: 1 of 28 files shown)


guacamole/src/licenses/dependencies/jsr250-api-1.0/CDDLv1.0.txt
```diff
@@ -0,0 +1,119 @@
...
+
+C. Any new file that is contributed or otherwise made available under the terms of this License.
+
+1.10. Original Software means the Source Code and Executable form of computer software code that is originally released under this License.
+
+1.11. Patent Claims means any patent claim(s), now owned or hereafter acquired, including without limitation, method, process, and apparatus claims, in any patent Licensable by grantor.
+
+1.12. Source Code means (a) the common form of computer software code in which modifications are made and (b) associated documentation included in or with such code.
+
+1.13. You (or Your) means an individual or a legal entity exercising rights under, and complying with all of the terms of, this License. For legal entities, You includes any entity which controls, is controlled by, or is under common control with You. For purposes of this definition, control means (a)�the power, direct or indirect, to cause the direction or management of such entity, whether by contract or otherwise, or (b)�ownership of more than fifty percent (50%) of the outstanding shares or beneficial ownership of such entity.
```
jmuehlner: I'm not sure what's with all the "�" characters in this diff, but they don't appear in the raw source, so I'm not worried about it.

mike-jumper: The document appears to be encoded in [ISO 8859-1](https://en.wikipedia.org/wiki/ISO/IEC_8859-1). Those characters are 0xA0, the ISO 8859-1 character for a non-breaking space.

PR #92, opened 2016-12-06
GUACAMOLE-136: Handle expired password reset within getUserContext()

mike-jumper:
The JDBC auth handles resetting of expired passwords within `authenticateUser()`, which doesn't play nicely with secondary authentication factors. The process ends up being:

1. User logs in
2. User is prompted to reset their password
3. User is verified with secondary authentication factors

Which is bad. Presumably the user's identity should be verified *before* allowing them to change their password. By moving the reset process to `getUserContext()`, the ordering of the process is fixed:

1. User logs in
2. User is verified with secondary authentication factors
3. User is prompted to reset their password


jmuehlner:
Makes sense to me.

PR #93, opened 2016-12-06
GUACAMOLE-136: Ensure field template is in DOM prior to invoking controller.

mike-jumper:
Lacking these changes, custom field type controllers cannot rely on their templates being present in the DOM, and things which would normally work for Angular directives fail to work with custom field types.


jmuehlner:
Ok, this makes sense to me.

PR #94, opened 2016-12-06
GUACAMOLE-136: Add support for two-factor authentication using Duo.

mike-jumper:
This change adds a new extension, guacamole-auth-duo, which implements two-factor authentication leveraging [Duo](https://duo.com/). They provide [Java and JavaScript libraries](https://github.com/duosecurity/duo_java/tree/56570d18a1f98642500e6c1158a3f41e448963ff) (albeit with some error-handling and lack-of-comments quirks) under a [BSD 3-clause license](https://github.com/duosecurity/duo_java/blob/56570d18a1f98642500e6c1158a3f41e448963ff/LICENSE).

Neither the Java nor the JavaScript libraries are present in Maven Central, and Duo has stated that they have no intent to do so (see: duosecurity/duo_java#1), thus they have been bundled in the source here. The top-level `LICENSE` has been updated accordingly.

With this extension in place, the overall process is as follows:

1. User logs in via any other installed authentication method.
2. If log in is successful, user is prompted to authenticate with Duo.
3. If Duo verifies the user's identity successfully, the user is allowed through. The previous authentication result is overridden otherwise.

The extension is configured with four new properties:

Name                  | Description
--------------------- | -----------
`duo-api-hostname`    | The API hostname provided by Duo (generally of the form: "api-XXXXXXXX.duosecurity.com").
`duo-integration-key` | The 20-character integration key ("ikey") provided by Duo.
`duo-secret-key`      | The 40-character secret key ("skey") provided by Duo.
`duo-application-key` | An arbitrary application key ("akey", must be *at least* 40 characters).

The "application key" is expected to be randomly generated by the administrator when configuring the extension (see [Duo's documentation on the subject](https://duo.com/docs/duoweb#1.-generate-an-akey)). All other values are provided by Duo.


LICENSE
```diff
@@ -217,5 +217,40 @@ Carlito (http://code.google.com/p/chromium/issues/detail?id=280557)
     Version: N/A
     From: 'tyPoland Lukasz Dziedzic' (http://www.lukaszdziedzic.eu/)
     License(s):
-        SIL Open Font (guacamole/src/dependencies/carlito/LICENSE)
+        SIL Open Font (guacamole/src/licenses/bundled/carlito/LICENSE)
```
jmuehlner: Cool

extensions/guacamole-auth-duo/src/main/java/org/apache/guacamole/auth/duo/AuthenticationProviderService.java
```diff
@@ -0,0 +1,109 @@
...
+import org.apache.guacamole.net.auth.AuthenticatedUser;
+import org.apache.guacamole.net.auth.Credentials;
+import org.apache.guacamole.net.auth.credentials.CredentialsInfo;
+import org.apache.guacamole.net.auth.credentials.GuacamoleInsufficientCredentialsException;
+
+/**
+ * Service providing convenience functions for the Duo AuthenticationProvider
+ * implementation.
+ */
+public class AuthenticationProviderService {
```
jmuehlner: AuthenticationProviderService is a very generic name that really means nothing to me. Could this be renamed to something more specific to what it does? Also, if I'm being pedantic, this doesn't actually provide multiple convenience methods, just one.

mike-jumper: OK. I'll rename it.

extensions/guacamole-auth-duo/src/main/resources/controllers/duoSignedResponseController.js
```diff
@@ -0,0 +1,78 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * Controller for the "GUAC_DUO_SIGNED_RESPONSE" field which uses the DuoWeb
+ * API to prompt the user for additional credentials, ultimately receiving a
+ * signed response from the Duo service.
+ */
+angular.module('guacDuo').controller('duoSignedResponseController', ['$scope',
```
mike-jumper: That would be ideal. We would need to manually inject our own `$element` within `formService`. I'm not sure if this is possible, but I'll give it a shot. If it works, I'll open another PR for that.

extensions/guacamole-auth-duo/src/main/java/org/apache/guacamole/auth/duo/DuoAuthenticationProvider.java
```diff
@@ -0,0 +1,100 @@
...
+    public AuthenticatedUser updateAuthenticatedUser(AuthenticatedUser authenticatedUser,
+            Credentials credentials) throws GuacamoleException {
+        return authenticatedUser;
+    }
+
+    @Override
+    public UserContext getUserContext(AuthenticatedUser authenticatedUser)
+            throws GuacamoleException {
+
+        UserVerificationService authProviderService =
```
mike-jumper: No... 😢

extensions/guacamole-auth-duo/src/main/resources/controllers/duoSignedResponseController.js
jmuehlner: Ok, sounds good.

jmuehlner:
Ok, everything looks good to me now.

extensions/guacamole-auth-duo/src/main/resources/controllers/duoSignedResponseController.js
mike-jumper: Awesome. It *is* possible (see PR #95).

mike-jumper:
OK. Rebased on top of latest changes, and added a commit to switch over to `$element`.

PR #96, opened 2016-12-10
GUACAMOLE-136: Remove DuoWeb Java API from codebase.

mike-jumper:
Given our own misgivings regarding the quality of DuoWeb's Java code, the numerous warnings from FindBugs regarding dodgy code and bad practice, and the external verifications we needed to add to prevent DuoWeb from dying with unchecked exceptions in the first place ... it *really* shouldn't be part of Guacamole.

This change removes the Java portion of DuoWeb from the codebase, replacing it with a re-implementation that I've written from scratch based on a high-level writeup of the request/response formats used by Duo:

1. The request/response itself is made up of two signed "cookies": one which was signed with the secret key ("skey") and another which was signed with the application key ("akey"). The cookies are separated with a colon.
2. Each signed cookie is in the format `PREFIX|DATA|SIGNATURE`, where `PREFIX` is one of "TX", for the cookie signed with the secret key within the request, "AUTH", for the cookie signed with the secret key in the response, or "APP", for the cookie signed with the application key. `DATA` is the base64-encoded data of the cookie itself (described below), and `SIGNATURE` is the lowercase hexadecimal representation of the HMAC-SHA1 signature of the UTF-8 bytes of `PREFIX|DATA`.
3. The cookie data is the string `USERNAME|IKEY|EXPIRES`, where `USERNAME` is the username of the user being validated, `IKEY` is the integration key provided by Duo ("ikey"), and `EXPIRES` is the timestamp after which the cookie should be considered invalid. The timestamp is standard UNIX epoch: the number of seconds elapsed since midnight of January 1, 1970.

The request is made up of the "TX" cookie followed by the "APP" cookie. The response is made up of "AUTH" followed by "APP". Assuming the request succeeds, the response is verified by checking:

1. Both signatures are correct.
2. Neither cookie is expired.
3. Both cookies contain the correct integration key.
4. The cookie usernames match the username of the user we are validating.

Only the JavaScript remains.


jmuehlner:
The code certainly looks better than before, and I don't see anything wrong with the implementation. Looks good! :shipit:

PR #97, opened 2016-12-12
Bypassing Tomcat landing page for Docker

efrecon:
When run within a Docker container, there is little point exhibiting the tomcat landing page. This slight modification to bash script that starts guacamole arranges for guacamole to be the root of the dockerized tomcat, while still being compatible with prior arrangements, i.e. still living under `/guacamole` as well.


mike-jumper:
I'm not a big fan of having multiple copies of the web application deployed under the same Tomcat, just so that it's available under `/` in addition to `/guacamole/`. I agree that backwards compatibility should be a priority here, but perhaps there is a better approach.

Have you considered using an environment variable to configure the webapp context path?

As for the PR itself, please take a look at our [contribution guidelines](https://github.com/apache/incubator-guacamole-server/blob/master/CONTRIBUTING). In particular, all changes to Guacamole really do need an associated JIRA issue so they can be properly tracked, and commits need to be tagged with that issue.

I would also recommend rebasing your changes such that you have a set of logical commits which embody your changes alone (rather than merge commits from merging upstream master back into your fork). In practice, working off a topic branch on your fork rather than your own master makes this easier.

efrecon:
Thanks for your input. You are right using an environment variable to control the context path is a much better idea since it would provide backwards compatibility (no env. variable set) while still provide the flexibility that was missing. Let me open a JIRA issue to continue the discussion over there.

necouchman:
Ping.

@efrecon: I believe you opened the JIRA issue for this, can you update this PR with that JIRA issue?  You'll also need to go back and edit your commits with the JIRA issue.

necouchman:
@efrecon: Ping...

necouchman:
@efrecon: Ping - please update this PR and your commits.

efrecon:
I never managed to finish this up properly, i.e. using environment variables.

necouchman:
@efrecon: If you're not planning on doing any further work, could you close the PR?

efrecon:
Closing it.

PR #98, opened 2016-12-18
GUACAMOLE-144: Do not list Category X licenses for dual-licensed dependencies

mike-jumper:
From @justinmclean regarding the [0.9.10-incubating (RC2) release vote](http://mail-archives.apache.org/mod_mbox/incubator-general/201612.mbox/%3CB6E93324-682C-4690-966B-EC0A4E39DDFE%40classsoftware.com%3E):

>
> ... some files (for instance [1][2]]) contain the text of the GPL license. GPL is Category X and is not allowed as a dependancy (unless it’s an optional component or build tools like autoconf in the sever artefact). I think this may be is OK as the software in question is dual license but if that is the case you should just include the text of the license you wish it to be licensed under no the text of both licenses. Can you please fix this in the next release.
>
> ...
>
> 1. ./guacamole/src/licenses/bundled/jaxb-impl-2.2.3-1/License.txt
> 2. ./guacamole/src/licenses/bundled/jersey-1.17.1/license.html
>

In addition to the above, there is also the EPL+LGPL dual-license of Logback (EPL is OK, but LGPL is Category X).

This change removes the GPL half of CDDL+GPL dual-licensed dependencies, and explicitly lists the license which Apache Guacamole is using for the dependencies in question. In the case of Logback, where we cannot remove the LGPL without altering the author's words, I have added the text "ECLIPSE PUBLIC LICENSE v1.0 CHOSEN".


guacamole/src/licenses/bundled/logback-1.1.2/LICENSE.txt
```diff
@@ -13,3 +13,5 @@ the Eclipse Foundation
 under the terms of the GNU Lesser General Public License version 2.1
 as published by the Free Software Foundation.
 
+** ECLIPSE PUBLIC LICENSE v1.0 CHOSEN. **
```
jmuehlner: Ok, I think that's obvious enough.

PR #99, opened 2017-01-04
GUACAMOLE-36: Support fully translatable messages

mike-jumper:
Extensions are already capable of throwing error messages which are translatable (the error message can be a translation key), but there is no way for substitution/pluralization to occur, and thus no way to include dynamic values within such messages.

This change defines a JSON format and mechanism for throwing translatable error messages from extensions, allowing for substitution of arbitrary values and pluralization within the JavaScript side of the web application.

### DIFF
--- a/guacamole-ext/src/main/java/org/apache/guacamole/language/Translatable.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/language/Translatable.java
@@ -1,0 +1,41 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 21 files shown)


guacamole-ext/src/main/java/org/apache/guacamole/language/Translatable.java
```diff
@@ -0,0 +1,41 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.guacamole.language;
+
+/**
+ * An interface which requires the definition of a translatable message that
```
jmuehlner: This comment is a little odd - could it be cleaned up a bit? It took me a few readthroughs to understand exactly what it's saying.

jmuehlner:
Ok, this all makes sense to me.

guacamole/src/main/webapp/app/rest/types/Error.js
```diff
@@ -43,6 +43,15 @@ angular.module('rest').factory('Error', [function defineError() {
         this.message = template.message;
```
jmuehlner: Where is this message field actually used? It looks like it's been replaced universally by translatableMessage.

mike-jumper: It's not used directly by the JavaScript anymore, but it continues to exist as a means of providing a human readable message alongside the machine-readable translation key. It can be helpful during debugging.

jmuehlner: Ah ok, that makes sense.

PR #100, opened 2017-01-04
GUACAMOLE-36: Add support for password policies

mike-jumper:
This change adds support to the database authentication backend for password policies defined within `guacamole.properties` which control the complexity of passwords allowed for database user accounts.

The restrictions allow for controlling:

1. Minimum password length
2. Required character classes (uppercase/lowercase, digits, symbols, etc.)
3. Whether the password can contain the username

The properties take the form:

```ini
# PostgreSQL properties
postgresql-user-password-min-length: 8
postgresql-user-password-prohibit-username: true
postgresql-user-password-require-multiple-case: true
postgresql-user-password-require-digit: true
postgresql-user-password-require-symbol: true
 
# MySQL properties
mysql-user-password-min-length: 8
mysql-user-password-prohibit-username: true
mysql-user-password-require-multiple-case: true
mysql-user-password-require-digit: true
mysql-user-password-require-symbol: true
```


jmuehlner:
Ok, I don't see any issues here. Everything looks alright to me.

PR #106, opened 2017-01-22
GUACAMOLE-163: Remove DEFAULT CURRENT_TIMESTAMP from password_date.

mike-jumper:
Specifying `DEFAULT CURRENT_TIMESTAMP` for a `DATETIME` column is not supported for MySQL 5.5 and older, breaking compatibility. This change removes `DEFAULT CURRENT_TIMESTAMP` entirely, as it's there purely for convenience. The database auth extension itself always specifies a non-`NULL` value when setting the password.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/schema/001-create-schema.sql
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/schema/001-create-schema.sql
@@ -85,7 +85,7 @@ CREATE TABLE `guacamole_user` (
   `username`      varchar(128) NOT NULL,
   `password_hash` binary(32)   NOT NULL,
   `password_salt` binary(32),
-  `password_date` datetime     NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `password_date` datetime     NOT NULL,
 
   -- Account disabled/expired status
   `disabled`      boolean      NOT NULL DEFAULT 0,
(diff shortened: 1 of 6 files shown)


patlachance:
New user creation process from WebUI doesn't work.

- environment setup using https://github.com/kevensen/openshift-kiosk/blob/master/openshift/guacamole-template.yaml
- container running: glyptodon/guacamole d38bea4
- mysql 5.7

Error from the guacamole_client:

> Error updating database.  Cause: java.sql.SQLException: Field 'password_date' doesn't have a default value
The error may involve org.apache.guacamole.auth.jdbc.user.UserMapper.insert-Inline
The error occurred while setting parameters
SQL: INSERT INTO guacamole_user (             username,             password_hash,             password_salt,             disabled,             expired,             access_window_start,             access_window_end,             valid_from,             valid_until,             timezone         )         VALUES (             ?,             ?,             ?,             ?,             ?,             ?,             ?,             ?,             ?,             ?         )
Cause: java.sql.SQLException: Field 'password_date' doesn't have a default value

- Workarounds:
  - change initdb.sql before initializing the MYSQL database or update the guacamole_user table on the running db instance with the following commands
  - ALTER TABLE guacamole_user MODIFY password_date DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP;

mike-jumper:
@patlachance - if you believe you've found a bug, the place to report it would be our JIRA: https://issues.apache.org/jira/browse/GUACAMOLE/

That said, it looks like you're simply mixing older releases of Guacamole with the new schema, which definitely won't work. Checking `UserMapper.xml` for MySQL, it definitely specifies that field in the `INSERT`:

https://github.com/apache/incubator-guacamole-client/blob/0.9.11-incubating/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/resources/org/apache/guacamole/auth/jdbc/user/UserMapper.xml#L146

If it's the downstream project you mentioned that set up this environment (installed a pre-0.9.11 Guacamole with a 0.9.11 schema), you should report this downstream.

PR #108, opened 2017-01-25
GUACAMOLE-177: Tie all layer property changes to display frame flush.

mike-jumper:


jmuehlner:
This looks good to me.

PR #109, opened 2017-01-25
GUACAMOLE-47: Get client hostname for use in guac RDP session

necouchman:
These changes allow for the client IP (and/or hostname) to be passed through as parameters to connections, similar to username and password.

I haven't updated any documentation, but it may be necessary given that some configuration changes might be necessary in Tomcat and any proxy servers in front of Tomcat in order to make this work properly.

### DIFF
--- a/guacamole-ext/src/main/java/org/apache/guacamole/token/StandardTokens.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/token/StandardTokens.java
@@ -45,12 +45,12 @@ public class StandardTokens {
     /**
      * The name of the client token added via addStandardTokens().
      */
-    public static final String CLIENT_HOST_TOKEN = "GUAC_REMHOST";
+    public static final String REMHOST_TOKEN = "GUAC_REMHOST";
 
     /**
      * The IP of the client token added via addStandardTokens().
      */
-    public static final String CLIENT_IP_TOKEN = "GUAC_REMIP";
+    public static final String REMIP_TOKEN = "GUAC_REMIP";
 
     /**
      * The name of the date token (server-local time) added via
@@ -129,8 +129,8 @@ public class StandardTokens {
         // Add client hostname and ip tokens
         HttpServletRequest request = credentials.getRequest();
         if (request != null) {
-            filter.setToken(CLIENT_HOST_TOKEN, request.getRemoteHost());
-            filter.setToken(CLIENT_IP_TOKEN, request.getRemoteAddr());
+            filter.setToken(REMHOST_TOKEN, request.getRemoteHost());
+            filter.setToken(REMIP_TOKEN, request.getRemoteAddr());
         }
 
         // Add any tokens which do not require credentials
--- a/guacamole-ext/src/main/java/org/apache/guacamole/token/StandardTokens.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/token/StandardTokens.java
@@ -42,6 +42,16 @@ public class StandardTokens {
     public static final String PASSWORD_TOKEN = "GUAC_PASSWORD";
 
     /**
+     * The name of the client token added via addStandardTokens().
+     */
+    public static final String CLIENT_HOST_TOKEN = "GUAC_CLIENT_HOST";
+
+    /**
+     * The IP of the client token added via addStandardTokens().
+     */
+    public static final String CLIENT_IP_TOKEN = "GUAC_CLIENT_IP";
+
+    /**
      * The name of the date token (server-local time) added via
      * addStandardTokens().
      */
@@ -114,6 +124,13 @@ public class StandardTokens {
         String password = credentials.getPassword();
         if (password != null)
             filter.setToken(PASSWORD_TOKEN, password);
+
+        // Add client hostname and ip tokens
+        HttpServletRequest request = credentials.getRequest();
+        if (request != null) {
+            filter.setToken(CLIENT_HOST_TOKEN, request.getRemoteHost());
+            filter.setToken(CLIENT_IP_TOKEN, request.getRemoteAddr());
+        }
 
         // Add any tokens which do not require credentials
         addStandardTokens(filter);
--- a/guacamole-ext/src/main/java/org/apache/guacamole/token/StandardTokens.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/token/StandardTokens.java
@@ -45,12 +45,12 @@ public class StandardTokens {
     /**
      * The name of the client token added via addStandardTokens().
      */
-    public static final String REMHOST_TOKEN = "GUAC_REMHOST";
+    public static final String REMHOST_TOKEN = "GUAC_CLIENT_HOSTNAME";
 
     /**
      * The IP of the client token added via addStandardTokens().
      */
-    public static final String REMIP_TOKEN = "GUAC_REMIP";
+    public static final String REMIP_TOKEN = "GUAC_CLIENT_ADDRESS";
 
     /**
      * The name of the date token (server-local time) added via
--- a/guacamole-ext/src/main/java/org/apache/guacamole/token/StandardTokens.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/token/StandardTokens.java
@@ -22,6 +22,7 @@ package org.apache.guacamole.token;
 import java.text.SimpleDateFormat;
 import java.util.Date;
 import org.apache.guacamole.net.auth.Credentials;
+import javax.servlet.http.HttpServletRequest;
 
 /**
  * Utility class which provides access to standardized token names, as well as
--- a/guacamole-ext/src/main/java/org/apache/guacamole/token/StandardTokens.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/token/StandardTokens.java
@@ -45,12 +45,12 @@ public class StandardTokens {
     /**
      * The name of the client token added via addStandardTokens().
      */
-    public static final String CLIENT_HOST_TOKEN = "GUAC_CLIENT_HOST";
+    public static final String CLIENT_HOST_TOKEN = "GUAC_REMHOST";
 
     /**
      * The IP of the client token added via addStandardTokens().
      */
-    public static final String CLIENT_IP_TOKEN = "GUAC_CLIENT_IP";
+    public static final String CLIENT_IP_TOKEN = "GUAC_REMIP";
 
     /**
      * The name of the date token (server-local time) added via
--- a/guacamole/src/main/java/org/apache/guacamole/rest/APIRequest.java
+++ b/guacamole/src/main/java/org/apache/guacamole/rest/APIRequest.java
@@ -72,7 +72,7 @@ public class APIRequest extends HttpServletRequestWrapper {
         if (request.getHeader("X-Guacamole-Client-Hostname") != null && !request.getHeader("X-Guacamole-Client-Hostname").isEmpty())
             this.remoteHost = request.getHeader("X-Guacamole-Client-Hostname");
         else if (request.getHeader("X-Forwarded-For") != null && !request.getHeader("X-Forwarded-For").isEmpty())
-            this.remoteHost = request.getHeader("X-Forwarded-For");
+            this.remoteHost = null;
         else if (request.getRemoteHost() != null && !request.getRemoteHost().isEmpty())
             this.remoteHost = request.getRemoteHost();
         else
--- a/guacamole/src/main/java/org/apache/guacamole/rest/APIRequest.java
+++ b/guacamole/src/main/java/org/apache/guacamole/rest/APIRequest.java
@@ -42,12 +42,12 @@ public class APIRequest extends HttpServletRequestWrapper {
     private final Map<String, String[]> parameters;
 
     /**
-     * The remote hostname that initiated the request.
+     * The hostname of the client that initiated the request.
      */
     private final String remoteHost;
 
     /**
-     * The remote ip address that initiated the request.
+     * The ip address of the client that initiated the request.
      */
     private final String remoteAddr;
 
@@ -68,10 +68,10 @@ public class APIRequest extends HttpServletRequestWrapper {
 
         super(request);
 
-        // Grab the remote host info.
+        // Grab the remote host info
         this.remoteHost = request.getRemoteHost();
 
-	// Grab the remote ip info.
+        // Grab the remote ip info
         this.remoteAddr = request.getRemoteAddr();
 
         // Copy parameters from given MultivaluedMap 
--- a/guacamole/src/main/java/org/apache/guacamole/rest/APIRequest.java
+++ b/guacamole/src/main/java/org/apache/guacamole/rest/APIRequest.java
@@ -69,16 +69,10 @@ public class APIRequest extends HttpServletRequestWrapper {
         super(request);
 
         // Grab the remote host info.
-        if (request.getRemoteHost() != null && !request.getRemoteHost().isEmpty())
-            this.remoteHost = request.getRemoteHost();
-        else
-            this.remoteHost = null;
+        this.remoteHost = request.getRemoteHost();
 
 	// Grab the remote ip info.
-        if(request.getRemoteHost() != null && !request.getRemoteAddr().isEmpty())
-            this.remoteAddr = request.getRemoteAddr();
-        else
-            this.remoteAddr = null;
+        this.remoteAddr = request.getRemoteAddr();
 
         // Copy parameters from given MultivaluedMap 
         this.parameters = new HashMap<String, String[]>(parameters.size());
--- a/guacamole/src/main/java/org/apache/guacamole/rest/APIRequest.java
+++ b/guacamole/src/main/java/org/apache/guacamole/rest/APIRequest.java
@@ -42,6 +42,16 @@ public class APIRequest extends HttpServletRequestWrapper {
     private final Map<String, String[]> parameters;
 
     /**
+     * The remote hostname that initiated the request.
+     */
+    private final String remoteHost;
+
+    /**
+     * The remote ip address that initiated the request.
+     */
+    private final String remoteAddr;
+
+    /**
      * Wraps the given HttpServletRequest, using the given MultivaluedMap to
      * provide all request parameters. All HttpServletRequest functions which
      * do not deal with parameter names and values are delegated to the wrapped
@@ -57,6 +67,29 @@ public class APIRequest extends HttpServletRequestWrapper {
             MultivaluedMap<String, String> parameters) {
 
         super(request);
+
+        // Try a few methods to get client info.
+        String clientHostname = "";
+        String clientAddress = "";
+        if(request.getHeader("X-Guacamole-Client-Hostname") != "") {
+            this.remoteHost = request.getHeader("X-Guacamole-Client-Hostname");
+        } else if(request.getHeader("X-Forwarded-For") != "") {
+            this.remoteHost = request.getHeader("X-Forwarded-For");
+        } else if(request.getRemoteHost() != "") {
+            this.remoteHost = request.getRemoteHost();
+        } else {
+            this.remoteHost = "";
+        }
+
+        if(request.getHeader("X-Guacamole-Client-IP") != "") {
+            this.remoteAddr = request.getHeader("X-Guacamole-Client-IP");
+        } else if(request.getHeader("X-Forwarded-For") != "") {
+            this.remoteAddr = request.getHeader("X-Forwarded-For");
+        } else if(request.getRemoteAddr() != "") {
+            this.remoteAddr = request.getRemoteAddr();
+        } else {
+            this.remoteAddr = "";
+        }
 
         // Copy parameters from given MultivaluedMap 
         this.parameters = new HashMap<String, String[]>(parameters.size());
@@ -99,6 +132,16 @@ public class APIRequest extends HttpServletRequestWrapper {
         // Otherwise, return first value
         return values[0];
 
+    }
+
+    @Override
+    public String getRemoteHost() {
+        return this.remoteHost;
+    }
+
+    @Override
+    public String getRemoteAddr() {
+        return this.remoteAddr;
     }
 
 }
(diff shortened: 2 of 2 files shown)


guacamole/src/main/java/org/apache/guacamole/rest/APIRequest.java
```diff
@@ -58,6 +68,27 @@ public APIRequest(HttpServletRequest request,
 
         super(request);
 
+        // Try a few methods to get client info.
+        if(request.getHeader("X-Guacamole-Client-Hostname") != null && request.getHeader("X-Guacamole-Client-Hostname") != "") {
```
mike-jumper: Why is a custom header being checked?

guacamole/src/main/java/org/apache/guacamole/rest/APIRequest.java
```diff
@@ -58,6 +68,27 @@ public APIRequest(HttpServletRequest request,
...
+            this.remoteHost = request.getHeader("X-Guacamole-Client-Hostname");
+        } else if(request.getHeader("X-Forwarded-For") != null && request.getHeader("X-Forwarded-For") != "") {
```
mike-jumper: You can't safely do something like `someString != ""` - this is a comparison which checks the value of the reference, not the value of the string itself. What you're looking for here is `!someString.isEmpty()`.

guacamole/src/main/java/org/apache/guacamole/rest/APIRequest.java
necouchman: Depending on how Guacamole + Tomcat is configured, namely behind a proxy, it can be difficult to get the original client information through.  By default, the remote host/IP info when Tomcat is behind an Apache httpd or nginx server is whatever the hostname and IP of the proxy is, rather than the client.  This is not the intended behavior of the module, since you want the hostname and/or IP of the actual client, not the proxy server.  I implemented it this way in order to give some options - you can set up custom headers that pass through, you can do some Apache+Tomcat configuration, or you can rely on the standard remote host/addr fields.  See comments on the JIRA issue: https://issues.apache.org/jira/browse/GUACAMOLE-47

guacamole/src/main/java/org/apache/guacamole/rest/APIRequest.java
necouchman: Got it, will fix this.

guacamole/src/main/java/org/apache/guacamole/rest/APIRequest.java
```diff
@@ -58,6 +68,27 @@ public APIRequest(HttpServletRequest request,
...
+        } else if(request.getRemoteHost() != null && request.getRemoteHost() != "") {
+            this.remoteHost = request.getRemoteHost();
+        } else {
+            this.remoteHost = "";
+        }
+
+        if(request.getHeader("X-Guacamole-Client-IP") != null && request.getHeader("X-Guacamole-Client-IP") != "") {
+            this.remoteAddr = request.getHeader("X-Guacamole-Client-IP");
+        } else if(request.getHeader("X-Forwarded-For") != null && request.getHeader("X-Forwarded-For") != "") {
+            this.remoteAddr = request.getHeader("X-Forwarded-For");
```
mike-jumper: There is no guarantee that `X-Forwarded-For` is an IP or a hostname. As written, the remote host and remote address tokens would both be populated with the same value, regardless of whether that value is actually what the token represents.

guacamole/src/main/java/org/apache/guacamole/rest/APIRequest.java
```diff
@@ -58,6 +68,27 @@ public APIRequest(HttpServletRequest request,
...
 
+        // Try a few methods to get client info.
+        if(request.getHeader("X-Guacamole-Client-Hostname") != null && request.getHeader("X-Guacamole-Client-Hostname") != "") {
+            this.remoteHost = request.getHeader("X-Guacamole-Client-Hostname");
+        } else if(request.getHeader("X-Forwarded-For") != null && request.getHeader("X-Forwarded-For") != "") {
+            this.remoteHost = request.getHeader("X-Forwarded-For");
+        } else if(request.getRemoteHost() != null && request.getRemoteHost() != "") {
+            this.remoteHost = request.getRemoteHost();
+        } else {
+            this.remoteHost = "";
```
mike-jumper: Why initialize to blank? Is there a reason you don't want to use `null` to represent the lack of a value?

necouchman: Yeah, that occurred to me as I was reviewing the other changes...will remove that.

guacamole/src/main/java/org/apache/guacamole/rest/APIRequest.java
```diff
@@ -58,6 +68,27 @@ public APIRequest(HttpServletRequest request,
...
+        }
+
+        if(request.getHeader("X-Guacamole-Client-IP") != null && request.getHeader("X-Guacamole-Client-IP") != "") {
```
mike-jumper: Please follow the same style as the other code in guacamole-client for the sake of readability. There should be a space between the `if` and the `(`, as `if` is not a function, and we try to avoid:

* Using braces when unnecessary (unless it hurts readability)
* Cuddling the brace with the `else` or `else if`

If unclear, we also have [general style guidelines](http://guacamole.incubator.apache.org/guac-style/), but the best way is usually to simply take a quick look around and make sure your code doesn't stand out.

guacamole/src/main/java/org/apache/guacamole/rest/APIRequest.java
necouchman: Yes, that's correct...I was going under the assumption that if you're relying on X-Forwarded-For (which is usually an IP address) you'd want that in both tokens.  I can remove the check from the hostname one so that the X-Forwarded-For only applies to the IP, or I could, instead, check to see if X-Forwarded-For is an IP-type value and assign it based on whether or not an IP is present??

guacamole/src/main/java/org/apache/guacamole/rest/APIRequest.java
necouchman: Removed the unnecessary braces and fixed if ( spacing in APIRequest.java.

guacamole-ext/src/main/java/org/apache/guacamole/token/StandardTokens.java
```diff
@@ -42,6 +43,16 @@
     public static final String PASSWORD_TOKEN = "GUAC_PASSWORD";
 
     /**
+     * The name of the client token added via addStandardTokens().
+     */
+    public static final String REMHOST_TOKEN = "GUAC_REMHOST";
```
mike-jumper: I'm not the biggest fan of "GUAC_REMHOST" and "GUAC_REMIP", given the verbosity of other tokens ("GUAC_USERNAME" and "GUAC_PASSWORD"), the ambiguity of "remote" in a multi-layered stack, and the specificity of "IP" compared to the "address" property it's pulled from.

What would you think of "GUAC_CLIENT_HOSTNAME" and "GUAC_CLIENT_ADDRESS"?

necouchman: I'm fine with that...I think the reason I settled on the shorter ones is I actually ran into an issue with a limit on the length of the tokens??  Can't remember exactly off the top of my head, but I think even GUAC_HOSTNAME and GUAC_ADDRESS were too long, but never fully tracked it down.  I'll try, again.

mike-jumper: Hm ... There should definitely be no limit.

necouchman: Yeah, had no problem with it this time around, so probably is was another issue that I confused for the length of the token.  Changed to your suggested values.

mike-jumper:
So, why the special guac-specific HTTP headers?

And why separate "hostname" and "address" tokens which, taking from `X-Forwarded-For` may end up with the same value, regardless of whether that value is a hostname or address?

necouchman:
So, to answer the question about guac-specific HTTP headers...this all depends on whether or not you're behind a proxy.  Here are the possible scenarios:
1) Tomcat is not behind a proxy, and the request.getRemoteHost() method returns the client hostname and the request.getRemoteAddr() method returns the client IP address.  This is the simplest scenario.
2) Tomcat is behind a proxy, and you've configured the X-Forwarded-For header on both the proxy (Nginx, Apache HTTPD) and the valve in Tomcat.  This results in the X-Forwarded-For header being set to the IP address of the original client rather than grabbing the IP and hostname of the proxy.  Setting both the hostname and address tokens to the IP here may not be the best idea - maybe in these scenario you want the hostname token to be null??
3) Tomcat is behind a proxy, and you either don't want to configure X-Forwarded-For, or you want to make sure you have both the hostname and the IP address of the client instead of just the IP (as available through X-Forwarded-For).  So, you configure the custom headers on the proxy side to pass through the original REMOTE_HOST and REMOTE_ADDR headers so that you can pick that information up on the other side of the proxy.

I evaluate these in reverse order because I assume that, with the custom headers you've done very specific configuration because you really want that information, and that should take precedence over more generic configurations, the simplest being that you're just running Tomcat by itself and have access to the headers.

Hopefully this makes sense??  I can drop the custom header configuration if that's the consensus - the side effect of this will essentially be that, if you're running Tomcat behind a proxy (as I am), you'll only ever have access to the client IP address and not the hostname.

mike-jumper:
> Tomcat is behind a proxy, and you either don't want to configure X-Forwarded-For, or you want to make sure you have both the hostname and the IP address of the client instead of just the IP (as available through X-Forwarded-For). So, you configure the custom headers on the proxy side to pass through the original REMOTE_HOST and REMOTE_ADDR headers so that you can pick that information up on the other side of the proxy.

Hm ... why would you not want to configure `X-Forwarded-For` on the proxy, but *want* to configure custom guac headers?

I'm also concerned about the potential for spoofing these headers. It's not a huge deal for something like the logging/reporting we're already doing, since [we perform verification that the string is indeed an IP address](https://github.com/apache/incubator-guacamole-client/blob/057a0922637c3f92c761cf26b1a6cb3d9eec8624/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/RemoteAuthenticatedUser.java#L87-L102)), but allowing the insertion of essentially arbitrary strings from any of three headers into the connection parameters of a connection would be a potential security vulnerability for any connection that happens to use those tokens. Any deployment of Guacamole would then be required to deploy a proxy to sanitize those headers from the original request, ensuring they come only from a trusted source.

There must be some way to lock things down such that it is *guaranteed* that these tokens contain what they are expected to contain (an IP address vs. a name), or reducing scope to simply IP if that is not possible, ideally using nothing but standard / de facto standard headers.

In fact ... given the issue above with security ... perhaps the best way forward would be to use the `getRemoteAddr()` and `getRemoteHost()` functions only, and not trust the headers at all?

necouchman:
> Hm ... why would you not want to configure X-Forwarded-For on the proxy, but want to configure custom guac headers?

Fair point.  I was trying to keep things consistent - that if you have a proxy, and you're going to configure a header to pass through the original client hostname, then you might as well just configure two of them and avoid using X-Forwarded-For altogether.  Also, using X-Forwarded-For requires additional Tomcat configuration to make it work, so it was sort of an end-around to avoid the Tomcat config changes and keep configuring it simpler.  Given the security issues below, though, maybe the Tomcat config changes aren't all bad...

> I'm also concerned about the potential for spoofing these headers. It's not a huge deal for something like the logging/reporting we're already doing, since we perform verification that the string is indeed an IP address), but allowing the insertion of essentially arbitrary strings from any of three headers into the connection parameters of a connection would be a potential security vulnerability for any connection that happens to use those tokens. Any deployment of Guacamole would then be required to deploy a proxy to sanitize those headers from the original request, ensuring they come only from a trusted source.

> There must be some way to lock things down such that it is guaranteed that these tokens contain what they are expected to contain (an IP address vs. a name), or reducing scope to simply IP if that is not possible, ideally using nothing but standard / de facto standard headers.

Yeah, I can definitely see the risk here - I had not thought through the header spoofing issue.  I can do some research to see about how to lock stuff down to certain values; however, since a hostname is going to be an unknown string, that one is probably a little hard to do.  I'll see what I can find on this.

For the X-Forwarded-For header, I think this is probably safe to use, since 1) it requires additional Tomcat configuration, and 2) part of that configuration involves telling Tomcat what hosts are valid proxies and are allowed to define the X-Forwarded-For header.  This configuration, along with what you'd be required to do on the Proxy side, seems like it *should* be safe to use within Guacamole without having to implement additional security around it.

For the custom headers, I'll do some research, but another possibility here would be to create configuration parameters in guacamole.properties that would allow a user to enable & define custom headers for this.  Maybe this is overkill, though, and I'm just trying too desperately to salvage the ability to get the real client hostname from behind a proxy :-).  I can yank the custom header code if it's going to be too risky and too complex.

> In fact ... given the issue above with security ... perhaps the best way forward would be to use the getRemoteAddr() and getRemoteHost() functions only, and not trust the headers at all?

Given the aforementioned requirements for using X-Forwarded-For, I think it should be safe to use.  The risk of ditching that and going with getRemoteAddr()/getRemoteHost() only is making this functionality useless for folks running Tomcat behind a reverse proxy.  As one of the guys who routinely runs Tomcat+Guacamole behind Apache httpd (for a variety of reasons), I'd rather not limit the functionality, but I also don't want to introduce code that's going to create vulnerabilities for the project.

necouchman:
Okay, here's a thought for security checks for the custom headers.  It's actually very similar to what Tomcat does with X-Forwarded-For:
- Create guacamole.properties option to enable custom headers.  Default to off.
- Create guacamole.properties option that sets the expected reverse proxy hosts.  Default to 127.0.0.1.
- In the code that checks for the custom headers, also check getRemoteAddr and make sure it is one of the values as configured in guacamole.properties.

Seems like that would at least lock things down for folks who aren't worrying at all about the custom headers and proxy configuration.  It doesn't really implement any checks to make sure that the values in those headers are the right type of values, not malicious, etc., but that's more a proxy-side configuration item.

mike-jumper:
> I can yank the custom header code if it's going to be too risky and too complex.

Yeah, I think the complexity of this change is expanding beyond what it needs to be, especially if `getRemoteAddr()` and `getRemoteHost()` will be sufficient for most cases while also being reliably secure. There's no reason things can't be expanded later if necessary.

Adding a bunch of supported HTTP headers, including custom headers, as well as new `guacamole.properties options ... is a bit much.

necouchman:
Fair enough.  I've removed the custom header code leaving only the XFF and getRemoteHost()/getRemoteAddr() code.  Note that I've also changed it such that, if the XFF header is used, since it only ever returns an IP address, the remote host token is set to null and only the remote address header is set.

mike-jumper:
Using XFF without validation is still a potential security vulnerability. Consider:

1. A user sets up Guacamole without a proxy (which is fine).
2. The administrator, in a reasonable attempt to organize things, sets of an RDP connection using `/var/lib/guacamole/rdp-drives/${GUAC_REMOTE_ADDR}` for [`drive-path`](http://guacamole.incubator.apache.org/doc/0.9.10-incubating/gug/configuring-guacamole.html#rdp-device-redirection).
3. A malicious user, suspecting this to be the case, manually sets his own `X-Forwarded-For` header to something like `../../../../../../../../`. Upon connecting, that user now has access to the entire filesystem of the Guacamole server (possibly with write privileges, depending on how guacd is run).

Unless proper validation is in place such that this header *cannot* contain anything but an IP address, it's too dangerous.

Please also remember that, just like with your previous pull request #110, commits need to be tagged with the JIRA issue, there shouldn't be merge commits here, and messages should be descriptive. Recent commits look OK except for the JIRA issue aspect, but there is still the old "Some tweaks to get the tokens working correctly." commit, as well as the merge from master. A rebase seems in order.

necouchman:
> Unless proper validation is in place such that this header cannot contain anything but an IP address, it's too dangerous.

The mitigating factor here should be the fact that you must configure Tomcat to pass through XFF in the first place, and that Tomcat does the validation based on the configuration you are required to supply.  The documentation is here: https://tomcat.apache.org/tomcat-7.0-doc/api/org/apache/catalina/valves/RemoteIpValve.html.  If I understand the documentation and other things I've read correctly, that valve has to be configured in order for the X-Forwarded-For header to be passed through at all, and Tomcat will validate that the header matches one of the internalProxies values specified there.  I can do some testing, but I'm pretty sure if you don't explicitly configure that in Tomcat, the header won't be passed through, which should mitigate that risk.

> Please also remember that, just like with your previous pull request #110, commits need to be tagged with the JIRA issue, there shouldn't be merge commits here, and messages should be descriptive. Recent commits look OK except for the JIRA issue aspect, but there is still the old "Some tweaks to get the tokens working correctly." commit, as well as the merge from master. A rebase seems in order.

Okay, I'll look at either amending the commits or rebasing it.  I had some trouble doing the rebase the last time around, which is why I ended up just closing the request and re-opening it with the clean branch.  I'll see what I can do to fix this.

necouchman:
>  If I understand the documentation and other things I've read correctly, that valve has to be configured in order for the X-Forwarded-For header to be passed through at all, and Tomcat will validate that the header matches one of the internalProxies values specified there.

Upon further reading, looks like I may have been mistaken on this.  I'll pull out the XFF code for now and just use getRemoteAddr/getRemoteHost.  Maybe I'll spend some more time in the future trying to figure out a secure way to do this, but I won't let it hold this up for now.

mike-jumper:
> The mitigating factor here should be the fact that you must configure Tomcat to pass through XFF in the first place

You do not need to configure Tomcat for it to pass `X-Forwarded-For` through. The configuration you refer to causes Tomcat to parse `X-Forwarded-For` and update the value returned by `getRemotAddr()`, but the original header is always accessible.

For example, I just tested on a deployment under Tomcat 8.5.8 with all default settings. Using a Chrome extension to manually set the `X-Forwarded-For` header:

extension-set-x-forwarded-for

I set a debug watch in Netbeans to see what value was returned:

debug-watch-header

The value is passed through as-is. In my case, because I do have a proxy in front of Guacamole, the address of the proxy itself is appended to it (part of the definition of `X-Forwarded-For`). For this to work as you intend (and without hurting security) you would need to both parse and validate the contents of the `X-Forwarded-For` header (see https://github.com/apache/incubator-guacamole-client/blob/057a0922637c3f92c761cf26b1a6cb3d9eec8624/guacamole/src/main/java/org/apache/guacamole/rest/auth/AuthenticationService.java#L80-L124).

Of course, even if Tomcat did handle this for you, can you assume the same about Jetty? GlassFish? Only if the Java servlet standard itself requires servlet containers to do such sanitization would this be a safe assumption.

mike-jumper:
Ah - never mind. I see you came to the same conclusion.

guacamole/src/main/java/org/apache/guacamole/rest/APIRequest.java
```diff
@@ -58,6 +68,18 @@ public APIRequest(HttpServletRequest request,
 
         super(request);
 
+        // Grab the remote host info.
+        if (request.getRemoteHost() != null && !request.getRemoteHost().isEmpty())
```
mike-jumper: Is it necessary to sanity check `getRemoteHost()` and `getRemoteAddr()` against `isEmpty()`, given that their return values are guaranteed by the standard? Assuming that the empty string, if it ever does occur, is a legitimate value, doesn't this boil down to:

    if (request.getRemoteHost() != null)
        this.remoteHost = request.getRemoteHost();
    else
        this.remoteHost = null;

and thus could simply be:

    this.remoteAddr = request.getRemoteAddr();

necouchman: Removed the unnecessary checks and just assigned both remote host and remote addr.

necouchman:
Okay, I've rebased and amended the commit messages with the JIRA prefix.  Let me know if there's anything else that needs cleaning up, code-wise or git-wise.

mike-jumper:
Looks much much better, but the merge commit and the "some tweaks" commit are still there. Really should rebase away the merge commit, and either reword the "some tweaks" commit or squash it into something relevant.

Beyond that, I only see one other style issue in the code. I'll comment on that momentarily.

guacamole/src/main/java/org/apache/guacamole/rest/APIRequest.java
```diff
@@ -58,6 +68,12 @@ public APIRequest(HttpServletRequest request,
 
         super(request);
 
+        // Grab the remote host info.
+        this.remoteHost = request.getRemoteHost();
+
+	// Grab the remote ip info.
```
mike-jumper: This poor comment is misaligned.

Also, to match the style of the rest of the guac codebase, one-line non-block comments should not end with a period (see the style of nearby lines). Same goes for the comment just above.

guacamole/src/main/java/org/apache/guacamole/rest/APIRequest.java
```diff
@@ -42,6 +42,16 @@
     private final Map<String, String[]> parameters;
 
     /**
+     * The remote hostname that initiated the request.
+     */
+    private final String remoteHost;
+
+    /**
+     * The remote ip address that initiated the request.
```
mike-jumper: Perhaps a bit of a semantic nitpick, but this is not the IP address that initiated the request, because IP addresses can't do things. It's the IP address of the client that initiated the request. Same thing for the hostname above.

guacamole/src/main/java/org/apache/guacamole/rest/APIRequest.java
necouchman: Fixed, I think - I had used a tab instead of spaces :-/.

guacamole/src/main/java/org/apache/guacamole/rest/APIRequest.java
necouchman: Tweaked the language.

guacamole/src/main/java/org/apache/guacamole/rest/APIRequest.java
mike-jumper: Ugh ... you monster.

I'll double-check that no further tabs exist in these changes.

mike-jumper:
Looks pretty decent now. I'll go ahead and merge, assuming the tabs are indeed gone.

PR #110, opened 2017-01-25
GUACAMOLE-29: Add Support for HTTP Authentication

necouchman:
Add a module to support HTTP-based authentication via REMOTE_USER (or custom-defined) header.


mike-jumper:
Can you clean up the commits here?

The commits for GUACAMOLE-29 should be tagged as such, but I only see the first commit tagged. There are also several commits from GUACAMOLE-47 which, though reverted, are still present along with the commits that revert them, as well as merge commits. All this can be fixed with a rebase.

Please also be sure to keep your messages descriptive (at a high level) of what you're accomplishing with that commit. "One more minor tweak" doesn't really provide any information to someone perusing the logs or checking `git blame`.

necouchman:
Closing...will re-open with new request once I fix the commit history.

PR #111, opened 2017-01-25
GUACAMOLE-195: Add module to support HTTP header-based authentication.

necouchman:
New pull request to merge in the new guacamole-auth-http module.


mike-jumper:
So, the idea behind this extension is to trust the content of the `REMOTE_USER` header, that value having been set by some other authentication service sitting in front of Guacamole?

necouchman:
Correct - the assumption being that, if you install this module you'll be putting something in front of it - either configuring Tomcat directly, or configuring your reverse proxy with settings that force authentication for the web location where the Guacamole client is running.  Obviously there's a substantial risk if you enable this module without configuring anything to protect the web location, but that's true for any application that uses HTTP BASIC authentication.

extensions/guacamole-auth-http/src/main/assembly/dist.xml
```diff
@@ -0,0 +1,53 @@
...
+    <!-- Output tar.gz -->
+    <formats>
+        <format>tar.gz</format>
+    </formats>
+
+    <!-- Include licenses and extension .jar -->
+    <fileSets>
+
+        <!-- Include licenses -->
+        <!--<fileSet>
```
mike-jumper: Why the commented-out section?

extensions/guacamole-auth-http/src/main/java/org/apache/guacamole/auth/http/AuthenticationProviderService.java
```diff
@@ -0,0 +1,93 @@
+/*
+ * Copyright (C) 2015 Glyptodon LLC
```
mike-jumper: If copying the `AuthenticationProviderService` from something else, should probably copy the Apache-licensed one.

extensions/guacamole-auth-http/src/main/java/org/apache/guacamole/auth/http/HTTPAuthenticationProvider.java
```diff
@@ -0,0 +1,110 @@
+/*
+ * Copyright (C) 2015 Glyptodon LLC
```
mike-jumper: Same thing here - the license boilerplate for the guac codebase is different now.

extensions/guacamole-auth-http/src/main/java/org/apache/guacamole/auth/http/HTTPAuthenticationProviderModule.java
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright (C) 2015 Glyptodon LLC
```
mike-jumper: Aaaaaand here.

extensions/guacamole-auth-http/src/main/java/org/apache/guacamole/auth/http/user/AuthenticatedUser.java
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright (C) 2015 Glyptodon LLC
```
mike-jumper: You probably get the idea.

pom.xml
```diff
@@ -53,6 +53,7 @@
         <module>extensions/guacamole-auth-jdbc</module>
         <module>extensions/guacamole-auth-ldap</module>
         <module>extensions/guacamole-auth-noauth</module>
+        <module>extensions/guacamole-auth-http</module>
```
mike-jumper: Please maintain alphabetical ordering of modules within the `pom.xml`.

extensions/guacamole-auth-http/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
mike-jumper: While there does need to be a copy of the Apache License included in the `.tar.gz` bundle (see the packaging of the other extensions), there shouldn't be a top-level copy. The guacamole-client source must have exactly one top-level `LICENSE` file, and that's already present.

extensions/guacamole-auth-http/src/main/assembly/dist.xml
necouchman: No clue...fixed.

extensions/guacamole-auth-http/LICENSE
necouchman: Removed.

extensions/guacamole-auth-http/.gitignore
```diff
@@ -0,0 +1,9 @@
+target/
+pom.xml.tag
```
mike-jumper: So ... what are all these files being ignored? A typical `.gitignore` just contains:

    target/
    *~

Though even that is probably not necessary due to the top-level `.gitignore` of guacamole-client.

pom.xml
necouchman: Reordered.

extensions/guacamole-auth-http/.gitignore
necouchman: Fixed - made the contents the same as the other extensions.

mike-jumper:
For modules which will be part of the guacamole-client codebase (which does make sense here), the Apache RAT plugin needs to be part of the build:

https://github.com/apache/incubator-guacamole-client/blob/c0e050c65add2a4bd83089cfd4e8527214328bb1/extensions/guacamole-auth-ldap/pom.xml#L98-L122

That plugin does license checks against the source of the project, and fails the build if files with unexpected licenses enter the codebase.

extensions/guacamole-auth-http/README.md
```diff
@@ -0,0 +1,2 @@
+# guacamole-auth-http
+Basic HTTP Authentication Module for Guacamole
```
mike-jumper: *Probably* don't need this virtually-empty `README.md`.

extensions/guacamole-auth-http/src/main/java/org/apache/guacamole/auth/http/AuthenticationProviderService.java
necouchman: Fixed.

extensions/guacamole-auth-http/src/main/java/org/apache/guacamole/auth/http/HTTPAuthenticationProvider.java
necouchman: Fixed.

extensions/guacamole-auth-http/src/main/java/org/apache/guacamole/auth/http/ConfigurationService.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * The Guacamole server environment.
+     */
+    @Inject
+    private Environment environment;
+
+    /**
+     * Returns the header of the HTTP server as configured with
+     * guacamole.properties used for HTTP basic authentication.
```
mike-jumper: Is it correct to refer to this as HTTP basic authentication?

The header may come from something that performs HTTP basic auth, but the extension itself isn't doing that. It's just consuming an arbitrary header, the default header `REMOTE_USER` not in fact being part of HTTP basic:

https://en.wikipedia.org/wiki/Basic_access_authentication

extensions/guacamole-auth-http/src/main/java/org/apache/guacamole/auth/http/HTTPAuthenticationProviderModule.java
necouchman: Fixed.

extensions/guacamole-auth-http/src/main/java/org/apache/guacamole/auth/http/user/AuthenticatedUser.java
necouchman: Fixed.

extensions/guacamole-auth-http/README.md
necouchman: Removed.

necouchman:
Added RAT and fixed a build issue from a leftover call to logger().

extensions/guacamole-auth-http/src/main/java/org/apache/guacamole/auth/http/ConfigurationService.java
necouchman: Removed the word basic from these comments.

mike-jumper:
So, I see that you're no longer referring to this as HTTP Basic auth, which is good, however:

1. Just removing the word "basic" doesn't seem to address the issue. "HTTP authentication" is mainly used to refer to the standard auth methods defined for HTTP (basic, digest, etc.) which is not what this extension implements. If the idea is that the extension facilitates auth via HTTP, then really any Guacamole extension qualifies as such, and this needs to be narrowed to what sets this extension apart. The title you specified for the PR says "HTTP header-based authentication" - perhaps that's the best way to describe it?
2. Since this isn't HTTP Basic, [GUACAMOLE-29](https://issues.apache.org/jira/browse/GUACAMOLE-29)  ("Add support for requesting HTTP Basic Authentication") is probably not the best home for these changes, no? The changes themselves look reasonable, but a proper JIRA issue is needed.

extensions/guacamole-auth-http/src/main/assembly/dist.xml
```diff
@@ -0,0 +1,53 @@
...
+        <format>tar.gz</format>
+    </formats>
+
+    <!-- Include licenses and extension .jar -->
+    <fileSets>
+
+        <!-- Include licenses -->
+        <fileSet>
+            <outputDirectory></outputDirectory>
+            <directory>src/licenses</directory>
```
mike-jumper: In order for the licenses to be included (and they need to be), they need to be part of the source tree. Here things are simpler, since this extension isn't bundling anything which hasn't already been bundled by other extensions, so the burden of research is lessened.  The contents of the `src/licenses/` directory for guacamole-auth-ldap is probably the best reference:

https://github.com/apache/incubator-guacamole-client/tree/d358449c7f1592e6b06aadc94dcf39dcc50c9671/extensions/guacamole-auth-ldap/src/licenses

I think your extension has the exact same dependencies (with the exception of jldap, of course).

mike-jumper: To clarify, the licenses I refer to above are the contents of the `src/licenses/` directory (as well as `DISCLAIMER` and `NOTICE`), which is absent here. Unlike the top-level license which you needed to remove previously, these licenses are what apply to the bundled `.tar.gz` produced by the build.

necouchman:
Hmmmm....guess I misunderstood what GUACAMOLE-29 was getting at - maybe just had it inverted or something like that, accepting existing HTTP header-based authentication rather than requesting it.  I'll open a new JIRA issue and a new pull request for the functionality.  Sorry about that.

mike-jumper:
Cool. Thanks, Nick.

There's no need for an entirely new pull request, by the way. It won't hurt anything, but you can simply rebase and tag the new issue once it exists (and edit the tag in the title of the PR).

extensions/guacamole-auth-http/src/main/assembly/dist.xml
necouchman: Added those licenses - copied from LDAP and removed jldap.

necouchman:
I've done the rebase and changed the title of the pull request.  The only thing I haven't figured out how to do is if there's a way to permanently rename the github.com branch, or if that even matters at this point.  The branch is still called GUACAMOLE-29 - if I can figure out a way to rename it, I will.

necouchman:
Oops...guess renaming a branch involves deleting the old one.  Let me know if I need to open a new pull request with a corrected branch name.

Also, I tweaked some of the comments to accurately reflect the HTTP *header* authentication nature of this extension - let me know if you think I should also rename classes or the entire module to something a little closer to that and I can rework it.

mike-jumper:
> Oops...guess renaming a branch involves deleting the old one. Let me know if I need to open a new pull request with a corrected branch name.

No worries - the branch name is strictly on your end. It has no bearing on the merge.

> Also, I tweaked some of the comments to accurately reflect the HTTP header authentication nature of this extension - let me know if you think I should also rename classes or the entire module to something a little closer to that and I can rework it.

Good point... "guacamole-auth-header", `org.apache.guacamole.auth.header`, `HTTPHeaderAuthenticationProvider`, etc.? @jmuehlner and @flangelo - any feelings here?

extensions/guacamole-auth-http/src/licenses/LICENSE
```diff
@@ -0,0 +1,253 @@
...
+Google Guice (https://github.com/google/guice)
+----------------------------------------------
+
+    Version: 3.0
+    From: 'Google Inc.' (http://www.google.com/)
+    License(s):
+        Apache v2.0 (bundled/guice-3.0/COPYING)
+
+
+JLDAP (http://www.openldap.org/jldap/)
```
mike-jumper: Missed a JLDAP.

necouchman: Removed.

necouchman:
Any more feedback on whether this extension should get renamed or not?  I'm happy to refactor as guacamole-auth-header, as suggested, or leave it as-is...

extensions/guacamole-auth-http/src/main/java/org/apache/guacamole/auth/http/AuthenticationProviderService.java
```diff
@@ -0,0 +1,89 @@
...
+     */
+    public AuthenticatedUser authenticateUser(Credentials credentials)
+            throws GuacamoleException {
+
+        // Pull HTTP header from request if present
+        HttpServletRequest request = credentials.getRequest();
+        if (request != null) {
+            String username = request.getRemoteUser();
+            if(username == null)
+                username = request.getHeader(confService.getHttpAuthHeader());
```
mike-jumper: The logic here could stand to be better commented. Beware the "if is not a function" thing again, too.

Beyond that, I'm a little worried about the behavior of the `http-auth-header` property with respect to `getRemoteUser()`. Presumably, the expectation is that if the admin specifies `http-auth-header: X-Foo-Bar` in their `guacamole.properties`, then the auth backend will look at that `X-Foo-Bar` header to determine the identity of the user, but that won't be the case if `getRemoteUser()` returns non-null.

For that matter, and of possible relevance to the renaming, is `getRemoteUser()` driven by HTTP headers?

I suppose that must be why you originally had this change listed as support for HTTP Basic, though I'm having a hard time finding what exactly drives `getRemoteUser()`. Do you have any information on this?

jmuehlner:
@mike-jumper I think that HTTPHeaderAuthenticationProvider is a reasonable name - HTTPAuthenticationProvider is a little overly generic for my tastes.

Similarly, guacamole-auth-header gives a better idea of what this extension does, in my opinion.

So yeah, renaming seems like a good idea. :)

necouchman:
Okay, I've gone through and renamed everything, built and tested to make sure it still works, and pushed the new code.

extensions/guacamole-auth-http/src/main/java/org/apache/guacamole/auth/http/AuthenticationProviderService.java
necouchman: Added some comments on the logic.  I'll dig into the getRemoteUser() stuff and see if I can find anything, but it may just be less confusing to the person operating it to just pull the getRemoteUser() code altogether and use strictly header-based authentication.  I'll see what I can find.

necouchman: So, Oracle's JavaEE documentation has this to say on the subject of HttpServletRequest getRemoteUser():

> Returns the login of the user making this request, if the user has been authenticated, or null if the user has not been authenticated. Whether the user name is sent with each subsequent request depends on the browser and type of authentication. Same as the value of the CGI variable REMOTE_USER.

This indicates to me that this function probably works off the REMOTE_USER header.

That said, it may still be a little confusing for the operator to set something up in guacamole.properties for a custom header, and then run into a situation where a different username gets pulled from getRemoteUser() (because it's looking at the REMOTE_USER header) and whatever the operator has configured as their header.  So, I think the two options would be:
1) Remove getRemoteUser() altogether and just work off the configured header, or default to REMOTE_USER if none is configured.
2) Reverse the logic such that the configured header is checked, first, and, if nothing is found or nothing configured, revert to getRemoteUser().  This may not be overly-useful though, and may not eliminate the confusion for the operator.

Thoughts?

mike-jumper: I'd say just remote `getRemoteUser()` altogether. If it does use the `REMOTE_USER` header, as you expect, then it's superfluous and a potential security issue for any deployment which explicitly specifies `http-auth-header` expecting the default `REMOTE_USER` to be overridden.

necouchman: Removed and pushed changes to this branch.

extensions/guacamole-auth-header/src/main/java/org/apache/guacamole/auth/header/AuthenticationProviderService.java
```diff
@@ -0,0 +1,88 @@
...
+    public AuthenticatedUser authenticateUser(Credentials credentials)
+            throws GuacamoleException {
+
+        // Pull HTTP header from request if present
+        HttpServletRequest request = credentials.getRequest();
+        if(request != null) {
+            // Get the username from the header configured in guacamole.properties
+            String username = request.getHeader(confService.getHttpAuthHeader());
+
+            if(username != null) {
```
mike-jumper: `if` is not a function please.

extensions/guacamole-auth-header/src/main/java/org/apache/guacamole/auth/header/AuthenticationProviderService.java
```diff
@@ -0,0 +1,88 @@
...
+     *     If an error occurs while authenticating the user, or if access is
+     *     denied.
+     */
+    public AuthenticatedUser authenticateUser(Credentials credentials)
+            throws GuacamoleException {
+
+        // Pull HTTP header from request if present
+        HttpServletRequest request = credentials.getRequest();
+        if(request != null) {
+            // Get the username from the header configured in guacamole.properties
```
mike-jumper: Please also keep in mind, from the style guidelines (and the general style of the code in this repository):

> Comment everything semantically and liberally, use blank lines to separate logical blocks. 

I like that this part is commented, but it looks awfully crowded lacking separation from the enclosing `if` block. Similarly, though I like the spacing before nested `if`, it's not commented semantically, and the closing half of things is crowded.

It looks decent as is, to be honest, but with minor changes it could be much more readable.

mike-jumper:
OK - looking through the rest of things, I'm not noticing anything else. Once the latest style nitpicks are addressed, I think it's ready for merge.

extensions/guacamole-auth-header/src/main/java/org/apache/guacamole/auth/header/AuthenticationProviderService.java
necouchman: Doh.  Fixed one of them but went the wrong way on it!  Fixed it the correct way this time.

extensions/guacamole-auth-header/src/main/java/org/apache/guacamole/auth/header/AuthenticationProviderService.java
necouchman: I added some spacing to it, so hopefully it looks better, now!

mike-jumper:
Looks good to me. :shipit:

PR #120, opened 2017-02-09
GUACAMOLE-201: Occasionally verify client responsiveness when intercepting streams.

mike-jumper:


guacamole/src/main/webapp/app/rest/services/tunnelService.js
```diff
@@ -215,6 +215,11 @@ angular.module('rest').factory('tunnelService', ['$injector',
             document.body.removeChild(iframe);
         };
 
+        // Acknowledge (and ignore) and received blobs
```
jmuehlner: Is this meant to say "any received blobs"?

mike-jumper: No. It's meant to say "Acknowledge (and ignore) any received blobs".

mike-jumper: ...but yes, you're right.

jmuehlner:
Ok, this makes sense to me.

PR #121, opened 2017-02-10
GUACAMOLE-204: Implement CAS Single Sign On Support

necouchman:
This pull request implements support for the Apereo CAS Single Sign On system.


necouchman:
Fixed a couple of issues with the initial pull request, including re-enabling Apache RAT in the pom.xml file, and removing unnecessary files from the base of the extension.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/CASAuthenticationProvider.java
```diff
@@ -0,0 +1,107 @@
...
+import org.apache.guacamole.net.auth.Credentials;
+import org.apache.guacamole.net.auth.UserContext;
+
+/**
+ * Guacamole authentication backend which authenticates users using an
+ * arbitrary external system implementing CAS. No storage for connections is
+ * provided - only authentication. Storage must be provided by some other
+ * extension.
+ *
+ * @author Michael Jumper
```
jmuehlner: I'm guessing you meant this to say "Nick Couchman"? Same for CASAuthenticationProviderModule.

necouchman: Well, since I had basically copied and pasted files that Mike wrote with almost no changes, I was trying to avoid plagiarizing someone else's work.  I've updated them to have my name...

extensions/guacamole-auth-cas/src/main/resources/cas.css
```diff
@@ -0,0 +1,23 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/* Hide login dialog */
+.login-ui div.login-dialog {
+    display: none;
```
mike-jumper: Why is the login dialog being hidden? Won't this break things for all other auth methods?

extensions/guacamole-auth-cas/src/main/resources/casConfig.js
```diff
@@ -0,0 +1,51 @@
...
+ * Config block which augments the existing routing, providing special handling
+ * for the "ticket=" fragments provided by OpenID Connect.
+ */
+angular.module('index').config(['$routeProvider',
+        function indexRouteConfig($routeProvider) {
+
+    var curPath = window.location.href;
+    var ticketPos = curPath.indexOf("?ticket=") + 8;
+    var hashPos = curPath.indexOf("#/");
+    if(ticketPos > 0 && ticketPos < hashPos) {
```
mike-jumper: `if(` 😭

extensions/guacamole-auth-cas/src/main/resources/cas.css
necouchman: Probably so, and, come to think of it, it really shouldn't be necessary since loading the form should redirect to the SSO page, then back to Guacamole after authentication.  I'll remove it, build it and make sure it still works, and commit after verifying.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/AuthenticationProviderService.java
```diff
@@ -0,0 +1,111 @@
...
+     *     given credentials.
+     *
+     * @throws GuacamoleException
+     *     If an error occurs while authenticating the user, or if access is
+     *     denied.
+     */
+    public AuthenticatedUser authenticateUser(Credentials credentials)
+            throws GuacamoleException {
+
+        String ticket = null;
```
necouchman: Fixed.

extensions/guacamole-auth-cas/src/main/resources/casConfig.js
```diff
@@ -0,0 +1,51 @@
...
+        var ticket = curPath.substring(ticketPos, hashPos);
+        var newPath = curPath.substring(0,ticketPos - 8) + '#/?ticket=' + ticket;
+        window.location=newPath;
```
mike-jumper: Is CAS OK with this sort of thing? I recall having difficulty with redirect URLs and OpenID as endpoints prohibited the presence of `#`.

Stylewise, should have spaces around the `=` like everywhere else.

extensions/guacamole-auth-cas/src/main/resources/casConfig.js
necouchman: -Sigh- I had even gone through and checked specifically for that.  Fixed.

mike-jumper:
Is there any documentation you can point to for CAS which could help put this PR into context?

extensions/guacamole-auth-cas/src/main/resources/casConfig.js
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Config block which augments the existing routing, providing special handling
+ * for the "ticket=" fragments provided by OpenID Connect.
+ */
+angular.module('index').config(['$routeProvider',
+        function indexRouteConfig($routeProvider) {
+
+    var curPath = window.location.href;
+    var ticketPos = curPath.indexOf("?ticket=") + 8;
+    var hashPos = curPath.indexOf("#/");
```
necouchman: I'll take a look - I actually meant to ask for help with this specifically.  This is pretty hacked together, but I fought with Angular for quite a while trying to pull the ticket= parameter out of the return URL and put it in the right place, and literally cutting it out and putting it at the right position ended up being the way I had to go.  I'll go back and review, but it seemed like I stumbled across a couple of limitations in Angular that prevented me from doing it in a more elegant way.

necouchman:
Here's the base documentation for CAS:
https://apereo.github.io/cas/4.2.x/index.html

If you're looking for specifics on the protocol, here is the protocol documentation:
https://apereo.github.io/cas/4.2.x/protocol/CAS-Protocol.html

And here's the github link for the java-specific implementation of the CAS client.  It has lots of information on parameters and what gets passed back and forth between client and server.
https://github.com/apereo/java-cas-client

extensions/guacamole-auth-cas/src/main/resources/casConfig.js
```diff
@@ -0,0 +1,51 @@
...
+}]);
+
+/**
+ * Config block which augments the existing routing, providing special handling
+ * for the "ticket=" fragments provided by OpenID Connect.
+ */
+angular.module('index').config(['$routeProvider',
+        function indexRouteConfig($routeProvider) {
+
+    var curPath = window.location.href;
```
mike-jumper: @jmuehlner would probably also point out that AngularJS wraps `window` and provides [`$window`](https://code.angularjs.org/1.3.16/docs/api/ng/service/$window) to help facilitate unit testing (whenever we get around to having such things at the JS level).

necouchman: I believe I have this implemented in the more proper AngularJS way.

extensions/guacamole-auth-cas/src/main/resources/cas.css
necouchman: Removed this and the CSS file altogether, and added a translation file just to blank out the LOGIN.FIELD_HEADER_TICKET text in the login box.  A login box is now displayed very briefly while the page redirects.

extensions/guacamole-auth-cas/src/main/resources/casConfig.js
necouchman: This was several weeks ago, but, IIRC, the reason I had to hack this together is that $routeParams only looks at everything past the #/ part of the URL.  When CAS redirects back to Guacamole, the URL comes in looking like this:
https://my.server.local/guacamole/?ticket=blahblahblah/#

$routeParams then pulls in nothing, because it sees nothing after the #.  I can't remember how much I played with window.location.search, but I'm looking into that now to see if I can pull this without having to use nasty code like strlen :-).

necouchman: So, I can definitely use $window.location.search to extract the ticket parameter from the current URL.  I'm still not entirely sure the most elegant way to then take that, strip it off of the current path, and add it on the end past the route parameters?

mike-jumper:
Thanks for the info, Nick! This helps a lot. I'll take a look in a bit.

mike-jumper:
Hey @necouchman - I'm picking this review back up finally, so stand by for further feedback. I plan to take a look over #122 again, soon, to help troubleshoot the issues you mentioned with the bouncycastle lib, but please kick me if I don't make good on that. I've been losing track of both of these.

In other news, I believe your Apache ID / account, etc. [has gone through successfully](https://people.apache.org/phonebook.html?uid=vnick), so please reset author on your commits to vnick@apache.org when you get a chance.

necouchman:
Sounds good, Mike - thanks.  I haven't had much time lately to work on it, but if you any feedback or changes I'll try to respond and make those.
And, yes, I got all my Apache account stuff squared away, so should be good to go there - thanks!
-Nick
 

    On Thursday, March 16, 2017 9:57 PM, Mike Jumper <notifications@github.com> wrote:
 

 Hey @necouchman - I'm picking this review back up finally, so stand by for further feedback. I plan to take a look over #122 again, soon, to help troubleshoot the issues you mentioned with the bouncycastle lib, but please kick me if I don't make good on that. I've been losing track of both of these.In other news, I believe your Apache ID / account, etc. has gone through successfully, so please reset author on your commits to vnick@apache.org when you get a chance.—

necouchman:
Okay, pinging on this one - any further issues/comments on this?

mike-jumper:
Thanks for the ping - taking a look now.

extensions/guacamole-auth-cas/pom.xml
```diff
@@ -0,0 +1,172 @@
...
+
+        <!-- Guacamole Extension API -->
+        <dependency>
+            <groupId>org.apache.guacamole</groupId>
+            <artifactId>guacamole-ext</artifactId>
+            <version>0.9.11-incubating</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+          <groupId>org.jasig.cas.client</groupId>
```
mike-jumper: Indentation here is wonky (should be four spaces, not two).

This is also crammed into the section labelled "Guacamole Extension API", which is incorrect. Should separate this section from the others with blank lines and add a human-readable comment noting what the dependency is, like the other deps.

extensions/guacamole-auth-cas/pom.xml
```diff
@@ -0,0 +1,172 @@
...
+          <artifactId>cas-client-core</artifactId>
```
mike-jumper: What license is `cas-client-core` under?

Just as with `guacamole-auth-header`, we'll need to include the licenses of all bundled dependencies, and ensure that the licensing is compliant with ASF policy (see https://www.apache.org/legal/resolved.html).

Some of these dependencies are the same as other extensions, so things like Guice should be easy to copy into place, but you'll have to track down the licensing of the CAS client, include it within `src/licenses/...` like the others, and so on.

Note also that the CAS client may itself pull in transitive dependencies which end up bundled, too, and we'll need to track down *those* licenses and do the same there.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/ConfigurationService.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * Returns the URI that the CAS service should redirect to after
+     * the authentication process is complete, as configured with
+     * guacamole.properties. This must be the full URL that a user would enter
+     * into their browser to access Guacamole.
+     *
+     * @return
+     *     The client secret to use when communicating with the CAS service,
```
mike-jumper: Looks like some copy pasta here. You mean the redirect URI?

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
```diff
@@ -0,0 +1,80 @@
...
+    public String processUsername(String ticket) throws GuacamoleException {
+        AttributePrincipal principal = null;
+
+        // Retrieve the configured CAS URL and establish a ticket validator
+        String casServerUrl = confService.getAuthorizationEndpoint();
+        Cas20ProxyTicketValidator sv = new Cas20ProxyTicketValidator(casServerUrl);
+        sv.setAcceptAnyProxy(true);
+        try {
+            String confRedirectURI = confService.getRedirectURI();
+            Assertion a = sv.validate(ticket, confRedirectURI);
```
mike-jumper: Can you clean up and comment the logic here? I see you say:

    // Retrieve the configured CAS URL and establish a ticket validator

but there seems to be more going on here than that. What is `sv`? The ticket validator? What are we doing with that ticket validator which results in us ultimately returning the username?

extensions/guacamole-auth-cas/src/main/resources/casConfig.js
```diff
@@ -0,0 +1,52 @@
...
+        templateUrl   : '',
+        controller    : 'guacCASController',
+        module        : 'guacCAS'
+    });
+
+}]);
+
+/**
+ * Config block which augments the existing routing, providing special handling
+ * for the "ticket=" fragments provided by OpenID Connect.
```
mike-jumper: You mean CAS?

extensions/guacamole-auth-cas/src/main/resources/casConfig.js
```diff
@@ -0,0 +1,52 @@
...
+ */
+angular.module('index').config(['$routeProvider','$windowProvider',
+        function indexRouteConfig($routeProvider,$windowProvider) {
+
+    var $window = $windowProvider.$get();
+    var curPath = $window.location.href;
+    var ticketPos = curPath.indexOf("?ticket=") + 8;
+    var hashPos = curPath.indexOf("#/");
+    if (ticketPos > 0 && ticketPos < hashPos) {
```
mike-jumper: The test `ticketPos > 0` will always be true, as you're adding `8` to the result of `indexOf()`, masking the potential `-1` return value indicating the absence of the string `"?ticket="`.

extensions/guacamole-auth-cas/src/main/resources/casConfig.js
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Config block which augments the existing routing, providing special handling
+ * for the "ticket=" fragments provided by OpenID Connect.
+ */
+angular.module('index').config(['$routeProvider','$windowProvider',
+        function indexRouteConfig($routeProvider,$windowProvider) {
+
+    var $window = $windowProvider.$get();
+    var curPath = $window.location.href;
+    var ticketPos = curPath.indexOf("?ticket=") + 8;
```
mike-jumper: Beware that, by convention, strings in JavaScript should be `'single quoted'`, not `"double quoted"`. They're both legal ... so all the more reason to pick one and only one for consistency's sake. We happen to have gone with the prevailing convention in JavaScript programming land of single quotes.

From http://guacamole.incubator.apache.org/guac-style/#naming:

> Prefer `'single quotes'` over `"double quotes"` for strings in JavaScript.

I had to get used to that one, too. You'll find much of the older guac JavaScript code using double quotes as I used to resist that convention. New code uses single quotes, though.

extensions/guacamole-auth-cas/src/main/resources/casConfig.js
```diff
@@ -0,0 +1,52 @@
...
+    var hashPos = curPath.indexOf("#/");
+    if (ticketPos > 0 && ticketPos < hashPos) {
+        var ticket = curPath.substring(ticketPos, hashPos);
+        var newPath = curPath.substring(0,ticketPos - 8) + '#/?ticket=' + ticket;
+        $window.location.href = newPath;
```
mike-jumper: There is a more Angular-y way to accomplish this, rather than manually parsing everything, since Angular already has to do this when parsing the URL to handle its own routing.

I'm actually not 100% positive this block of code will work as expected - it's a config block which doesn't actually do any configging. If you want an example of an approach which embraces Angular's own routing, take a peek at the WIP OpenID support:

https://github.com/mike-jumper/incubator-guacamole-client/blob/9deaac26df8220efeca90a0c9f8e9bdde39c86ff/extensions/guacamole-auth-openid/src/main/resources/openidConfig.js#L34-L52

I suspect that's what you were already looking at when you first started. ;)

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/form/CASTicketField.java
```diff
@@ -0,0 +1,97 @@
...
+import java.math.BigInteger;
+import java.net.URLEncoder;
+import java.security.SecureRandom;
+import org.apache.guacamole.form.Field;
+
+
+/**
+ * Field definition which represents the ticket returned by an CAS service.
+ * Within the user interface, this will be rendered as an appropriate "Log in
+ * with ..." button which links to the CAS service.
```
mike-jumper: > Within the user interface, this will be rendered as an appropriate "Log in with ..." button which links to the CAS service.

Is this the case? The code looks like it takes the user to the CAS service immediately. If it does take the user to the CAS service immediately ... should there be a button? What would be the ideal user experience?

extensions/guacamole-auth-cas/src/main/resources/translations/en.json
```diff
@@ -0,0 +1,7 @@
+{
+
+    "LOGIN" : {
+        "FIELD_HEADER_TICKET" : ""
```
mike-jumper: What's the purpose of defining an empty translation string for the field header?

mike-jumper:
OK - done with first round of reviewing.

As a side note: now would be a good time to reset the author of your existing commits here such that they all use your Apache email address. You can edit your GitHub info such that it's associated with your GitHub account along with all the other emails you use, but your commits against Guacamole should use the Apache one.

extensions/guacamole-auth-cas/pom.xml
necouchman: Should be fixed.

extensions/guacamole-auth-cas/pom.xml
necouchman: Good question...will have to check on this.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
necouchman: Beefed up those comments a bit.

extensions/guacamole-auth-cas/src/main/resources/casConfig.js
necouchman: Uh...yeah.

extensions/guacamole-auth-cas/src/main/resources/casConfig.js
necouchman: Fixed here, need to check other JS source files to make sure.

extensions/guacamole-auth-cas/src/main/resources/casConfig.js
necouchman: Added some logic to check for the -1 and return if not found, then proceed on.

extensions/guacamole-auth-cas/src/main/resources/casConfig.js
necouchman: Yeah, I tried to figure that out, but ran into a significant road-block going that route, and, after hours of searching and trying, resorted to this admittedly messy way of accomplishing this.  Here's the issue...

When CAS redirects back to Guacamole, the URL that comes back is something like this:
http://guacamole.example.com/guacamole?ticket=<BIG LONG STRING>

What happens is that Angular then appends the trailing slash and the hash tag onto the end of that:
http://guacamole.example.com/guacamole?ticket=<BIG LONG STRING>/#/

So, if I try to use the Angular built-in routing functions, it's only looking for the ticket= after the trailing slash.  It doesn't process anything ahead of that slash.  I tried several incantations, including putting a trailing slash in the redirect URI, but just ended up with this:
http://guacamole.example.com/guacamole/?ticket=<BIG LONG STRING>/#/

I also tried putting the #/ in the CAS redirect URI, and that didn't work, either.  I think the core of the issue is that, during authentication, it isn't just an iFrame or something like that that accomplishes the authentication (as in the case of Duo) - it actually redirects completely away from the site, then back, with this ticket= parameter in the URL.  Since the ticket= parameter is placed there before Angular ever initializes, Angular believes that it is part of the URL it should be using, appends it /#/ to the end, and goes on.  So, I have to manually move the ?ticket= chunk to the end.

I'm definitely open to suggestions and help with this one - I banged my head on a kitchen table for quite a while trying to figure out where I went wrong, and came away with the belief that the way CAS interacts with the page here just doesn't place well with Angular.

I suppose I could try to go the iFrame route, instead of a full redirect, and see if that changes things?  Maybe it would end up looking a lot more like the Duo module...and it might dovetail nicely with the suggestion below about having a separate button for the user to click to authenticate with CAS, rather than just having it redirect to CAS authentication automatically.  Based on how I've explained CAS to you, does that sound more feasible (and desirable) than redirecting completely away??

extensions/guacamole-auth-cas/src/main/resources/translations/en.json
necouchman: Since in the current implementation the redirect to CAS is transparent, and the ticket is handled transparently, this just serves to make sure any passing UI components are essentially blank as the user gets redirected to the authentication page and back.  Maybe it's unnecessary...

necouchman:
Thanks, Mike.  I've corrected most of the stuff, and have changed my commit e-mail address for this repo.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/ConfigurationService.java
necouchman: Fixed.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/form/CASTicketField.java
necouchman: Good question.  It seems like the default behavior by all of the applications I have integrated with CAS is that, as soon as the application login page is displayed, you are redirected transparently to the CAS login page, then, once you've successfully authenticated, you're redirected back to the application page where the ticket is validated and the application allows you in.  My guess is that most of the applications that people want SSO integration with in CAS they want people to be able to hit a page, login and then, after that, just get transparently logged into each application without having to do anything - including click another button to trigger the CAS SSO login.

This is a little different from something like OpenID (or Google, Yahoo, Live, etc.), where so many different providers are integrating with the SSO system, and you want some clue that your data in the SSO system is going to be accessed by the application you're trying to use.  In the case of CAS, the target is more organizations that are intentionally building and integrating applications into an SSO framework, and trying to make it as easy as possible for users to interact with those applications.

That said, I'm not opposed to have a single SSO button somewhere that allows the user to bypass username/password authentication to Guacamole and proceed to SSO, if that's what we think should happen?

So, the short answer to "What would be the ideal user experience?"  That depends :-).

extensions/guacamole-auth-cas/pom.xml
necouchman: JASIG CAS client is licensed under Apache 2.0.  So that works well :-).

Here are the dependencies that get pulled in by either org.jasig.cas-client or org.jasig.cas-client-core:
org.springframework.spring-core (Apache 2.0)
org.springframework.spring-context (Apache 2.0)
org.springframework.spring-test (Apache 2.0)
org.springframework.spring-beans (Apache 2.0)
log4j.log4j (Apache 2.0)
junit.junit (EPL 1.0)
org.slf4j.slf4j-api (MIT)
org.slf4j.jcl-over-slf4j (MIT)
org.slf4j.slf4j-simple (MIT)
javax.servlet.servlet-api (CDDL/GPL2.0)
xml-security.xmlsec (Apache 2.0)
commons-codec.commons-codec (Apache 2.0)

necouchman:
Any further issues with these changes?

mike-jumper:
> I've corrected most of the stuff, and have changed my commit e-mail address for this repo.

You'll still need to reset the author of the past commits, though, as changing the email address for the repository only affects commits going forward.

> Any further issues with these changes?

I'll take another look now, but off the top of my head:

I don't see the licenses, etc. for known dependencies common with other extensions (like Guice), their transitive dependencies, and the CAS library dependency and its transitive dependencies. Those will need to be added just as with the other extensions before we can bring this into the main codebase.

extensions/guacamole-auth-cas/src/main/resources/casConfig.js
mike-jumper: > I suppose I could try to go the iFrame route ...

Nooooooooo

> Maybe it would end up looking a lot more like the Duo module

Hopefully not - we were forced to use an `<iframe>` there because that's the only way their "Web SDK" works, but I cringe at being forced to do that here.

I'll comment on the new changes so things remain visible - had to dig a bit to find the old comments. ;)

extensions/guacamole-auth-cas/src/main/resources/casConfig.js
```diff
@@ -0,0 +1,55 @@
...
+    });
+
+}]);
+
+/**
+ * Config block which augments the existing routing, providing special handling
+ * for the "ticket=" parameter provided by the CAS authentication process.
+ */
+angular.module('index').config(['$routeProvider','$windowProvider',
+        function indexRouteConfig($routeProvider,$windowProvider) {
```
mike-jumper: Issues with the query parameter vs. Angular's routing notwithstanding, the following issues remain:

1. This is a `.config()` block which doesn't actually do any configuration. Besides being not very Angular-y, this will only run once when the application first loads. Any time after that will not work as expected, including when navigating from page to page or after logging out.
2. As you've noted, it's sloppy.

Have you had any luck using the same technique as my WIP OpenID auth?

https://github.com/mike-jumper/incubator-guacamole-client/blob/9deaac26df8220efeca90a0c9f8e9bdde39c86ff/extensions/guacamole-auth-openid/src/main/resources/openidConfig.js#L34-L52

If not ... there must be some reasonable way of doing this. Let me know if the above doesn't work (and an example of the URL used by CAS for its redirect), and I'll dig in to see if this can be fixed globally somehow.

extensions/guacamole-auth-cas/src/main/resources/translations/en.json
mike-jumper: Hm ... This issue will likely bite me, as well, with the OpenID dev. In fact, I suspect it may have already and I just don't remember. It's been a bit since I actively worked with that code.

I wonder if perhaps we should modify the handling of fields such that fields without defined templates do not render at all... That way automatically-populated fields required for credentials, etc. could be defined without requiring repeated CSS or empty translation string hackery.

mike-jumper: I took a quick stab at simply not rendering fields or their headers if they have no associated content: mike-jumper/incubator-guacamole-client@6c5cdae

If that solves the case you're working around here, I'll open a pull request for that commit.

extensions/guacamole-auth-cas/src/main/resources/casConfig.js
necouchman: 1) Okay, I'll figure out some other block type to move it to.
2) Yeah.

No, the block of code you highlighted there falls into the hole I mentioned before.  Specifically:
- The comment indicates that it transforms /#/id_token= to /#/?id_token=.  The situation I end up with is that I need to transform ?id_token=/#/ to /#/?id_token=.
- Because the ?id_token= parameter gets placed before the /#/, the $routeProvider.when block never actually executes, because the $routeProvider code only ever looks at anything passed the hash tag.
- I can't remember for sure if this was true or not, but I believe that this also means that the $location.search(params) line also does not work, because it's looking at anything trailing the #/ on the URL, not what occurs in front of it.

I'm happy to continue trying stuff to find a better solution to this - let me know if you have other suggestions!

extensions/guacamole-auth-cas/src/main/resources/translations/en.json
necouchman: I think that helps, although the redirect happens quickly enough that it's a little hard to tell.  I'll at least remove the empty translation information from my extension.

necouchman:
> I don't see the licenses, etc. for known dependencies common with other extensions (like Guice), their ?
transitive dependencies, and the CAS library dependency and its transitive dependencies. Those will need to be added just as with the other extensions before we can bring this into the main codebase.

I've gone through the dependencies and added the licenses.  I did want to ask how far down the chain I need to chase dependencies and add licenses - for example, cas depends on log4j, which depends on several base Java libraries (see in the Compile Dependencies on this page: https://mvnrepository.com/artifact/log4j/log4j/1.2.15 ) - do I need to run those down, as well?  That's fine if I do, just want to verify before I go to the trouble.

Also, to this point I've been chasing down these dependencies just by going through the Maven POM files for each of them and looking at the dependencies.  Is there a better way to do this?  Or is it just a manual process?

extensions/guacamole-auth-cas/src/main/resources/casConfig.js
necouchman: > This is a .config() block which doesn't actually do any configuration. Besides being not very Angular-y, this will only run once when the application first loads. Any time after that will not work as expected, including when navigating from page to page or after logging out.

What would be the proper place in Angular to do something like this?  I took a stab at creating a new controller in the index module and putting it there, but I'm guessing the controller functions don't get run automatically (need to be called), so is there a service/controller/directive/etc. that will get run automatically, or some place I should call the controller?

necouchman: So, still not sure this is the way to do this, but I moved this out of the config() block and into a run() block.  Code should reflect changes.

mike-jumper:
> Also, to this point I've been chasing down these dependencies just by going through the Maven POM files for each of them and looking at the dependencies. Is there a better way to do this?

Thankfully, yes. You can use [the `dependency:list` goal](https://maven.apache.org/plugins/maven-dependency-plugin/list-mojo.html):

     mvn dependency:list -DincludeScope=runtime

That will produce a list of all runtime dependencies, including transitive dependencies, which are what end up exploded inside the extension.

Beware that you can't necessarily trust the license information in a dependency's pom, nor can you look to the project's current source or website (projects sometimes change licenses). Once the dependencies are known, the tricky part is tracking down the source of each (at the specific point in time of the version being used) and determining the license used.

mike-jumper:
> I've gone through the dependencies and added the licenses. I did want to ask how far down the chain I need to chase dependencies and add licenses - for example, cas depends on log4j, which depends on several base Java libraries (see in the Compile Dependencies on this page: https://mvnrepository.com/artifact/log4j/log4j/1.2.15 ) - do I need to run those down, as well?

Yep. Anything which is ultimately bundled with the extension (and thus will be distributed in the `.tar.gz` included with each release) needs to be tracked down. The exception to this would be things which we will not bundle ourselves (like the WIP RADIUS extension, as we recently figured out), but the preference is definitely to be able to bundle.

mike-jumper:
OK - I've reread through these changes, and tentatively I think they're fine. I think we're good to move forward once we find a way to clean up that JavaScript:

https://github.com/necouchman/incubator-guacamole-client/blob/3ebe96a8688338d8c7fbf2cdf89ef34d0f107702/extensions/guacamole-auth-cas/src/main/resources/casIndexTicket.js#L28-L40

I'll try installing CAS and this extension myself, and tinker with the above. Perhaps a solution will present itself, or perhaps the main webapp can be modified to clean up its own URLs after the redirect.

mike-jumper:
I ran the Maven dependency list plugin myself to check the dependencies, licenses, etc. The following are the only dependencies that would end up bundled:

```
[INFO]    com.google.inject.extensions:guice-multibindings:jar:3.0:compile
[INFO]    javax.inject:javax.inject:jar:1:compile
[INFO]    com.google.inject:guice:jar:3.0:compile
[INFO]    org.jasig.cas.client:cas-client-core:jar:3.4.1:compile
[INFO]    aopalliance:aopalliance:jar:1.0:compile
[INFO]    org.slf4j:slf4j-api:jar:1.7.7:compile
```

Of the above, SLF4J is actually already included in the Guacamole webapp. We should exclude it from inclusion in the extension (declare it as provided in the `pom.xml`), to avoid conflicts and to make life easier.

I'll comment on the above specifically in another review.

extensions/guacamole-auth-cas/src/licenses/LICENSE
```diff
@@ -0,0 +1,294 @@
...
+
+AOP Alliance (http://aopalliance.sourceforge.net/)
+--------------------------------------------------
+
+    Version: 1.0
+    From: 'AOP Alliance' (http://aopalliance.sourceforge.net/members.html)
+    License(s):
+        Public Domain (bundled/aopalliance-1.0/LICENSE)
+
+Commons Codec (http://commons.apache.org/codec/)
```
mike-jumper: What pulls this in? I didn't see it in the list when I ran the Maven dependency plugin.

extensions/guacamole-auth-cas/src/licenses/LICENSE
```diff
@@ -0,0 +1,294 @@
...
+
+Google Guice (https://github.com/google/guice)
+----------------------------------------------
+
+    Version: 3.0
+    From: 'Google Inc.' (http://www.google.com/)
+    License(s):
+        Apache v2.0 (bundled/guice-3.0/COPYING)
+
+Javax Servlet API (http://servlet-spec.java.net/)
```
mike-jumper: This is not bundled, but already provided by the servlet container hosting the Guacamole webapp. If it does turn up within the extension, it'll need to be excluded to avoid conflicts.

3.0.1? Is this a transitive dependency of the CAS client library?

extensions/guacamole-auth-cas/src/licenses/LICENSE
```diff
@@ -0,0 +1,294 @@
...
+
+JSR-330 / Dependency Injection for Java (http://code.google.com/p/atinject/)
+----------------------------------------------------------------------------
+
+    Version: 1
+    From: 'JSR-330 Expert Group' (https://jcp.org/en/jsr/detail?id=330)
+    License(s):
+        Apache v2.0 (bundled/javax.inject-1/LICENSE-2.0.txt)
+
+JUnit (http://junit.org)
```
mike-jumper: This is not bundled. It might be a transitive dependency, but only for running the build-time unit tests of that dependency.

extensions/guacamole-auth-cas/src/licenses/LICENSE
```diff
@@ -0,0 +1,294 @@
...
+
+slf4j (http://www.slf4j.org)
+----------------------------
+
+    Version: 1.7.1
+    From: 'Quality Open Software' (http://www.qos.ch)
+    License(s):
+        MIT (bundled/slf4j/LICENSE)
+
+Spring Framework (https://projects.spring.io/spring-framework/)
```
mike-jumper: Is this bundled? I don't see it as a dependency (even transitive), nor do I see anything springy in the resulting `.jar`.

extensions/guacamole-auth-cas/src/licenses/LICENSE
```diff
@@ -0,0 +1,294 @@
...
+---------------------------------------------------------------
+
+    Version: 3.0.5.RELEASE
+    From: 'Spring by Pivotal' (https://spring.io)
+    License(s):
+        Apache v2.0 (bundled/springframework/LICENSE-2.0.txt)
+
+XML Security (http://santuario.apache.org)
```
mike-jumper: What pulls this in? I don't see this in the dependency list produced by Maven either.

extensions/guacamole-auth-cas/src/licenses/LICENSE
```diff
@@ -0,0 +1,294 @@
...
+
+LOG4J (http://logging.apache.org/log4j/1.2/)
+--------------------------------------------
+
+    Version: 1.2.15
+    From: 'Apache Software Foundation' (https://www.apache.org)
+    License(s):
+        Apache v2.0 (bundled/log4j/LICENSE-2.0.txt)
+
+slf4j (http://www.slf4j.org)
```
mike-jumper: If this does end up bundled, we should exclude it. It will be provided at runtime by the Guacamole webapp, thus there is no need for the extension to include it, too:

https://github.com/apache/incubator-guacamole-client/blob/e8f4cd55155fef8e256b81b8047799eedcc009bd/guacamole/pom.xml#L266-L270

extensions/guacamole-auth-cas/src/licenses/LICENSE
```diff
@@ -0,0 +1,294 @@
...
+
+JUnit (http://junit.org)
+------------------------
+
+    Version: 4.8.2
+    From: 'JUnit' (http://junit.org)
+    License(s):
+        CPL v1.0 (bundled/junit/LICENSE-1.0.txt)
+
+LOG4J (http://logging.apache.org/log4j/1.2/)
```
mike-jumper: Are you sure? As with some of the other dependencies, I don't see this bundled nor listed by Maven.

extensions/guacamole-auth-cas/src/licenses/NOTICE
```diff
@@ -0,0 +1,5 @@
+Apache Guacamole
+Copyright 2016 The Apache Software Foundation
```
mike-jumper: 2017 ;)

extensions/guacamole-auth-cas/src/main/resources/guac-manifest.json
```diff
@@ -0,0 +1,19 @@
+{
+
+    "guacamoleVersion" : "0.9.11-incubating",
```
mike-jumper: Correct at the time this PR was originally opened, but needs to be `0.9.12-incubating` now.

extensions/guacamole-auth-cas/pom.xml
```diff
@@ -0,0 +1,175 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0
+                        http://maven.apache.org/maven-v4_0_0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>org.apache.guacamole</groupId>
+    <artifactId>guacamole-auth-cas</artifactId>
+    <packaging>jar</packaging>
+    <version>0.9.11-incubating</version>
```
mike-jumper: Now that 0.9.12-incubating is out, this needs to be updated to 0.9.12-incubating.

extensions/guacamole-auth-cas/pom.xml
```diff
@@ -0,0 +1,175 @@
...
+            <artifactId>guacamole-common</artifactId>
+            <version>0.9.10-incubating</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <!-- Guacamole Extension API -->
+        <dependency>
+            <groupId>org.apache.guacamole</groupId>
+            <artifactId>guacamole-ext</artifactId>
+            <version>0.9.11-incubating</version>
```
mike-jumper: 0.9.12-incubating

extensions/guacamole-auth-cas/src/main/resources/guac-manifest.json
necouchman: Updated.

extensions/guacamole-auth-cas/pom.xml
necouchman: Updated.

extensions/guacamole-auth-cas/pom.xml
necouchman: Updated.  Do I need to rebase on the 0.9.12-incubating code?  As it is this extension won't build because it'll require the updated dependency, which won't be there unless I rebase.

extensions/guacamole-auth-cas/src/licenses/NOTICE
necouchman: Updated

extensions/guacamole-auth-cas/src/licenses/LICENSE
necouchman: It's a dependency listed in the cas-client POM.  This is one of the reasons I was asking how deep I needed to go.  I'll pull it out.

extensions/guacamole-auth-cas/src/licenses/LICENSE
necouchman: Also listed in cas-client POM, but it has "compile" as the scope, so probably only needed for compile-time.  Removed.

extensions/guacamole-auth-cas/src/licenses/LICENSE
necouchman: Removed.

extensions/guacamole-auth-cas/src/licenses/LICENSE
necouchman: Removed.

extensions/guacamole-auth-cas/src/licenses/LICENSE
necouchman: It's an optional dependency listed in the cas-client-core POM.  So, guacamole-auth-cas -> cas-client -> cas-client-core -> commons-codec.  I'll remove it.

extensions/guacamole-auth-cas/src/licenses/LICENSE
necouchman: Another one listed for cas-client in its POM file.  Removed.

extensions/guacamole-auth-cas/src/licenses/LICENSE
necouchman: Listed as optional runtime dependency in the cas-client-core POM.

necouchman:
Okay, cleaned up licensing stuff, updated years, and updated versions.  Let me know if you come up with something for the JavaScript issue or have a direction you want me to run.  Also let me know if I should rebase the entire extesion on the 0.9.12-incubating code base.

mike-jumper:
I think I finally have a clean solution to the Angular parameter issues. I've written a script which reformats the URL when the page is loaded if query parameters are present in a location where Angular will not see them:

mike-jumper/incubator-guacamole-client@be2a40657448a93033f319e7ba9389551454bd5e

Let me know if the above solves the problem for you, and I'll open a pull request for that improvement.

> Also let me know if I should rebase the entire extesion on the 0.9.12-incubating code base.

Yes, definitely. Things won't build properly otherwise.

necouchman:
Really sorry for the delay in getting back to this - my day job has been really busy these days and haven't had much of a chance to work on this.

With the change, above, the CAS authentication extension works fine and redirects correctly, so I think you can go ahead and do a pull request for that.

Now I have to dig into the corner of my brain where the rebase knowledge went and rebase this extension on the current version...I think master is up to 0.9.13, now??

necouchman:
Okay, extension has been rebased to current master, and I've removed the extension-specific URL parameter rewriting in anticipation of your commit going through.  Once you do push that through I'll rebase it once more with that commit, and this issue should be taken care of.

mike-jumper:
> Really sorry for the delay in getting back to this - my day job has been really busy these days and haven't had much of a chance to work on this.

No worries.

> With the change, above, the CAS authentication extension works fine and redirects correctly, so I think you can go ahead and do a pull request for that.

Excellent - will do.

> Now I have to dig into the corner of my brain where the rebase knowledge went and rebase this extension on the current version...I think master is up to 0.9.13, now??

The version number on master is still 0.9.12-incubating, though we should probably finalize scope and bump versions soon.

necouchman:
This has been rebased to the master with the recent commit for URL parameters.  Any other concerns for this PR?  Anything else that needs work before it gets committed?

mike-jumper:
Looks good to me! Brace for merge...

zhangjianpinghik:
the redirect_url is in the guacamole config file, but it is not convenient, the best way is that always auto redirect to the url which user access before, i hope you can improve that.

necouchman:
> the redirect_url is in the guacamole config file, but it is not convenient, the best way is that always auto redirect to the url which user access before, i hope you can improve that.

Yes, I know that it is sub-optimal - I need to do some testing to see if there's any way to accomplish this without explicitly specifying the redirect URL.  When I get a chance, I'll try some things out.  I don't know when that will be.

mike-jumper:
If the redirect URL is not explicitly specified by the administrator, and is instead detected client-side or through inspecting the HTTP request, would that not open us up to redirect-based attacks, potentially allowing malicious sites to steal the CAS token?

https://www.owasp.org/index.php/Unvalidated_Redirects_and_Forwards_Cheat_Sheet

necouchman:
> If the redirect URL is not explicitly specified by the administrator, and is instead detected client-side or through inspecting the HTTP request, would that not open us up to redirect-based attacks, potentially allowing malicious sites to steal the CAS token?

Yeah, definitely need to carefully consider that and avoid introducing security holes...

Looking at that page, however, it looks like most of the examples of "dangerous" redirects are ones based on things like GET parameters and the like - things that could be easily spoofed by altering the URL or accepting user input.  It seems like if we calculate the URL of the Guacamole application, from within the Java code (inside the servlet), it should avoid most, if not all, of the pitfalls listed on that page, correct?

Security issues aside, I've run into problems actually coming up with the correct URL for the web application, particularly when Tomcat is behind a reverse proxy, like Apache or Nginx, so I need to make sure I can get that to work reliably in all cases.

mike-jumper:
> It seems like if we calculate the URL of the Guacamole application, from within the Java code (inside the servlet), it should avoid most, if not all, of the pitfalls listed on that page, correct?

Yes - if the default value for the redirect URI is calculated purely from server-side values (not the HTTP request, nor generated client-side), I think this should be safe.

As you say, though, it's likely impossible to determine the correct value for deployments of guac behind reverse proxies.

saydulk:
Hello how i can get Guacamole working with CAS for SAML authentication?

necouchman:
@saydulk Three things:
1. This is not a support forum.  If you have questions or need help, please use the mailing list: http://guacamole.apache.org/support/
2. If you're using CAS, why not use CAS natively instead of SAML?  There's already a CAS authentication module: http://guacamole.apache.org/doc/gug/cas-auth.html
3. This PR is still a work-in-progress - it has not been completed/approved/merged, so SAML support is not present, yet, in Guacamole.

PR #125, opened 2017-02-28
GUACAMOLE-225: Bump version numbers to 0.9.12-incubating (guacamole-common unchanged).

mike-jumper:


mike-jumper:
Chose wrong merge base initially, but should be fine now.

**NOTE:** Intended for `staging/0.9.12-incubating`.

necouchman:
Sorry if I missed this conversation at some point in the past, but why is guacamole-common not changing with the rest of it?  I noticed it is still on 0.9.10 while everything else was on 0.9.11 - any particular reason it isn't advancing with everything else to 0.9.12?

extensions/guacamole-auth-duo/pom.xml
```diff
@@ -209,19 +209,11 @@
 
     <dependencies>
 
-        <!-- Guacamole Java API -->
-        <dependency>
```
jmuehlner: Why are these dependency blocks being removed?

mike-jumper: Similar to apache/incubator-guacamole-manual#23, I removed the explicit dependencies on guacamole-common to rely instead on the transitive dependency pulled via guacamole-ext.

Think that's a bad idea/practice?

mike-jumper:
> Sorry if I missed this conversation at some point in the past, but why is guacamole-common not changing with the rest of it?

Because guacamole-common has not changed (see title). Can confirm this via git:

    $ git fetch upstream
    remote: Counting objects: 115, done.
    remote: Compressing objects: 100% (56/56), done.
    remote: Total 77 (delta 44), reused 11 (delta 4)
    Unpacking objects: 100% (77/77), done.
    From https://git-wip-us.apache.org/repos/asf/incubator-guacamole-client
       6a2235e..f887730  master     -> upstream/master
    $ git diff 0.9.11-incubating..upstream/staging/0.9.12-incubating guacamole-common
    $

extensions/guacamole-auth-duo/pom.xml
jmuehlner: No, makes sense to me - it's just good to have some information on why these are being removed in a PR that's about bumping version numbers. 😃

mike-jumper: Ah, OK.

I removed them as a part of the version number bump because of the need to check whether those numbers get bumped as well in each case. Normally, I'm an extreme proponent of "include what you use", but guacamole-common in this case is not independently used by the code; it's only used as required to interact with parts of guacamole-ext.

There's some discussion on this here: http://stackoverflow.com/questions/4226756/maven-should-i-keep-or-remove-declared-dependencies-that-are-also-transitives

IMHO: it's not a logical dependency of these extensions.

PR #126, opened 2017-03-02
GUACAMOLE-229: Remove intervals properly when XHR is polling.

teschnei:
https://issues.apache.org/jira/browse/GUACAMOLE-229

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Tunnel.js
+++ b/guacamole-common-js/src/main/webapp/modules/Tunnel.js
@@ -393,7 +393,7 @@ Guacamole.HTTPTunnel = function(tunnelURL, crossDomain) {
                 if (pollingMode === POLLING_ENABLED) {
                     if (xmlhttprequest.readyState === 3 && !interval)
                         interval = setInterval(parseResponse, 30);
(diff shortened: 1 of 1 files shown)


mike-jumper:
Thanks, @teschnei.

The changes here look fine code-wise, but can you clarify the commit message a bit? Just saying "fixed [thing which was apparently broken]" is too vague. Reading the message, it's not clear what high level issue is being fixed.

If you could also format the message such that it doesn't stand out next to the rest of the history (ie: starts with a capital letter, ends with a period), that would be nice. Looking at the history for this file, for example, you see:

    GUACAMOLE-229: fixed clearing intervals when polling
    GUACAMOLE-44: Expose tunnel UUID to JavaScript. Document allowed internal use of the empty opcode.
    GUACAMOLE-1: Relicense C and JavaScript files.
    GUAC-791: Fix type expressions to comply with JSDoc 3.
    GUAC-1430: Do not move on to the next chained tunnel on upstream timeout.
    ...

The style of the GUACAMOLE-229 message is completely different from the rest

teschnei:
Certainly! I will amend it when I am next able to.

teschnei:
Amended

PR #127, opened 2017-03-07
GUACAMOLE-230: Provide oncursor event handler with canvas having exact dimensions

mike-jumper:
Guacamole implements its layers using a `canvas` wrapped in a `div`, with that outer `div` providing the clipping area. There is thus no internal requirement that the `canvas` itself be the same dimensions as the layer it belongs to, and recent changes ([GUACAMOLE-187](https://issues.apache.org/jira/browse/GUACAMOLE-187) via #113) leveraged this to reduce the number of expensive resize operations attempted when a layer is being continuously resized.

Unfortunately, as described in [GUACAMOLE-230](https://issues.apache.org/jira/browse/GUACAMOLE-230):

> The mouse cursor renders extremely small when set via CSS, at least on Chrome. This seems to be due to layer dimensions now rounding up to the nearest multiple of 64. On high-resolution devices, Chrome assumes that 64x64 cursor image must also be high-resolution, and uses a pixel size inappropriate for the connection. This does not occur for smaller cursor images.

This change adds a new `toCanvas()` function to `Guacamole.Layer` which produces an entirely new canvas representing the contents of the layer. Unlike `getCanvas()`, which returns the internal `canvas` backing the layer itself, `toCanvas()` is guaranteed to return an entirely independent `canvas` which has the same dimensions as the layer. By using `toCanvas()` rather than `getCanvas()` when invoking the `oncursor` handler, cursors now work as they did previously.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Display.js
+++ b/guacamole-common-js/src/main/webapp/modules/Display.js
@@ -423,7 +423,7 @@ Guacamole.Display = function() {
 
             // Fire cursor change event
             if (guac_display.oncursor)
-                guac_display.oncursor(cursor.getCanvas(), hotspotX, hotspotY);
+                guac_display.oncursor(cursor.toCanvas(), hotspotX, hotspotY);
 
         });
     };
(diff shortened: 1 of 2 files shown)


mike-jumper:
**NOTE:** Intended for `staging/0.9.12-incubating`. This is a fix for a regression introduced by changes made for the 0.9.12-incubating release.

jmuehlner:
Ok, this makes sense to me.

PR #128, opened 2017-03-11
GUACAMOLE-231: Update software cursor location when a "mouse" instruction is received.

mike-jumper:
This change adds client support for the "mouse" instruction added via apache/incubator-guacamole-server#76.


mike-jumper:
**NOTE:** Intended for `staging/0.9.12-incubating` (regression due to [GUACAMOLE-117](https://issues.apache.org/jira/browse/GUACAMOLE-177)).

PR #129, opened 2017-03-17
GUACAMOLE-243: Ignore LDAPReferralExceptions thrown when iterating over LDAPSearchResults

alt36:


extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
```diff
@@ -101,21 +102,24 @@ private void putAllUsers(Map<String, User> users, LDAPConnection ldapConnection,
...
+                        logger.warn("Queried user is missing the username attribute \"{}\".", usernameAttribute);
+                        continue;
+                    }
+
+                    // Store user using their username as the identifier
+                    String identifier = username.getStringValue();
+                    if (users.put(identifier, new SimpleUser(identifier)) != null)
+                        logger.warn("Possibly ambiguous user account: \"{}\".", identifier);
+
+                } catch (LDAPReferralException e) {
```
mike-jumper: Please do not cuddle the `catch`. For more information, see the general style of other code in the codebase, or the overview of the style we use at http://guacamole.incubator.apache.org/guac-style/.

The most important thing here style-wise is consistency.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
```diff
@@ -259,8 +263,13 @@ private String generateLDAPQuery(String username)
 
             // Add all DNs for found users
             while (results.hasMore()) {
-                LDAPEntry entry = results.next();
-                userDNs.add(entry.getDN());
+                try {
+                    LDAPEntry entry = results.next();
+                    userDNs.add(entry.getDN());
+                } catch (LDAPReferralException e) {
```
mike-jumper: Same thing here - please do not cuddle the `catch`. See other code in the codebase for examples of style.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
```diff
@@ -101,21 +102,24 @@ private void putAllUsers(Map<String, User> users, LDAPConnection ldapConnection,
...
+                        continue;
+                    }
+
+                    // Store user using their username as the identifier
+                    String identifier = username.getStringValue();
+                    if (users.put(identifier, new SimpleUser(identifier)) != null)
+                        logger.warn("Possibly ambiguous user account: \"{}\".", identifier);
+
+                } catch (LDAPReferralException e) {
+                    logger.debug("Ignoring LDAP Referral: \"{}\".", e.toString());
```
mike-jumper: As mentioned in the comments on [GUACAMOLE-243](https://issues.apache.org/jira/browse/GUACAMOLE-243), ignoring all LDAP referrals is not really a solution unless there is an option to follow them.

Warning that an LDAP referral has occurred and isn't being followed is good, but:

1. Such a warning would naturally need to be at the "warning" log level.
2. The details of the exception should be logged at the "debug" level (ie: `logger.debug("Some contextual message about what just happened.", e)`, *not* `e.toString()` or there won't be any stacktraces at the debug level).

Normally, non-debug log messages use `e.getMessage()` to pull some additional human-readable context while excluding the stacktrace, but if the `LDAPReferralException` will only ever occur for non-followed referrals, that may be just noise here.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
```diff
@@ -259,8 +263,13 @@ private String generateLDAPQuery(String username)
 
             // Add all DNs for found users
             while (results.hasMore()) {
-                LDAPEntry entry = results.next();
-                userDNs.add(entry.getDN());
+                try {
+                    LDAPEntry entry = results.next();
+                    userDNs.add(entry.getDN());
+                } catch (LDAPReferralException e) {
+                    logger.debug("Ignoring LDAP Referral: \"{}\".", e.toString());
```
mike-jumper: As above, probably not a good idea to just ignore all referrals.

alt36: Fair enough - my thought process was that at present, it is simply not possible to log in via LDAP if a referral is returned. This makes it impossible to use LDAP auth against an AD where the search base dn is the root of the AD, because at present the LDAPReferralException will be rethrown and not dealt with in any way. This was a v0.1 attempt to at least improve the situation a little. I don't disagree that including the capability to (probably optionally) follow the referral is a better solution, but I'm afraid I'll have to leave that to someone more familiar with the code, then.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
alt36: Thanks for the feedback - I honestly thought I was following what I'd seen elsewhere in the code, but evidently I wasn't paying close enough attention! I'll read the style guide more carefully next time :)

mike-jumper:
@alt36 - I'm OK with these changes providing the issues with the cuddly catches and the log message/levels are fixed. From the discussion on the JIRA issue, though we'll need to keep working on the issue before it can be considered complete, there's definitely no problem with properly logging the fact that referrals are being ignored, and simply ignoring them universally for the time being.

alt36:
@mike-jumper : e.getMessage() is rather unhelpful for the LDAPReferralException, returning just the string "Referral" with no further info. So I've gone for a warning message with a little more context about where the exception is being generated from in this version.

mike-jumper:
@alt36 - looks good, but by convention there still need to be `logger.debug()` calls for the caught exceptions (so that stack traces are logged when debug-level logging is enabled). From the relevant earlier review comment:

> 2. The details of the exception should be logged at the "debug" level (ie: logger.debug("Some contextual message about what just happened.", e), ...

Example of this in practice:

https://github.com/apache/incubator-guacamole-client/blob/f8877304e0950653ca2b5e5abf4b475fce809796/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java#L198-L199

The idea is that the default logging behavior should provide enough information for the administrator to troubleshoot issues and obtain context for failures, with additional low-level information around those failures (including stack traces) available at the debug level.

necouchman:
Also worth noting for the `logger.debug()` section that the LDAPReferralException class has a getFailedReferral() method that should return the referral that it's choking on.  This might be useful to throw into the debug information - less for the case where you want to ignore referrals entirely, and more for situations where you expect referrals to be followed but it fails on one.

mike-jumper:
> Also worth noting for the `logger.debug()` section that the LDAPReferralException class has a getFailedReferral() method that should return the referral that it's choking on. This might be useful to throw into the debug information

FYI - I just checked the SLF4J documentation regarding logging of exceptions, since the `Logger` class only provides explicit functions for logging exceptions which do not handle parameterization with `{}` like the others (they take only a `String` and a `Throwable`). Apparently, doing what you suggest is indeed possible: https://www.slf4j.org/faq.html#paramException

The `Throwable` just needs to be the last object in the parameter list.

That said, if `getFailedReferral()` returns information relevant to the LDAP directory, that may be worth logging at the warning level with the original message. The debug level is mainly aimed at assisting those debugging the guac code itself.

necouchman:
Ping.
@alt36: Adam, any progress here?

necouchman:
@alt36 Ping.

necouchman:
@alt36 Ping...

necouchman:
@alt36: Any progress, Adam?

mike-jumper:
It looks like this PR has bit the dust.

For reference, @necouchman has resumed general development of [GUACAMOLE-243](https://issues.apache.org/jira/browse/GUACAMOLE-243) with a different approach via #202, which has been included in scope of the upcoming 0.9.14-incubating release. Once that PR is merged, this one should be automatically closed.

@alt36, if you're around, please feel free to manually close this PR.

alt36:
Thanks @necouchman - I've not had the time to revisit this PR, and I much prefer your solution to the approach I was taking!

PR #131, opened 2017-03-18
GUACAMOLE-244: Support configuration of alias dereferencing

necouchman:
This adds the necessary code to allow the user to configure how the LDAP extension behaves with regard to dereferencing aliases.

### DIFF
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
@@ -241,17 +241,23 @@ public class ConfigurationService {
             "never"
         );
 
-        if (derefAliases == "always")
+        if (derefAliases.equals("always"))
             return 3;
 
-        else if (derefAliases == "finding")
+        else if (derefAliases.equals("finding"))
             return 2;
 
-        else if (derefAliases == "searching")
+        else if (derefAliases.equals("searching"))
             return 1;
 
-        else
+        else if (derefAliases.equals("never"))
             return 0;
+        
+        else {
+            logger.error("Invalid value given for ldap-dereference-aliases.");
+            logger.debug("Received {} but expected one of the following: always, finding, searching, never.", derefAliases);
+            throw new GuacamoleException("Invalid valid for ldap-dereference-aliases.");
+        }
 
     }
 
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
@@ -20,15 +20,23 @@
 package org.apache.guacamole.auth.ldap;
 
 import com.google.inject.Inject;
+import com.novell.ldap.LDAPSearchConstraints;
 import java.util.Collections;
 import java.util.List;
 import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.environment.Environment;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Service for retrieving configuration information regarding the LDAP server.
  */
 public class ConfigurationService {
+
+    /**
+     * Logger for this class.
+     */
+    private final Logger logger = LoggerFactory.getLogger(ConfigurationService.class);
 
     /**
      * The Guacamole server environment.
@@ -261,6 +269,32 @@ public class ConfigurationService {
             logger.debug("Received {} but expected one of the following: always, finding, searching, never.", derefAliases);
             throw new GuacamoleException("Invalid valid for ldap-dereference-aliases.");
         }
+
+    }
+
+    /**
+     * Returns a set of LDAPSearchConstraints to apply globally
+     * to all LDAP searches rather than having various instances
+     * dispersed throughout the code.  Currently contains the
+     * maximum number of LDAP results to return in a search, as
+     * well as whether or not aliases should be dereferenced
+     * during LDAP operations.
+     *
+     * @return
+     *     A LDAPSearchConstraints object containing constraints
+     *     to be applied to all LDAP search operations.
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed.
+     */
+    public LDAPSearchConstraints getLDAPSearchConstraints() throws GuacamoleException {
+
+        LDAPSearchConstraints constraints = new LDAPSearchConstraints();
+
+        constraints.setMaxResults(getMaxResults());
+        constraints.setDereference(getDereferenceAliases());
+
+        return constraints;
 
     }
 
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
@@ -223,4 +223,36 @@ public class ConfigurationService {
         );
     }
 
+    /**
+     * Returns whether or not LDAP aliases will be dereferenced,
+     * as configured with guacamole.properties.
+     * By default they will never be dereferenced.
+     *
+     * @return
+     *     An integer representing the status of of alias
+     *     dereferencing, as configured in guacamole.properties.
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed.
+     */
+    public int getDereferenceAliases() throws GuacamoleException {
+        String derefAliases = environment.getProperty(
+            LDAPGuacamoleProperties.LDAP_DEREFERENCE_ALIASES,
+            "never"
+        );
+
+        if (derefAliases == "always")
+            return 3;
+
+        else if (derefAliases == "finding")
+            return 2;
+
+        else if (derefAliases == "searching")
+            return 1;
+
+        else
+            return 0;
+
+    }
+
 }
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
@@ -224,7 +224,7 @@ public class ConfigurationService {
      * @throws GuacamoleException
      *     If guacamole.properties cannot be parsed.
      */
-    public int getMaxResults() throws GuacamoleException {
+    private int getMaxResults() throws GuacamoleException {
         return environment.getProperty(
             LDAPGuacamoleProperties.LDAP_MAX_SEARCH_RESULTS,
             1000 
@@ -234,7 +234,7 @@ public class ConfigurationService {
     /**
      * Returns whether or not LDAP aliases will be dereferenced,
      * as configured with guacamole.properties.  The default
-     * behavior if not explicityly defined is to never 
+     * behavior if not explicitly defined is to never 
      * dereference them.
      *
      * @return
@@ -244,21 +244,17 @@ public class ConfigurationService {
      * @throws GuacamoleException
      *     If guacamole.properties cannot be parsed.
      */
-    public DereferenceAliases getDereferenceAliases() throws GuacamoleException {
+    private DereferenceAliasesMode getDereferenceAliases() throws GuacamoleException {
         return environment.getProperty(
             LDAPGuacamoleProperties.LDAP_DEREFERENCE_ALIASES,
-            DereferenceAliases.NEVER
+            DereferenceAliasesMode.NEVER
         );
 
     }
 
     /**
      * Returns a set of LDAPSearchConstraints to apply globally
-     * to all LDAP searches rather than having various instances
-     * dispersed throughout the code.  Currently contains the
-     * maximum number of LDAP results to return in a search, as
-     * well as whether or not aliases should be dereferenced
-     * during LDAP operations.
+     * to all LDAP searches.
      *
      * @return
      *     A LDAPSearchConstraints object containing constraints
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
@@ -225,12 +225,15 @@ public class ConfigurationService {
 
     /**
      * Returns whether or not LDAP aliases will be dereferenced,
-     * as configured with guacamole.properties.
-     * By default they will never be dereferenced.
+     * as configured with guacamole.properties.  The default
+     * behavior if not explicityly defined is to never 
+     * dereference them.
      *
      * @return
-     *     An integer representing the status of of alias
-     *     dereferencing, as configured in guacamole.properties.
+     *     An integer value that maps to the JLDAP constants
+     *     for dereferencing - 0 is DEREF_NEVER, 1 is DEREF_SEARCHING,
+     *     2 is DEREF_FINDING, and 3 is DEREF_ALWAYS - as configured
+     *     in guacamole.properties.
      *
      * @throws GuacamoleException
      *     If guacamole.properties cannot be parsed.
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
@@ -233,8 +233,8 @@ public class ConfigurationService {
 
     /**
      * Returns whether or not LDAP aliases will be dereferenced,
-     * as configured with guacamole.properties.  The default
-     * behavior if not explicitly defined is to never 
+     * as configured with guacamole.properties. The default
+     * behavior if not explicitly defined is to never
      * dereference them.
      *
      * @return
@@ -249,7 +249,6 @@ public class ConfigurationService {
             LDAPGuacamoleProperties.LDAP_DEREFERENCE_ALIASES,
             DereferenceAliasesMode.NEVER
         );
-
     }
 
     /**
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
@@ -238,37 +238,17 @@ public class ConfigurationService {
      * dereference them.
      *
      * @return
-     *     An integer value that maps to the JLDAP constants
-     *     for dereferencing - 0 is DEREF_NEVER, 1 is DEREF_SEARCHING,
-     *     2 is DEREF_FINDING, and 3 is DEREF_ALWAYS - as configured
-     *     in guacamole.properties.
+     *     The behavior for handling dereferencing of aliases
+     *     as configured in guacamole.properties.
      *
      * @throws GuacamoleException
      *     If guacamole.properties cannot be parsed.
      */
-    public int getDereferenceAliases() throws GuacamoleException {
-        String derefAliases = environment.getProperty(
+    public DereferenceAliases getDereferenceAliases() throws GuacamoleException {
+        return environment.getProperty(
             LDAPGuacamoleProperties.LDAP_DEREFERENCE_ALIASES,
-            "never"
+            DereferenceAliases.NEVER
         );
-
-        if (derefAliases.equals("always"))
-            return 3;
-
-        else if (derefAliases.equals("finding"))
-            return 2;
-
-        else if (derefAliases.equals("searching"))
-            return 1;
-
-        else if (derefAliases.equals("never"))
-            return 0;
-        
-        else {
-            logger.error("Invalid value given for ldap-dereference-aliases.");
-            logger.debug("Received {} but expected one of the following: always, finding, searching, never.", derefAliases);
-            throw new GuacamoleException("Invalid valid for ldap-dereference-aliases.");
-        }
 
     }
 
@@ -292,7 +272,7 @@ public class ConfigurationService {
         LDAPSearchConstraints constraints = new LDAPSearchConstraints();
 
         constraints.setMaxResults(getMaxResults());
-        constraints.setDereference(getDereferenceAliases());
+        constraints.setDereference(getDereferenceAliases().DEREF_VALUE);
 
         return constraints;
 
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/DereferenceAliases.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/DereferenceAliases.java
@@ -1,0 +1,71 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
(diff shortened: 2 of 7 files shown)


extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
```diff
@@ -223,4 +223,36 @@ public int getMaxResults() throws GuacamoleException {
...
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed.
+     */
+    public int getDereferenceAliases() throws GuacamoleException {
+        String derefAliases = environment.getProperty(
+            LDAPGuacamoleProperties.LDAP_DEREFERENCE_ALIASES,
+            "never"
+        );
+
+        if (derefAliases == "always")
```
mike-jumper: Java does not do string comparisons in this way. You need to use `.equals()`.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
```diff
@@ -223,4 +223,36 @@ public int getMaxResults() throws GuacamoleException {
...
+     *     An integer representing the status of of alias
+     *     dereferencing, as configured in guacamole.properties.
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed.
+     */
+    public int getDereferenceAliases() throws GuacamoleException {
+        String derefAliases = environment.getProperty(
+            LDAPGuacamoleProperties.LDAP_DEREFERENCE_ALIASES,
+            "never"
```
mike-jumper: Perhaps it would be better to implement a `GuacamoleProperty` subclass which actually (and strictly) parses these values? The list of possible values here is begging for an `enum`.

I see `never` is never explicitly handled below, and any mistake in the spelling of the other values will result in the property silently assuming you meant `never`, rather than throwing an explicit exception.

It would be better if incorrect values result in an exception (which will ultimately be logged).

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
```diff
@@ -153,4 +153,14 @@ private LDAPGuacamoleProperties() {}
 
     };
 
+    /**
+     * The behavior of alias dereferncing for the LDAP connections.
```
mike-jumper: dereferencing*

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
```diff
@@ -223,4 +223,36 @@ public int getMaxResults() throws GuacamoleException {
...
+        );
+
+        if (derefAliases == "always")
+            return 3;
```
mike-jumper: [`LDAPSearchConstraints`](https://www.novell.com/documentation/developer/jldap/jldapenu/api/com/novell/ldap/LDAPSearchConstraints.html) defines constants for these values (`DEREF_ALWAYS`, `DEREF_FINDING`, ...). We should be using those, rather than hard-coding their low-level integer values.

That said, I'm not sure this is really a good way to achieve this. We're essentially marrying Guacamole's configuration to the internals of JLDAP, which definitely violates separation of concerns. If the internals of JLDAP happened to just make that much sense, then OK, but "always", "finding", "searching", and "never" don't really make intuitive sense.

If `ConfigurationService` will ultimately return the integer that `LDAPSearchConstraints` requires, that's OK - it's internal to the extension, and `ConfigurationService` itself separates that concern from that of the configuration, but you don't want to violate that separation by essentially mapping internals directly to `guacamole.properties`. Ideally, you'd determine a sensible set of values based on the behaviors that the LDAP auth should support, and then translate those into the values required by JLDAP.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
```diff
@@ -88,6 +88,7 @@ private void putAllUsers(Map<String, User> users, LDAPConnection ldapConnection,
             // Set search limits
             LDAPSearchConstraints constraints = new LDAPSearchConstraints();
             constraints.setMaxResults(confService.getMaxResults());
+            constraints.setDereference(confService.getDereferenceAliases());
```
mike-jumper: If we keep building a new `LDAPSearchConstraints` object, would it make more sense to provide a function which does so within `ConfigurationService`, rather than repeatedly calling the same setters on a new `LDAPSearchConstraints` prior to each LDAP query?

I'm also not clear on why `setMaxResults()` is only applied here. If it should apply universally, then I would lean even more strongly in the direction described above. We can abstract away the configuration properties which apply to LDAP constraints within the internals of `ConfigurationService` and just rely on the service to return an appropriate constraints object.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
```diff
@@ -223,4 +223,36 @@ public int getMaxResults() throws GuacamoleException {
         );
     }
 
+    /**
+     * Returns whether or not LDAP aliases will be dereferenced,
+     * as configured with guacamole.properties.
+     * By default they will never be dereferenced.
+     *
+     * @return
+     *     An integer representing the status of of alias
```
mike-jumper: "An integer representing the status of alias dereferencing" is not clear in itself. The possible values of that integer, and the semantics of each, are not documented. Just documenting the values is insufficient, though. More on this below, but this would be a perfect spot to use an `enum`, that way all values are very strictly defined and documented.

mike-jumper:
@necouchman, can you also explain what these changes are aimed at? I get that they're meant to allow "alias dereferencing" to be configured, but I'm vague on what that actually means for LDAP in practice.

The [GUACAMOLE-244](https://issues.apache.org/jira/browse/GUACAMOLE-244) issue in JIRA, titled "Allow Configuration of Alias Dereferencing", currently contains the following description:

> Allow Alias Dereferencing Behavior to be Configured by User

So ... any clarification would be helpful.

necouchman:
Sorry, one of the comments on JIRA should have some more detailed information.  Basically, in my past history of running LDAP servers, I've run into a couple of situations where the ability to configure how the LDAP client handled dereferencing of aliases was almost, if not, essential, to being able to use LDAP.

Consider an LDAP tree where the main user account section is ou=People,dc=example,dc=com, and a separate section specifically for users with VPN access is ou=VPN_Users,dc=example,dc=com.  The entries in the ou=VPN_Users section are aliases to the actual object in the ou=People section, allowing the tree to remain organized, but providing a separate section that can be queried specifically for users who need VPN access.

Here are the following situations:
- If you query the base of the tree, at dc=example,dc=com, alias dereferencing is enabled, you'll come away with duplicate objects - for example, cn=Nick Couchman,ou=VPN_Users,dc=example,dc=com will show up, as will cn=Nick Couchman,ou=People,dc=example,dc=com.  For some LDAP clients that base their naming on the cn= property, for example, this will cause errors because of the duplicate accounts.  Being able to turn off alias dereferencing in these situations is very useful, as it allows the tree to be organized and divided for certain applications but also searchable a the root level for others.
- Conversely, if you wanted, for example, to allow access to Guacamole for the folks who are aliased in the ou=VPN_Users area, and not have to move their accounts, you would need alias dereferencing to be enabled and then you'd point the base of your search at the ou=VPN_Users,dc=example,dc=com part of the LDAP tree, and it would appear as if those users were in that location, even though they are simply linked to the real objects.

Hopefully this helps - also see the first comment on the JIRA issue.  I was trying to keep the description a little more succinct.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
necouchman: I think that approach probably makes sense.  I think this setMaxResults was a recent-ish change to the code to allow searching on large LDAP trees and avoid whatever built-in default existed in the JLDAP code, but I could be wrong about that.  Anyway, probably makes sense to move the constraints to a more global area and apply them universally.  I'll see what kind of progress I can make on that...though any specific help/suggestions would be welcome.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
necouchman: I'll look at a couple of other implementations of LDAP clients and see what they do for this type of behavior.  The "finding" and "searching" terms are a little strange to me - I think in the past I've usually run across either two or three objects - usually on, off, and maybe something in between.  Four is a little odd...I'll have to see what the more general consensus is among LDAP clients.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
necouchman: Doh.

necouchman: Fixed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
necouchman: Yes, probably a good idea...will work on that.

I did fix the checking of the string value, in the meantime, such that it expects one of the four values and throws an error if it doesn't get what it expects.  Still would be more elegant and proper with an enum.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
necouchman: Beefed up the commentary some, hopefully it's clearer this way, though will probably end up changing with implementation of an enum type.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
necouchman: Trusty VIM spell checker failed me...oh, wait, nevermind.  Even if it was there, it wouldn't like dereferencing, anyway.  Fixed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
necouchman: Create getLDAPSearchConstraints in the ConfigurationService class which pulls in getMaxResults() and getDereferenceAliases() and sets up the constraints.  Changed code in UserService and ConnectionService to use this function from confService instead of setting up their own each time.

necouchman:
Hopefully I've taken care of must of the concerns - knocked out the easy stuff...on to the enum :-).

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
mike-jumper: The `derefAliases.equals()` comparisons will also throw a `NullPointerException` when `derefAliases` is `null`, which will be the case if that property is not specified.

In any case, this and more can be resolved once things are cleaned up and migrated to an `enum`. There's an example of this being done elsewhere in the LDAP auth, actually. Take a peek at [`EncryptionMethodProperty`](https://github.com/apache/incubator-guacamole-client/blob/1a621886c65d5e99ec44841da41d137000a1c9f4/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/EncryptionMethodProperty.java).

mike-jumper:
> Hopefully this helps - also see the first comment on the JIRA issue.

The comment on JIRA definitely helps. Thanks.

> I was trying to keep the description a little more succinct.

Succinctness is good, but the description of an issue should probably describe the issue. ;) I'll copy what you've written here, etc. into the issue description.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
necouchman: Okay, the latest commit takes a stab at an implementation of the new enum property.  I referenced the EncryptionMethod and EncryptionMethodProperty classes in doing the implementation.  It's slightly different since I'm mapping actual values for the JLDAP library instead of using encryption method to determine default port, but I think it'll do the trick.

Also, I looked up the OpenLDAP values for the DEREF setting in ldap.conf, and it matches JLDAP exactly, so I'm going to stick with the never/searching/finding/always values at this point in time, unless someone would like to suggest other options.  Looks like it's more standard than I have experienced - I usually just toggle between never and always.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/DereferenceAliases.java
```diff
@@ -0,0 +1,71 @@
...
+ * under the License.
+ */
+
+package org.apache.guacamole.auth.ldap;
+
+/**
+ * Acceptable values for configuring the dereferencing of aliases in
+ * talking to LDAP servers.
+ */
+public enum DereferenceAliases {
```
mike-jumper: The name of this class kind of throws me at first glance. It sounds like it represents a collection of aliases or all possible legal aliases. The convention that type names are nouns leads to an incorrect interpretation of what this actually is.

Perhaps something like `DereferenceAliasesMode` or `AliasDereferencingMode` or anything-else-that's-a-noun would be more clear?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/DereferenceAliases.java
```diff
@@ -0,0 +1,71 @@
...
+     * Aliases are dereferenced below the base object, but not to locate
+     * the base object itself.  So, if the base object is itself an alias
+     * the search will not complete.
+     */
+    SEARCHING(1),
+
+    /**
+     * Aliases are only dereferenced to locate the base object, but not
+     * after that.  So, a search against a base object that is an alias will
+     * find any subordinates of the real object the aliase references, but
```
mike-jumper: alias*

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
```diff
@@ -153,4 +153,14 @@ private LDAPGuacamoleProperties() {}
 
     };
 
+    /**
+     * The behavior of alias dereferencing for the LDAP connections.
```
mike-jumper: But what does this mean?

It's clear to me since I've read what you've written in JIRA, etc., but the comment doesn't tell me what changing this setting does. I would have to dig through other parts of the code to find out that "alias dereferencing" is actually a consideration for queries against LDAP, but the comment as written implies this aliases have something to do with the LDAP connection itself.

In your other PR for the documentation covering this new property (apache/incubator-guacamole-manual#38), you say:

> Controls whether or not the LDAP connection follows (dereferences) aliases as it searches the tree.

Something like that would be perfect. Descriptive, accurate, yet not overly verbose.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
```diff
@@ -223,4 +231,51 @@ public int getMaxResults() throws GuacamoleException {
...
+            LDAPGuacamoleProperties.LDAP_DEREFERENCE_ALIASES,
+            DereferenceAliases.NEVER
+        );
+
+    }
+
+    /**
+     * Returns a set of LDAPSearchConstraints to apply globally
+     * to all LDAP searches rather than having various instances
+     * dispersed throughout the code.  Currently contains the
```
mike-jumper: I would recommend, simply:

"Returns a set of LDAPSearchConstraints which should be applied to all LDAP searches."

That a function should be used "rather than [copying the internals of the function everywhere]" applies to all functions by definition, and goes without saying. There's no need to enumerate the bad practices we avoid by having functions.

That said ... if this function effectively replaces the `getDereferenceAliases()` and `getMaxResults()` functions, and it no longer makes sense to call those functions externally, perhaps those functions should be removed (or declared `private`)?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/DereferenceAliases.java
```diff
@@ -0,0 +1,71 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.guacamole.auth.ldap;
+
+/**
+ * Acceptable values for configuring the dereferencing of aliases in
+ * talking to LDAP servers.
```
necouchman: Hopefully that looks better.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/DereferenceAliases.java
necouchman: Chose DereferenceAliasesMode.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/DereferenceAliases.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+public enum DereferenceAliases {
+
+    /**
+     * Never dereference aliases.  This is the default.
+     */
+    NEVER(0),
```
necouchman: Replaced with the LDAPSearchConstraints constants.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/DereferenceAliases.java
necouchman: Fixed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/DereferenceAliases.java
```diff
@@ -0,0 +1,71 @@
...
+    FINDING(2),
+
+    /**
+     * Aliases will always be dereferenced, both to locate the base object
+     * and when handling results returned by the search.
+     */
+    ALWAYS(3);
+
+    /**
+     * The integer value that the enum represents, which is used in
```
necouchman: Reworded.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
```diff
@@ -223,4 +231,51 @@ public int getMaxResults() throws GuacamoleException {
         );
     }
 
+    /**
+     * Returns whether or not LDAP aliases will be dereferenced,
+     * as configured with guacamole.properties.  The default
+     * behavior if not explicityly defined is to never
```
necouchman: Fixed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
necouchman: Fixed comments, made getDereferenceAliases and getMaxResults private - I can remove them and just put the logic into the constraints function if that makes more sense.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
necouchman: Done.

mike-jumper:
Looks good to me. @jmuehlner - any comment?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/DereferenceAliasesMode.java
```diff
@@ -0,0 +1,74 @@
...
+import com.novell.ldap.LDAPSearchConstraints;
+
+/**
+ * Data type that handles acceptable values for configuring
+ * alias dereferencing behavior when querying LDAP servers.
+ */
+public enum DereferenceAliasesMode {
+
+    /**
+     * Never dereference aliases.  This is the default.
```
jmuehlner: This comment and the 3 following ones all have unnecessary extra spaces.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
```diff
@@ -216,11 +224,54 @@ public EncryptionMethod getEncryptionMethod() throws GuacamoleException {
...
+    private int getMaxResults() throws GuacamoleException {
         return environment.getProperty(
             LDAPGuacamoleProperties.LDAP_MAX_SEARCH_RESULTS,
             1000 
         );
     }
 
+    /**
+     * Returns whether or not LDAP aliases will be dereferenced,
+     * as configured with guacamole.properties.  The default
```
jmuehlner: Extra space here too.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/DereferenceAliasesProperty.java
```diff
@@ -0,0 +1,61 @@
...
+package org.apache.guacamole.auth.ldap;
+
+import org.apache.guacamole.GuacamoleException;
+import org.apache.guacamole.GuacamoleServerException;
+import org.apache.guacamole.properties.GuacamoleProperty;
+
+/**
+ * A GuacamoleProperty with a value of DereferenceAliases. The possible strings
+ * "never", "searching", "finding", and "always" are mapped to their values as a
+ * DereferenceAliases enum.  Anything else results in a parse error.
```
jmuehlner: Extra space here too.

jmuehlner:
Aside from the tiniest nitpick about extra spaces in comments, everything looks good to me!

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/DereferenceAliasesMode.java
necouchman: I'm guessing you were referring to spaces at the end of the sentences?  I've reduced those all to one - once upon a time it was considered correct to have two spaces, and I guess I revert to that every now and then, at the cost of being consistent about it.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
necouchman: Removed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/DereferenceAliasesProperty.java
necouchman: Removed.

mike-jumper:
Cool, then. Moving ahead with the merge.

PR #132, opened 2017-03-21
GUACAMOLE-101: Allow Filtering of LDAP Users

necouchman:
Took a stab at implementing two LDAP filter configuration properties for user and connection searches.

This was a very quick stab at implementing it - there are a couple of things that might be worth adding:
- Error checking on the LDAP query - not sure if there's an easy way to parse out the filters to make sure that they're valid before actually running the searches??  The LDAPException should handle this - there are specific messages for filter errors.
- The JIRA issue mentions implementing multiple search bases - this does not do anything with that.

### DIFF
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
@@ -275,12 +275,12 @@ public class ConfigurationService {
     /**
      * Returns the search filter that should be used when querying the
      * LDAP server for Guacamole users.  If no filter is specified,
-     * a default of objectClass=* is returned.
+     * a default of "(objectClass=*)" is returned.
      *
      * @return
      *     The search filter that should be used when querying the
      *     LDAP server for users that are valid in Guacamole, or
-     *     objectClass=* if not specified.
+     *     "(objectClass=*)" if not specified.
      *
      * @throws GuacamoleException
      *     If guacamole.properties cannot be parsed.
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
@@ -295,12 +295,12 @@ public class ConfigurationService {
     /**
      * Returns the search filter that should be used when querying the 
      * LDAP server for Guacamole connections.  If no filter is specified,
-     * null is returned.
+     * the default of objectClass=guacConfigGroup is returned.
      * 
      * @return
      *     The search filter that should be used when querying the 
      *     LDAP server for connections for Guacamole, or 
-     *     null if no filter is specified. 
+     *     objectClass=guacConfigGroup if no filter is specified. 
      *
      * @throws GuacamoleException
      *     If guacamole.properties cannot be parsed.
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
@@ -292,24 +292,4 @@ public class ConfigurationService {
         );
     }
 
-    /**
-     * Returns the search filter that should be used when querying the 
-     * LDAP server for Guacamole connections.  If no filter is specified,
-     * the default of objectClass=guacConfigGroup is returned.
-     * 
-     * @return
-     *     The search filter that should be used when querying the 
-     *     LDAP server for connections for Guacamole, or 
-     *     objectClass=guacConfigGroup if no filter is specified. 
-     *
-     * @throws GuacamoleException
-     *     If guacamole.properties cannot be parsed.
-     */
-    public String getConnectionSearchFilter() throws GuacamoleException {
-        return environment.getProperty(
-            LDAPGuacamoleProperties.LDAP_CONNECTION_SEARCH_FILTER,
-            "(objectClass=guacConfigGroup)"
-        );
-    }
-
 }
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
@@ -270,7 +270,46 @@ public class ConfigurationService {
         constraints.setDereference(getDereferenceAliases().DEREF_VALUE);
 
         return constraints;
+    }
 
+    /**
+     * Returns the search filter that should be used when querying the
+     * LDAP server for Guacamole users.  If no filter is specified,
+     * a default of objectClass=* is returned.
+     *
+     * @return
+     *     The search filter that should be used when querying the
+     *     LDAP server for users that are valid in Guacamole, or
+     *     objectClass=* if not specified.
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed.
+     */
+    public String getUserSearchFilter() throws GuacamoleException {
+        return environment.getProperty(
+            LDAPGuacamoleProperties.LDAP_USER_SEARCH_FILTER,
+            "(objectClass=*)"
+        );
+    }
+
+    /**
+     * Returns the search filter that should be used when querying the 
+     * LDAP server for Guacamole connections.  If no filter is specified,
+     * null is returned.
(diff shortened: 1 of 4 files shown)


jmuehlner:
Hey @necouchman, looks like you've got some conflicts arising from your previous change. :)

necouchman:
So, I think it should be good, now.  I don't know if I did this in the proper way, but I rebased my branch from the current apache master, which merged in the changes from GUACAMOLE-244, and then resolved those conflicts.  It currently says no conflicts, so I think it succeeded, but let me know if there's a preferred method of handling these situations.

necouchman:
Any further comments/changes on this?

mike-jumper:
Hey @necouchman - I'll take a quick look.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
```diff
@@ -270,7 +270,46 @@ public LDAPSearchConstraints getLDAPSearchConstraints() throws GuacamoleExceptio
...
+     *     LDAP server for connections for Guacamole, or 
+     *     objectClass=guacConfigGroup if no filter is specified. 
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed.
+     */
+    public String getConnectionSearchFilter() throws GuacamoleException {
+        return environment.getProperty(
+            LDAPGuacamoleProperties.LDAP_CONNECTION_SEARCH_FILTER,
+            "(objectClass=guacConfigGroup)"
```
mike-jumper: If we're going to allow connection filters to be defined, `(objectClass=guacConfigGroup)` should probably be of the filter in all cases. Unlike the objects returned by the user search, the connection search really must be strictly `guacConfigGroup` objects as defined by the schema modifications included with the LDAP extension.

Since connections returned by the search using this filter will already be limited by membership (users must be explicit members of the group to see the connection), this probably simply isn't necessary and can be simply removed.

Given that the summary of [GUACAMOLE-101](https://issues.apache.org/jira/browse/GUACAMOLE-101) is "Allow arbitrary filtering of LDAP users", it is probably also out of scope.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
```diff
@@ -270,7 +270,46 @@ public LDAPSearchConstraints getLDAPSearchConstraints() throws GuacamoleExceptio
         constraints.setDereference(getDereferenceAliases().DEREF_VALUE);
 
         return constraints;
+    }
+
+    /**
+     * Returns the search filter that should be used when querying the
+     * LDAP server for Guacamole users.  If no filter is specified,
+     * a default of objectClass=* is returned.
```
mike-jumper: a default of "(objectClass=*)" is returned.

The parenthesis are a critical part of LDAP filter syntax.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
```diff
@@ -270,7 +270,46 @@ public LDAPSearchConstraints getLDAPSearchConstraints() throws GuacamoleExceptio
...
+     *
+     * @return
+     *     The search filter that should be used when querying the
+     *     LDAP server for users that are valid in Guacamole, or
+     *     objectClass=* if not specified.
```
mike-jumper: Same here - should be "(objectClass=*)".

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
necouchman: Fixed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
necouchman: Fixed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
necouchman: Removed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
```diff
@@ -85,11 +85,20 @@ private void putAllUsers(Map<String, User> users, LDAPConnection ldapConnection,
 
         try {
 
+            // Build a filter using the configured or default user search filter
+            // to find all user objects in the LDAP tree
+            StringBuilder userSearchFilter = new StringBuilder();
+            userSearchFilter.append("(&");
+            userSearchFilter.append(confService.getUserSearchFilter());
+            userSearchFilter.append("(" + escapingService.escapeLDAPSearchFilter(usernameAttribute) + "=*)");
```
mike-jumper: If using a `StringBuilder` (good), you shouldn't be doing string concatenation with `+`. The inline concatenation here will actually result in Java creating a temporary `StringBuilder` for the concatenation operation.

You should either do the whole thing with a single line of chained `+`:

    "(&" + confService.getUserSearchFilter() + "(" + escapingService.escapeLDAPSearchFilter(usernameAttribute) + "=*))"

or use only the `StringBuilder` you've created:

    StringBuilder userSearchFilter = new StringBuilder();
    userSearchFilter.append("(&");
    userSearchFilter.append(confService.getUserSearchFilter());
    userSearchFilter.append("(");
    userSearchFilter.append(escapingService.escapeLDAPSearchFilter(usernameAttribute));
    userSearchFilter.append("=*))");

FYI: In the case of the construction of the other query below (```StringBuilder ldapQuery = ...```), the use of a `StringBuilder` is absolutely necessary, since parts of the concatenation operation are conditional, and using Java's `+` operator would result in unnecessary creation of temporary `String` and `StringBuilder` objects for intermediate results.

In this case, you're safe either way, but using both is wasteful.

necouchman: Switched it all to StringBuilder, as you suggested.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
```diff
@@ -85,11 +85,22 @@ private void putAllUsers(Map<String, User> users, LDAPConnection ldapConnection,
...
 
+            // Build a filter using the configured or default user search filter
+            // to find all user objects in the LDAP tree
+            StringBuilder userSearchFilter = new StringBuilder();
+            userSearchFilter.append("(&");
+            userSearchFilter.append(confService.getUserSearchFilter());
+            userSearchFilter.append("(");
+            userSearchFilter.append(escapingService.escapeLDAPSearchFilter(usernameAttribute));
+            userSearchFilter.append("=*)");
+            userSearchFilter.append(")");
```
mike-jumper: >     userSearchFilter.append("=*)");
>     userSearchFilter.append(")");

Really should be ```userSearchFilter.append("=*))")``` - no need for splitting across multiple `append()` calls what is otherwise a single string constant.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
```diff
@@ -85,11 +85,22 @@ private void putAllUsers(Map<String, User> users, LDAPConnection ldapConnection,
...
+         
+
```
mike-jumper: Please kill one of these blank lines - separating things by two blanks is somewhat unusual compared to the surrounding code.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
necouchman: Fixed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
necouchman: Fixed.

mike-jumper:
Thanks, @necouchman! Looks great!

:shipit:

PR #133, opened 2017-03-23
GUACAMOLE-252: Ensure display container is explicitly sized to cover the vertical area of its container.

mike-jumper:
It seems MS Edge does not always stretch `display: table-cell` elements to fit the vertical space of their `display: table` parents. This seems like a layout bug to me, but I can't find any confirmation from standards one way or the other.

In any case, explicitly specifying `height: 100%` makes things work as expected across the board.


jmuehlner:
Glad the fix was simple! :shipit:

PR #138, opened 2017-04-11
GUACAMOLE-250: Add tunnel implementation which reads static files via HTTP.

mike-jumper:
Part of the body of changes intended to make up [GUACAMOLE-250](https://issues.apache.org/jira/browse/GUACAMOLE-250) (in-browser support for playback of recordings), this change adds a new `Guacamole.Tunnel` implementation, `Guacamole.StaticHTTPTunnel`, which downloads static files via HTTP, parsing the instructions therein as if received over a normal Guacamole connection.

`Guacamole.StaticHTTPTunnel` does not enforce any delay on how quickly those instructions stream through. If code needs the original framerate to be preserved, it is up to that code to enforce such a delay through its own buffering mechanism.


jmuehlner:
This all looks reasonable to me.

PR #139, opened 2017-04-12
GUACAMOLE-250: Add support for saving/restoring Guacamole.Client state.

mike-jumper:
Adding functions for exporting/importing client state allow snapshots of Guacamole sessions to be taken, such that instructions following those snapshots can be replayed with the same effect.

This mainly has practical uses for [GUACAMOLE-250](https://issues.apache.org/jira/browse/GUACAMOLE-250), where these opaque state objects will form the basis of auto-generated keyframes, but is required in any case where a Guacamole session may need to be rewound (otherwise, the entire session would need to be replayed to produce the same state).

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Client.js
+++ b/guacamole-common-js/src/main/webapp/modules/Client.js
@@ -133,6 +133,127 @@ Guacamole.Client = function(tunnel) {
     }
 
     /**
+     * Returns an opaque representation of Guacamole.Client state which can be
+     * later imported through a call to importState(). This object is
+     * effectively an independent, compressed snapshot of protocol and display
+     * state.
+     *
+     * @returns {Object}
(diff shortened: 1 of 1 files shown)


guacamole-common-js/src/main/webapp/modules/Client.js
```diff
@@ -133,6 +133,124 @@ Guacamole.Client = function(tunnel) {
...
+
+            var importLayer = state.layers[key];
+            var layer = getLayer(index);
+
+            // Initialize new layer with imported data
+            display.resize(layer, importLayer.width, importLayer.height);
+            display.setChannelMask(layer, Guacamole.Layer.SRC);
+            display.draw(layer, 0, 0, importLayer.url);
+
+            // Set position if not a buffer nor the default layer
```
jmuehlner: This comment seems a bit inaccurate.

PR #141, opened 2017-04-13
GUACAMOLE-250: Invoke client onsync only after display has flushed.

mike-jumper:
The `onsync` handler of `Guacamole.Client` is intended to be invoked when a sync is received. Unfortunately, there is no guarantee that the display has actually finished rendering things at that time, so no display-related action can be performed as a result of that handler.

This change modifies the call to `onsync` so that there is such a guarantee. It will only be invoked once display flush has completed.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Client.js
+++ b/guacamole-common-js/src/main/webapp/modules/Client.js
@@ -1359,15 +1359,15 @@ Guacamole.Client = function(tunnel) {
                     currentTimestamp = timestamp;
                 }
 
+                // If received first update, no longer waiting.
+                if (currentState === STATE_WAITING)
+                    setState(STATE_CONNECTED);
+
+                // Call sync handler if defined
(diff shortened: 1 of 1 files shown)


jmuehlner:
🎄

mike-jumper:
OK - I'm clearly unsure about this.

The originally reasoning behind doing this more-or-less still applies, but I'm encountering issues in implementing GUACAMOLE-250 due to the way display state is asynchronous relative to client state. Exporting client state via `exportState()` at any given time, even within the now-synchronous `onsync` handler, will only represent the layers defined within the client at that moment. Layers which are no longer defined as far as the client's protocol stream is concerned may still be visible within the display because the display has not finished flushing frames.

Making `onsync` synchronous with the display flush doesn't actually solve the problem it's meant to solve. Invoking `flatten()`, for example, *may* result in an image which doesn't contain what the user currently sees.

I'm looking into other ways to solve this.

PR #142, opened 2017-04-14
GUACAMOLE-250: Unblock image rendering tasks if image decode fails.

mike-jumper:
Currently, Guacamole will permanently block the display from all further updates if image decoding fails (bad PNG, unsupported image format, etc.). This is potentially bad, as it would cause the connection to become visibly unresponsive.

This shouldn't happen in practice unless something is going horribly wrong within guacd, but it can easily happen when playing back screen recordings if those recordings contain images not supported by the browser handling playback (playing back a recording containing WebP images on Firefox, for example).


jmuehlner:
Looks good.

PR #143, opened 2017-04-14
GUACAMOLE-250: Export client state asynchronously

mike-jumper:
Guacamole's protocol state and display state are asynchronous with respect to each other. A layer or buffer may have been deleted using a ["dispose" instruction](http://guacamole.incubator.apache.org/doc/gug/protocol-reference.html#dispose-instruction), for example, and thus cannot be referenced as far as the protocol is concerned, yet still exist within the display because the display has not yet flushed (it can be blocked by image decoding). This means that the current implementation of `exportState()` will not always function correctly, especially if the browser is decoding images slowly.

This change modifies `exportState()` such that it only returns the state object asynchronously (through invoking a callback), and populates the protocol and display portions of that state when they are available. A snapshot of layers and buffers present with respect to the protocol is taken immediately, and the data related to the layers and buffers in the snapshot is added to the state object once the display is ready.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Client.js
+++ b/guacamole-common-js/src/main/webapp/modules/Client.js
@@ -196,8 +196,7 @@ Guacamole.Client = function(tunnel) {
             }
 
             // Invoke callback now that the state is ready
-            if (callback)
-                callback(state);
+            callback(state);
 
         });
 
--- a/guacamole-common-js/src/main/webapp/modules/Client.js
+++ b/guacamole-common-js/src/main/webapp/modules/Client.js
@@ -133,16 +133,18 @@ Guacamole.Client = function(tunnel) {
     }
 
     /**
-     * Returns an opaque representation of Guacamole.Client state which can be
+     * Produces an opaque representation of Guacamole.Client state which can be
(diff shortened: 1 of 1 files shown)


mike-jumper:
This solves the problem that #141 was intended to solve.

guacamole-common-js/src/main/webapp/modules/Client.js
```diff
@@ -151,39 +153,53 @@ Guacamole.Client = function(tunnel) {
...
-                exportLayer.matrix = layer.matrix;
-                exportLayer.parent = getLayerIndex(layer.parent);
             }
 
-            // Store exported layer
-            state.layers[key] = exportLayer;
-
-        }
+            // Invoke callback now that the state is ready
+            if (callback)
```
jmuehlner: Is there any point in even using this function with no callback? Can we make the callback required?

mike-jumper: It is required in this case, actually. I can get rid if that check.

mike-jumper: Check removed.

PR #147, opened 2017-04-17
GUACAMOLE-208: Add exception classes for newly-defined status codes.

mike-jumper:
The newly-defined `UPSTREAM_NOT_FOUND`, `UPSTREAM_UNAVAILABLE`, `SESSION_CLOSED`, `SESSION_CONFLICT`, and `SESSION_TIMEOUT` (from #123) are missing corresponding `GuacamoleException` classes. This change defines those.

This change also adds an exception for the old `RESOURCE_CLOSED` status code, which has existed for some time but had no corresponding `GuacamoleException` class.

### DIFF
--- a/guacamole-common/src/main/java/org/apache/guacamole/GuacamoleResourceClosedException.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/GuacamoleResourceClosedException.java
@@ -1,0 +1,69 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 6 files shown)


julianmsanchez:
Hello,

Thanks for this great improvement. Nevertheless despite i'm using 0.9.13-incubating the session gets reconnected automatically once the session is expired. There is any flag to set in order to use this improvement?

Best regards,

Julian.

mike-jumper:
@julianmsanchez, please don't comment on an old pull request if you have a question about Guacamole (even if relevant to that pull request). The place for new questions and discussion is the mailing list:

http://guacamole.incubator.apache.org/support/#mailing-lists

GitHub pull requests are great for code review, but they don't have as much visibility as the main mailing lists, particularly after they're closed.

PR #148, opened 2017-04-21
GUACAMOLE-250: Add example project demonstrating in-browser recording playback.

mike-jumper:
This change adds a barebones example project demonstrating use of the new `Guacamole.SessionRecording` and `Guacamole.StaticHTTPTunnel` objects for in-browser playback of recorded sessions.


jmuehlner:
Nice!

PR #149, opened 2017-04-21
GUACAMOLE-267: Take remote desktop errors into account for balancing groups.

mike-jumper:
This change adds a new object to guacamole-common, `FailoverGuacamoleSocket`, which watches for "error" instructions related to the upstream remote desktop while the Guacamole connection process occurs, and modifies the balancing group behavior of the database auth to use this class where applicable.

If an error is encountered, `FailoverGuacamoleSocket` intercepts that error, rethrowing it as a `GuacamoleUpstreamException` (or subclass). In the case of the database auth, this results in the connection transparently failing over to the next available connection in the group. From the user's perspective, everything just magically works.


jmuehlner:
Ok, everything looks good to me.

PR #151, opened 2017-05-01
GUACAMOLE-289: Allow extensions to provide their own REST services.

mike-jumper:
This changes adds a `getResource()` function to both `AuthenticationProvider` and `UserContext` which can return an arbitrary REST service annotated with JSR-311 annotations (JAX-RS). The resource returned by `AuthenticationProvider` is available without authenticating with the extension, while the resource returned by `UserContext` implicitly requires authentication (there would be no `UserContext` instance to return it otherwise).


mike-jumper:
For reference, `getResource()` provides the following behavior:

 Class                   | Authenticated?   | Resource URL   
------------------------ | ---------------- | -------------
`UserContext`            | Yes (implicitly) | `.../api/session/ext/IDENTIFIER/`
`AuthenticationProvider` | No               | `.../api/ext/IDENTIFIER/`

where "IDENTIFIER" is the identifier of the authentication provider.

PR #152, opened 2017-05-06
GUACAMOLE-281: Consider template GUACAMOLE_HOME in Docker image configuration sanity checks.

mike-jumper:
The sanity check which verifies whether a Guacamole Docker container has at least one configured authentication mechanism shouldn't fail if a custom `GUACAMOLE_HOME` is being used.

### DIFF
--- a/guacamole-docker/bin/start.sh
+++ b/guacamole-docker/bin/start.sh
@@ -427,7 +427,7 @@ fi
 # Validate that at least one authentication backend is installed
 #
 
-if [ -z "$INSTALLED_AUTH" ]; then
+if [ -z "$INSTALLED_AUTH" -a -z "$GUACAMOLE_HOME_TEMPLATE" ]; then
     cat <<END
 FATAL: No authentication configured
 -------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


guacamole-docker/bin/start.sh
```diff
@@ -427,15 +427,15 @@ fi
 # Validate that at least one authentication backend is installed
 #
 
-if [ -z "$INSTALLED_AUTH" ]; then
+if [ -z "$INSTALLED_AUTH" -a -z "$GUACAMOLE_HOME_TEMPLATE" ]; then
```
jmuehlner: Is this supposed to be "-o" instead of "-a" here?

jmuehlner: Don't you need an "or" or something here?

mike-jumper: Nope. The sanity check should fail only if both `$INSTALLED_AUTH` *and* `$GUACAMOLE_HOME_TEMPLATE` are blank. If either is defined, then the sanity check should pass.

jmuehlner: Ah right

PR #153, opened 2017-05-06
GUACAMOLE-189: Add support for per-connection guacd instances

mike-jumper:
This change adds an API-level object describing the connection to guacd, `GuacamoleProxyConfiguration`, and modifies the database authentication to support defining guacd's hostname, port, and encryption method on a per-connection basis.

Besides allowing for deployments leveraging multiple guacds for different subsets of machines, this is a necessary step for [the Guacamole X.Org driver](https://issues.apache.org/jira/browse/GUACAMOLE-168), which itself acts as guacd.


jmuehlner:
Looks good to me... @necouchman any concerns?

necouchman:
Looks good to me!

PR #154, opened 2017-05-07
GUACAMOLE-294: Fix an incorrectly positioned bracket.

chrisgavin:
I spotted this line where `deltaY` is strictly compared against false, which should never be true because `deltaY` is a number. I believe the correct behavior is the one expressed on line 153, where the result of the `guacTouchDrag` function is compared against false.

---

I found this issue through [lgtm.com](https://lgtm.com/projects/g/apache/incubator-guacamole-client/alerts/), a service that analyzes open-source code to look for potential problems. (Full disclosure: I work for the company that operates it.)

### DIFF
--- a/guacamole/src/main/webapp/app/touch/directives/guacTouchDrag.js
+++ b/guacamole/src/main/webapp/app/touch/directives/guacTouchDrag.js
@@ -168,7 +168,7 @@ angular.module('touch').directive('guacTouchDrag', [function guacTouchDrag() {
                     // Signal end of drag gesture
                     if (inProgress && guacTouchDrag) {
                         $scope.$apply(function dragComplete() {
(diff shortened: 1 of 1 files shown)


mike-jumper:
Hi @chrisgavin,

Please see our [contribution guidelines](https://github.com/apache/incubator-guacamole-client/blob/master/CONTRIBUTING), specifically that all changes need a corresponding issue in [JIRA](https://issues.apache.org/jira/browse/GUACAMOLE/), with that issue tagged in the commit message:

> ...
>  All changes to Guacamole must have corresponding issues in JIRA so the change can be properly tracked:
>
>        https://issues.apache.org/jira/browse/GUACAMOLE/
> ... Commits should be made in logical units
    and must reference the JIRA issue number:

    $ git commit -m "GUACAMOLE-123: High-level message describing the changes."

PR #156, opened 2017-05-15
GUACAMOLE-204: Reformat URLs as necessary for AngularJS to read all query parameters.

mike-jumper:


necouchman:
Looks good to me, and tested it with the CAS extension and it works for me.

jmuehlner:
@necouchman Do you have write access to the repos yet? If so you could merge this one. :)

necouchman:
@jmuehlner I think so, but I've not tried it, yet.  I suppose I should learn how to do it - someone want to give me a run-down, and I can try it out?!

mike-jumper:
@necouchman http://guacamole.incubator.apache.org/pull-requests/ ;)

Follow the above and you *should* be fine. If unsure, definitely ask. The big things to remember are:

1. Use `--no-ff` to force a merge commit.
2. Be sure to write a commit message describing what you're merging and tagging the JIRA issue, just like all other commits.
3. If there is a release underway, watch out for whether things go to `master` or to the staging branch and then from there to `master`. There isn't a release underway at the moment, so you don't need to worry about that ... yet. :)
4. Never force push to the main repository.

necouchman:
Cool :-).  Done!

PR #157, opened 2017-05-15
GUACAMOLE-301: Clear login interface if navigation succeeds.

mike-jumper:
From [GUACAMOLE-301](https://issues.apache.org/jira/browse/GUACAMOLE-301):

>
> The current login screen logic displays the prompt if credentials are requested and clears the prompt once login is successful, but this is partially incorrect. If the user navigates to a page which does not require authentication (such as one exposed by an extension), the login prompt will persist unless the entire interface is reloaded.
>
> Because visiting a page will immediately result in credentials being requested if login is required, the login prompt should simply be cleared upon successful navigation.
>


necouchman:
Looks good to me.

PR #158, opened 2017-05-22
GUACAMOLE-292: Restrict editing of attributes based on permissions.

mike-jumper:
Previously, Guacamole users within the database authentication backend could edit any of their own user attributes so long as they had `UPDATE` permission on their own user object. This change modifies this behavior, restricting editing of attributes related to scheduling to users with `ADMINISTER` permission on the user object, and applies that same logic across all object management pages.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
@@ -97,9 +97,12 @@ public abstract class ModeledDirectoryObjectService<InternalType extends Modeled
      *
      * @return
      *     An object which is backed by the given model object.
+     *
+     * @throws GuacamoleException
+     *     If the object instance cannot be created.
      */
     protected abstract InternalType getObjectInstance(ModeledAuthenticatedUser currentUser,
-            ModelType model);
+            ModelType model) throws GuacamoleException;
 
     /**
      * Returns an instance of a model object which is based on the given
@@ -113,9 +116,12 @@ public abstract class ModeledDirectoryObjectService<InternalType extends Modeled
      *
      * @return
      *     A model object which is based on the given object.
+     *
+     * @throws GuacamoleException
+     *     If the model object instance cannot be created.
      */
     protected abstract ModelType getModelInstance(ModeledAuthenticatedUser currentUser,
-            ExternalType object);
+            ExternalType object) throws GuacamoleException;
 
     /**
      * Returns whether the given user has permission to create the type of
@@ -199,9 +205,12 @@ public abstract class ModeledDirectoryObjectService<InternalType extends Modeled
      * @return
      *     A collection of objects which are backed by the models in the given
      *     collection.
+     *
+     * @throws GuacamoleException
+     *     If any of the object instances cannot be created.
      */
     protected Collection<InternalType> getObjectInstances(ModeledAuthenticatedUser currentUser,
-            Collection<ModelType> models) {
+            Collection<ModelType> models) throws GuacamoleException {
 
         // Create new collection of objects by manually converting each model
         Collection<InternalType> objects = new ArrayList<InternalType>(models.size());
--- a/guacamole/src/main/webapp/app/form/directives/form.js
+++ b/guacamole/src/main/webapp/app/form/directives/form.js
@@ -55,7 +55,16 @@ angular.module('form').directive('guacForm', [function form() {
              *
              * @type Object.<String, String>
              */
-            model : '='
+            model : '=',
+
+            /**
+             * Whether the contents of the form should be restricted to those
+             * fields/forms which match properties defined within the given
+             * model object. By default, all fields will be shown.
+             *
+             * @type Boolean
+             */
+            modelOnly : '='
 
         },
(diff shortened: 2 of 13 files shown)


extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
```diff
@@ -97,9 +97,12 @@
      *
      * @return
      *     An object which is backed by the given model object.
+     *
+     * @throws GuacamoleException
+     *     If the object instance cannot be created.
      */
     protected abstract InternalType getObjectInstance(ModeledAuthenticatedUser currentUser,
-            ModelType model);
+            ModelType model) throws GuacamoleException;
```
jmuehlner: Why are these throwing exceptions now?

guacamole/src/main/webapp/app/manage/controllers/manageUserController.js
```diff
@@ -261,6 +261,31 @@ angular.module('manage').controller('manageUserController', ['$scope', '$injecto
     };
 
     /**
+     * Returns whether the current user can change/set all user attributes for
+     * the user being edited within the given data source, regardless of
+     * whether those attributes are already explicitly associated with that
+     * user.
+     *
+     * @param {String} [dataSource]
```
jmuehlner: What's the reason for this dataSource param? I don't see it used anywhere.

guacamole/src/main/webapp/app/manage/controllers/manageUserController.js
```diff
@@ -225,8 +225,8 @@ angular.module('manage').controller('manageUserController', ['$scope', '$injecto
     };
 
     /**
-     * Returns whether the current user can change attributes associated with
-     * the user being edited within the given data source.
+     * Returns whether the current user can change attributes explicitle
```
jmuehlner: Typo here - "explicitle"

guacamole/src/main/webapp/app/form/directives/form.js
```diff
@@ -163,6 +181,59 @@ angular.module('form').directive('guacForm', [function form() {
...
+             *     The field to check.
+             *
+             * @returns {Boolean}
+             *     true if the given field should be visible, false otherwise.
+             */
+            $scope.isVisible = function isVisible(field) {
+
+                // Forms with valuesOnly set should display only fields with
+                // associated values in the model object
+                if ($scope.valuesOnly)
```
jmuehlner: I don't see "valuesOnly" ever being used - is there a plan to use it later?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
mike-jumper: Ah - because `getObjectInstance()` now needs to call `hasObjectPermission()`, at least in the user case. See:

https://github.com/mike-jumper/incubator-guacamole-client/blob/232935fe2cd81077c81bed6079707e3a41ceb9fb/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/UserService.java#L161-L162

guacamole/src/main/webapp/app/manage/controllers/manageUserController.js
mike-jumper: I'll remove that/those.

guacamole/src/main/webapp/app/form/directives/form.js
mike-jumper: There was in the early days of the user profile development, but I think that all got reverted. I'll double-check.

mike-jumper: Yep. It's definitely not needed by these changes, and the changes which *did* need it were ultimately reverted.

guacamole/src/main/webapp/app/manage/controllers/manageUserController.js
mike-jumper: Wait ... it *is* used here. It's not used in the other controllers though. Did you mean those?

jmuehlner: I mean that it's never passed in...

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
jmuehlner: Ok

guacamole/src/main/webapp/app/manage/controllers/manageUserController.js
mike-jumper: Fair enough, and true of all the past similar `can*()` functions, too. I was going to argue that this is the pattern used by similar things here, but that's silly. It's wrong and we should stop wronging it.

mike-jumper: OK - rebased and removed.

PR #159, opened 2017-05-27
GUACAMOLE-292: Add base support for user profiles.

mike-jumper:
This change adds reserved attribute names for each user's full name, email address, organization, and organizational role, exposing those attributes within the user menu if set. A new "EMAIL" field type has been added to allow the definition of attributes which accept only email addresses as values.

The database authentication has been modified to add columns to `guacamole_user` for these attributes, and allows the properties to be set by any user with "UPDATE" permission on the user object.


jmuehlner:
Looks good to me.

PR #160, opened 2017-05-31
GUACAMOLE-203: Add support for SSH Keep Alive

necouchman:
This is for the minor changes required on the client side to support configuring the SSH Keep Alive interval for SSH connections.


necouchman:
Another commit to add support for keepalive configuration for SFTP connections in VNC and RDP.

necouchman:
Ping...any takers for review on this?

necouchman:
@mike-jumper: Is there any way with the implementation of the NUMERIC_FIELD type to force it to be >= 0?

necouchman:
@mike-jumper Ping...any further action needed, here?  Any comments on my previous question Re: NUMERIC_FIELD and forcing values to be 0 or positive?

necouchman:
Okay, negative numbers dealt with on Guacamole server side, so nothing really to do, here, although it still seems like adding options for these NUMERIC fields for forcing value ranges might be worthwhile.

guacamole/src/main/webapp/translations/en.json
```diff
@@ -355,6 +355,7 @@
         "FIELD_HEADER_SERVER_LAYOUT"   : "Keyboard layout:",
         "FIELD_HEADER_SFTP_DIRECTORY"   : "Default upload directory:",
         "FIELD_HEADER_SFTP_HOSTNAME"    : "Hostname:",
+        "FIELD_HEADER_SFTP_KEEPALIVE"   : "SFTP Keepalive Interval:",
```
mike-jumper: For consistency's sake, the field headers should be sentence case, not title case.

guacamole/src/main/webapp/translations/en.json
```diff
@@ -531,6 +533,7 @@
         "FIELD_HEADER_RECORDING_PATH" : "Recording path:",
         "FIELD_HEADER_SFTP_DIRECTORY"   : "Default upload directory:",
         "FIELD_HEADER_SFTP_HOSTNAME"    : "Hostname:",
+        "FIELD_HEADER_SFTP_KEEPALIVE"   : "SFTP Keepalive Interval:",
```
mike-jumper: `FIELD_HEADER_SFTP_KEEPALIVE` will not work, as the parameter has been renamed to `sftp-server-alive-interval`. This will need to be changed to `FIELD_HEADER_SFTP_SERVER_ALIVE_INTERVAL`.

guacamole/src/main/webapp/translations/en.json
necouchman: Corrected.

guacamole/src/main/webapp/translations/en.json
necouchman: Fixed, x2.

PR #161, opened 2017-06-01
GUACAMOLE-102: Support Connection Load Balancing

necouchman:
Okay, this is a LONG shot, but I've taken a stab at a basic implementation of a weighted load balancing algorithm in the Guacamole client.  This is likely to need a lot of work before it's actually ready for merging - I'm acknowledging that up-front :-).

This adds the table column for the connection weight into the DB schema, the necessary JDBC calls to grab, set, and update connection weight, and then the updates to the compare function to sort connections based on a "Weighted Least Connection" algorithm.

The algorithm itself has some nuances that are worth describing.  First, I tried to incorporate as much of what was discussed in the JIRA issue as possible.  Not setting the weight is acceptable, and will result in the algorithm calculating based on a weight of 1 when no weight is explicitly assigned.  I don't know if this is ideal, but kind of necessary to do something like this when doing a WLC calculation.  Weights of 0 and negative values cause the connection to be eliminated from the group - the thought is that these values can be used when updating weights externally/dynamically to denote systems that have not been polled yet (0) or that are administratively disabled (-1, etc.).  For the purposes of sorting the connections, the 0 and negative values are also replaced with a 1 so that the compare function works as expected, but are those connections will later be bypassed by the acquire() method.

The compare() method now adds one to the number of active connections of both items during comparison in order to avoid a zero value.  The rationale behind this is that it allows servers with higher weights and zero connections to be preferred over servers with lower weights and zero connections - without this addition anything with zero connections will end up basically equivalent regardless of its configured weight rather than actually being evaluated by weight.

The overall algorithm was taken from a combination of the LVS wiki page describing how they do WLC and a Citrix article on WLC connections.

This is a pretty simple initial implementation - I have some goals for building upon this in the future, but those are probably better discussed on the JIRA issue and not in this PR.

Fire away...

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
@@ -175,8 +175,8 @@ public class ConnectionModel extends ChildObjectModel {
      * Sets the connection weight.
      *
      * @param connectionWeight
-     *     The weight of the connection.  null is acceptable, -1 indicates the
-     *     connection should not be used.
+     *     The weight of the connection.  null is acceptable, negative values
+     *     indicate that the connection should not be used.
      */
     public void setConnectionWeight(Integer connectionWeight) {
         this.connectionWeight = connectionWeight;
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
@@ -55,6 +55,13 @@ public class ConnectionModel extends ChildObjectModel {
     private Integer maxConnectionsPerUser;
 
     /**
+     * The weight of the connection for the purposes of calculating
+     * WRR algorithm.  null indicates nothing has been set, -1 indicates
+     * the system is unavailable.
+     */
+    private Integer connectionWeight;
+
+    /**
      * The identifiers of all readable sharing profiles associated with this
      * connection.
      */
@@ -162,6 +169,29 @@ public class ConnectionModel extends ChildObjectModel {
      */
     public Integer getMaxConnectionsPerUser() {
         return maxConnectionsPerUser;
+    }
+
+    /**
+     * Sets the connection weight.
+     *
+     * @param connectionWeight
+     *     The weight of the connection.  null is acceptable, -1 indicates the
+     *     connection should not be used.
+     */
+    public void setConnectionWeight(Integer connectionWeight) {
+        this.connectionWeight = connectionWeight;
+    }
+
+    /**
+     * Returns the connection weight used in calculating the
+     * WRR algorithm.
+     *
+     * @return
+     *     The connection weight.  Null indicates no weight has been set,
+     *     -1 indicates that the system is unavailable.
+     */
+    public Integer getConnectionWeight() {
+        return connectionWeight;
     }
 
     /**
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
@@ -175,8 +175,8 @@ public class ConnectionModel extends ChildObjectModel {
      * Sets the connection weight.
      *
      * @param connectionWeight
-     *     The weight of the connection.  null is acceptable, negative values
-     *     indicate that the connection should not be used.
+     *     The weight of the connection.  null is acceptable, and anything
+     *     less than one will prevent the connection from being used.
      */
     public void setConnectionWeight(Integer connectionWeight) {
         this.connectionWeight = connectionWeight;
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
@@ -172,22 +172,25 @@ public class ConnectionModel extends ChildObjectModel {
     }
 
     /**
-     * Sets the connection weight.
+     * Sets the connection weight for load balancing.
      *
      * @param connectionWeight
-     *     The weight of the connection.  null is acceptable, and anything
-     *     less than one will prevent the connection from being used.
+     *     The weight of the connection used in load balancing. 
+     *     The value is not required for the connection (null), and
+     *     values less than 1 will prevent the connection from being
+     *     used.
      */
     public void setConnectionWeight(Integer connectionWeight) {
         this.connectionWeight = connectionWeight;
     }
 
     /**
-     * Returns the connection weight used in calculating the
-     * weighted algorithms.
+     * Returns the connection weight used in applying weighted
+     * load balancing algorithms.
      *
      * @return
-     *     The connection weight as an Integer.
+     *     The connection weight used in applying weighted
+     *     load balancing aglorithms.
      */
     public Integer getConnectionWeight() {
         return connectionWeight;
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
@@ -56,8 +56,8 @@ public class ConnectionModel extends ChildObjectModel {
 
     /**
      * The weight of the connection for the purposes of calculating
-     * WRR algorithm.  null indicates nothing has been set, -1 indicates
-     * the system is unavailable.
+     * WLC algorithm.  null indicates nothing has been set, and anything less
+     * than 1 eliminates the system from being used for connections.
      */
     private Integer connectionWeight;
 
@@ -187,10 +187,7 @@ public class ConnectionModel extends ChildObjectModel {
      * weighted algorithms.
      *
      * @return
-     *     The connection weight as an int.  If the weight is
-     *     null a default weight of 1 is returned.  Zero and
-     *     negative numbers are used to indicate the system is
-     *     unavailable.
+     *     The connection weight as an Integer.
      */
     public Integer getConnectionWeight() {
         return connectionWeight;
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
@@ -191,6 +191,8 @@ public class ConnectionModel extends ChildObjectModel {
      *     -1 indicates that the system is unavailable.
      */
     public Integer getConnectionWeight() {
+        if (connectionWeight == null)
+            return 1;
         return connectionWeight;
     }
 
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
@@ -192,10 +192,8 @@ public class ConnectionModel extends ChildObjectModel {
      *     negative numbers are used to indicate the system is
      *     unavailable.
      */
-    public int getConnectionWeight() {
-        if (connectionWeight == null)
-            return 1;
-        return connectionWeight.intValue();
+    public Integer getConnectionWeight() {
+        return connectionWeight;
     }
 
     /**
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
@@ -184,16 +184,18 @@ public class ConnectionModel extends ChildObjectModel {
 
     /**
      * Returns the connection weight used in calculating the
-     * WRR algorithm.
+     * weighted algorithms.
      *
      * @return
-     *     The connection weight.  Null indicates no weight has been set,
-     *     -1 indicates that the system is unavailable.
+     *     The connection weight as an int.  If the weight is
+     *     null a default weight of 1 is returned.  Zero and
+     *     negative numbers are used to indicate the system is
+     *     unavailable.
      */
-    public Integer getConnectionWeight() {
+    public int getConnectionWeight() {
         if (connectionWeight == null)
             return 1;
-        return connectionWeight;
+        return connectionWeight.intValue();
     }
 
     /**
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
@@ -117,6 +117,11 @@ public class ModeledConnection extends ModeledChildDirectoryObject<ConnectionMod
     public static final String MAX_CONNECTIONS_PER_USER_NAME = "max-connections-per-user";
 
     /**
+     * The connection weight for the WRR algorithm.
+     */
+    public static final String CONNECTION_WEIGHT = "connection-weight";
+
+    /**
      * All attributes related to restricting user accounts, within a logical
      * form.
      */
@@ -265,6 +270,9 @@ public class ModeledConnection extends ModeledChildDirectoryObject<ConnectionMod
             }
         }
 
+        // Set connection weight
+        attributes.put(CONNECTION_WEIGHT, NumericField.format(getModel().getConnectionWeight()));
+
         return attributes;
     }
 
@@ -309,6 +317,13 @@ public class ModeledConnection extends ModeledChildDirectoryObject<ConnectionMod
         // Unimplemented / unspecified
         else
             getModel().setProxyEncryptionMethod(null);
+
+        // Translate connection weight attribute
+        try { getModel().setConnectionWeight(NumericField.parse(attributes.get(CONNECTION_WEIGHT))); }
+        catch (NumberFormatException e) {
+            logger.warn("Not setting the connection weight: {}", e.getMessage());
+            logger.debug("Unable to parse numeric attribute.", e);
+        }
 
     }
 
@@ -394,6 +409,25 @@ public class ModeledConnection extends ModeledChildDirectoryObject<ConnectionMod
             encryptionMethod != null ? encryptionMethod : defaultConfig.getEncryptionMethod()
         );
 
+    /** 
+     * Returns the weight of the connection, or the default.
+     *  
+     * @return
+     *     The weight of the connection.
+     *  
+     * @throws GuacamoleException
+     *     If an error occurs while parsing the concurrency limit properties
+     *     specified within guacamole.properties.
+     */
+    public int getConnectionWeight() throws GuacamoleException {
+
+        // Pull default from environment if weight is unset
+        Integer value = getModel().getConnectionWeight();
+        if (value == null)
+            return environment.getDefaultConnectionWeight();
+
+        // Otherwise use defined value
+        return value;
     }
 
 }
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
@@ -412,7 +412,8 @@ public class ModeledConnection extends ModeledChildDirectoryObject<ConnectionMod
     }
 
     /** 
-     * Returns the weight of the connection, or the default.
+     * Returns the weight of the connection, or a default
+     * of 1 if the weight is undefined.
      *  
      * @return
      *     The weight of the connection.
@@ -420,8 +421,10 @@ public class ModeledConnection extends ModeledChildDirectoryObject<ConnectionMod
      */
     public int getConnectionWeight() {
 
-        // Return the connection weight
-        return getModel().getConnectionWeight();
+        Integer connectionWeight = getModel().getConnectionWeight();
+        if(connectionWeight == null)
+            return 1;
+        return connectionWeight;
 
     }
 
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
@@ -117,7 +117,7 @@ public class ModeledConnection extends ModeledChildDirectoryObject<ConnectionMod
     public static final String MAX_CONNECTIONS_PER_USER_NAME = "max-connections-per-user";
 
     /**
-     * The connection weight for the WRR algorithm.
+     * The connection weight for weighted algorithms.
      */
     public static final String CONNECTION_WEIGHT = "weight";
 
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
@@ -418,7 +418,7 @@ public class ModeledConnection extends ModeledChildDirectoryObject<ConnectionMod
      *     The weight of the connection.
      *  
      */
-    public Integer getConnectionWeight() {
+    public int getConnectionWeight() {
 
         // Return the connection weight
         return getModel().getConnectionWeight();
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
@@ -131,7 +131,7 @@ public class ModeledConnection extends ModeledChildDirectoryObject<ConnectionMod
     ));
 
     /**
-     * All attributes related to to load balancing in a logical form.
+     * All attributes related to load balancing in a logical form.
      */
     public static final Form LOAD_BALANCING = new Form("load-balancing", Arrays.<Field>asList(
         new NumericField(CONNECTION_WEIGHT)
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
@@ -409,6 +409,7 @@ public class ModeledConnection extends ModeledChildDirectoryObject<ConnectionMod
             port             != null ? port             : defaultConfig.getPort(),
             encryptionMethod != null ? encryptionMethod : defaultConfig.getEncryptionMethod()
         );
+    }
 
     /** 
      * Returns the weight of the connection, or the default.
@@ -417,7 +418,7 @@ public class ModeledConnection extends ModeledChildDirectoryObject<ConnectionMod
      *     The weight of the connection.
      *  
      */
-    public int getConnectionWeight() {
+    public Integer getConnectionWeight() {
 
         // Return the connection weight
         return getModel().getConnectionWeight();
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
@@ -119,7 +119,7 @@ public class ModeledConnection extends ModeledChildDirectoryObject<ConnectionMod
     /**
      * The connection weight for the WRR algorithm.
      */
-    public static final String CONNECTION_WEIGHT = "connection-weight";
+    public static final String CONNECTION_WEIGHT = "weight";
 
     /**
      * All attributes related to restricting user accounts, within a logical
@@ -127,7 +127,8 @@ public class ModeledConnection extends ModeledChildDirectoryObject<ConnectionMod
      */
     public static final Form CONCURRENCY_LIMITS = new Form("concurrency", Arrays.<Field>asList(
         new NumericField(MAX_CONNECTIONS_NAME),
-        new NumericField(MAX_CONNECTIONS_PER_USER_NAME)
+        new NumericField(MAX_CONNECTIONS_PER_USER_NAME),
+        new NumericField(CONNECTION_WEIGHT)
     ));
 
     /**
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
@@ -127,7 +127,13 @@ public class ModeledConnection extends ModeledChildDirectoryObject<ConnectionMod
      */
     public static final Form CONCURRENCY_LIMITS = new Form("concurrency", Arrays.<Field>asList(
         new NumericField(MAX_CONNECTIONS_NAME),
-        new NumericField(MAX_CONNECTIONS_PER_USER_NAME),
+        new NumericField(MAX_CONNECTIONS_PER_USER_NAME)
+    ));
+
+    /**
+     * All attributes related to to load balancing in a logical form.
+     */
+    public static final Form LOAD_BALANCING = new Form("load-balancing", Arrays.<Field>asList(
         new NumericField(CONNECTION_WEIGHT)
     ));
 
@@ -137,6 +143,7 @@ public class ModeledConnection extends ModeledChildDirectoryObject<ConnectionMod
      */
     public static final Collection<Form> ATTRIBUTES = Collections.unmodifiableCollection(Arrays.asList(
         CONCURRENCY_LIMITS,
+        LOAD_BALANCING,
         GUACD_PARAMETERS
     ));
 
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
@@ -117,7 +117,7 @@ public class ModeledConnection extends ModeledChildDirectoryObject<ConnectionMod
     public static final String MAX_CONNECTIONS_PER_USER_NAME = "max-connections-per-user";
 
     /**
-     * The connection weight for weighted algorithms.
+     * The connection weight attribute used for weighted load balancing algorithms.
      */
     public static final String CONNECTION_WEIGHT = "weight";
 
@@ -412,11 +412,13 @@ public class ModeledConnection extends ModeledChildDirectoryObject<ConnectionMod
     }
 
     /** 
-     * Returns the weight of the connection, or a default
-     * of 1 if the weight is undefined.
+     * Returns the weight of the connection used in applying weighted
+     * load balancing algorithms, or a default of 1 if the 
+     * attribute is undefined.
      *  
      * @return
-     *     The weight of the connection.
+     *     The weight of the connection used in applying weighted
+     *     load balancing algorithms.
      */
     public int getConnectionWeight() {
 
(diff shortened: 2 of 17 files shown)


extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/RestrictedGuacamoleTunnelService.java
```diff
@@ -179,9 +187,27 @@ protected ModeledConnection acquire(RemoteAuthenticatedUser user,
...
+                    weightB = b.getConnectionWeight().intValue();
+                // In all other cases assign 1 for sorting.
+                else
+                    weightB = 1;
+
+                // Get current active connections, add 1 to both to avoid calculations with 0.
+                int connsA = getActiveConnections(a).size() + 1;
+                int connsB = getActiveConnections(b).size() + 1;
+
+                return (connsA * 10000 / weightA) - (connsB * 10000 / weightB);
```
mike-jumper: What's the significance of 10000 here?

necouchman: There's no significance to the 10000 itself, that number is arbitrary.  The overall idea, though, is to get the values high enough that the final calculation returns an integer.  Because the number of connections is going to be a relatively low number (between 1 and 100, maybe 1000), and the weights can be relatively high (100, 1000, 10000, etc.), the idea is to get the numerator sufficiently high so that an integer calculation will actually happen.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/RestrictedGuacamoleTunnelService.java
```diff
@@ -179,9 +187,27 @@ protected ModeledConnection acquire(RemoteAuthenticatedUser user,
 
             @Override
             public int compare(ModeledConnection a, ModeledConnection b) {
-
-                return getActiveConnections(a).size()
-                     - getActiveConnections(b).size();
+                
+                int weightA, weightB;
+                // Check if weight of a is non-null and retrieve it.
+                if (a.getConnectionWeight() != null && a.getConnectionWeight().intValue() > 0)
```
mike-jumper: Checking for `null`, using a default value, etc. seems the sort of logic that should be within `ModeledConnection`, rather than repeated here. That way you could simply call `a.getWeight()` or `a.getConnectionWeight()` and it will always just work?

Only the underlying model object (and the code applying attribute changes) need actually worry about such internal concerns.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/RestrictedGuacamoleTunnelService.java
mike-jumper: > ... the idea is to get the numerator sufficiently high so that an integer calculation will actually happen.

Ick. =/

With normal WLC, the idea is that a server with weight N should be allocated N times as many connections as a server with weight 1, correct? That way weights can be calculated based on the proportion of resources available on each server relative to some arbitrary baseline?

I agree that multiplying each of the compared values by a constant will not alter the above relationship, but I'm not a fan of the arbitrariness of that constant here.

If that constant were instead `weightB * weightA`, this would become:

    return (connsA * weightB * weightA / weightA) - (connsB * weightB * weightA / weightB);

which clearly reduces to:

    return (connsA * weightB) - (connsB * weightA);

which is nice and integery. You won't get the same values, *but* you will get the same sign, which is all that matters here.

necouchman: Yes, clearly not the best code ever written.

> With normal WLC, the idea is that a server with weight N should be allocated N times as many connections as a server with weight 1, correct?

Well, I'd say that the idea of weight in general is as you say - but the WLC also factors in the number of current connections.  There's also WRR, which accounts for the last time a system was used for a connection combined with weight.  But, yes, that's the overall idea...

> return (connsA * weightB) - (connsB * weightA);

Works for me, I'll update the code.  I agree that the actual value that is returned or computed does not matter - the sign and the ability to get a meaningful integer value are really all that I was going for, and that seems to solve it more elegantly :-).  I'll update the code, test a few times and commit the changes.

Thanks for the help!

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/RestrictedGuacamoleTunnelService.java
necouchman: Put the null check in the getConnectionWeight method, and then removed the checking for > 0.  Since anything < 1 gets eliminated from consideration later on, it doesn't really matter how those connections sort, so there isn't much point in checking.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/RestrictedGuacamoleTunnelService.java
necouchman: Ran through a few test cases with the updated code and that calculation seems to work well, so I've pushed the changes.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/RestrictedGuacamoleTunnelService.java
```diff
@@ -180,8 +188,15 @@ protected ModeledConnection acquire(RemoteAuthenticatedUser user,
             @Override
             public int compare(ModeledConnection a, ModeledConnection b) {
 
-                return getActiveConnections(a).size()
-                     - getActiveConnections(b).size();
+                // Get connection weight for the two systems being compared.                
+                int weightA = a.getConnectionWeight().intValue();
+                int weightB = b.getConnectionWeight().intValue();
+
+                // Get current active connections, add 1 to both to avoid calculations with 0.
```
mike-jumper: Why? What do you have against 0? ;)

Especially considering that you:

> ...  removed the checking for > 0. Since anything < 1 gets eliminated from consideration later on, it doesn't really matter how those connections sort, so there isn't much point in checking.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/RestrictedGuacamoleTunnelService.java
```diff
@@ -180,8 +188,15 @@ protected ModeledConnection acquire(RemoteAuthenticatedUser user,
             @Override
             public int compare(ModeledConnection a, ModeledConnection b) {
 
-                return getActiveConnections(a).size()
-                     - getActiveConnections(b).size();
+                // Get connection weight for the two systems being compared.                
+                int weightA = a.getConnectionWeight().intValue();
```
mike-jumper: Why not return `int`?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/RestrictedGuacamoleTunnelService.java
necouchman: It seems like this might actually matter for number of connections.  If you have connections A, B, and C, with weights 1, 2, and 3 respectively, and no connections to any of them, then, ideally, with a weighted algorithm, you'd like C to get the first connection because it has the larger weight.  If the number of active connections is 0, then the net weight for a host evaluates to zero, which messes up this balancing case a little bit.  It may not be the end of the world, but it's less than ideal.  By adding one to both of the connections being compared, connections with 0 active connections will be "correctly" (or, maybe, more ideally?) evaluated for their weight.

The reason that I removed the check for the weight calculation is that items with a weight < 1 are removed, anyway, so the sorting for those items doesn't matter, at all.  This isn't true for active connections, obviously, so it seems like adding one here makes the weight behave as expected when a connection has zero active connections.

Or am I worrying too much about a corner case?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/RestrictedGuacamoleTunnelService.java
necouchman: Yeah, that makes sense...and then we don't need all those assignments and can easily fit everything into a single return statement...

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/RestrictedGuacamoleTunnelService.java
mike-jumper: Nope - sorting connections with zero active connections by their weight makes sense, but if that's what you're aiming for here, then perhaps that's what the comment should describe?

The thing that wouldn't make sense is arbitrarily avoiding zero.

mike-jumper: Actually ... doesn't adding 1 to both upset the intended property that a server with weight N will be given N times as many connections as a server with weight 1?

You'd expect a server with weight 4 to be given twice as many connections as a server with weight 2, but this would result in a server with weight 4 receiving 1.6666... times as many connections as a server with weight 2.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
```diff
@@ -165,6 +172,33 @@ public Integer getMaxConnectionsPerUser() {
...
+     * @return
+     *     The connection weight as an int.  If the weight is
+     *     null a default weight of 1 is returned.  Zero and
+     *     negative numbers are used to indicate the system is
+     *     unavailable.
+     */
+    public int getConnectionWeight() {
+        if (connectionWeight == null)
+            return 1;
+        return connectionWeight.intValue();
```
mike-jumper: Java will do this automatically, by the way. If you have something like:

    Integer foo = 2; // Automatically wrapped within Integer by the Java compiler
    int bar = foo; // Automatically unwrapped by the Java compiler

things happen magically thanks to autoboxing / autounboxing: https://docs.oracle.com/javase/tutorial/java/data/autoboxing.html

mike-jumper: Shouldn't this logic be within `ModeledConnection`? The model object itself should be relatively logic-free, being the pure (ish) intermediary between the database tables and Java.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
```diff
@@ -393,6 +409,19 @@ public GuacamoleProxyConfiguration getGuacamoleProxyConfiguration()
             port             != null ? port             : defaultConfig.getPort(),
             encryptionMethod != null ? encryptionMethod : defaultConfig.getEncryptionMethod()
         );
+    }
+
+    /** 
+     * Returns the weight of the connection, or the default.
```
mike-jumper: "or the default" is unclear here.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
```diff
@@ -393,6 +409,19 @@ public GuacamoleProxyConfiguration getGuacamoleProxyConfiguration()
...
+     *  
+     * @return
+     *     The weight of the connection.
+     *
```
mike-jumper: Looks like a blank line snuck in here.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/RestrictedGuacamoleTunnelService.java
```diff
@@ -180,8 +187,8 @@ protected ModeledConnection acquire(RemoteAuthenticatedUser user,
             @Override
             public int compare(ModeledConnection a, ModeledConnection b) {
 
-                return getActiveConnections(a).size()
-                     - getActiveConnections(b).size();
+                return ((getActiveConnections(a).size() + 1) * b.getConnectionWeight() -
```
mike-jumper: Repeating from an older comment for convenience: though that `+ 1` does result in the connection weight having an effect when the number of connections is zero, it also offsets the math such that the proportions governing weights in WLC are wrong, particularly for smaller numbers of connections.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/RestrictedGuacamoleTunnelService.java
```diff
@@ -193,6 +200,12 @@ public int compare(ModeledConnection a, ModeledConnection b) {
         // Return the first unreserved connection
         for (ModeledConnection connection : sortedConnections) {
 
+            // If connection weight is zero or negative, this host is disabled and should not be used.
+            if (connection.getConnectionWeight() < 1) {
+                logger.warn("Weight for {} is < 1, connection will be skipped.", connection.getName());
```
mike-jumper: Perhaps this should instead be logged at the debug level?

If zero and negative weights are intended mechanisms for disabling connection, logging this at the warning level will result in the logs being filled with messages warning the admin that they're using a feature as intended.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/schema/upgrade/upgrade-pre-0.9.12.sql
```diff
@@ -0,0 +1,25 @@
...
+-- specific language governing permissions and limitations
+-- under the License.
+--
+
+--
+-- Add per-user password set date
+--
+
+ALTER TABLE guacamole_connection
+    ADD COLUMN connection_weight int(11);
```
mike-jumper: With 0.9.12 already out, and 0.9.13 pending, this should be `upgrade-pre-0.9.14.sql`.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/MySQLEnvironment.java
```diff
@@ -88,6 +88,12 @@
     private int DEFAULT_MAX_CONNECTIONS = 0;
 
     /**
+     * The default value for the connection weight for a connection in
+     * a balancing group.
+     */
+    private int DEFAULT_CONNECTION_WEIGHT = 0;
```
mike-jumper: Shouldn't the default connection weight be 1?

mike-jumper:
Is there an updated `ConnectionMapper.xml` for MySQL? I only see the one for PostgreSQL.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
necouchman: I think I ran into an issue with the automatic (un)boxing stuff and went that route, but it seems to be fine, now, so I've cleaned that up.  Also moved the null check/default code to the ModeledConnection class.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/MySQLEnvironment.java
necouchman: Removed this completely - the DB-specific default connection weight was not being used at all.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/schema/upgrade/upgrade-pre-0.9.12.sql
necouchman: Fixed.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/RestrictedGuacamoleTunnelService.java
necouchman: Changed to debug.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
necouchman: Removed.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
necouchman: Cleaned up.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/RestrictedGuacamoleTunnelService.java
necouchman: Okay, so a couple of follow-up questions:
- How much does it actually mess up the math?  Since basically all we're doing is sorting the array based on the sign and an integer value, does it actually mess up the calculation enough to even throw off the sorting?
- Is it a lose-lose situation?  I'm trying to think of another way to run the calculation so as to account for zero connections but not throw off the math, and nothing is come to mind.  Is it simply a case where we have to choose one path or the other, and accept that one of the cases is going to be flawed?
- If there isn't a way to resolve it so that both cases work, any preferences on which one to stick with the win on?  I would assume we'd want to ignore the 0 connection case in favor of the other ones being correct (if it matters enough to worry about it), but any reason we'd want 0 to be correct over lower connection counts?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
mike-jumper: Right so, to be clear:

1. It makes sense that `getConnectionWeight()` of `ConnectionModel` returns `Integer`. Since this is a model object and the value can indeed be `null`, this actually should be the case.
2. Within `ModeledConnection`, a higher-level object which is expected to abstract away some of the low level detail, it makes sense for `getConnectionWeight()` to return `int`, automatically handling things like that `null` check and the default value.

It's definitely not safe to simply keep this particular function as returning `int` when `connectionWeight` can be `null`, but that's OK because this particular function should be returning `Integer`. It's the `ModeledConnection` version that should handle the other aspects, always returning a sensible `int`.

necouchman:
> Is there an updated ConnectionMapper.xml for MySQL? I only see the one for PostgreSQL.
Oops, forgot that one...should be added, now.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
necouchman: Got it - should be cleared up with the latest round of updates.  Thanks!

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/RestrictedGuacamoleTunnelService.java
mike-jumper: > How much does it actually mess up the math? Since basically all we're doing is sorting the array based on the sign and an integer value, does it actually mess up the calculation enough to even throw off the sorting

Quite a bit, actually.

With the `+ 1`, the calculations effectively assume that each server has one more connection than they actually do. Depending on the weights involved, that one extra connection can unbalance things considerably:

comparator-comparison

Take a look at the 21 connection case above (row 24). Without the `+ 1`, the number of connections allocated to A and B stays around the expected 10:1 ratio... but *with* the `+ 1`, the ratio gets all the way up to 20:1 before things are rebalanced properly.

> Is it a lose-lose situation? I'm trying to think of another way to run the calculation so as to account for zero connections but not throw off the math, and nothing is come to mind. Is it simply a case where we have to choose one path or the other, and accept that one of the cases is going to be flawed?

The only way would be to compare the weights as a means of breaking ties. That way, servers are always ordered based on a correct calculation of their relative load, and a sensible choice is made in the case that each server is under equivalent load.

> If there isn't a way to resolve it so that both cases work, any preferences on which one to stick with the win on? I would assume we'd want to ignore the 0 connection case in favor of the other ones being correct (if it matters enough to worry about it), but any reason we'd want 0 to be correct over lower connection counts?

The one that's actually correct. ;)

In this case, though, it seems we can have it both ways - we just need to make sure the math is sound.

necouchman: Okay, that makes sense.  So I guess what we're going for here is keeping the ratio under the specified limit - that is, if 10:1 is the configured ratio (based on configured weights of the servers), then work up to that, then drop back down.  It's still, mathematically, unbalanced (you end up with a 5.5:1 ratio), but it keeps it "in bounds" so to speak.

I've tweaked the code such that it computes the net weight of each of the connections, and, if they're equal, returns the difference between the configured weight of each system, otherwise returns the difference in the net weight.  Does that look more like what we're going for in a WLC algorithm?

mike-jumper: Yep - it looks much better now.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
```diff
@@ -55,6 +55,13 @@
     private Integer maxConnectionsPerUser;
 
     /**
+     * The weight of the connection for the purposes of calculating
+     * WRR algorithm.  null indicates nothing has been set, -1 indicates
```
mike-jumper: WLC, not WRR, right?

Is it -1 that is intended to be the documented way to indicate the system is unavailable, or zero?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
```diff
@@ -165,6 +172,31 @@ public Integer getMaxConnectionsPerUser() {
...
+    public void setConnectionWeight(Integer connectionWeight) {
+        this.connectionWeight = connectionWeight;
+    }
+
+    /**
+     * Returns the connection weight used in calculating the
+     * weighted algorithms.
+     *
+     * @return
+     *     The connection weight as an int.  If the weight is
```
mike-jumper: I'm guessing this is from the previous version of this function? It currently just returns the connection weight, with the handling of default value, null, etc. being (correctly) elsewhere.

mike-jumper:
Do you think it's strange that we're using both zero and negative values equivalently to denote a disabled connection?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
necouchman: Yep...replaced a few instances of that, already, but missed this one.  Should be fixed.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
necouchman: Yes, removed most of that comment.

necouchman:
> Do you think it's strange that we're using both zero and negative values equivalently to denote a disabled connection?

I think most places I've replaced mentions of "0 or negative" with just anything less than 1.  It's really just more of an administrative thing - should not really matter to the algorithm itself.  I'll try to go back and look and make sure any remaining instances of "0 or negative values" are cleaned up.

necouchman:
Okay, finished cleaning up a couple of those comments.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
```diff
@@ -165,6 +172,28 @@ public Integer getMaxConnectionsPerUser() {
...
+    public void setConnectionWeight(Integer connectionWeight) {
+        this.connectionWeight = connectionWeight;
+    }
+
+    /**
+     * Returns the connection weight used in calculating the
+     * weighted algorithms.
+     *
+     * @return
+     *     The connection weight as an Integer.
```
mike-jumper: To avoid confusion, I suggest removing "as an Integer" here, as that wording implies that the connection weight is a different type under other circumstances. If the intent is to document the return type, that's already inherently documented by the function prototype.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/RestrictedGuacamoleTunnelService.java
```diff
@@ -180,8 +187,24 @@ protected ModeledConnection acquire(RemoteAuthenticatedUser user,
             @Override
             public int compare(ModeledConnection a, ModeledConnection b) {
 
-                return getActiveConnections(a).size()
-                     - getActiveConnections(b).size();
+                // Active connections
+                int Ca = getActiveConnections(a).size();
```
mike-jumper: Any reason the variables here don't follow the `headlessCamelCase` convention used by the other Java code?

Depending on why you went this route, I'm not sure this needs to be changed, but this function is slightly hard to read with how similar all the variables look.

Anyway, not necessarily an issue worth further iterations, but I'm interested to hear what you were thinking at the time and what your thoughts are now.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/RestrictedGuacamoleTunnelService.java
```diff
@@ -180,8 +187,24 @@ protected ModeledConnection acquire(RemoteAuthenticatedUser user,
...
+                int Cb = getActiveConnections(b).size();
+
+                // Assigned weight
+                int Wa = a.getConnectionWeight();
+                int Wb = b.getConnectionWeight();
+
+                // Net weight of connections
```
mike-jumper: Is "net weight" the standard term for this part of the calculations for WLC?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/resources/translations/en.json
```diff
@@ -19,6 +19,7 @@
 
         "FIELD_HEADER_MAX_CONNECTIONS"          : "Maximum number of connections:",
         "FIELD_HEADER_MAX_CONNECTIONS_PER_USER" : "Maximum number of connections per user:",
+        "FIELD_HEADER_WEIGHT"                   : "Connection Weight for Load Balancing:",
```
mike-jumper: For the sake of consistency, this should follow the same style as the other headers (sentence case, not title case).

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
```diff
@@ -117,12 +117,18 @@
...
+
+    /**
      * All attributes related to restricting user accounts, within a logical
      * form.
      */
     public static final Form CONCURRENCY_LIMITS = new Form("concurrency", Arrays.<Field>asList(
         new NumericField(MAX_CONNECTIONS_NAME),
-        new NumericField(MAX_CONNECTIONS_PER_USER_NAME)
+        new NumericField(MAX_CONNECTIONS_PER_USER_NAME),
+        new NumericField(CONNECTION_WEIGHT)
```
mike-jumper: Perhaps this should go in a different `Form`? As is, it will be rendered as one of several fields in a group labeled "Concurrency Limits", which is not what this relates to.

This would also help clean up the wording of the field itself, currently "Connection Weight for Load Balancing", as locating the field within a section titled "Load Balancing" (or similar) would remove the need to further qualify "Connection Weight" with the phrase "for Load Balancing".

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/RestrictedGuacamoleTunnelService.java
necouchman: Both the lack of camel case and the "Net Weight" came from referencing the following pages:
http://kb.linuxvirtualserver.org/wiki/Weighted_Least-Connection_Scheduling
https://www.citrix.com/blogs/2010/10/01/load-balancing-weights/
https://www.citrix.com/blogs/2010/09/02/load-balancing-least-connections/

(as did the initial, and now-removed, concept of multiplying by the arbitrary 10K number).  Anyway, I don't know that "Net Weight" is actually called out in those, but the Nw abbreviation is used in the Citrix page.  I can go with something else - computed weight or calculated weight - if that makes more sense.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
```diff
@@ -165,6 +172,28 @@ public Integer getMaxConnectionsPerUser() {
     }
 
     /**
+     * Sets the connection weight.
+     *
+     * @param connectionWeight
+     *     The weight of the connection.  null is acceptable, and anything
```
necouchman: Hopefully the updated commentary here looks better.  I didn't know exactly how to describe this since it should be coming from the DB - let me know if you're looking for something more specific.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
```diff
@@ -165,6 +172,28 @@ public Integer getMaxConnectionsPerUser() {
...
+     *     less than one will prevent the connection from being used.
+     */
+    public void setConnectionWeight(Integer connectionWeight) {
+        this.connectionWeight = connectionWeight;
+    }
+
+    /**
+     * Returns the connection weight used in calculating the
+     * weighted algorithms.
```
necouchman: Took your suggestion and reworded this.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
necouchman: Removed.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
```diff
@@ -117,12 +117,18 @@
     public static final String MAX_CONNECTIONS_PER_USER_NAME = "max-connections-per-user";
 
     /**
+     * The connection weight for weighted algorithms.
```
necouchman: Cleaned this up.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
```diff
@@ -393,6 +409,21 @@ public GuacamoleProxyConfiguration getGuacamoleProxyConfiguration()
             port             != null ? port             : defaultConfig.getPort(),
             encryptionMethod != null ? encryptionMethod : defaultConfig.getEncryptionMethod()
         );
+    }
+
+    /** 
+     * Returns the weight of the connection, or a default
```
necouchman: Clarified.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/PostgreSQLGuacamoleProperties.java
```diff
@@ -157,6 +157,15 @@ private PostgreSQLGuacamoleProperties() {}
 
     };
 
+    public static final IntegerGuacamoleProperty
+            POSTGRESQL_DEFAULT_CONNECTION_WEIGHT =
```
necouchman: Oops...removed.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/MySQLGuacamoleProperties.java
```diff
@@ -175,6 +175,18 @@ private MySQLGuacamoleProperties() {}
     };
 
     /**
+     * The connection weight for connections in balancing groups.
```
necouchman: Removed.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/RestrictedGuacamoleTunnelService.java
necouchman: Changed the variable names to be more consistent with standards.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/RestrictedGuacamoleTunnelService.java
necouchman: Re-worded this.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/resources/translations/en.json
necouchman: Should be cleaned up - also added a section for the soon-to-be added form...

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
necouchman: That's fine with me - I was trying to avoid creating too many section breaks, especially for sections with a single item in them, and it seemed like this still fell under the heading of "Concurrency Limits", since it is used to impact how connections are concurrently applied to systems, but I'll break it out into another section.

necouchman: Moved to a separate section/form.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
```diff
@@ -126,11 +131,19 @@
     ));
 
     /**
+     * All attributes related to to load balancing in a logical form.
```
mike-jumper: "to to"

👗

mike-jumper:
Cool - other than the silly typo noted above, things look good to me. I'll be happy to merge once there's no related tutu. ;)

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
necouchman: Fine, since you don't like tu-tus, I've removed it.

necouchman:
Okay, I think it's cleaned up!  Thanks, again, for the help on this one...good educational experience for me!

PR #163, opened 2017-06-04
GUACAMOLE-314: Update extension API version sanity check to 0.9.13-incubating.

mike-jumper:
Lacking this, none of the extensions whose versions were recently bumped will load.

Whoops.


mike-jumper:
**NOTE:** Intended for `staging/0.9.13-incubating`.

PR #164, opened 2017-06-04
GUACAMOLE-189: Define guacd override columns for guacamole_connection table in main schema init script.

mike-jumper:


mike-jumper:
**NOTE:** Intended for `staging/0.9.13-incubating`.

PR #165, opened 2017-06-05
GUACAMOLE-284: Apply database account restrictions when a database account is required.

mike-jumper:
These changes modify the enforcement of database account restrictions (disabled accounts, scheduling, password expiration) such that they apply even if authentication did not come from the database, so long as [the `mysql-user-required` or `postgresql-user-required` properties](http://guacamole.incubator.apache.org/doc/gug/jdbc-auth.html#jdbc-auth-restrict) are set.


jmuehlner:
Ok, looks reasonable to me.

PR #166, opened 2017-06-08
GUACAMOLE-317: Add support for failover-only connections.

mike-jumper:


extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/resources/org/apache/guacamole/auth/jdbc/connection/ConnectionMapper.xml
```diff
@@ -222,6 +228,7 @@
             proxy_port               = #{object.proxyPort,jdbcType=INTEGER},
             proxy_encryption_method  = #{object.proxyEncryptionMethod,jdbcType=VARCHAR},
             connection_weight        = #{object.connectionWeight,jdbcType=INTEGER}
```
necouchman: Missing comma, I believe?

mike-jumper: OK - should be fully commatose now.

PR #167, opened 2017-06-15
GUACAMOLE-323: Provide login help text during CAS redirect.

necouchman:
Quick fix for JIRA issue 323, where no helpful error message is displayed during redirect to CAS login.  This seems to do the trick.


necouchman:
One other note on this...the login page still seems to display the goofy LOGIN.FIELD_HEADER_TICKET for the "hidden" ticket field.  There was some work done around the RADIUS authentication to try to hide that - was there some trick to get that to not display?

mike-jumper:
Other than hiding that field entirely via CSS, nope, though if we go that route, it might be a reasonable solution to just have the field itself render as that helpful message. The Duo authentication takes an approach like this, with the field being styled such that it covers the entire view.

necouchman:
I was looking back at the commit you made that attempted to fix that, and it looks like it successfully hides the field itself, but maybe not the header.  Actually, the rendering in the browser is a little weird - it's almost like that one field has an entire copy of the index.html DOM nested inside it.  It's a little hard to explain - if I figure out an useful way to export the output from the Chrome element inspector I'll send it over.

necouchman:
guac-odd-login-field

Sorry about the image format, but hopefully you can see what I mean...

necouchman:
Okay, well, I followed your lead on what you did in the Duo module, adding a translation file, setting the header to an empty string, and then providing the translation for the help text.  I'm still not sure what's going on with all that nested content inside the field there - if you have any insight to that, I'd appreciate it.  Otherwise this PR should take care of this JIRA issue.

mike-jumper:
@necouchman - can you change the merge base of the PR to `staging/0.9.13-incubating`, since this relates to a bug in the pending release?

necouchman:
> can you change the merge base of the PR to staging/0.9.13-incubating, since this relates to a bug in the pending release?

Will do.

mike-jumper:
That login dialog that you just posted ... is that how things currently look? Are you still seeing the weird inclusion of index.html within the field?

I'm not sure what's going on there, but the result of rendering the insufficient credentials auth response shouldn't look that way. The same mechanism is used for the database auth's password reset feature, and things render very differently:

password-reset

necouchman:
The last login dialog that I posed is how it currently visually appears on the page, which is I think is fine.

However, if I look at the Chrome Developer Console and dig into the elements, all of the extra index.html stuff still shows up "inside" that element (as per the image I posted previously).  Steps to replicate it are pretty easy:
- Check out currently Guacamole source.
- In the CAS extension directory, edit the casController.js file and comment out the window.location line to prevent the redirect.
- Build, copy over CAS extension, and deploy Guacamole client.
- Go to the login page - you should see the dialog above - and then dig into the Chrome Developer console and drill down into that element.

I'll rebase the PR.

necouchman:
Okay, I believe I've rebased correctly such that only the commits relevant to this PR will get merged.

necouchman:
Ping...any further action needed on this one, or is it ready to be merged?  I think it's holding up 0.9.13 release, correct?

mike-jumper:
Correct.

I'll go ahead and merge this. I'm a little concerned over the behavior you're seeing with `index.html` getting re-included, and with the login screen not being styled differently for the insufficient credentials case, but other issues can be opened for those if those problems recur during testing. I see no issue with the code itself.

PR #169, opened 2017-06-26
GUACAMOLE-335: Use /etc/guacamole for GUACAMOLE_HOME if not overridden and ~/.guacamole does not exist.

mike-jumper:


jmuehlner:
Sounds like a good idea!

PR #170, opened 2017-06-26
GUACAMOLE-334: Add support for downloading connection history search results as CSV.

mike-jumper:


jmuehlner:
Seems reasonable.

PR #173, opened 2017-07-06
GUACAMOLE-338: Expand connections/groups by default if any of their descendants are selected.

mike-jumper:


guacamole/src/main/webapp/app/groupList/types/GroupListItem.js
```diff
@@ -187,7 +187,7 @@ angular.module('groupList').factory('GroupListItem', ['ConnectionGroup', functio
             dataSource : dataSource,
 
             // Type information
-            expandable : includeSharingProfiles,
+            expandable : includeSharingProfiles !== false,
```
jmuehlner: Why this change?

jmuehlner:
This looks like a good idea to me!

guacamole/src/main/webapp/app/groupList/types/GroupListItem.js
mike-jumper: Sharing profiles are supposed to be included by default (the function is documented as `includeSharingProfiles` defaulting to `true`), but not specifying that parameter makes `includeSharingProfiles` falsey. The `expandable` property needs to be `true` unless `includeSharingProfiles` is explicitly `false`.

necouchman:
Looks good to me...I'll get these commits into the master branch shortly.

PR #174, opened 2017-07-07
GUACAMOLE-341: Make SSO Authentication work and pass through username…

dotmaster:
… to RDP session

To make tokenfilter work with auth-header and noauth module. (needed e.g. for Kerberos authentication through an Apache/Nginx Reverse Proxy, that passes REMOTE_USER header), username must be set in the credentials object, because it is added to the Tokenfilter only if username is not null in the credentials object.
basically make $
{GUAC_USERNAME}
be replaced with the credentials passed though the REMOTE_USER variable.
See also:
https://github.com/glyptodon/guacamole-client/blob/b26a664d66fd14a22eb7300d29aa20390cf408ec/guacamole-ext/src/main/java/org/apache/guacamole/token/StandardTokens.java#L118
https://github.com/glyptodon/guacamole-client/blob/0.9.12-incubating/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleAuthenticationProvider.java#L170

### DIFF
--- a/extensions/guacamole-auth-header/src/main/java/org/apache/guacamole/auth/header/AuthenticationProviderService.java
+++ b/extensions/guacamole-auth-header/src/main/java/org/apache/guacamole/auth/header/AuthenticationProviderService.java
@@ -71,6 +71,9 @@ public class AuthenticationProviderService {
 
             // Get the username from the header configured in guacamole.properties
             String username = request.getHeader(confService.getHttpAuthHeader());
+            
+            //write username to the credentials object to make tokenfilter work
+            credentials.setUsername(username);
 
             if (username != null) {
                 AuthenticatedUser authenticatedUser = authenticatedUserProvider.get();
--- a/extensions/guacamole-auth-header/src/main/java/org/apache/guacamole/auth/header/AuthenticationProviderService.java
+++ b/extensions/guacamole-auth-header/src/main/java/org/apache/guacamole/auth/header/AuthenticationProviderService.java
@@ -72,7 +72,7 @@ public class AuthenticationProviderService {
             // Get the username from the header configured in guacamole.properties
             String username = request.getHeader(confService.getHttpAuthHeader());
             
-            //write username to the credentials object to make tokenfilter work
+            //  Write username to the credentials object to make tokenfilter work
             credentials.setUsername(username);
 
             if (username != null) {
--- a/extensions/guacamole-auth-header/src/main/java/org/apache/guacamole/auth/header/AuthenticationProviderService.java
+++ b/extensions/guacamole-auth-header/src/main/java/org/apache/guacamole/auth/header/AuthenticationProviderService.java
@@ -71,11 +71,10 @@ public class AuthenticationProviderService {
 
             // Get the username from the header configured in guacamole.properties
             String username = request.getHeader(confService.getHttpAuthHeader());
-            
-            //  Write username to the credentials object to make tokenfilter work
-            credentials.setUsername(username);
 
             if (username != null) {
+                //  Write username to the credentials object to make tokenfilter work
+                credentials.setUsername(username);
                 AuthenticatedUser authenticatedUser = authenticatedUserProvider.get();
                 authenticatedUser.init(username, credentials);
                 return authenticatedUser;


necouchman:
Good call!  I suspect this also needs to be done in the guacamole-auth-cas module, as well, as I probably failed to put it there, too :-).

Any reason not to place it inside the `if (username != null)` block?  It should be safe to set the username to null in the credentials object, but I don't know that there's much point, and since the check is already there for the authenticatedUser initialization, seems like a good place for it.

extensions/guacamole-auth-header/src/main/java/org/apache/guacamole/auth/header/AuthenticationProviderService.java
```diff
@@ -71,6 +71,9 @@ public AuthenticatedUser authenticateUser(Credentials credentials)
 
             // Get the username from the header configured in guacamole.properties
             String username = request.getHeader(confService.getHttpAuthHeader());
+            
+            //write username to the credentials object to make tokenfilter work
```
mike-jumper: Please follow the existing code style. Compare:

>     //write username to the credentials object to make tokenfilter work

vs. other comments in incubator-guacamole-client:

>     // Get the username from the header configured in guacamole.properties

Relative to the established style, this comment is missing the space following `//`, and should be in sentence case (first letter capitalized).

dotmaster:
Hi @necouchman, yes no reason why not to put it in the if block. I have moved it there.

extensions/guacamole-auth-header/src/main/java/org/apache/guacamole/auth/header/AuthenticationProviderService.java
dotmaster: @mike-jumper changed the comment style to the coding style guide.

necouchman:
Looks good to me.

extensions/guacamole-auth-header/src/main/java/org/apache/guacamole/auth/header/AuthenticationProviderService.java
```diff
@@ -73,6 +73,8 @@ public AuthenticatedUser authenticateUser(Credentials credentials)
             String username = request.getHeader(confService.getHttpAuthHeader());
 
             if (username != null) {
+                //  Write username to the credentials object to make tokenfilter work
```
mike-jumper: "to make tokenfilter work" sounds like this is a stopgap measure / hack just to get things working for a very specific case, which isn't really the kind of change that should ever be made to any part of the Guacamole codebase.

If `TokenFilter` is not working as expected because this value isn't populated, and the fact that it isn't populated is a bug, then that's a legitimate reason to make this change, and there's no need to call out the specific things that break because it's absent. If it's necessary, it's necessary.

If this truly is a hack, however, it might be worth instead looking into ways that `TokenFilter` could be modified to not depend in the username in the `Credentials` object. Grabbing the identifier of the `AuthenticatedUser` might be an alternative, though I haven't thought through whether that would go against the defined semantics of the `GUAC_USERNAME` token.

mike-jumper:
Please remember that each commit needs to tag the corresponding issue in JIRA, and that each commit message should describe at a high-level the change being made. The recently-added pair of commits are both missing the JIRA issue and have the same low-level message:

duplicate-commits-missing-tags

If two commits are so close to identical that they need the same message, they should probably be combined into one commit. Regardless of whether it makes sense to combine the commits, the message really needs to describe the nature of the change itself; something like "Update NameOfSomeFile.java" does not add any information beyond what git already records automatically. Commit messages should be aimed at describing the aspects which the various automated systems can't determine for you.

necouchman:
@dotmaster Any progress on this - opening a JIRA issue and tagging the commits correctly?

necouchman:
@dotmaster Ping...

necouchman:
@dotmaster: Any progress on fixing your commit messages?

extensions/guacamole-auth-header/src/main/java/org/apache/guacamole/auth/header/AuthenticationProviderService.java
necouchman: @mike-jumper Any thoughts here on whether it's preferable to set the username inside this module, or modify code elsewhere to use the AuthenticatedUser identifier?  Going the AuthenticatedUser route looks like it would require one of the following approaches:
- Another method in StandardTokens to be able to pass in the username token, specifically, with its own identifier, and then changes to the various places that StandardTokens is used to add both the credentials and then, alternatively the AuthenticatedUser code.
- Checks around the existing StandardTokens uses that make sure the Credentials object has a valid username, and then code to create a new object or modify the existing one around there using the AuthenticatedUser object.

It seems to me that setting it up inside the authentication module is the right way to go - it results in the fewest places that have to be reworked, and makes it available across the various places where those Credentials objects are used.  This module (auth-header) needs the fix, as will the CAS module.  I think those are the only two at the moment - any additional SSO-type modules would also have to keep it in mind (SAML, OAuth).

necouchman:
@dotmaster: Can you please update the commits with the JIRA issue, and also see Mike's previous comment about potentially combining the last two commits?

necouchman:
@dotmaster: Please update...

necouchman:
@dotmaster Ping.

mike-jumper:
This PR seems to have died, but I think I may have a general solution to this problem, automatically pulling the `GUAC_USERNAME` token from the `AuthenticatedUser`.

PR #176, opened 2017-07-16
GUACAMOLE-346: Avoid blocking main thread when seeking.

mike-jumper:
For sufficiently large recordings, the existing implementation of `Guacamole.SessionRecording` can potentially block the main thread to the point that the script appears to have hung. This change modifies `Guacamole.SessionRecording` such that seek operations are performed asynchronously, with completion of the seek reported through a callback.

The playback example has been modified accordingly, providing a notification and cancel button for in-progress seek operations.

### DIFF
--- a/doc/guacamole-playback-example/src/main/webapp/index.html
+++ b/doc/guacamole-playback-example/src/main/webapp/index.html
@@ -31,7 +31,13 @@
         <div class="player">
 
             <!-- Player display -->
-            <div id="display"></div>
+            <div id="display">
+                <div class="notification-container">
+                    <div class="seek-notification">
+                        <p>Seek in progress... Click "play" to cancel.</p>
+                    </div>
+                </div>
+            </div>
 
             <!-- Player controls -->
             <div class="controls">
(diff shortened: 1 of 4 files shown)


necouchman:
Nothing jumps out at me, but this is complex enough that it might need another reviewer.  @jmuehlner?  I'm happy to do the merge work on it...

doc/guacamole-playback-example/src/main/webapp/index.html
```diff
@@ -28,10 +28,19 @@
     <body>
 
         <!-- Guacamole recording player -->
-        <div class="player">
+        <div id="player">
```
jmuehlner: Not that I have a problem with making this an ID, but you could have used getElementsByClassName instead of changing this, right?

mike-jumper: Eh ... now that the example is referring to the player as a single, distinct object within the JS, that would require doing something like `getElementsByClassName('player')[0]`, which is semantically wrong.

It might make sense to apply the player functionality as a behavior to absolutely all elements having the "player" class, but that would make the example overly complex IMHO.

jmuehlner: Makes sense to me.

jmuehlner:
Ok, this all looks good to me. 👍

PR #177, opened 2017-07-17
GUACAMOLE-339: Add Remote Host to Connection History Table

necouchman:
This pull request adds the necessary fields and methods to the JDBC module to track the remote host in the connection history table, and to display it in the various history output tables in the client interface.


extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/schema/001-create-schema.sql
```diff
@@ -322,6 +322,7 @@ CREATE TABLE `guacamole_connection_history` (
   `history_id`           int(11)      NOT NULL AUTO_INCREMENT,
   `user_id`              int(11)      DEFAULT NULL,
   `username`             varchar(128) NOT NULL,
+  `remote_host`          varchar(128) DEFAULT NULL,
```
mike-jumper: Do you think we may need more that 128 characters?

Googling around a bit, the maximum length of any particular domain name in DNS is 253 characters, while fully-qualified hostnames are generally limited to 255 characters on UNIX and others. Based on that, I'd say this should be bumped up to 256.

necouchman: Done.

PR #178, opened 2017-07-23
GUACAMOLE-345: Modify schema scripts to be compatible with PostgreSQL 8.

mike-jumper:
There were two changes necessary to achieve compatibility:

1. Name all indexes. PostgreSQL 8 does not generate names automatically. See: https://www.postgresql.org/docs/8.4/static/sql-createindex.html
2. Replace hex literals with calls to `decode()` PostgreSQL 8 does not provide hex literals (only octal). See: https://www.postgresql.org/docs/8.4/static/datatype-binary.html


jmuehlner:
Looks good to me!

PR #179, opened 2017-08-03
GUACAMOLE-358: Fix issue with 404 errors from REST API

necouchman:
This pull request adds a new SimpleUserContext to the CAS extension, instead of returning null, which seems to resolve the issue reported in this JIRA issue with errors trying to access the Settings page in Guacamole.

### DIFF
--- a/guacamole/src/main/webapp/app/settings/directives/guacSettingsPreferences.js
+++ b/guacamole/src/main/webapp/app/settings/directives/guacSettingsPreferences.js
@@ -192,6 +192,9 @@ angular.module('settings').directive('guacSettingsPreferences', [function guacSe
                 $scope.canChangePassword = PermissionSet.hasUserPermission(permissions,
                         PermissionSet.ObjectPermissionType.UPDATE, username);
                         
+            })
+            .error(function permissionsFailed(error) {
+                $scope.canChangePassword = false;
             });
 
             /**


mike-jumper:
`getUserContext()` is supposed to be able to return `null` though: http://guacamole.incubator.apache.org/doc/guacamole-ext/org/apache/guacamole/net/auth/AuthenticationProvider.html#getUserContext-org.apache.guacamole.net.auth.AuthenticatedUser-

If returning `null` from `getUserContext()` results in the interface failing, it sounds to me like that is the bug to be addressed.

necouchman:
Hmmm...okay.  Well, if you look at the JIRA issue, you'll see the 404 error that gets returned from the current implementation of the CAS extension.  The error thrown is:
`
{"message":"Session not associated with authentication provider \"cas\".","translatableMessage":
Unknown macro: {"key"},"statusCode":null,"expected":null,"type":"NOT_FOUND"}
`
I ran into a similar issue on another extension I was messing with, and, in both cases, assigning *something* to the userContext object took care of the problem.  This is thrown in the GuacamoleSession class, in the getUserContext method.

mike-jumper:
Well, I have no doubt that this solves the problem, but since extensions are supposed to be able to return `null` for exactly this sort of case, I think we need to address it from that perspective. It's not a bug in the CAS extension, but rather a bug in the way the webapp is handling `null` return values from `getUserContext()`.

Should we be creating a skeleton `UserContext` automatically for any extensions which return `null` for `getUserContext()`? Or is it wrong for the part of the code making these calls to assume that there will be data associated with that auth provider?

necouchman:
> but rather a bug in the way the webapp is handling null return values from getUserContext()

Is it a bug in the webapp for handling null return values, or is it a bug in the GuacamoleSession code causing the exception to be thrown in the first place?

> Or is it wrong for the part of the code making these calls to assume that there will be data associated with that auth provider?

I would vote going the direction of allowing userContext to be null and handling the cases where it is null, instead of populating userContext with a skeleton, but I don't know if there are other implications to going that route??

jmuehlner:
@necouchman 

I'll have to dig into this code a bit more to come up with a more comprehensive answer, but I have to agree that not allowing the returned `UserContext` to be `null` probably isn't a good choice, since that is done by other extensions as well.

I think your suggested approach sounds reasonable.

necouchman:
Okay, so, on the JavaScript side, I believe the reason for the eternal spinning gear is this bit of code:

            // Retrieve current permissions
            permissionService.getPermissions(dataSource, username)
            .success(function permissionsRetrieved(permissions) {

                // Add action for changing password if permission is granted
                $scope.canChangePassword = PermissionSet.hasUserPermission(permissions,
                        PermissionSet.ObjectPermissionType.UPDATE, username);

            });

            /**
             * Returns whether critical data has completed being loaded.
             *
             * @returns {Boolean}
             *     true if enough data has been loaded for the user interface to be
             *     useful, false otherwise.
             */
            $scope.isLoaded = function isLoaded() {

                return $scope.canChangePassword !== null
                    && $scope.languages         !== null;

            };

So, basically the getPermissions call fails with the same 404 error as the getuser call, and, when it does, in this case, the isLoaded() function never resolves because the $scope.canChangePassword object continues to be null.

necouchman:
Okay, so I've rolled back the change to the CAS module and, instead, applied a change to the guacSettingsPreferences.js file that allows the settings page to load even if 404 errors are thrown and the permissions retrieval fails.

jmuehlner:
Does this same problem with the permissions call 404ing occur elsewhere, by the way? Is it only on this one page?

necouchman:
> Does this same problem with the permissions call 404ing occur elsewhere, by the way? Is it only on this one page?

The 404 error occurs in a couple of pages (it actually occurs during login if you watch the JavaScript console); however, the Preferences tab of the Settings page is the only place where I've seen the eternal spinning gear show up.  I'll ask on the JIRA issue and see if anyone else on the ticket has any feedback on that.

necouchman:
> I'll ask on the JIRA issue and see if anyone else on the ticket has any feedback on that.

Feedback from the user who opened the JIRA issue is that the Settings/Preferences area is the only place he has noticed any adverse behavior in the UI.

necouchman:
@mike-jumper @jmuehlner Any additional feedback?

guacamole/src/main/webapp/app/settings/directives/guacSettingsPreferences.js
```diff
@@ -192,6 +192,9 @@ angular.module('settings').directive('guacSettingsPreferences', [function guacSe
                 $scope.canChangePassword = PermissionSet.hasUserPermission(permissions,
                         PermissionSet.ObjectPermissionType.UPDATE, username);
                         
+            })
+            .error(function permissionsFailed(error) {
+                $scope.canChangePassword = false;
```
mike-jumper: The more I think on this, the more I think that 404 response is indeed the correct response, and that what you have here is the most correct approach. My original thoughts were to modify the REST API to ensure that the authenticated user can always be queried, but if an extension does not expose data at all, even data related to the authenticated user, 404 is really exactly what should be seen. It's the client-side which needs to be modified to properly understand and deal with that response. So ... looks good.

My only concern with these changes as they stand is that absolutely all errors from the REST API here will be interpreted as successful indications that the user lacks permission.

necouchman: So is there some additional checking I should do on the contents of the error object to see what's happening there, and do some interpretation?  Are there other specific errors that you'd expect to see coming out of this particular REST call?

mike-jumper: Eh ... unless @jmuehlner disagrees, I withdraw my concerns. I think these changes are fine.

There are ways to explicitly check for and handle the error response, as done in other parts of the admin interface, such that the user will see some useful error message describing the failure...  but in this case, I can't think of any such message which would be appropriate or helpful. It's just the password edit screen.

If the rest of the webapp's interface handled such errors and always warned the user when the service is unavailable (for example), that would be another matter, but that's not currently done, and doing so here would thus be out of scope.

PR #122, opened 2017-02-10
GUACAMOLE-197: Implement Support for RADIUS Authentication

necouchman:
This PR adds a RADIUS authentication module, allowing users to authenticate against RADIUS via a variety of protocols.  It includes support for TLS as well as for 2-factor via the radius AccessChallenge packet.  It uses JRadius as the RADIUS implementation.

PR #180, opened 2017-08-11
GUACAMOLE-355: Catch CAS Errors and Display Them

necouchman:
This PR catches the errors thrown by the CAS module and rethrows them as Guacmaole exceptions, and displays an error message in the login box.

### DIFF
--- a/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
+++ b/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
@@ -113,7 +113,7 @@ public class TicketValidationService {
         }
         catch (Throwable t) {
             logger.error("Error validating ticket with CAS server: {}", t.getMessage());
-            throw new GuacamoleInsufficientCredentialsException("Error validating ticket with CAS server.", t, CredentialsInfo.USERNAME_PASSWORD);
+            throw new GuacamoleInsufficientCredentialsException("CAS login failed.", CredentialsInfo.USERNAME_PASSWORD);
         }
 
     }
--- a/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
+++ b/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
@@ -34,7 +34,7 @@ import org.apache.guacamole.GuacamoleServerException;
 import org.apache.guacamole.auth.cas.conf.ConfigurationService;
 import org.apache.guacamole.net.auth.Credentials;
 import org.apache.guacamole.net.auth.credentials.CredentialsInfo;
-import org.apache.guacamole.net.auth.credentials.GuacamoleInsufficientCredentialsException;
+import org.apache.guacamole.net.auth.credentials.GuacamoleInvalidCredentialsException;
 import org.jasig.cas.client.authentication.AttributePrincipal;
 import org.jasig.cas.client.validation.Assertion;
 import org.jasig.cas.client.validation.Cas20ProxyTicketValidator;
@@ -113,7 +113,7 @@ public class TicketValidationService {
         }
         catch (Throwable t) {
             logger.error("Error validating ticket with CAS server: {}", t.getMessage());
-            throw new GuacamoleInsufficientCredentialsException("CAS login failed.", CredentialsInfo.USERNAME_PASSWORD);
+            throw new GuacamoleInvalidCredentialsException("CAS login failed.", CredentialsInfo.USERNAME_PASSWORD);
         }
 
     }
--- a/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
+++ b/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
@@ -33,6 +33,8 @@ import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.GuacamoleServerException;
 import org.apache.guacamole.auth.cas.conf.ConfigurationService;
 import org.apache.guacamole.net.auth.Credentials;
+import org.apache.guacamole.net.auth.credentials.CredentialsInfo;
+import org.apache.guacamole.net.auth.credentials.GuacamoleInsufficientCredentialsException;
 import org.jasig.cas.client.authentication.AttributePrincipal;
 import org.jasig.cas.client.validation.Assertion;
 import org.jasig.cas.client.validation.Cas20ProxyTicketValidator;
@@ -108,6 +110,10 @@ public class TicketValidationService {
         } 
         catch (TicketValidationException e) {
             throw new GuacamoleException("Ticket validation failed.", e);
+        }
+        catch (Throwable t) {
+            logger.error("Error validating ticket with CAS server: {}", t.getMessage());
+            throw new GuacamoleInsufficientCredentialsException("Error validating ticket with CAS server.", t, CredentialsInfo.USERNAME_PASSWORD);
         }
 
     }


extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
```diff
@@ -75,6 +84,10 @@ public String processUsername(String ticket) throws GuacamoleException {
         catch (TicketValidationException e) {
             throw new GuacamoleException("Ticket validation failed.", e);
         }
+        catch (Throwable t) {
+            logger.error("Error validating ticket with CAS server: {}", t.getMessage());
+            throw new GuacamoleInsufficientCredentialsException("Error validating ticket with CAS server.", t, CredentialsInfo.USERNAME_PASSWORD);
```
necouchman: Still not sure this is the right exception to throw.  It's the one that works - that actually displays an error on the dialog box - but still seems a little weird to throw Insufficient instead of Invalid.  However, throwing Invalid and getting the error would require some additional reworking of the index controller.

guacamole/src/main/webapp/app/login/styles/dialog.css
```diff
@@ -115,13 +115,6 @@
     background-image: url("images/guac-tricolor.png");
 }
 
-.login-ui.continuation div.login-dialog {
-    border-right: none;
-    border-left: none;
-    box-shadow: none;
-    max-width: 6in;
-}
-
```
necouchman: So, I removed all of this because I wasn't sure that the continuation piece of this actually does anything other than make the dialog box look a little goofy.  I can understand hiding the logo and version, which I did leave in place; however, this just takes off half of the borders, the shadow, and makes the box show up as an odd width (relative to the "normal" login box).  I also noticed this when working on the RADIUS module - when it prompts for the additional credentials from RADIUS.

Any reason not to pull this out?

mike-jumper: The main other piece of code which leverages the authentication continuation process is the database password reset. Does the password reset still look OK with these changes?

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
mike-jumper: Internal errors really shouldn't be exposed to users of the system. If the ticket can't be validated, then that should definitely be logged for the administrator's sake, but such internals shouldn't make their way to the interface.

Why not let the interface continue rendering all failures as "Invalid login"?

necouchman: Okay, I can make the error more generic.  However, throwing GuacamoleInvalidCredentialsException does not display the red "Invalid Login" error message when the username/password field is displayed - I think it has something to do with the fact that the CAS authentication redirects away from the Guacamole page and then back to it.  So, unless there's some trick to getting that to display, I have to stick with the InsufficientCredentials exception.

guacamole/src/main/webapp/app/login/styles/dialog.css
necouchman: I'll give it a try...

necouchman:
@mike-jumper Any additional feedback here?

guacamole/src/main/webapp/app/login/styles/dialog.css
mike-jumper: Ah, OK. Well, this doesn't look bad to me (though that text is a bit squished). Based on the screenshot, I don't have any major objection to the style changes if they are necessary.

necouchman: I could leave in the max-width option there and that would un-squish things a bit, if that's preferable.  I think the main thing was the shadow/box effect.

mike-jumper:
So ... the more-generic error message is an improvement, to be sure, but:

1. The login process already provides its own generic "Invalid login" message.
2. The "insufficient credentials" exception represents different semantics, and it definitely makes me squirm to use it as a vehicle to pass a user-facing message.

Perhaps I'm not understanding the root of these changes. Is the "invalid login" message not being displayed at all in the CAS case? If that's the case, then we should try to figure out what is resulting in that part of the system failing (as the auth system is supposed to render all auth-related errors as "invalid login").

Was the "invalid credentials" exception not working somehow?

necouchman:
> Perhaps I'm not understanding the root of these changes. Is the "invalid login" message not being displayed at all in the CAS case? If that's the case, then we should try to figure out what is resulting in that part of the system failing (as the auth system is supposed to render all auth-related errors as "invalid login").

When I use invalid login, the error message does not get displayed at all.  I believe this is because of the full redirection over to the CAS login page that happens, and then the redirect back.  The way the Guacamole webapp is currently set up, it's not supposed to display the initial Invalid Login error message the first time the page loads, as this would result in an error message every single time you loaded the login page, even without the CAS module, since the way the login is triggered in the first place is to throw the 403 error and send an invalid exception.

I'm not sure if there's a good way to detect whether it's a redirect in the CAS module or something like that, so that the behavior only happens when the CAS module is loaded, but, as it stands today, throwing the Invalid Login exception results in the login page being shown, but no error message.

guacamole/src/main/webapp/app/login/styles/dialog.css
necouchman: Added back the max-width tag so that the message isn't quite so squished, but kept the shadow/box around it.

necouchman:
@mike-jumper Any further insight into this or suggestions on which direction to go?

mike-jumper:
I'm still confused about what behavior this change is aiming for. The [GUACAMOLE-355](https://issues.apache.org/jira/browse/GUACAMOLE-355) issue in JIRA states:

> As reported by Kaushnik on the dev mailing list, attempting to use the guacamole-auth-cas module with an incorrect CAS authentication endpoint results in very generic and not very useful 500 Internal Server errors rather than meaningful messages that point the user in the right direction for correcting the error. ...

It is the admin that needs to be pointed in the correct direction (via proper log messages and, ideally, proper exceptions behind them), while the user should not be exposed to internals.

If the server-side log messages do not properly describe the problem, then that needs to be fixed for sure, but then why the focus on client-side messages?

necouchman:
> If the server-side log messages do not properly describe the problem, then that needs to be fixed for sure, but then why the focus on client-side messages?

I guess my reasoning for the user getting some sort of useful message is that they're the ones who are going to hit the web URL, get redirected to the expected SSO page, try to log in, and then be taken back to the Guacamole login page with no explanation as to why the SSO login did not work.  I'm not saying that the user needs a ton of detail on why the login failed - certainly it should be a fairly generic error message - but I think they need some sort of indication as to the fact that the SSO login failed that they can then report back to the admin, who can then look at the logs and figure out why.  If the user hits the web UI, gets taken to SSO, logs in, then gets another login box with no error at all, it seems to me this is going to lead to confusion and lost time on the part of the user figuring out that something is actually wrong in the first place, let alone that it's an SSO issue.  They're just going to say "It isn't working" - my favorite type of error report :-).

mike-jumper:
> ...  They're just going to say "It isn't working" - my favorite type of error report :-).

Well ... mine, too, but we can't log internal details in a user-visible manner, and one of those details is the nature of the SSO system. It wouldn't be OK to display an error like "MySQL database unreachable" or "Unable to connect to LDAP server", for example.

Assuming that there is a message which is specific enough to be useful, but generic enough to not reveal internal details to users ... there has to be a better way than manually throwing a `GuacamoleInsufficientCredentialsException` requesting a username/password pair, which is semantically incorrect. There *is* a mechanism for generic error reporting built into REST services, and I think the token service used for authentication will produce those sorts of error responses (JSON which can be parsed by the client-side JavaScript).

For normal username/password authentication, the accepted safe behavior is to handle internal failures as if the username/password are invalid. This may seem innocuous (why not tell the user that something is failing?), but doing otherwise can reveal system internals which could lead to the exploitation of a vulnerability. A malicious user that gets "Invalid login" when trying to log in as "foobar", but "Internal error" when trying to log in as "; DROP TABLE foo; --" would successfully determine that the server likely vulnerable to SQL injection.

What would be the SSO or CAS equivalent to login failure?

necouchman:
> ...This may seem innocuous (why not tell the user that something is failing?), but doing otherwise can reveal system internals which could lead to the exploitation of a vulnerability. A malicious user that gets "Invalid login" when trying to log in as "foobar", but "Internal error" when trying to log in as "; DROP TABLE foo; --" would successfully determine that the server likely vulnerable to SQL injection.

That makes sense.

> What would be the SSO or CAS equivalent to login failure?

Probably just something like SSO Login Failed or Invalid SSO Login would do the trick - something to indicate that the SSO login is failing, which would help the user help the admin pinpoint the failure, but avoid exposing the nature of the failure to an attacker.

mike-jumper:
Not sure... Even stating "SSO" here makes me uneasy. Stating "SSO Login Failed" does provide additional information, which could be of potential use to an attacker if they are able to trigger that failure.

If the goal is that the user should be able to point the admin toward the cause of the failure, that seems in conflict with security best practices. A user should not be able to distinguish between Guacamole-side failure of SSO and refusal of the SSO system to authenticate the user. Only the admin should be able to distinguish these cases via access to the logs.

In the case of failure, why not discard the received CAS token entirely (if any), just as would be done if the token itself were invalid, and send the user back to the CAS system to re-authenticate?

Or would that result in a redirect loop...

necouchman:
> Or would that result in a redirect loop...

Well, it kind of depends on the nature of the failure.  If the failure is similar to the events that precipitated this particular PR, then, yes, it would result in a redirect loop - the user would be sent to login, which succeeds, then back to Guacamole, which would do the ticket validation and fail because the URL is wrong, which would then clear the token and send them back to the CAS login, which would report success since the user is already logged in, send them back to Guac...etc.

Other than that, I see some of the points about providing information to potential attackers.  I'm not sure I see the risk in very generic things like stating that the SSO login failed, but I'll defer to your expertise on it, and I'd rather get this PR through in some form that makes error handling a little more sane for these situations.

So, should I just make it throw a GuacamoleInvalidCredentialsException and ask for a username and password?  Or is it worth implementing a new exception class?  Or stick with InsufficientCredentials?

-Nick

necouchman:
@mike-jumper Ping - any feedback on my last question?

mike-jumper:
Is it possible that CAS may produce an invalid token under normal circumstances? That is, can there be a case where the proper behavior for receipt of an invalid token would be to redirect back to CAS for a fresh token, and replacing that behavior with an error screen would break that?

necouchman:
> Is it possible that CAS may produce an invalid token under normal circumstances? That is, can there be a case where the proper behavior for receipt of an invalid token would be to redirect back to CAS for a fresh token, and replacing that behavior with an error screen would break that?

Yes, both the tokens and the actual CAS login can expire, which results in users being redirected back to the CAS page either for the token refresh or for a CAS login.  The main situation would, I think, be the later, where the application (Guacamole) requests a ticket and then it gets pushed back to the CAS login screen because the CAS session has timed out.

mike-jumper:
Per [recent discussion on the mailing lists](https://lists.apache.org/thread.html/d75b09f4636789bd38ed2eff40ed27dd36faa7fbf200c701d202a4b3@%3Cdev.guacamole.apache.org%3E), the release process for 0.9.14-incubating has begun, and the issue associated with this change ([GUACAMOLE-355](https://issues.apache.org/jira/browse/GUACAMOLE-355)) is within scope. Please update the merge base of this PR to point to `staging/0.9.14-incubating`.

necouchman:
So, I went to the `GuacamoleInvalidCredentialsException` route here for dealing with SSO failures.  The user won't get any sort of error message on the screen, but the admin will get one logged in the log file and the user will get the normal Guacamole username/password dialog.

mike-jumper:
Sounds perfect. It's unfortunate that exception handling within the CAS library has to be so crazy.

guacamole/src/main/webapp/app/login/styles/dialog.css
```diff
@@ -116,9 +116,6 @@
 }
 
 .login-ui.continuation div.login-dialog {
-    border-right: none;
-    border-left: none;
-    box-shadow: none;
```
mike-jumper: Is this still necessary?

necouchman: For this particular request, no, it doesn't matter, so I'll pull that change.  However, it may make it into the RADIUS change (#122), as having these entries here makes the subsequent login dialogs look a little odd.

necouchman:
Hopefully good, now.

PR #181, opened 2017-08-12
GUACAMOLE-38: Implement on-demand connections

necouchman:
This is an extension that implements the on-demand (ad-hoc, "QuickConnect") feature, which places a connection bar on the home page and allows users to type in a URI to establish a connection.  Everything is in-memory, so connections do not persist after logout, and there are no records kept of the connections.

The original JIRA issue was closed as a duplicate of parameter prompting, but reopened it as I think they're different enough, and went ahead and wrote it, anyway, and figured I'd see if there was any interest in merging it into the code.


extensions/guacamole-auth-quickconnect/src/main/resources/controllers/quickconnectController.js
```diff
@@ -0,0 +1,186 @@
...
+         * [5] - Username
+         * [7] - Password
+         * [8] - Hostname
+         * [10] - Port
+         * [11] - Path
+         * [13] - Document
+         * [15] - Parameters
+         * [17] - JS Route
+         */
+        var regexURL = /^(((rdp|ssh|telnet|vnc)?):\/)?\/?((.*?)(:(.*?)|)@)?([^:\/\s]+)(:([^\/]*))?((\/\w+\/)*)([-\w.\/]+[^#?\s]*)?(\?([^#]*))?(#(.*))?$/g;
```
mike-jumper: Whoa.

necouchman: Ha.  Yeah.  If you have suggestions for better ways to parse out a URI, I'm open.  That regex right there is the result of a combo of Google searches and working with the regex on regex101.com.

necouchman: Obviously it would be better if the available protocols there were parsed from a REST API check of supported protocols instead of hard-coded.

mike-jumper: On a similar note, any reason why you chose to handle this client-side vs. server-side? I'm not sure whether one is better than the other at the moment, but I am curious. Might be worth considering.

necouchman: I think the reason I did it that way was because I was trying to leverage as much as possible of the existing REST endpoints for creating/saving the connections, which seems to be easier if the connection attributes and parameters are parsed out on the client-side and then fed to the existing API.

I suppose I could leverage the extension-specific REST API and feed the entire connection string to a REST endpoint and then parse it out there.  Or is there some other way I'm not thinking about that already exists to get the data to the server-side for parsing by the extension?

necouchman: > I'm not sure whether one is better than the other at the moment, but I am curious.

Well, since Java already has a URI class and is capable of parsing out the URI without the...colorful...regex, I think I might be leaning toward redoing it server-side.  JavaScript does have some basic parsing available through the DOM, and some external libraries available for it, but it doesn't seem nearly as robust as the Java library.

necouchman:
@mike-jumper: Is this something you're interested in having in the Guacamole client (with changes), or is this something that seems out-of-scope to you?  I'm happy to close out this PR and maintain this extension myself - I find it very useful, but that may just be me.  If there's not interest in having it as part of the guacamole-client codebase I'll be happy to close it out so that it doesn't distract from other PRs.

mike-jumper:
It seems in-scope. I don't think it would make sense for this to be the default behavior of the webapp, but I definitely think it will be a welcome extension.

Have things changed in your approach since your initial push? I see some new commits have appeared.

necouchman:
Cool, I'll continue to push forward with this.

I've made some progress, but nothing worth any more review at this point - will ping back once I finish it up.  I just wanted to make sure this was worth the continued review before I pushed on.

Thanks!

necouchman:
@mike-jumper: Okay, I think this is ready for a review, now - the parsing has been implemented server-side, and seems to be functional.  Let me know what needs to be tweaked.

necouchman:
@mike-jumper Ping, when you get a chance to review.  Thx...

necouchman:
@mike-jumper <Gentle Nudge>

Or, @jmuehlner, @ceharris, @flangelo - anyone else care to review/comment?

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectDirectory.java
```diff
@@ -0,0 +1,158 @@
...
+     *     The QuickConnection object to add to the tree.
+     *
+     * @return
+     *     The connectionId of the object added to the directory.
+     *
+     * @throws GuacamoleException
+     *     If an error is encountered adding the object to the
+     *     directory.
+     */
+    public String put(QuickConnection object) throws GuacamoleException {
```
ceharris: Replace `object` with `connection`?

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnection.java
```diff
@@ -0,0 +1,181 @@
...
+
+    @Override
+    public void setAttributes(Map<String, String> attributes) {
+        // Do nothing - there are no attributes
+    }
+
+    @Override
+    public GuacamoleTunnel connect(GuacamoleClientInformation info)
+            throws GuacamoleException {
+
```
ceharris: This is not a comment about the quick connect extension so much as a comment about whether there is a generalization here that should be abstracted. There's a body of about 40 lines of code here that has nothing to do with quick connect. Isn't this behavior (using the environment to get the proxy config and figure out what's needed to create an appropriate Guacamole socket) something that should be either pulled up into the abstract base type or otherwise moved into a reusable component?

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectDirectory.java
necouchman: Yep, you/re right - this should be fixed, now.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnection.java
necouchman: Yes, you're probably right, and it's probably already implemented in some other connection.  Let me see if there's some other class that I should be `extend`ing or `implement`ing that already has this rather than having it in here.

necouchman: So, I think when I originally wrote this I copied the `org.apache.guacamole.net.auth.simple.SimpleConnection` class over to this QuickConnection class, instead of just extending it.  I've refactored it to just extend the SimpleConnection object and override where I need to instead of duplicaitng the code.  Thanks for pointing that out!

necouchman:
Thanks @ceharris - I've tweaked the couple of items you pointed out there!  Let me know if anything else needs attention.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectDirectory.java
```diff
@@ -0,0 +1,158 @@
...
+
+    /**
+     * Returns the current counter and then increments it.
+     *
+     * @return
+     *     An Integer representing the next available connection
+     *     ID to get used when adding connections.
+     */
+    private Integer getNextConnectionID() {
+        return CONNECTION_ID++;
```
mike-jumper: This is not threadsafe. If multiple calls to `add()` overlap, this value will not be updated as expected. You will need to generate these IDs in a threadsafe manner, or switch to a mechanism that guarantees uniqueness without depending on synchronous updates ([`UUID.randomUUID()`](https://docs.oracle.com/javase/7/docs/api/java/util/UUID.html#randomUUID()) for example).

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectDirectory.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    @Override
+    public void add(Connection object) throws GuacamoleException {
+
+        put(new QuickConnection(object));
```
mike-jumper: Out of curiosity, is it an intended feature that these connections persist in memory until the end of the session? Or is that a side effect of the implementation, with the only goal being to provide some way to connect in an ad hoc manner?

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectDirectory.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /**
+     * Returns the current counter and then increments it.
+     *
+     * @return
+     *     An Integer representing the next available connection
+     *     ID to get used when adding connections.
+     */
+    private Integer getNextConnectionID() {
```
mike-jumper: Really should use `int` unless there is a reason a wrapper class like `Integer` is necessary.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectDirectory.java
ceharris: You want to be careful with `UUID.randomUUID()`. The underlying implementation uses `SecureRandom` which can block indefinitely waiting for sufficient entropy (e.g. in the kernel implementation of `/dev/random`). On a system with low activity this can be significant concern. 

`AtomicInteger.getAndIncrement` will give the same semantics as your use of the post-increment operator with thread safety. Performance is generally better than what you can do with your own `ReentrantLock` or (not recommended) an approach using `synchronized`.

mike-jumper: Not persisting connections at all, and instead dynamically generating the returned `Connection` within `get()` (using the URL as the identifier) would be another option, assuming that persisting the connections in memory is not an intended feature.

necouchman: Went the AtomicInteger route - having the connections persist in-memory is (to me) a desired functionality, so that accomplishes thread-safeness without losing that.  Also cleaned up the code (like my oddly-named all-caps CONNECTION_ID) a little.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectDirectory.java
necouchman: It was probably initially a side-effect, but one that I liked so I made it intentional.  How's that?! :-)

Obviously the main goal is to connect in an ad-hoc manner; however, I think it's also useful to be able to quickly retrieve those connections in the same session - i.e., I get kicked out of my session due to a network issue and want to quickly reconnect without typing the entire connection string, again.

Also, in one iteration of the implementation I had a little box where it looked for other available connection directories, and, if you had permission, allowed you to use the ad-hoc box to create a connection within the directory.  I don't think I kept it, here - needs a little more work - but it might be something that gets tacked (back) on in the future.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectDirectory.java
necouchman: Switched to `int`.  I think the reason I went `Integer` is because there were a couple of places where it gets converted to a `String`.  I've replaced those with `Integer.toString()` calls with the primitive, so accomplishes the same thing.

necouchman:
Okay, more tweaks - let me know what else needs to be cleaned up!

necouchman:
@mike-jumper These last updates look okay?  Anything else I need to work on?

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
```diff
@@ -0,0 +1,169 @@
...
+            port = DEFAULT_URI_PORT;
+
+        if (query != null && !query.equals(""))
+            paramList = Arrays.asList(query.split("&"));
+
+        if (userInfo != null && !userInfo.equals("")) {
+            String[] authenticators = userInfo.split(":");
+            if (authenticators.length > 0 && authenticators[0] != null)
+                username = authenticators[0];
+            if (authenticators.length > 1 && authenticators[1] != null)
```
mike-jumper: What if `authenticators` is something like "myusername:myPassword:That:Has:Colons"? Rather than manually parse everything with `split()`, I suggest defining (and pre-compiling) regular expressions. It would avoid recompiling the pattern each time `split()` is invoked, and may be more robust.

Alternatively, though I haven't looked into this, there may be a better class for parsing URIs which handles all of this. It doesn't look like anything we're doing here is non-standard with respect to URI format.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
```diff
@@ -0,0 +1,169 @@
...
+        URI qcUri;
+        try {
+            qcUri = new URI(uri);
+        }
+        catch (URISyntaxException e) {
+            throw new GuacamoleClientException("Invalid URI Syntax", e);
+        }
+        String protocol = qcUri.getScheme();
+        String host = qcUri.getHost();
+        Integer port = qcUri.getPort();
```
mike-jumper: You should avoid using wrapper classes like this unnecessarily.

extensions/guacamole-auth-quickconnect/src/main/resources/templates/quickconnectField.html
```diff
@@ -0,0 +1,9 @@
+<meta name="before" content=".recent-connections" />
+<div class="header" ng-controller="quickconnectController">
+    <div class="quickconnect-container">
+        <form ng-submit="quickConnect()">
+            <input type=text class="quickconnect-field" placeholder="Enter Connection URI" ng-model="uri" />
```
mike-jumper: Strings which will be displayed to the user need to be defined as translation strings, rather than hard-coded in the HTML.

extensions/guacamole-auth-quickconnect/src/main/resources/styles/quickconnect.css
```diff
@@ -0,0 +1,48 @@
...
+    padding: 0.5em;
+    padding-left: 2.25em;
+    width: 100%;
+    max-width: none;
+    border: 0;
+    border-left: 1px solid rgba(0,0,0,0.125);
+    box-sizing: border-box;
+}
+
+.quickconnect-field input[type="submit"] {
```
mike-jumper: Is there an `<input type="submit" ...>`? I don't see one in the HTML.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/rest/QuickConnectREST.java
```diff
@@ -0,0 +1,102 @@
...
+     *     The UserContext object associated with this REST endpoint
+     *
+     * @throws GuacamoleException
+     *     If the UserContext is unavailable or the directory object
+     *     cannot be retrieved.
+     */
+    public QuickConnectREST(QuickConnectUserContext userContext)
+            throws GuacamoleException {
+        this.userContext = userContext;
+        this.directory = (QuickConnectDirectory)this.userContext.getConnectionDirectory();
```
mike-jumper: If usages of `QuickConnectUserContext` must rely on `getConnectionDirectory()` returning a `QuickConnectDirectory`, it would be better to avoid requiring typecasting and explicitly declare that as the return type.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/rest/QuickConnectREST.java
```diff
@@ -0,0 +1,102 @@
...
+     *
+     * @throws GuacamoleException
+     *     If an error is encountered parsing the URI.
+     */
+    @POST
+    @Path("create")
+    public String create(@FormParam("uri") String uri) 
+            throws GuacamoleException {
+
+        if (directory == null)
```
mike-jumper: Will this ever be the case?

extensions/guacamole-auth-quickconnect/pom.xml
```diff
@@ -0,0 +1,209 @@
...
+            <groupId>org.apache.guacamole</groupId>
+            <artifactId>guacamole-ext</artifactId>
+            <version>0.9.14</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <!-- Jersey - JAX-RS Implementation -->
+        <dependency>
+            <groupId>com.sun.jersey</groupId>
+            <artifactId>jersey-server</artifactId>
```
mike-jumper: You actually don't need to pull in `jersey-server` just for JAX-RS annotations. You can pull just the API containing the annotations, and declare the scope as "provided" so it doesn't get bundled:

    
    <dependency>
        <groupId>javax.ws.rs</groupId>
        <artifactId>jsr311-api</artifactId>
        <version>1.1.1</version>
        <scope>provided</scope>
    </dependency>

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectAuthenticationProvider.java
```diff
@@ -0,0 +1,91 @@
...
+     *     Returns null, which causes the client to move on to the next
+     *     module.
+     */
+    @Override
+    public AuthenticatedUser authenticateUser(Credentials credentials)
+        throws GuacamoleException {
+
+        String username = credentials.getUsername();
+        if (username == null || username.isEmpty())
+            throw new GuacamoleInvalidCredentialsException("You must login.", CredentialsInfo.USERNAME_PASSWORD);
```
mike-jumper: If this extension will not perform authentication, why does it request a username and password?

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectAuthenticationProvider.java
```diff
@@ -0,0 +1,91 @@
...
+import org.apache.guacamole.net.auth.simple.SimpleAuthenticationProvider;
+import org.apache.guacamole.net.auth.UserContext;
+import org.apache.guacamole.protocol.GuacamoleConfiguration;
+
+/**
+ * Class providing the necessary hooks into the Guacamole Client authentication
+ * process so that the QuickConnect functionality can be initialized and be used
+ * throughout the web client.
+ */
+public class QuickConnectAuthenticationProvider extends SimpleAuthenticationProvider {
```
mike-jumper: As you have your own underlying `UserContext` implementation (a good thing), you should not be extending `SimpleAuthenticationProvider` here.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectConnectionGroup.java
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.guacamole.net.auth.AbstractConnectionGroup;
+import org.apache.guacamole.net.auth.ConnectionGroup;
+import org.apache.guacamole.protocol.GuacamoleClientInformation;
+
+/**
+ * Provides a very simple, single-level connection group used
+ * for temporarily storing the QuickConnections created by
+ * users.
+ */
+class QuickConnectConnectionGroup extends AbstractConnectionGroup {
```
mike-jumper: Not necessarily an issue, but to confirm: is this intentionally package-private?

necouchman: Well, it's never expected to be used outside this package, but neither are a bunch of the other classes, so I should probably be consistent about it...

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectAuthenticationProvider.java
necouchman: ...even though doing so greatly reduces the amount of code I have to write to overwrite methods that are doing exactly the same thing as `SimpleAuthenticationProvider` - like `shutdown()` and `decorate()` and `redecorate()` and probably a couple of others that I would basically just be copying/pasting code out of `SimpleAuthenticationProvider` to duplicate that functionality?

mike-jumper: I get where you're coming from, and reducing duplication is good, but what you need is an actual base implementation of `AuthenticationProvider` which guarantees the default behavior you're looking for. `SimpleAuthenticationProvider` is not guaranteed or intended to provide this, and is only defined with respect to `getAuthorizedConfigurations()`.

A base, empty implementation of `AuthenticationProvider` would probably be a good thing (similar to `AbstractUser` and others). If you're looking to avoid duplication, perhaps we should:

* Write an `AbstractAuthenticationProvider` (perhaps also `AbstractUserContext` if it helps)
* Refactor `SimpleAuthenticationProvider` to build off that if appropriate.
* Avoid using `SimpleAuthenticationProvider` here

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectAuthenticationProvider.java
necouchman: Good question, to which I do not have an answer.  Removed.

extensions/guacamole-auth-quickconnect/pom.xml
necouchman: Good to know.  Changed.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.Arrays;
+import java.util.List;
+import org.apache.guacamole.GuacamoleClientException;
+import org.apache.guacamole.GuacamoleException;
+import org.apache.guacamole.protocol.GuacamoleConfiguration;
+
+public class QCParser {
+
+    /**
+     * The default protocol to parse to if one is undefined.
```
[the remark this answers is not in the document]
necouchman: Cleaned this up.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
```diff
@@ -0,0 +1,169 @@
...
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Arrays;
+import java.util.List;
+import org.apache.guacamole.GuacamoleClientException;
+import org.apache.guacamole.GuacamoleException;
+import org.apache.guacamole.protocol.GuacamoleConfiguration;
+
+public class QCParser {
```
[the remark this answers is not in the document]
necouchman: Documented.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
```diff
@@ -0,0 +1,169 @@
...
+
+    /**
+     * The default host to use if one is not defined.
+     */
+    public static final String DEFAULT_URI_HOST = "localhost";
+
+    /**
+     * The default port to use if one is not defined.
+     */
+    public static final Integer DEFAULT_URI_PORT = 22;
```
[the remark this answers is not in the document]
necouchman: Removed.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
necouchman: I haven't found a class that parses the username and password into separate components - even the URI and URL classes that have methods for grabbing the "user-info" part of the string (including both username and password) don't seem to have methods for splitting it out to individual components.  I'll work on the regex route.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectAuthenticationProvider.java
necouchman: Ah, okay, that makes sense.  I guess I'll just switch it over to `implements AuthenticationProvider` for now - I think it would probably make sense at some point in the future to create the `AbstractAuthenticationProvider` and `AbstractUserContext` implementations, but it seems like that might be more appropriate in a separate change.

necouchman: Implemented.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/rest/QuickConnectREST.java
necouchman: Certainly should not.  Removed.

extensions/guacamole-auth-quickconnect/src/main/resources/templates/quickconnectField.html
necouchman: Done.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/rest/QuickConnectREST.java
necouchman: Fixed.  I think I was originally avoiding implementing my own connection directory and trying to use one of the provided ones, and that didn't pan out and I forgot to go back and fix stuff.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
necouchman: Switched to int.

extensions/guacamole-auth-quickconnect/src/main/resources/styles/quickconnect.css
necouchman: Nope.  Removed.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
necouchman: Okay, implemented regex and Pattern/Matcher for this.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
```diff
@@ -0,0 +1,169 @@
...
+
+        if (port == -1 || port < 1)
+            port = DEFAULT_URI_PORT;
+
+        if (query != null && !query.equals(""))
+            paramList = Arrays.asList(query.split("&"));
+
+        if (userInfo != null && !userInfo.equals("")) {
+            String[] authenticators = userInfo.split(":");
+            if (authenticators.length > 0 && authenticators[0] != null)
```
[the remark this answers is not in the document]
necouchman: Should be cleaned up, now, with move to regex.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
```diff
@@ -0,0 +1,169 @@
...
+            host = DEFAULT_URI_HOST;
+
+        if (port == -1 || port < 1)
+            port = DEFAULT_URI_PORT;
+
+        if (query != null && !query.equals(""))
+            paramList = Arrays.asList(query.split("&"));
+
+        if (userInfo != null && !userInfo.equals("")) {
+            String[] authenticators = userInfo.split(":");
```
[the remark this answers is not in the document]
necouchman: Switched to RegEx.

necouchman:
Okay, next round completed.  What's next? :-)

necouchman:
> What's next? :-)

Refactoring this based on the recent changes:
- `AbstractUserContext` and `AbstractAuthenticationModule` implementatoins
- Deprecation of `Simple*Directory` classes in favor of `SimpleDirectory<*>`

necouchman:
Okay @mike-jumper and @ceharris (and whomever else wants to review it!), this has been rebased onto the master and reworked with the recent `AbstractUserContext` and `AbsractAuthenticationProvider` classes, and also the removal of the deprecated `Simple*Directory` classes.  Net reduction of 150 lines of code :-).  Let me know what else needs to be cleaned up, here.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectAuthenticationProviderModule.java
```diff
@@ -0,0 +1,81 @@
...
+     * class.
+     *
+     * @param authProvider
+     *     The authentication provider for which injection is being
+     *     configured.
+     *
+     * @throws GuacamoleException
+     *     If an error occurs while retrieving the server environment.
+     */
+    public QuickConnectAuthenticationProviderModule(AuthenticationProvider authProvider)
```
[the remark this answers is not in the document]
mike-jumper: Actually, I also don't see `@Inject` anywhere. Is Guice being used here at all? Perhaps this is unneeded.

extensions/guacamole-auth-quickconnect/src/licenses/NOTICE
```diff
@@ -0,0 +1,5 @@
+Apache Guacamole
+Copyright 2017 The Apache Software Foundation
```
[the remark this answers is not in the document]
necouchman: Just reliving the glory days.  Fixed.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectAuthenticationProvider.java
```diff
@@ -0,0 +1,56 @@
...
+        return "quickconnect";
+    }
+
+    @Override
+    public UserContext getUserContext(AuthenticatedUser authenticatedUser)
+            throws GuacamoleException {
+
+        if (userContext == null)
+            userContext = new QuickConnectUserContext(this,
+                    authenticatedUser.getIdentifier());
```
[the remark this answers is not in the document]
necouchman: Should be good, now.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectAuthenticationProviderModule.java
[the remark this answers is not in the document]
necouchman: Removed the whole class.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectUserContext.java
```diff
@@ -0,0 +1,149 @@
...
+        // preexisting connections, and the single root group.
+        this.self = new SimpleUser(username,
+            Collections.<String>emptyList(),
+            Collections.singleton(DEFAULT_ROOT_CONNECTION_GROUP)
+        );
+
+        // Initialize each of the directories associated with the userContext.
+        this.userDirectory = new SimpleDirectory<User>(self);
+        this.connectionDirectory = new QuickConnectDirectory(this.rootGroup);
+        this.connectionGroupDirectory = new SimpleDirectory<ConnectionGroup>(Collections.singleton(this.rootGroup));
```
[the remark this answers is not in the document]
necouchman: Well we should probably not bother doing that, then.  Removed.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnection.java
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * A type of Connection specific to this authentication extension.
+ */
+public class QuickConnection extends SimpleConnection {
+
+    /**
+     * Number of active connections.
+     */
+    private int activeConnections;
```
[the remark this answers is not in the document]
necouchman: If by "this" you mean `this` as in the entire class, then, yes.  Removed this and just using `SimpleConnection` class.  I was thinking about doing something with active connections in the QuickConnect extension once upon a time, but it doesn't seem like it has much value.

extensions/guacamole-auth-quickconnect/src/main/resources/controllers/quickconnectController.js
```diff
@@ -0,0 +1,75 @@
...
+     */
+    var ACKNOWLEDGE_ACTION = {
+        name        : "MANAGE_CONNECTION.ACTION_ACKNOWLEDGE",
+        // Handle action
+        callback    : function acknowledgeCallback() {
+            guacNotification.showStatus(false);
+        }
+    };
+
+    $scope.uri = null;
```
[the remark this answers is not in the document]
necouchman: Documented.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
```diff
@@ -0,0 +1,187 @@
...
+     * for the configuration based on the protocol, host, user
+     * and port in the configuration, and return the string value.
+     *
+     * @param config
+     *     The GuacamoleConfiguration object to use to generate
+     *     the name.
+     *
+     * @return
+     *     The String value of the name that is generated, or
+     *     null if no config is provided.
```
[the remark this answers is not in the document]
necouchman: Reworded.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
```diff
@@ -0,0 +1,187 @@
...
+            return null;
+
+        String protocol = config.getProtocol();
+        String host = config.getParameter("hostname");
+        String port = config.getParameter("port");
+        String user = config.getParameter("username");
+
+        String name = "";
+
+        if (protocol != null && !protocol.equals(""))
```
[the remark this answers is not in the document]
necouchman: Changed it throughout the class.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
```diff
@@ -0,0 +1,187 @@
...
+            name += protocol + "://";
```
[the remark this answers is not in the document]
necouchman: Will work on this...

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
```diff
@@ -0,0 +1,187 @@
...
+        if (username != null && username.length() > 0)
+            qcConfig.setParameter("username", username);
+
+        if (password != null && password.length() > 0)
+            qcConfig.setParameter("password", password);
+
+        if (paramList != null) {
+            for (String parameter : paramList) {
+                String[] paramArray = parameter.split("=");
+                qcConfig.setParameter(paramArray[0],paramArray[1]);
```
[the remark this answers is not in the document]
necouchman: Added the limit parameter to this split, which will make sure the second `=` is part of the value.  This doesn't totally solve the issue, as you mention below...

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
```diff
@@ -0,0 +1,187 @@
...
+        List<String> paramList = null;
+
+        if (protocol == null || protocol.equals(""))
+            protocol = DEFAULT_URI_PROTOCOL;
+
+        if (host == null || host.equals(""))
+            host = DEFAULT_URI_HOST;
+
+        if (query != null && !query.equals(""))
+            paramList = Arrays.asList(query.split("&"));
```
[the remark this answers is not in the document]
necouchman: Yeah, good point.  Any pointers for how to handle this?

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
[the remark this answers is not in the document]
necouchman: StringBuilder built.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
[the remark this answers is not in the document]
necouchman: Okay, I took a stab at handling encoded parameters.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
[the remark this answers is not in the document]
necouchman: Okay, hopefully between this tweak (adding the `limit` parameter) and the changes to address encoding, this is properly handled, now.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
```diff
@@ -0,0 +1,204 @@
...
+            qcConfig.setParameter("port", Integer.toString(port));
+
+        if (username != null && !username.isEmpty())
+            qcConfig.setParameter("username", username);
+
+        if (password != null && !password.isEmpty())
+            qcConfig.setParameter("password", password);
+
+        if (paramList != null) {
+            for (String parameter : paramList) {
```
ceharris: It might make sense and would aid readability here if the parsing of the query were delegated to another method. It also seems like the kind of code that has lots of edge cases that would benefit greatly from some unit tests.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
```diff
@@ -164,23 +165,23 @@ public static String getName(GuacamoleConfiguration config)
         String port = config.getParameter("port");
         String user = config.getParameter("username");
 
-        String name = "";
+        StringBuilder name = new StringBuilder("");
```
ceharris: Just a minor nit, but `new StringBuilder()` starts out empty.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
necouchman: I can definitely break out the query parsing into its own method.

Never done unit tests before...another new experience for me :-).  I know there are a couple of examples in the rest of the Guacamole code that I can probably use as a starting place.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
```diff
@@ -131,7 +141,14 @@ public static GuacamoleConfiguration getConfiguration(String uri)
         if (paramList != null) {
             for (String parameter : paramList) {
                 String[] paramArray = parameter.split("=", 2);
-                qcConfig.setParameter(paramArray[0],paramArray[1]);
+                try {
+                    qcConfig.setParameter(URLDecoder.decode(paramArray[0], "UTF-8"),
```
ceharris: Might make sense to apply `URLDecoder.decode` to the entire query string once.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
necouchman: Okay, basic unit testing implemented - that wasn't too bad!

Let me know if those tests look okay...

necouchman:
Okay, I think I cleaned up most of the stuff from both @mike-jumper and @ceharris for the latest round, including handling encoded parameters and some unit tests for parameter parsing.  Also tweaked a few other things on my own along the way.  Let me know what else needs to be tweaked.

extensions/guacamole-auth-quickconnect/src/licenses/bundled/aopalliance-1.0/LICENSE
```diff
@@ -0,0 +1,4 @@
+From http://aopalliance.sourceforge.net/:
```
mike-jumper: Is this still bundled? Looking at the `pom.xml`, I no longer see any dependencies that are not marked as provided.

extensions/guacamole-auth-quickconnect/src/licenses/LICENSE
```diff
@@ -0,0 +1,221 @@
...
+==============================================================================
+
+APACHE GUACAMOLE SUBCOMPONENTS
+
+Apache Guacamole includes a number of subcomponents with separate copyright
+notices and license terms. Your use of these subcomponents is subject to the
+terms and conditions of the following licenses.
+
+
+JUnit
```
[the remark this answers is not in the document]
necouchman: Ah, yes, makes sense.  Fixed.

necouchman:
I haven't looked closely or tried it, but does anything in here need to be changed to go along with the changes for GUACAMOLE-526 that were recently merged?

necouchman:
Well, now I have a regression somewhere in my code where it errors out trying to open the connection initially when it is created.  So, get to go track that one down, tomorrow, too... :-/.

necouchman:
So, my regression appears to be some disconnect between how I've implemented the `QuickConnectREST` class and what `requestService` expects for data and returns.  The current implementation of the `create` method in `QuickConnectREST` returns a string representing the identifier of the connection that was just created:

https://github.com/apache/guacamole-client/pull/181/files#diff-6f01d367b00060b398b2a82c49ea11daR71

And the implementation of `requestService` appears to expect an object called `data` to be present:

https://github.com/apache/guacamole-client/blob/9d09f0bb0a6bed3b355572644faf374d5ca14f8b/guacamole/src/main/webapp/app/rest/services/requestService.js#L51

So, do I need to do something else in my REST class - create an object type or something like that - for this to return, or is there something else I should be doing to "wrap" this response in such a way that requestService can pass it through correctly?

necouchman:
> So, my regression appears to be some disconnect between how I've implemented the QuickConnectREST class and what requestService expects for data and returns.

Nevermind...figured it out.  Should be updated for the recent GUACAMOLE-526 changes.

mike-jumper:
I'll re-review.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectDirectory.java
```diff
@@ -0,0 +1,127 @@
...
+     *     An int representing the next available connection
+     *     identifier to be used when adding connections.
+     */
+    private int getNextConnectionID() {
+        return connectionId.getAndIncrement();
+    }
+
+    @Override
+    public void add(Connection connection) throws GuacamoleException {
+        connections.put(connection.getIdentifier(), connection);
```
mike-jumper: This is not threadsafe, and may behave unexpectedly if multiple requests to add a new connection are serviced for the same user at the same time. I recommend switching from `HashMap` to `ConcurrentHashMap`.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectDirectory.java
```diff
@@ -0,0 +1,127 @@
...
+
+        // Get the next connection identifier.
+        String connectionId = Integer.toString(getNextConnectionID());
+
+        // Generate a name for the configuration.
+        String name = QCParser.getName(config);
+
+        // Create a new connection and set the parent identifier.
+        Connection connection = new SimpleConnection(name, connectionId, config);
+        connection.setParentIdentifier(ROOT_IDENTIFIER);
```
mike-jumper: The `QuickConnectUserContext` uses `DEFAULT_ROOT_CONNECTION_GROUP`, which is not guaranteed to match `ROOT_IDENTIFIER` here. As the identifier of the root group is in fact `DEFAULT_ROOT_CONNECTION_GROUP`, that should be the value used here.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectUserContext.java
```diff
@@ -0,0 +1,123 @@
...
+        // single root identifier.
+        this.rootGroup = new QuickConnectionGroup(
+            DEFAULT_ROOT_CONNECTION_GROUP,
+            DEFAULT_ROOT_CONNECTION_GROUP
+        );
+
+        // Initialize the user to a SimpleUser with the provided username,
+        // no connections, and the single root group.
+        this.self = new SimpleUser(username,
+            Collections.<String>emptyList(),
```
mike-jumper: This will only be correct when no connections exist. Once at least one connection is added, the permissions stored within this `SimpleUser` will not match up with the connections they are supposed to be able to read.

I'd recommend instead returning a new `SimpleUser` from `self()`, using `connectionDirectory.getIdentifiers()` to pull the set of readable connections.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectionGroup.java
```diff
@@ -0,0 +1,112 @@
...
+     *     The name of the QuickConnectionGroup.
+     *
+     * @param identifier
+     *     The identifier of the QuickConnectionGroup.
+     */
+    public QuickConnectionGroup(String name, String identifier) {
+
+        setName(name);
+        setIdentifier(identifier);
+        setType(ConnectionGroup.Type.ORGANIZATIONAL);
```
mike-jumper: Though older code within Guacamole does this, it's bad practice to invoke overridable functions of the class being instantiated within its own constructor. If the setters belong to the superclass, `super` should be used explicitly. Some IDEs will warn of this.

https://stackoverflow.com/questions/3404301/whats-wrong-with-overridable-method-calls-in-constructors

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
```diff
@@ -0,0 +1,233 @@
...
+        // Break out individual components of the URI.
+        String protocol = qcUri.getScheme();
+        String host = qcUri.getHost();
+        int port = qcUri.getPort();
+        String userInfo = qcUri.getUserInfo();
+        String query = qcUri.getQuery();
+
+        String username = null;
+        String password = null;
+        Map<String, String> queryParams = null;
```
mike-jumper: There are a few cases here of the general form:

```java
String foo = null;

if (someCondition) {
    foo = somethingThatNeverReturnsNull();
}

if (foo != null) {
    doMoreThings(foo);
}
```

This is really taking what is effectively one condition and splitting it across two separate `if` blocks. The unnecessary initialization and unnecessary split can both be removed by collapsing into:

```java
if (someCondition) {
    String foo = somethingThatNeverReturnsNull();
    doMoreThings(foo);
}
```

The logic is a bit tangled otherwise.

mike-jumper:
Except for the above, things look generally good to me. I only have remaining nitpicks regarding the quickconnect UI itself, which looks a bit clunky IMHO:

quickconnect-interface

Perhaps there's a better way to style this?

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectDirectory.java
necouchman: So, do I need to inject or pass in the `UserContext` object, here?  The `DEFAULT_ROOT_CONNECTION_GROUP` string is protected in the class, so I can't just use it...

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectionGroup.java
necouchman: Okay, but if the superclass doesn't have a constructor that calls these, how should I handle that?  Override the individual methods inside the class itself and then call them?  I'm looking at `AbstractConnectionGroup` and `ConnectionGroup` and `Identifiable` and not seeing a constructor that actually takes these arguments and initializes the various objects?

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
necouchman: Okay, I'll take a look at this section and try to untangle it.

necouchman:
> Perhaps there's a better way to style this?

I'm definitely open to suggestions - of the long list of things I'm not great at, GUI/graphical design is definitely toward the top of that list.  I welcome any suggestions or help with that.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectDirectory.java
necouchman: Switched.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectDirectory.java
necouchman: Done.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectUserContext.java
necouchman: I believe I've done this as you suggested.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
necouchman: Untangled.

necouchman:
@mike-jumper 
> Perhaps there's a better way to style this?

So, I tweaked the style a little bit to clean up some of the issues, but not sure it addresses all of the concerns.  I had tried to style it as much like the current interface as possible, but, as mentioned before, I'm open to suggestions - move it somewhere else, resize it, etc.  I do think it needs to be in a pretty obvious place on the screen - I wouldn't bury it in a menu - since the idea is that it allows for creating connections really quickly, so seems like minimizing mouse clicks to get there would be good.  Anyway, let me know what else should be tweaked, there.

mike-jumper:
Looks much much better. I'll re-review and retest. Assuming nothing new pops up, I think we're probably good to go here.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectionGroup.java
mike-jumper: We probably should provide a constructor, but you can get around this for the time being by using `super.setName()`, `super.setIdentifier()`, etc., as that will invoke a function against the superclass which is definitely initialized at this point.

necouchman: Done.

necouchman:
Fixed that final issue with the superclass methods.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
```diff
@@ -0,0 +1,232 @@
...
+        // Look for the username and password and parse them out.
+        if (userInfo != null && !userInfo.isEmpty()) {
+
+            Matcher userinfoMatcher = userinfoPattern.matcher(userInfo);
+            if (userinfoMatcher.matches()) {
+                String username = userinfoMatcher.group(USERNAME_GROUP);
+                String password = userinfoMatcher.group(PASSWORD_GROUP);
+
+                if (username != null && !username.isEmpty())
+                    qcConfig.setParameter("username", username);
```
mike-jumper: I see `parseQueryString()` automatically handling URL decoding using `URLDecoder.decode()`. Does that need to happen here, too, with the username and password?

extensions/guacamole-auth-quickconnect/src/test/java/org/apache/guacamole/auth/quickconnect/utility/QCParserTest.java
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Class to test methods in the QCParser utility class.
+ */
+public class QCParserTest {
+
+    /**
+     * Verify that the parseQueryString() method functions as designed.
+     */
+    @Test
+    public void testParseQueryString() throws UnsupportedEncodingException {
```
mike-jumper: I suggest adding a unit test for the overall URI parsing performed by `QCParser.getConfiguration()`, which is decently complex. The parsing of query string parameters is only one piece of that pie.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
necouchman: I've added the `URLDecoder.decode()` method to this, and I also broke out the parsing of the username/password into its own method, as it was gaining the same level of complexity that the query string parser had.

extensions/guacamole-auth-quickconnect/src/test/java/org/apache/guacamole/auth/quickconnect/utility/QCParserTest.java
necouchman: Will work this up, and will also add a unit test for the new `parseUserInfo()` method.

necouchman: Tests should be implemented.

necouchman:
Addressed the most recent couple of concerns, including adding unit tests for both the `parseUserInfo()` method and the `getConfiguration()` method.  Let me know if those tests need to process any additional corner cases.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
```diff
@@ -184,6 +184,43 @@ public static GuacamoleConfiguration getConfiguration(String uri)
...
+        Map<String, String> userInfoMap = new HashMap<String, String>();
+        Matcher userinfoMatcher = userinfoPattern.matcher(userInfo);
+
+        if (userinfoMatcher.matches()) {
+            String username = URLDecoder.decode(
+                    userinfoMatcher.group(USERNAME_GROUP), "UTF-8");
+            String password = URLDecoder.decode(
+                    userinfoMatcher.group(PASSWORD_GROUP), "UTF-8");
+
+            if (username != null && !username.isEmpty())
```
mike-jumper: Under what circumstances will `username` or `password` be null? `URLDecoder.decode()` is not documented as returning null.

necouchman: Okay, moved the null check to prior to decoding it, so that should be solved.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
```diff
@@ -184,6 +184,43 @@ public static GuacamoleConfiguration getConfiguration(String uri)
...
+                userInfoMap.put("username", username);
+
+            if (password != null && !password.isEmpty())
+                userInfoMap.put("password", password);
```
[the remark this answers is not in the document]
necouchman: Okay, passed the config object through to the method and just set it directly.

mike-jumper:
Just tested the latest from this PR, and I'm currently seeing an issue where entering a nonsensical URI like "sadjfhlaksjdhsadf" succeeds, creating a new `ssh://localhost/` connection.

necouchman:
> Just tested the latest from this PR, and I'm currently seeing an issue where entering a nonsensical URI like "sadjfhlaksjdhsadf" succeeds, creating a new ssh://localhost/ connection.

Hmmm...yeah, this case maybe it makes more sense to not have either a default protocol or a default host and require those items to be present at a minimum.  I'll ditch those defaults and throw some errors, instead.

necouchman:
Okay, got rid of the defaults, and also implemented a `QuickConnectException` class to handle error messages out of this module.  Unfortunately right now it's just throwing a standard Tomcat 500 Internal Server Error instead of passing back the error, so I have to track down why that's happening.  More to come...

necouchman:
Okay, @mike-jumper, I could use a little help, here, with this `QuickConnectException` class.  I've written it to extend `GuacamoleClientException` and implement `Translatable` so that it *should* provide a useful error message to the user when they enter an invalid URI.  However, instead, I'm getting a blank "ERROR" dialog, and the following in my catalina.out file:

    22-May-2018 22:20:28.708 SEVERE [http-nio-127.0.0.1-8080-exec-352] com.sun.jersey.spi.container.ContainerResponse.mapMappableContainerException The exception contained within MappableContainerException could not be mapped to a response, re-throwing to the HTTP container
    org.apache.guacamole.auth.quickconnect.QuickConnectException: URI must be absolute.
	at org.apache.guacamole.auth.quickconnect.utility.QCParser.getConfiguration(QCParser.java:83)
	at org.apache.guacamole.auth.quickconnect.rest.QuickConnectREST.create(QuickConnectREST.java:77)

I expect it to, instead, provide the proper HTTP status code (should be 400, in this case, which is set/overriden in the `GuacamoleClientException` method `getStatus()`.  What am I missing, here, that is causing it to just rethrow as HTTP 500 instead of using the assigned status code and provide the proper JSON that should include the error ("URI must be absolute.")??

mike-jumper:
I think this is due to the custom REST services not being intercepted in the same way that Guacamole's REST services are intercepted. There's a `RESTExceptionWrapper` thing that gets magicked around all methods that throw `GuacamoleException` and have JAX-RS annotations:

https://github.com/apache/guacamole-client/blob/1710c31fc2c0c8f56326e4175d13b21640d477ba/guacamole/src/main/java/org/apache/guacamole/rest/RESTServiceModule.java#L87-L90

For the same magic to apply to extensions, we'll need to (somehow) add the same method intercepting as part of the extension load process.

mike-jumper:
I think the magic will need to be added to the extension load process. That sort of introspection and run-time class modification probably should be done one time during load, not once per request.

necouchman:
> I think the magic will need to be added to the extension load process.

So, in the extension itself or in the extension loader?  Or both?

mike-jumper:
In the main extension loader. I'd think we should just somehow let the same magic apply to classes within extensions.

Probably not the sort of thing that should be the duty of the extensions, nor part of the public API.

necouchman:
Okay, I can't remember if the main extension loader is in the guacamole part of the project or guacamole-ext?  I'm guessing this could mean moving some of those classes over from guacamole to guacamole-ext??

necouchman:
Okay, I'll take a look and see if I can figure it out, open a new JIRA issue for it, and either submit a new PR or come back and beg for help because I'm in over my head :-D.

Guessing this PR needs to wait for that issue to be resolved, though, so that we avoid introducing already broken features :-/.

necouchman:
This appears to be working fine after implementing an `ExceptionWrapper` class, so hopefully good to go after changes for GUACAMOLE-566 get a PR and merge.

necouchman:
Okay, I've rebased this onto the current master after the changes in #291.  It didn't need it to function properly, but seems a little cleaner.

I had gone through and cleaned up a few things - the REST endpoint now returns JSON instead of just a string, and I touched up some comments.  Anything else need attention, here?

mike-jumper:
Code looks good to me. I'll retest.

@necouchman - can you update the merge base of the PR to `staging/1.0.0` now that the release process is underway?

mike-jumper:
Things seem to work as expected. Once the merge base is updated, I'll be happy to merge.

necouchman:
Merge base updated.

PR #182, opened 2017-08-16
GUACAMOLE-363: Add support for SQL Server authentication

necouchman:
This pull request implements a SQL Server (sqlserver) module for the JDBC authentication backend, allowing the Guacamole database to be hosted in SQL Server.

I went the route of creating a guacamole schema and renaming the tables a bit - I have a feeling I'll end up changing that to be more consistent with the other modules, but SQL Server uses a default "dbo" schema that's just different, anyway, so I took a stab at it this way.


mike-jumper:
> I went the route of creating a guacamole schema and renaming the tables a bit - I have a feeling I'll end up changing that to be more consistent with the other modules, ...

Yup - I agree this needs to change. The schema needs to be as consistent as possible across the various guacamole-auth-jdbc-* implementations.

> ... but SQL Server uses a default "dbo" schema that's just different, anyway, so I took a stab at it this way.

I ran across similar uncertainty when implementing the PostgreSQL authentication after having used nothing but MySQL for a while, as Postgres also uses a default schema called "public".

Is creating tables within the "dbo" schema frowned upon in SQL Server circles? Rather than renaming the tables, perhaps the schema name should be a configuration parameter?

necouchman:
> Yup - I agree this needs to change. The schema needs to be as consistent as possible across the various guacamole-auth-jdbc-* implementations.
Okay, I've normalized the table names and just stuck with the dbo schema for the time being.

> Is creating tables within the "dbo" schema frowned upon in SQL Server circles? Rather than renaming the tables, perhaps the schema name should be a configuration parameter?

I did a little research on this, and the consensus I was able to find is that it is "best practice" to use a separate schema, primarily for the purpose of security and permissions management.  Not sure how applicable it is in this use case, but probably a good idea in the future to add the support across all of the JDBC modules for a custom schema.  It's actually going to be necessary in the Oracle version of the JDBC module I'm working on, as Oracle doesn't really have a single, database-wide public schema - it very much relies on schemas based on the name of the user accessing the database.

However, the changes for passing the schema over are going to be a little more involved that I had hoped - since you cannot call non-static methods from the XML mapper files, retrieving the schema name either has to be in a static method somewhere, or we have to use the dynamic XML generation feature of MyBatis, or have to pass the schema into the mapper as a parameter somehow.  In any case, I think that probably warrants its own JIRA issue and pull request to change that functionality.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerAuthenticationProviderModule.java
```diff
@@ -0,0 +1,91 @@
...
+        // Use UTF-8 in database
+        driverProperties.setProperty("characterEncoding", "UTF-8");
+
+    }
+
+    @Override
+    public void configure(Binder binder) {
+
+        // Bind SQLServer-specific properties
+        JdbcHelper.SQL_Server_2005_MS_Driver.configure(binder);
```
mike-jumper: Is this going to be an issue with other versions of SQL Server? Why this particular driver vs. jTDS?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/resources/org/apache/guacamole/auth/jdbc/connection/ConnectionMapper.xml
```diff
@@ -0,0 +1,235 @@
...
+                    column="connection_id" foreignColumn="primary_connection_id">
+            <result column="sharing_profile_id"/>
+        </collection>
+
+    </resultMap>
+
+    <!-- Select all connection identifiers -->
+    <select id="selectIdentifiers" resultType="string">
+        SELECT connection_id 
+        FROM [dbo].[guacamole_connection]
```
mike-jumper: Out of curiosity, does SQL Server require that the schema be specified with each query? I don't recall having to do this in the past, but it's been a rather long time since I had to touch SQL Server.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/schema/001-create-schema.sql
```diff
@@ -0,0 +1,558 @@
...
+ * The system permission data type.
+ */
+CREATE TYPE [dbo].[guacamole_system_permission] FROM [nvarchar](32) NOT NULL
+GO
+
+/**
+ * The connection_group table stores organizational and balancing groups.
+ */
+SET ANSI_NULLS ON
+SET QUOTED_IDENTIFIER ON
```
mike-jumper: >     SET ANSI_NULLS ON
>     SET QUOTED_IDENTIFIER ON

I see this quite a lot here. Why is this incantation necessary?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/schema/001-create-schema.sql
```diff
@@ -0,0 +1,558 @@
...
+
+    CONSTRAINT [PK_guacmaole_connection_group] PRIMARY KEY CLUSTERED
+        ([connection_group_id] ASC)
+        WITH (PAD_INDEX = OFF,
+            STATISTICS_NORECOMPUTE = OFF,
+            IGNORE_DUP_KEY = OFF,
+            ALLOW_ROW_LOCKS = ON,
+            ALLOW_PAGE_LOCKS = ON)
+       ON [PRIMARY]
+) ON [PRIMARY]
```
mike-jumper: Though older SQL Server may not require semicolons in all cases, newer versions do, and it's a bad practice that can lead to difficult-to-track-down bugs.

http://www.dbdelta.com/always-use-semicolon-statement-terminators/

Better that we stick with ANSI as far as possible.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/schema/001-create-schema.sql
necouchman: So, most of the actual text here, minus the formatting, came from using SQL Server Management Studio to generate the output of the database minus any actual data.  As such, I took the file they spit out, re-arranged it, styled it to match Guacamole style, and that's what the .sql files you see here are.  I'm not entirely certain why SSMS found it necessary to spit this out so frequently, but I think that at least some of those options are only good in the current batch (up until the next "GO" statement), and I did run into issues trying to batch too many things together (got messages from SQL Server that certain queries had to be in their own batch).

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/resources/org/apache/guacamole/auth/jdbc/connection/ConnectionMapper.xml
necouchman: I think it depends.  I definitely do not consider myself an expert on SQL Server; however, unlike PostgreSQL, I believe SQL Server is a little further down the road toward Oracle of having certain schemas associated with certain users.  When you create a user object in a SQL Server database you can associate a default schema with that user.  So, if an admin create a guacamole user in for the database and sets the schema to something other than dbo, but loads the schema using the sa account or something similar, there could be some confusion for the admin about which schema the tables are in.  This forcibly eliminates that confusion.

I could go the route of just assuming the default schema, if you think that's better??  It doesn't matter all that much to me...

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerAuthenticationProviderModule.java
necouchman: I'll have to give the SQL_Server_jTDS driver a try - I didn't even try that one.  I tried the SQL_Server_MS_Driver (sans 2005), and it had something silly like the order of the class name was slightly different (com.microsoft.sqlserver.jdbc vs. com.microsoft.jdbc.sqlserver).  I'll try the jTDS one and see what happens.  I'm using Microsoft's JAR file, so not sure if the jTDS one will require that, some other JAR file, or is all built in??  I'll find out, I guess.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/schema/001-create-schema.sql
necouchman: Okay, I've added those to the code.  You know, it's almost hard to believe that I was just using Microsoft-generated code - that Microsoft would do something that didn't conform to standards!

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerAuthenticationProviderModule.java
necouchman: So, I tried a route that allows for both of them - added another option to the configuration to specify the JTDS driver instead of the Microsoft one, but kept the Microsoft one as the default.  How's that look?

necouchman: One other comment on the jTDS driver.  I'm all for using the OSS alternatives to the commercials, but it looks like development on the jTDS driver has languished a bit.  The last release was sometime in 2013, while the last commit was toward the end of 2016.  I'm all for supporting use of this driver via the change I made above, but I'd hesitate to make it the default, particularly when SQL Server 2016 has come out since the last jTDS release, and SQL Server 2017 (SQL Server for Linux) is just out.  I can be talked out of that position, but that's my current train of thought.

necouchman:
@mike-jumper @jmuehlner Any additional changes you think should be made here?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/resources/org/apache/guacamole/auth/jdbc/connection/ConnectionMapper.xml
mike-jumper: It's not that I think we need to require the default schema vs. something else, but rather that I'm surprised that each query would need to specify that schema. I would have expected that the schema would be specified at the JDBC connection level, and would then effect all queries which do not explicitly specify a different schema.

In the case of the PostgreSQL implementation, use of the default `public` schema was due to not specifying the schema at all, which was probably mostly due to my inexperience with PostgreSQL at the time. If something about PostgreSQL then required that I hard-code the `public` schema into all queries, I think I'd start getting worried.

I'm definitely far from an expert in SQL Server, but it feels like hard-coding the schema into each query would be bad. I don't know what the alternative would be. If the expectation for SQL Server databases is that the application itself dictates the schema name, etc. (not just the tables), then perhaps I'm steering things in the wrong direction here by pushing use of the default schema.

necouchman: That's fair.  Let me poke around a little more and see if I can figure out a way to set the schema search path (similar to Postgres) at the connection level.  I glanced when I was initially doing it and didn't see any parameter for it, but maybe it's just a little more obscure than that.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerAuthenticationProviderModule.java
```diff
@@ -0,0 +1,102 @@
...
+        myBatisProperties.setProperty("JDBC.password", environment.getSQLServerPassword());
+        myBatisProperties.setProperty("JDBC.autoCommit", "false");
+        myBatisProperties.setProperty("mybatis.pooled.pingEnabled", "true");
+        myBatisProperties.setProperty("mybatis.pooled.pingQuery", "SELECT 1");
+
+        // Use UTF-8 in database
+        driverProperties.setProperty("characterEncoding", "UTF-8");
+
+        // Capture whether or not to use the JTDS driver.
+        this.useJTDSDriver = environment.getSQLServerJTDSDriver();
```
mike-jumper: From http://www.mybatis.org/guice/jdbc-helper.html, I see the following SQL Server drivers supported by MyBatis out of the box:

* `SQL_Server_DataDirect`
* `SQL_Server_jTDS`
* `SQL_Server_MS_Driver`
* `SQL_Server_2005_MS_Driver`

Any reason the configuration here is being limited to 2005-specific Microsoft driver and jTDS?

necouchman: No reason, no...I can add them all.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/resources/org/apache/guacamole/auth/jdbc/connection/ConnectionMapper.xml
necouchman: Well, according to research, it isn't possible to set the default schema or schema search path in the connection string for SQL Server - it's a per-user thing.  So, I think we'll just have to go the route of not explicitly specifying the schema and include instructions that the account that is logging in to SQL Server needs to have its default schema set properly.  Sound okay?  Counterproposals?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerAuthenticationProviderModule.java
necouchman: Okay, pushed commit that updates this property to be a string property and look for those values, and use the MS 2005 driver by default.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/resources/org/apache/guacamole/auth/jdbc/connection/ConnectionMapper.xml
necouchman: Okay, all of the explicit schema calls have been removed, so it'll just rely on the default schema for the user account being used for guacamole.  I'll just need to document this when I write the documentation for this module.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/schema/001-create-schema.sql
```diff
@@ -0,0 +1,558 @@
...
+/**
+ * The connection_parameter table stores parameters for
+ * connection objects.
+ */
+SET ANSI_NULLS ON;
+SET QUOTED_IDENTIFIER ON;
+CREATE TABLE [guacamole_connection_parameter](
+    [connection_id] [int] NOT NULL,
+    [parameter_name] [nvarchar](128) NOT NULL,
+    [parameter_value] [nvarchar](max) NOT NULL,
```
mike-jumper: Why `max`?

necouchman: max = 4000, which is less than what Postgres and MySQL use (4096).

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerAuthenticationProviderModule.java
```diff
@@ -0,0 +1,107 @@
...
+     */
+    private final Properties myBatisProperties = new Properties();
+
+    /**
+     * SQLServer-specific driver configuration properties.
+     */
+    private final Properties driverProperties = new Properties();
+
+    /**
+     * Whether or not to use JTDS Driver
```
mike-jumper: Now that the property has been generalized to cover all SQL Server drivers supported by MyBatis, this is no longer correct.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerAuthenticationProviderModule.java
```diff
@@ -0,0 +1,107 @@
...
+        // Bind SQLServer-specific properties
+        // Look at the property to choose the correct driver.
+        if (sqlServerDriver.equals(SQLServerEnvironment.SQLSERVER_DRIVER_JTDS))
+            JdbcHelper.SQL_Server_jTDS.configure(binder);
+        else if (sqlServerDriver.equals(SQLServerEnvironment.SQLSERVER_DRIVER_DATADIRECT))
+            JdbcHelper.SQL_Server_DataDirect.configure(binder);
+        else if (sqlServerDriver.equals(SQLServerEnvironment.SQLSERVER_DRIVER_MS))
+            JdbcHelper.SQL_Server_MS_Driver.configure(binder);
+        else
+            JdbcHelper.SQL_Server_2005_MS_Driver.configure(binder);
```
mike-jumper: I'm curious - why is the 2005 driver the default? Is it known to be more common and/or more compatible?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerEnvironment.java
```diff
@@ -0,0 +1,357 @@
...
+    public String getSQLServerPassword() throws GuacamoleException {
+        return getRequiredProperty(SQLServerGuacamoleProperties.SQLSERVER_PASSWORD);
+    }
+
+    /**
+     * Returns whether or not to use the SourceForge JTDS driver for more
+     * generic JTDS connections instead of the Microsoft-provided JDBC driver.
+     *
+     * @return
+     *     True if the JTDS driver should be used; false by default.
```
mike-jumper: This (and the rest of this comment) is no longer correct due to the migration to a generalized property covering all supported SQL Server drivers.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerEnvironment.java
```diff
@@ -0,0 +1,357 @@
...
+     *
+     * @throws GuacamoleException
+     *     If an error occurs while retrieving the property value, or if the
+     *     value was not set, as this property is required.
+     */
+    public String getSQLServerDriver() throws GuacamoleException {
```
mike-jumper: This may be more clear and easier to maintain if an `enum` with strict values is returned, rather than relying on external parsing of a string value.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerAuthenticationProviderModule.java
necouchman: It's more compatible with the most recent versions of Microsoft SQL Server.  The other driver is/appears to be older than the 2005 one - at least, the latest JDBC driver you download from Microsoft seems to be the "2005 driver."  The jTDS one hasn't seen a lot of development recently.  The DataDirect one is a third-party one from Progress.  So, I went with the one that, by all appearances, is going to be most likely to support the most recent versions of SQL Server.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerEnvironment.java
```diff
@@ -0,0 +1,357 @@
...
+                DEFAULT_MAX_CONNECTIONS       = 1;
+                DEFAULT_MAX_GROUP_CONNECTIONS = 0;
+            }
+            else {
+                DEFAULT_MAX_CONNECTIONS       = 0;
+                DEFAULT_MAX_GROUP_CONNECTIONS = 0;
+            }
+
+            // Warn of deprecation
+            logger.warn("The \"{}\" property is deprecated. Use \"{}\" and \"{}\" instead.",
```
mike-jumper: The "disallow simultaneous connections", etc. properties continue to exist in the MySQL and PostgreSQL auth (with deprecation warnings) because they were at one time supported, and users are expected to migrate to the new properties when possible. This doesn't make sense here, though, since all SQL Server properties are new. If the properties shouldn't be used, they shouldn't be added.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerEnvironment.java
necouchman: What's the best way to do this as a property?  Stick with a string property and then translate to an enum, or implement a new property type?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerEnvironment.java
necouchman: Okay, will remove them.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerEnvironment.java
```diff
@@ -0,0 +1,357 @@
...
+
+        }
+
+        // Check driver property is one of the acceptable values.
+        String driver = getProperty(SQLServerGuacamoleProperties.SQLSERVER_DRIVER);
+        if (driver != null && !(driver.equals(SQLSERVER_DRIVER_JTDS) ||
+                                driver.equals(SQLSERVER_DRIVER_DATADIRECT) ||
+                                driver.equals(SQLSERVER_DRIVER_MS) ||
+                                driver.equals(SQLSERVER_DRIVER_MS_2005)))
+            logger.warn("{} property has been set to an invalid value.  The default Microsoft 2005 driver will be used.",
```
mike-jumper: Rather than read and parse the driver property just to determine whether it's valid, only to parse it a second time later in independent code, it would make more sense to handle this within a common function that encapsulates that parsing, ideally masking the value behind an `enum` which Java can helpfully check at compile time.

For example:

https://github.com/apache/incubator-guacamole-client/blob/1c0ee41d0ecd5bc4a3550804b74b73b901e074c2/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/EncryptionMethodProperty.java

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerEnvironment.java
mike-jumper: The cleanest way is to implement a new property type. Parsing logic is kept separate, and usage of the property ends up simplified.

See: https://github.com/apache/incubator-guacamole-client/blob/1c0ee41d0ecd5bc4a3550804b74b73b901e074c2/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/EncryptionMethodProperty.java

necouchman: Nevermind...see your comment below, will reimplement it similar to LDAP encryption property.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/schema/001-create-schema.sql
mike-jumper: On recent versions of SQL Server, values up to 8000 are allowed, while a `varchar(max)` indicates the value should can grow up to 2 GB. According to https://docs.microsoft.com/en-us/sql/t-sql/data-types/char-and-varchar-transact-sql (emphasis added):

> varchar [ ( n | max ) ] Variable-length, non-Unicode string data. n defines the string length and can be a value from 1 through 8,000. **max indicates that the maximum storage size is 2^31-1 bytes (2 GB)**. The storage size is the actual length of the data entered + 2 bytes. The ISO synonyms for varchar are charvarying or charactervarying.

Not sure if the definition of this has changed between various versions of SQL Server, but assuming that's the case, that possibility may be another reason  to avoid `max`.

Are you testing primarily against SQL Server 2005 and its limit is 4000?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerEnvironment.java
necouchman: Implemented.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerEnvironment.java
necouchman: Believe I got all this cleaned out.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerEnvironment.java
necouchman: Implemented.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerEnvironment.java
necouchman: Cleaned up.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerAuthenticationProviderModule.java
necouchman: Cleaned up.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/schema/001-create-schema.sql
necouchman: Nope, I'm testing against SQLServer on Linux, or SQLServer 2017.  When I tried to use the 4096 value, I was told it was above the max.

<Shrug>?!

necouchman: Ah, actually, the issue is that I'm using nvarchar instead of varchar, because nvarchar supports storing unicode values.

necouchman:
Ping @mike-jumper...anything else that needs attention, here?

mike-jumper:
Rechecking now.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/schema/001-create-schema.sql
mike-jumper: I'm still concerned about the use of `max` here, as the documentation states that this is equivalent to a maximum storage size of 2 GB, not the intended 4 KB.

If the intent is to have a `[nvarchar](4000)`, why not use `[nvarchar](4000)`?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/schema/001-create-schema.sql
```diff
@@ -0,0 +1,558 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * List for permission data type.
+ */
+CREATE RULE [guacamole_permission_list]
```
mike-jumper: I don't see `guacamole_permission_list` anywhere else in this script. Does SQL Server somehow magically tie this to the `guacamole_permission` type?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerAuthenticationProviderModule.java
```diff
@@ -0,0 +1,116 @@
...
+            case DATA_DIRECT:
+                JdbcHelper.SQL_Server_DataDirect.configure(binder);
+                break;
+
+            case MICROSOFT_LEGACY:
+                JdbcHelper.SQL_Server_MS_Driver.configure(binder);
+                break;
+
+            case MICROSOFT_2005:
+            default:
```
mike-jumper: Since the `default` case will only match if we add a new SQL Server driver type without actually implementing that type, silently failing over to the 2005 driver is dangerous behavior. I would recommend either:

1. Bailing out with a hard and unmistakable [`UnsupportedOperationException`](https://docs.oracle.com/javase/7/docs/api/java/lang/UnsupportedOperationException.html).
2. Documenting for humans and compilers that this condition is expected to be impossible through an `assert(false)`.

Example of the above:

https://github.com/apache/incubator-guacamole-client/blob/1c0ee41d0ecd5bc4a3550804b74b73b901e074c2/guacamole/src/main/java/org/apache/guacamole/tunnel/TunnelRequestService.java#L184-L186

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerEnvironment.java
```diff
@@ -0,0 +1,254 @@
...
+    /**
+     * The default port to connect to, if SQLSERVER_PORT is not specified.
+     */
+    private static final int DEFAULT_PORT = 1433;
+
+    /**
+     * Whether a database user account is required by default for authentication
+     * to succeed.
+     */
+    private static final boolean DEFAULT_USER_REQUIRED = true;
```
mike-jumper: The default for the "*-user-required" properties should be `false`:

https://github.com/apache/incubator-guacamole-client/blob/1c0ee41d0ecd5bc4a3550804b74b73b901e074c2/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/MySQLEnvironment.java#L53

https://github.com/apache/incubator-guacamole-client/blob/1c0ee41d0ecd5bc4a3550804b74b73b901e074c2/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/PostgreSQLEnvironment.java#L53

See also:

http://guacamole.incubator.apache.org/doc/gug/jdbc-auth.html#jdbc-auth-restrict

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerGuacamoleProperties.java
```diff
@@ -0,0 +1,211 @@
...
+        public String getName() { return "sqlserver-user-required"; }
+
+    };
+
+    /**
+     * Whether or not multiple users accessing the same connection at the same
+     * time should be disallowed.
+     */
+    public static final BooleanGuacamoleProperty
+            SQLSERVER_DISALLOW_SIMULTANEOUS_CONNECTIONS =
```
mike-jumper: Since this property is not actually used, it shouldn't be defined.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerGuacamoleProperties.java
```diff
@@ -0,0 +1,211 @@
...
+        public String getName() { return "sqlserver-disallow-simultaneous-connections"; }
+
+    };
+
+    /**
+     * Whether or not the same user accessing the same connection or connection
+     * group at the same time should be disallowed.
+     */
+    public static final BooleanGuacamoleProperty
+            SQLSERVER_DISALLOW_DUPLICATE_CONNECTIONS =
```
mike-jumper: Same here - since this property is unused, it shouldn't be defined.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/schema/002-create-admin-user.sql
```diff
@@ -0,0 +1,46 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * Create the default admin user account and set up full privileges.
+ */
+INSERT INTO [guacamole_user] (username, password_hash, password_date)
+VALUES ('guacadmin', HASHBYTES('SHA2_256', 'guacadmin'), getdate());
```
mike-jumper: Any reason to not use the same hash+salt values as the other user creation scripts?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/schema/001-create-schema.sql
necouchman: Negative...I assumed that, when I exported the database using SQL Server Management Studio, it would generate the commands to bind these two things together.  Silly me.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerAuthenticationProviderModule.java
necouchman: I went the throw an exception route.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerEnvironment.java
necouchman: Oops, not sure why that was like that.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerGuacamoleProperties.java
necouchman: Removed.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerGuacamoleProperties.java
necouchman: Removed.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/schema/002-create-admin-user.sql
necouchman: I was having trouble getting it to work trying to paste in the raw values.  I'll try some variations on it, again, and see if I can make it work.

necouchman: Well, whatever I did last time was wrong, but I think I got it this time.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/schema/001-create-schema.sql
necouchman: Should be fixed.

necouchman:
Here is Microsoft's page about ANSI_NULLS:

https://docs.microsoft.com/en-us/sql/t-sql/statements/set-ansi-nulls-transact-sql

The items worth noting are, first, that in a future version of SQL Server, it will always be ON and OFF will not be allowed.  Second, it impacts how SQL Server handles WHERE conditions that involve comparing to NULL values (WHERE column = NULL, WHERE column <> NULL).  With it on, it complies to ISO standards; with it off, it does not.  The good news is you can turn it on at the DB level, so I've done that.

Here is the page on QUOTED_IDENTIFIER:
https://docs.microsoft.com/en-us/sql/t-sql/statements/set-quoted-identifier-transact-sql

From what I can tell, ON is the default behavior, and I don't know of any reason to turn it off.  I've also set this at the top of the file for DB-level rather than doing it for each transaction.

For the rest of them, on the individual constraints, for example, I went through and checked the defaults for those and just removed the explicit specification if it was the same as the default.

Let me know what else needs attention.

necouchman:
Sorry about that - pushed the commits before I had reviewed them.  I made the changes and rebased the commit, so should be good, now.  Let me know if I missed anything.

I also tried to brush up the formatting on the schema files a little bit, but it's just so...different...from the other SQL platforms that it's hard to do.  Maybe a SQL Server expert will jump on and help contribute :-).

PR #185, opened 2017-09-04
GUACAMOLE-310: Fix clipboard handling of newlines.

mike-jumper:
As reported in [GUACAMOLE-310](https://issues.apache.org/jira/browse/GUACAMOLE-310), newlines are being stripped from clipboard data copied using the [W3C Clipboard API](https://www.w3.org/TR/clipboard-apis/). This makes clipboard use cumbersome for users where the browser provides direct integration of the clipboard, such as IE11 or Chrome with the [Clipboard Permission Manager](https://chrome.google.com/webstore/detail/clipboard-permission-mana/ipbhneeanpgkaleihlknhjiaamobkceh) extension.

This is due to newlines being stripped from `textContent` of a `<div>` despite `white-space: pre` being set. As this did indeed work before, this seems to be a recent browser behavior change. By reverting back to using a `<textarea>` the clipboard event target, newlines work as expected.

### DIFF
--- a/guacamole/src/main/webapp/app/clipboard/services/clipboardService.js
+++ b/guacamole/src/main/webapp/app/clipboard/services/clipboardService.js
@@ -135,14 +135,23 @@ angular.module('clipboard').factory('clipboardService', ['$injector',
      */
     var selectAll = function selectAll(element) {
 
-        // Generate a range which selects all nodes within the given element
-        var range = document.createRange();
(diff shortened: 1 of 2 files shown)


necouchman:
Looks good to me.  I can merge it to master here, shortly.

mike-jumper:
Thanks, @necouchman!

PR #186, opened 2017-09-04
GUACAMOLE-128: Fix clipboard bounce.

mike-jumper:
Under IE11, the clipboard integration code can use an unusual amount of CPU, to the point of crippling the browser. This is due to a number of factors:

1. Copy/paste within content-editable `<div>` in IE11 is extremely slow (this is partly addressed by #185).
2. Attempts to read/write the clipboard result in events which the Guacamole interface listens on to determine whether to re-read the clipboard, resulting in a never-ending read/write cycle, even though the clipboard is not actually changing.
3. Depending on timing, clipboard read attempts may overlap. Since all read operations share the same event target, this may result in clipboard contents being duplicated (pasted multiple times within the event target prior to reading the event target contents).

This change switches event handling order to ensure that read/write attempts will not themselves result in yet more attempts, and modifies `getLocalClipboard()` to consistently return the same promise while a read attempt is still in progress (thus avoiding overlap).

### DIFF
--- a/guacamole/src/main/webapp/app/clipboard/services/clipboardService.js
+++ b/guacamole/src/main/webapp/app/clipboard/services/clipboardService.js
@@ -86,6 +86,7 @@ angular.module('clipboard').factory('clipboardService', ['$injector',
     };
 
     // Prevent events generated due to execCommand() from disturbing external things
+    clipboardContent.addEventListener('cut',   stopEventPropagation);
     clipboardContent.addEventListener('copy',  stopEventPropagation);
     clipboardContent.addEventListener('paste', stopEventPropagation);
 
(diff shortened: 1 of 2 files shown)


necouchman:
Looks good to me - seems pretty straight-forward.  Will merge here in a few.

limenghui614:
After using RDP connection, tens of thousands of lines of data copied from file A to file B in the target machine will be disconnected. 'check_memory: func xzalloc size 413042130' will be reported. How to solve this problem?

mike-jumper:
@limenghui614, please do not use ancient pull requests as a place to ask questions. Use the user@guacamole.apache.org mailing list instead.

PR #187, opened 2017-09-21
GUACAMOLE-385: Add a convenience filter to cache the contents of HTTP requests

nacx:
Some implementations of the HTTP servlet request do not allow a free use of the getInputStream() and getParameter() methods. This is especially annoying in Apache Tomcat >= 7 and Servlet 3.0, where invoking the getParameter() method consumes the InputStream (making later uses of it fail) and consuming the InputStream makes later getParameter() calls fail to find the parameters, as the request body has been consumed.

This causes issues for tunnel servlets that need to access the request parameters, commonly used to pass authentication tokens or additional info provided in the "tunnel.connect()" call, as the tunnel needs later access to the request input stream. Applications using the popular Spring framework may be affected too, since some of its common filters also access the request parameters.

This PR adds a convenience filter that caches the request contents to allow free access to its body and parameters, so servlets and filters can access the request information normally. Without this caching, if the request parameters are accessed in the tunnel servlet or a previous filter, the `doWrite` method in the base HTTP tunnel servlet always gets an already consumed InputStream and does not write anything to the socket, causing an eventual timeout in the client.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Tunnel.js
+++ b/guacamole-common-js/src/main/webapp/modules/Tunnel.js
@@ -306,7 +306,7 @@ Guacamole.HTTPTunnel = function(tunnelURL, crossDomain) {
             var message_xmlhttprequest = new XMLHttpRequest();
             message_xmlhttprequest.open("POST", TUNNEL_WRITE + tunnel.uuid);
             message_xmlhttprequest.withCredentials = withCredentials;
-            message_xmlhttprequest.setRequestHeader("Content-type", "application/x-www-form-urlencoded; charset=UTF-8");
+            message_xmlhttprequest.setRequestHeader("Content-type", "application/octet-stream");
 
             // Once response received, send next queued event.
             message_xmlhttprequest.onreadystatechange = function() {


necouchman:
Ignasi,
Thanks for the contribution.  A few things need to be cleaned up, though.  First, please file a JIRA issue for this.  All Pull Requests must have an associated JIRA issue.  Once you have the JIRA issue, change the title of the pull request so that it is prepended with the JIRA issue (GUACAMOLE-XXX: Add a convenience...), and then update your commits with that same tag.

Second, please check your code against the style guidelines and make sure everything is consistent.

You can find the documentation for contributing, pull requests, and style, here:
http://guacamole.incubator.apache.org/open-source/

nacx:
@necouchman Done. Please, let me know if there is still any style inconsistency. Thanks!

mike-jumper:
I'm concerned about this filter. Caching all request bodies sounds potentially dangerous, particularly for large requests, and re-implementing the parsing of HTTP parameters which the servlet API should be doing for us makes me uneasy.

@nacx, can you provide a concrete example for a case where this would be necessary? Is the need isolated to `tunnel.connect()` for the HTTP tunnel?

nacx:
Thanks for reviewing @mike-jumper!

I know it is not an ideal solution, but a limitation of a flaky JEE spec. According to the documentation of the [getParameter](http://docs.oracle.com/javaee/7/api/javax/servlet/ServletRequest.html#getParameter-java.lang.String-) method:

>If the parameter data was sent in the request body, such as occurs with an HTTP POST request, then reading the body directly via getInputStream() or getReader() can interfere with the execution of this method.

That is the spec. It leaves it up to the implementation to decide what that *can interfere* means. In the case of Tomcat versions <= 6, calling the `getParameter` method internally called `getInputStream`, so just overriding that method worked. Starting from Tomcat 7, that changed and both methods directly access (and consume) the internal buffer of data so you need to cache both if you want those methods to behave consistently. It is a problem of the spec itself, that does not offer a consistent definition.

The problem in the HTTP tunnel servlet is [here](https://github.com/apache/incubator-guacamole-client/blob/master/guacamole-common/src/main/java/org/apache/guacamole/servlet/GuacamoleHTTPTunnelServlet.java#L466-L467). When getting the input stream there, if someone (at any point) has read any request parameter, the input stream is already consumed, and the writer has nothing to write to the socket, causing the mentioned timeout. It can be easily reproduced by calling, for example `tunnel.connect("foo=bar")` on the client, and have the tunnel servlet calling `request.getParameter("foo")` in the `doConnect` method, before creating the socket and tunnel. That simple thing breaks the protocol causing the console to be unresponsive.

But let me remark that it is not a problem of someone implementing a servlet that reads parameters. If you use the default `tunnel.connect()` without parameters and a servlet that does not inspect the request at all, but use a framework (such as Spring) that configures filters that at some point read the request parameters, the same behavior will appear. It is so flaky (because of the vague spec) that IMO guacamole should provide a default workaround, especially given the pain that is to debug this kind of issues.

This said, the only alternative approach would be to avoid calling `request.getInputStream()` directly in the base tunnel servlet class, but I don't know if this is possible, as it is passing everything directly to the socket and that's probably a requirement to support arbitrary new protocols.

I agree that the cache is not a perfect solution but I think it is the only viable one that is left at this point. But we don't need to cache *all* requests. As I mentioned in the javadoc, only the requests that go to the tunnel servlet should be cached. The Guacamole requests I've seen so far are not big, so I'm pretty confident this won't present memory issues (but I could be wrong as I'm still getting familiar with Guacamole :)). An improvement could be to just cache `write` requests. The write operation is in the query line and we don't need to read the body to determine the operation, so we could add some logic to only cache POST requests for the *write* operation (this, however, couples the filter to an internal behavior of the HTTP tunnel servlet, so I decided to not apply this, given the small request assumption I made before).

Does this make sense?

nacx:
Also note that using it is completely optional. It's just there to help users that experience the same issues when using frameworks/whatever that inspect the request parameters before the requests reach the tunnel servlet.

mike-jumper:
> ... It can be easily reproduced by calling, for example tunnel.connect("foo=bar") on the client, and have the tunnel servlet calling request.getParameter("foo") in the doConnect method, before creating the socket and tunnel. That simple thing breaks the protocol causing the console to be unresponsive. ...

This should not be the case - reading from the parameters, etc. should be safe from within `doConnect()`. The places where this is definitely *not* safe would be within the actual internal read/writes of the HTTP tunnel.

If this is true, and the connect functionality breaks with parameters, then that sounds like the thing that needs to be addressed, most likely within the HTTP tunnel itself.

Adding overall caching of requests in front of the tunnel doesn't seem a good solution for this.

> ... The write operation is in the query line and we don't need to read the body to determine the operation, so we could add some logic to only cache POST requests for the write operation ...

Caching the body of the write operation would definitely hurt things. The HTTP tunnel depends on streaming the write operation by reading from that input stream.

mike-jumper:
> ... It's just there to help users that experience the same issues when using frameworks/whatever that inspect the request parameters before the requests reach the tunnel servlet.

Would such frameworks not render `getInputStream()` useless for all servlet applications? In the case of the framework that you're using, is there no way to configure that framework to not eat the input stream for absolutely all requests?

nacx:
>This should not be the case - reading from the parameters, etc. should be safe from within doConnect(). The places where this is definitely not safe would be within the actual internal read/writes of the HTTP tunnel.

My apologies, that was a bad example and I didn't explain it properly. I does not affect the `doConnect` method, so let's forget about that. The failure happens as I mentioned in the second example, in write requests.

>Caching the body of the write operation would definitely hurt things. The HTTP tunnel depends on streaming the write operation by reading from that input stream.

Could you elaborate on what *would definitely hurt things* exactly means? Take into account that the current implementation just caches the body of the requests if they are `application/x-www-form-urlencoded`, that is, we're not parsing and caching arbitrary request bodies, but only those that are form-params (and I think it is safe to expect that those kind of requests are not significantly big).

Also, the `cache` term is usually read with skepticism. In this case, we are only eagerly reading the form-params in the body to make sure the parameters and the request input stream can be accessed more than once (something the servlet spec leaves simply undefined). It is not a big cache or a complex one we need to invalidate, etc. We don't need to deal with the complexity of caches; we're just eagerly reading it to allow multiple later reads.

>Would such frameworks not render getInputStream() useless for all servlet applications? In the case of the framework that you're using, is there no way to configure that framework to not eat the input stream for absolutely all requests?

Most frameworks, say Spring, JAX-RS implementations, etc, that do this kind of request *inspection* already deserialize the requests in their own request objects and perform no later access to the raw request input stream.

Regarding the configuration, yes, it could be done. But there are very common patterns where that would not be desirable. For example, it is a common practice to use HTTP filters to secure access to applications, and it is a very common practice to map security access filters to `<url-pattern>/*</url-pattern>`. It is fair to assume that access to the tunnel servlet should be secure too, and that users may configure in front of it a set of filters they're not directly in control of (being the Spring Security filters, or whatever is offered by the framework or their custom app).

In my opinion, having a requirement in Guacamole that says that you can't put anything that reads the request parameters in front of the tunnel servlet is a very strong requirement that will affect adoption. Having an optional workaround users can use if needed would be very valuable, especially taking into account the pain that is debugging this thing and putting together a fix that works.

mike-jumper:
> Could you elaborate on what would definitely hurt things exactly means? Take into account that the current implementation just caches the body of the requests if they are `application/x-www-form-urlencoded`, that is, we're not parsing and caching arbitrary request bodies, but only those that are form-params (and I think it is safe to expect that those kind of requests are not significantly big).

The HTTP tunnel depends on reading the POST body as a stream via `getInputStream()`, writing the Guacamole protocol instructions within that body (received from the client) to the socket connected to guacd:

https://github.com/apache/incubator-guacamole-client/blob/1c0ee41d0ecd5bc4a3550804b74b73b901e074c2/guacamole-common/src/main/java/org/apache/guacamole/servlet/GuacamoleHTTPTunnelServlet.java#L473-L480

This new filter is written to pre-read the contents of the entire input stream during processing of the request:

https://github.com/nacx/incubator-guacamole-client/blob/b7e903d53e30e73c59bad435bebfa0e27934557a/guacamole-common/src/main/java/org/apache/guacamole/servlet/GuacamoleHTTPCachingFilter.java#L130

providing those contents as a `ByteArrayInputStream` to callers of `getInputStream()`:

https://github.com/nacx/incubator-guacamole-client/blob/b7e903d53e30e73c59bad435bebfa0e27934557a/guacamole-common/src/main/java/org/apache/guacamole/servlet/GuacamoleHTTPCachingFilter.java#L150

while also masking exceptions which would otherwise be directly available to readers of the input stream:

https://github.com/nacx/incubator-guacamole-client/blob/b7e903d53e30e73c59bad435bebfa0e27934557a/guacamole-common/src/main/java/org/apache/guacamole/servlet/GuacamoleHTTPCachingFilter.java#L139-L142

This concerns me because:

* This caching introduces an additional layer of latency.
* The body can potentially be quite large, particularly if file uploads, audio input, etc. are active.
* The filter implementation handles its own parsing of query parameters, etc., removing some of the benefit of depending on well-established servlet API implementations from Tomcat, Jetty, etc.

> In my opinion, having a requirement in Guacamole that says that you can't put anything that reads the request parameters in front of the tunnel servlet is a very strong requirement that will affect adoption. Having an optional workaround users can use if needed would be very valuable ...

I have to disagree here.

Providing some sort of a workaround for such cases could be useful for users of such frameworks, yes, however I do not believe a filter which caches request bodies and masks exceptions is a good solution which belongs within the upstream API. It seems like the sort of solution that should not be endorsed.

Ultimately, Guacamole by its nature depends on receiving data as a stream. Interfering with that stream interferes with one of the core requirements for reasonably-performant remote desktop. If Spring Security filters (and similar parts of other frameworks) interfere with normal use of the standard `getInputStream()`, I'd expect it would be within the scope of those frameworks to provide their own workaround to allow servlets which read the POST body to continue to function. I'm not sure the Guacamole API is the place for this sort of workaround to live, and the implication that all mainstream applications / APIs which use `getInputStream()` should provide their own workarounds for frameworks which kill `getInputStream()` does not seem reasonable to me.

If, on the other hand, there is some way to modify the HTTP tunnel such that:

* It is compatible with frameworks that munch the input stream.
* Performance is not harmed for users that do not use such frameworks, if/when Guacamole is forced to fallback to HTTP.

that would certainly be very welcome.

mike-jumper:
For what it's worth, the mainline web application does use JAX-RS, and filters everything through Guice:

https://github.com/apache/incubator-guacamole-client/blob/0e2da5a2a0da10572909b4958cc79a3d6be1d740/guacamole/src/main/webapp/WEB-INF/web.xml#L31-L39

with Guice properly handling even the HTTP tunnel:

https://github.com/apache/incubator-guacamole-client/blob/0e2da5a2a0da10572909b4958cc79a3d6be1d740/guacamole/src/main/java/org/apache/guacamole/tunnel/TunnelModule.java#L88-L107

mike-jumper:
@nacx, assuming that you have a deployment which fails unless your filter is in place, can you please try:

1. Removing your filter
2. Modifying `Tunnel.js` within guacamole-common, specifying "application/octet-stream" for the content type, rather than "application/x-www-form-urlencoded": https://github.com/apache/incubator-guacamole-client/blob/0e2da5a2a0da10572909b4958cc79a3d6be1d740/guacamole-common-js/src/main/webapp/modules/Tunnel.js#L306-L309

I suspect we shouldn't be using that content type at all in this case, and we are erroneously informing downstream consumers of tunnel requests, including filters, that the request body contains parameters that should be parsed.

nacx:
This makes much more sense. I assumed file uploads were not going to be cached since I assumed they wouldn't use the form-params media type. I wasn't sure if changing the media type for all write requests was a good thing (as I'm still quite new to Guacamole) but given your comments, I agree it makes sense.

I've tried the fix and it works like a charm, so I've updated the PR. Many thanks for your review and constructive discussion!

mike-jumper:
Thanks!

PR #188, opened 2017-09-24
GUACAMOLE-392: add basic travis-ci integration

chipitsine:


.travis.yml
```diff
@@ -0,0 +1,14 @@
...
+
+install: true
+
+cache:
+  directories: [ $HOME/.m2 ]
+
+language: java
+
+script:
+  - mvn install -DskipTests=true -Drat.skip=true -Dmaven.javadoc.skip=true -B -V
```
mike-jumper: As with apache/incubator-guacamole-server#110, I'm not confident Travis CI is a good choice. Depending on the result of discussion, if we do choose to go with Travis, skipping parts of the build shouldn't be done, especially tests and the RAT check. Catching test failures is one of the key points of using CI.

chipitsine: as for skipping RAT check, I left that to those who understands how to deal with that :) I do not insist in skipping actually

mike-jumper:
@chipitsine - as with apache/guacamole-server#110, please close this PR when you can. We are now using Jenkins for CI.

chipitsine:
closing in favour of jenkins

PR #189, opened 2017-09-25
GUACAMOLE-395: Populate "expired" property of Guacamole users defined via MySQL / MariaDB.

mike-jumper:
The mapping for the `expired` column of `guacamole_user` is missing in the read direction for MySQL. For reference, see the corresponding PostgreSQL mapping:

https://github.com/apache/incubator-guacamole-client/blob/b90a98946359c9b0b7e4af8571034961d47322fa/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/resources/org/apache/guacamole/auth/jdbc/user/UserMapper.xml#L26-L44

The column is queried in each `SELECT`. The value is simply dropped upon being queried due to that part of the result being unmapped, leaving the `expired` property of each user at the initial value of `false`.


necouchman:
Looks fine, will merge here in a few...

PR #190, opened 2017-09-25
GUACAMOLE-393: Add extension logout/shutdown hooks

mike-jumper:
This change modifies the Guacamole extension API, adding the following additional hooks:

Class / Interface        | Function       | Description
------------------------ | -------------- | -----------
`AuthenticationProvider` | `shutdown()`   | The web application is being shut down, and the `AuthenticationProvider` should clean up any resources which will not be automatically freed.
`AuthenticatedUser`      | `invalidate()` | The user associated with the `AuthenticatedUser` is logging out, or their session has been automatically invalidated.
`UserContext`            | `invalidate()` | The user associated with the `UserContext` is logging out, or their session has been automatically invalidated.

Existing extensions have been updated accordingly.


necouchman:
Looks pretty straight-forward, no issues, here.  Will merge here in a few minutes.

PR #191, opened 2017-09-25
GUACAMOLE-394: Refactor extension API to define user history

mike-jumper:
This change:

* Extracts an `ActivityRecord` base interface from the `ConnectionRecord` interface, leveraging that to represent both user and connection history entries.
* Deprecates the connection-specific `ConnectionRecordSet` interface in favor of a new, generic `ActivityRecordSet` interface.
* Adds history retrieval functions for `User` objects, analogous to those already available for `Connection` objects.
* Adds a `getLastActive()` function to both `User` and `Connection` with similar semantics for each.

Existing extensions have been updated accordingly, though the database auth backend has not been updated to actually record these history entries.


guacamole-ext/src/main/java/org/apache/guacamole/net/auth/ActivityRecordSet.java
```diff
@@ -0,0 +1,128 @@
...
+     *
+     * @throws GuacamoleException
+     *      If an error occurs while retrieving the records within this set.
+     */
+    Collection<RecordType> asCollection() throws GuacamoleException;
+
+    /**
+     * Returns the subset of records which contain the given value. The
+     * properties and semantics involved with determining whether a particular
+     * record "contains" the given value is implementation dependent. This
```
jmuehlner: It doesn't look like any implementations actually mutate the `ActivityRecordSet` when `contains()` is called. Do you expect that this will be the case? If so, why?

mike-jumper: Yep - since most of the `ActivityRecordSet` functions return a possibly-new `ActivityRecordSet` based off the original, it's likely that explicitly allowing these functions to mutate the original will simplify implementations, as well as free implementations to make use of optimizations that rely on shared state.

The JDBC version actually does mutate the original:

https://github.com/apache/incubator-guacamole-client/blob/1c0ee41d0ecd5bc4a3550804b74b73b901e074c2/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionRecordSet.java#L77-L82

jmuehlner: Ahhh ok. I gotcha. This makes sense to me.

jmuehlner:
Ok, I think this all makes sense to me. Any concerns @necouchman?

necouchman:
Good by me.  I can merge it here, shortly.

PR #192, opened 2017-09-25
GUACAMOLE-210: Add support for single sign-on with OpenID Connect

mike-jumper:
A mere two thousand lines of new code, this tiny pull request adds a new extension, "guacamole-auth-openid", which provides supports for single sign-on with OpenID Connect.

Similar to the existing CAS extension, the OpenID extension requires several properties which define how Guacamole should connect to the OpenID service. When properly configured, the Guacamole login screen will redirect users to the OpenID service, which then generates a token authenticating the user and redirects them back to Guacamole. Once the token has been validated, and the identity of the user has been determined, the user is allowed in.


extensions/guacamole-auth-openid/src/main/java/org/apache/guacamole/auth/openid/AuthenticationProviderService.java
```diff
@@ -0,0 +1,132 @@
...
+     *     given credentials.
+     *
+     * @throws GuacamoleException
+     *     If an error occurs while authenticating the user, or if access is
+     *     denied.
+     */
+    public AuthenticatedUser authenticateUser(Credentials credentials)
+            throws GuacamoleException {
+
+        String username = null;
```
necouchman: Nitpick, but isn't a Java String already null by default?  Any reason to explicitly set it?  I'm sure I'm guilty of the same, just curious...

extensions/guacamole-auth-openid/src/main/java/org/apache/guacamole/auth/openid/AuthenticationProviderService.java
```diff
@@ -0,0 +1,132 @@
...
+                username = tokenService.processUsername(token);
+        }
+
+        // If the username was successfully retrieved from the token, produce
+        // authenticated user
+        if (username != null) {
+
+            // Create corresponding authenticated user
+            AuthenticatedUser authenticatedUser = authenticatedUserProvider.get();
+            authenticatedUser.init(username, credentials);
```
necouchman: If this SSO provider works the same as the Header and CAS provider do (Redirect to SSO login, then back to Guacamole, skipping the Guacamole username/password login), then I believe this will encounter the same challenge as those do with the username value (and password value) missing from the credentials object, and, thus, the `${GUAC_USERNAME}` token not functioning correctly.

I'm not sure we ever agreed upon the proper solution for that, but just FYI.

extensions/guacamole-auth-openid/src/main/java/org/apache/guacamole/auth/openid/AuthenticationProviderService.java
mike-jumper: Not exactly. Only the properties of an object have default values. For example:

    public class Foo {

        private String username; // This is null by default

        private int foo; // This is 0 by default

    }

but:

    public void foo() {

        String username; // This is not initialized, and the compiler will warn us if we try to use it

        int foo; // This is not initialized either

    }

It's good to avoid unnecessary initialization since it allows the compiler to catch cases where we didn't mean to use a variable in a particular place. In this case, initialization is necessary as a catch-all representing that the username could not be retrieved, for any one of several reasons.

mike-jumper:
> I guess there's one other big-picture question - I'm not familiar with OpenID Connect, specifically, but is there something about the OpenID Connect implementation of OAuth2 that requires this module to be specific to OpenID and not support OAuth2 more generically?

That's a common point of confusion.

While OpenID explicitly deals with identity, OAuth and OAuth 2.0 are authorization frameworks. It is true that OpenID Connect is built upon OAuth 2.0, but OAuth is not a generic version of an identity layer. On the contrary, it is OpenID which is the generic standard for identity.

See also: https://oauth.net/articles/authentication/

extensions/guacamole-auth-openid/src/main/java/org/apache/guacamole/auth/openid/AuthenticationProviderService.java
mike-jumper: Good point - I'll fix that.

extensions/guacamole-auth-openid/src/main/java/org/apache/guacamole/auth/openid/AuthenticationProviderService.java
necouchman: Got it, thanks for the clarification.

necouchman:
Sounds good - thanks for clearing up the relationship between OpenID and OAuth.  Once that username issue is tweaked I'll be happy to merge it to master.

mike-jumper:
I think I have a general solution for the username issue via #193.

necouchman:
> I think I have a general solution for the username issue via #193.

Will go ahead and merge this since that PR has been merged, too.

PR #193, opened 2017-09-26
GUACAMOLE-341: Automatically pull GUAC_USERNAME token from AuthenticatedUser.

mike-jumper:
This change is meant to take the place of the spot fix implemented by #174, automatically pulling the "GUAC_USERNAME" token from the identifier of the `AuthenticatedUser` object if the associated `Credentials` to not have an explicit username set.


necouchman:
Looks fine to me...will merge here in a few.

necouchman:
One post-merge question...is this going to impact at all how the current work on CAS ClearPass integration works, or is this for username, only?

mike-jumper:
This just adds one additional location from which the username will be pulled, and does not affect the password. If the username is present on the `Credentials` at all, that value will be used instead of the `AuthenticatedUser` identifier.

PR #194, opened 2017-10-06
GUACAMOLE-221: Support for Connection Prompts

necouchman:
First, this is a looooooooooong shot.  I make no claims about the quality of this code, other than to say it works.  This will likely require significant cleanup to get it in shape enough to actually merge - if it's worth merging.  If it needs to be thrown out and started from scratch, just let me know...

There was talk on the JIRA issue about how to accomplish this, and one of the ideas tossed around was throwing an exception at connection time if prompts were required.  There were several reasons I ended up not going this direction, but at least one of them was because the tunnel connection, when it throws exceptions, does not throw them in JSON format, it actually generates the standard Tomcat HTML pages, which would have to be changed.  Another reason was the complexity of dealing with starting the tunnel, dealing with the exception, and then restarting the tunnel.

I also considered going the route of using a REST endpoint to create the connection, getting the UUID and prompts, and then passing through the UUID and prompt values to the tunnel connection. I was able to get this to work; however, it seemed overly-complex.

So, after a few months of working on this off-and-on, this really seems like the simplest solution - grab the list of prompts ahead of time, evaluate whether anything is required, present the prompts, then pass everything through to the tunnel connection.


guacamole-ext/src/main/java/org/apache/guacamole/token/PromptEntry.java
```diff
@@ -0,0 +1,157 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.guacamole.token;
```
necouchman: Not sure this is the right place for this...??

guacamole-ext/src/main/java/org/apache/guacamole/token/PromptEntry.java
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * The Field data associated with the prompt.
+     */
+    private Field field;
+
+    /**
+     * The entire value for the parameter being prompted.
+     */
+    private String value;
```
necouchman: Not sure I even use this anywhere...need to double-check this is really worthwhile.

guacamole-ext/src/main/java/org/apache/guacamole/token/StandardTokens.java
```diff
@@ -51,6 +51,16 @@
...
+     * The token that triggers prompts for numeric fields.
+     */
+    public static final String PROMPT_TOKEN_NUMERIC = "-1";
+
+    /**
+     * The token that triggers prompts for string fields.
+     */
+    public static final String PROMPT_TOKEN_STRING = "GUAC_PROMPT";
+
+    /**
```
necouchman: Are these "standard" tokens?

guacamole/src/main/java/org/apache/guacamole/rest/connection/ConnectionResource.java
```diff
@@ -193,4 +207,52 @@ public ConnectionResource(@Assisted UserContext userContext,
...
+        // Get the connection configuration parameters and then filter for prompts.
+        Map <?, String> parameters = connection.getConfiguration().getParameters();
+        Map <String, List<String>> prompts = TokenFilter.getPrompts(parameters);
+
+        // Get protocol information for the configured protocol.
+        Collection<Form> myForms = environment.getProtocol(connection.getConfiguration().getProtocol()).getConnectionForms();
+
+        // Loop through each prompt and find the matching Field object for the
+        // parameter.
+        for (Map.Entry<String, List<String>> entry : prompts.entrySet()) {
```
necouchman: This is a lot of code for a rest endpoint, not sure this shouldn't be reorganized and put somewhere else??

guacamole/src/main/webapp/app/client/types/ManagedClient.js
```diff
@@ -503,27 +504,70 @@ angular.module('client').factory('ManagedClient', ['$rootScope', '$injector',
...
+
+        // When we've received connection data and prompts, display prompts to user.        
+        $q.all([gettingConnectionData,gettingConnectionPrompts])
+        .then(function connectClient(clientData) {
+
+            var connData = clientData[0].data;
+            var connPrompts = clientData[1].data;
+
+            // Display the prompts, then process them
+            guacPrompt.getUserInput(connPrompts,connData)
```
necouchman: Yeah, this is about where the code starts to get fun.  This changes some of the asynchronous nature of the JavaScript in here, which has some potential performance impacts, and I'm not sure this is real clean.  Suggestions?

necouchman:
Well, upon further self-review, this implementation only works for the JDBC module (not for LDAP or basic file), which is less-than-desirable and probably indicates that I'm relying on something I shouldn't be relying on to make it function.  I'll have to go back and see if there's a cleaner way to integrate it at a level that will work across all connection types, regardless of how they're configured.

I'd still be grateful for any other perspectives on the code for others that are willing to review...

necouchman:
Well, added a filter for the SimpleConnection object during the connect() method that should allow that to work, and rebased on current master to resolve conflicts.  Still not sure this is entirely the right way to go with all of this, but it's further than I've managed to get before.

guacamole/src/main/webapp/app/prompt/services/guacPrompt.js
```diff
@@ -0,0 +1,149 @@
...
+     * When the prompt object is displayed, prevent keypress and
+     * mouse clicks from propagating through to the client so that
+     * the prompt dialog can be completed.
+     */
+    $window.addEventListener('input', preventEvent, true);
+    $window.addEventListener('keydown', preventEvent, true);
+    $window.addEventListener('keypress', preventEvent, true);
+    $window.addEventListener('keyup', preventEvent, true);
+    $window.addEventListener('click', preventEvent, false);
+
```
necouchman: And then there's this, which I think I need to replace with the keyboard disable stuff Mike sent me...

lorodoes:
Would this allow for Username and password for SSH/RDP and password for VNC to be used for a prompt?

Also, what about SSH keys?

necouchman:
> Would this allow for Username and password for SSH/RDP and password for VNC to be used for a prompt?

Yes.

> Also, what about SSH keys?

Yes.

lorodoes:
So, if i automated adding the hosts in to the database or user-mappings.xml, this would allow for a prompt to occur during connection so that the guacamole box would never permanently have the SSH keys correct?

Also, would it be possible to get a patch file so I can try this in my dev/test environment?

necouchman:
> So, if i automated adding the hosts in to the database or user-mappings.xml, this would allow for a prompt to occur during connection so that the guacamole box would never permanently have the SSH keys correct?

Yes, that's the idea - also, specifically for NLA-enabled RDP connections, at present you have to either use the GUAC_USERNAME/GUAC_PASSWORD tokens (which is fine, but doesn't always work depending on authentication method), or embed the credentials in the connection.

> Also, would it be possible to get a patch file so I can try this in my dev/test environment?

You can check out the branch from my github repo (this PR should point to it) and build it from there.  Keep in mind a couple of things:
- This code hasn't been reviewed by anyone else, yet, so it is probably fraught with bugs and things that need fixing, and it's quite possible that the other folks on the project will look at it and tell me that it needs to be scrapped and started from scratch.  It's very early in the development of this type of feature.
- It's based a few commits back in the master repo, so probably contains some bugs that have since been fixed in other areas.  I probably need to rebase it, here, soon, which will take care of that, but that's the status at the moment.

necouchman:
> It's based a few commits back in the master repo, so probably contains some bugs that have since been fixed in other areas. I probably need to rebase it, here, soon, which will take care of that, but that's the status at the moment.

Rebased to the current master, and it still appears to compile.  So, this one is out of the way - but, it's still unreviewed code.  You've been warned :-).

lorodoes:
So, if I am understanding this correctly. I have to do restful API requests to get the prompt and then supply the information when the prompt is given via the API is that correct?

necouchman:
> So, if I am understanding this correctly. I have to do restful API requests to get the prompt and then supply the information when the prompt is given via the API is that correct?

Yes - all of that is included in the current pull request.  There's a new REST endpoint that enumerates the prompts, and then you feed those into the tunnel?connect request.  If you're just using the Guacamole Client you shouldn't have to do anything special, it *should* just work.  If you're writing your own extension or customizing the client, then, yes, you'll need to look at how the prompts are handled.

Again, I would caution that this is very preliminary and that I believe the likelihood of this changing significantly after others review it is very high.  At this point, though, most of the focus and effort is going in to the final stuff that needs to be done for the 0.9.14 release, which is at least one of the reasons this isn't getting a lot of attention.

guacamole-ext/src/main/java/org/apache/guacamole/token/PromptEntry.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.apache.guacamole.form.Field;
+
+/**
+ * A class that collects all of the information required to
+ * to display a prompt to the user during client connection.
+ */
+public class PromptEntry {
```
mike-jumper: So, I'm confused by the overall purpose of the `PromptEntry` class? What does it represent? What are these mysterious string positions?

necouchman: So, the main point is to provide a data structure that can then be presented by the REST API such that the front-end application can provide a list of fields where prompts are necessary.  The `Field` is tracked so that we know both the name and the type of entry this should be (text vs. password vs. number, etc.), and the `positions` is so that you can cope with the following possible scenarios:
- The entire field (e.g. username) is being prompted.
- One or more positions within a field are being prompted.  The idea here is that there may be some text within a parameter, such as a filesystem location or directory path, that I want to hard-code as the administrator, while I want the user to be able to provide some sort of input to the field.  As an example, let's say I want to prompt the user for a folder within their home directory that they want to pass through to a RDP connection - the text in the connection configuration might be:

`/home/${GUAC_USERNAME}/${GUAC_PROMPT}`

Or, perhaps you want to allow them to enter a hostname for a server, but within a defined range of servers.  You might do:

`vdi-${GUAC_PROMPT}.example.com`

or:

`${GUAC_PROMPT}.vdi.example.com`

Thus, the user would not be allowed to override everything about that parameter, just provide some input within a scope that the administrator has defined.

mike-jumper: > As an example, let's say I want to prompt the user for a folder within their home directory that they want to pass through to a RDP connection - the text in the connection configuration might be:
>
>     /home/${GUAC_USERNAME}/${GUAC_PROMPT}

If a malicious user entered "../../" for that, they would gain access to the root directory.

> Thus, the user would not be allowed to override everything about that parameter, just provide some input within a scope that the administrator has defined.

This may not actually end up being what happens, depending on the semantics of the parameter.

I'm still unclear as to why `positions` is necessary, or what its values are intended to be. Can you describe how `positions` would be used in the examples you provided?

necouchman:
> If a malicious user entered "../../" for that, they would gain access to the root directory.

Yes, those situations would have to be handled.

> I'm still unclear as to why positions is necessary, or what its values are intended to be. Can you describe how positions would be used in the examples you provided?

The idea was that there is a possibility, however remote, that someone could specify multiple `${GUAC_PROMPT}` tokens within the same field.  This can certainly be done using the "normal" tokens - you could have a path that contains `/home/${GUAC_USERNAME}/tmp/${GUAC_USERNAME}` that would substitute the username twice.  So, going with the token idea, the thought was, what if someone entered something like this:
`/home/${GUAC_USERNAME}/${GUAC_PROMPT}/${GUAC_PROMPT}`

The `positions` field keeps track of how many times the prompt occurs for a given field so that multiple prompts can be presented and then filled in.  This may not be a case worth dealing with - maybe this is too far-fetched to worry about.  I was just going for consistency of behavior across token-like objects.

lorodoes:
So, if I want to use this function with the client I should replace the username and password/SSH key with ${GUAC_PROMPT}?

necouchman:
> So, if I want to use this function with the client I should replace the username and password/SSH key with ${GUAC_PROMPT}?

That's the idea, yes, and, if you work from this branch or apply the diff, that's what should happen.  I warn, again, that this probably has a long way to go before it actually gets merged - and I think there's still a possibility this PR gets closed out and development moves in another direction for this type of functionality.  It needs to be done, I'm just not 100% sure this is the route we will ultimately take.

You've been warned :-).

guacamole-ext/src/main/java/org/apache/guacamole/token/StandardTokens.java
```diff
@@ -50,6 +50,16 @@
...
 
+    /**
+     * The token that triggers prompts for numeric fields.
+     */
+    public static final String PROMPT_TOKEN_NUMERIC = "-1";
+
+    /**
+     * The token that triggers prompts for string fields.
+     */
+    public static final String PROMPT_TOKEN_STRING = "GUAC_PROMPT";
```
uvtrip: In case the whole string needs to be prompt for, maybe you can also check for null values
(same for numeric field as well)
instead of hardcoded -1 or string?

necouchman: If the whole string needs to be prompted, then you just put ```${GUAC_PROMPT}``` in for the whole string.  In Guacamole null or empty values indicate the absence of a parameter and delete that parameter from the underlying database table (in the JDBC modules, anyway), so prompting for an empty value would not be practical and would interfere with some of the basic ways in which parameters are evaluated, aside from which it isn't really necessary.

guacamole/src/main/webapp/app/client/types/ManagedClient.js
```diff
@@ -503,27 +504,70 @@ angular.module('client').factory('ManagedClient', ['$rootScope', '$injector',
...
+
+                // Create a parameter string from the received data
+                var userData = '';
+                for (var key in data) {
+                    var param = data[key];
+                    for (var idx in param) {
+                        var inst = param[idx];
+                        if (userData != '')
+                            userData += '&';
+                        userData += key + '[' + idx + ']=' + inst;
```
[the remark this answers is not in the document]
necouchman: Yep, you're right - will add that in.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleConnection.java
```diff
@@ -98,6 +99,10 @@ public void setAttributes(Map<String, String> attributes) {
     public GuacamoleTunnel connect(GuacamoleClientInformation info)
             throws GuacamoleException {
 
+        // Filter in connection-time information
+        TokenFilter tokenFilter = new TokenFilter();
+        tokenFilter.filterPrompts(config.getParameters(),info.getParameters());
```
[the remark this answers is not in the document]
necouchman: Yeah, I think I see your point, here.  Will see what options there are to take care of this.

guacamole/src/main/webapp/app/prompt/services/guacPrompt.js
```diff
@@ -0,0 +1,149 @@
...
+            event.stopPropagation();
+    };
+
+    /**
+     * When the prompt object is displayed, prevent keypress and
+     * mouse clicks from propagating through to the client so that
+     * the prompt dialog can be completed.
+     */
+    $window.addEventListener('input', preventEvent, true);
+    $window.addEventListener('keydown', preventEvent, true);
```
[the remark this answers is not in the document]
necouchman: Ok.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleConnection.java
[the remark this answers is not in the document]
uvtrip: i think you should save the prompts per each configuration, and maybe calculate them in advance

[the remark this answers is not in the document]
necouchman: I thought about that, and one of my earlier implementation attempts actually did function this way.  The reason I opted for this sort of inline processing method was to be consistent with the way other tokens are processed, which is done at connection time rather than being done ahead of time.  I think this is probably mostly due to the fact that the GUAC_DATE and GUAC_TIME tokens are expected to be processed in real-time so that their values are current.

That said, the prompts don't necessarily have to work the same as the other tokens if there's a compelling reason to do something else.  I went back and forth on this a lot, and eventually ended up at this point.

I think for this particular issue you raise I'll just make a copy of the config object and filter that one for prompts, and it can get discarded after the connection is finished.

guacamole/src/main/webapp/app/client/types/ManagedClient.js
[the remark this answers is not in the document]
necouchman: Should be encoded, now, in the `getConnectString()` method.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleConnection.java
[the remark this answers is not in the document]
necouchman: Okay, I'm now copying the config in this method, so that the values are placed into that non-persistent config, which will then get discarded after the connection finishes.

guacamole/src/main/webapp/app/client/types/ManagedClient.js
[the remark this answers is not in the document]
necouchman: Slight change here - I had encoded everything, including parameter name, which obviously didn't work.  Instead I should do the encoding where you suggested it :-).  Should be fixed and working, now.

guacamole/src/main/webapp/app/prompt/services/guacPrompt.js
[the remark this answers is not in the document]
necouchman: Well, this is going to be every so slightly more difficult than I had originally thought due to the way the callbacks work here for the prompt.  I'll see what I can figure out...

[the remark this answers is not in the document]
necouchman: I'm having a couple of issues getting this to work nicely - one issue in particular, with variable scope.  Also, escape key doesn't seem to be making it through - not sure what's going on, but it may be getting captured and handled elsewhere in the code such that it doesn't flow down.

I'll see what I can do, but this also may end up just being a nice-to-have - the Guacamole notification dialogs (after a disconnect, for example) currently do not handle Enter/Escape to dismiss them, so I don't think it'll be too terrible if it doesn't work, here, either.

justvanbloom:
Great. Is it possible to use the Connection promts with VNC? so maybe we get apple vnc working? that would be nice!

necouchman:
> Is it possible to use the Connection promts with VNC?

Yes, the prompting should work with any connection type.

markusheiliger:
@mike-jumper Any plans to merge this PR in the near future? I need this for my solution.

mike-jumper:
@markusheiliger, no, as Nick notes above:

> I warn, again, that this probably has a long way to go before it actually gets merged - and I think there's still a possibility this PR gets closed out and development moves in another direction for this type of functionality. It needs to be done, I'm just not 100% sure this is the route we will ultimately take.

jeremylarose:
What's the hold up on this?  It seems like an extremely major addition.  It essentially allows the use of NLA RDP connections as opposed to two less secure options of NOT using NLA or storing passwords in the database.

necouchman:
@jeremylarose 
The hold-up is exactly that it is a major change, needs a lot of review, and will probably change significantly before it gets merged in.  We have other priorities that are higher on our list of things to work on than getting this support in there, so it hasn't been completed, yet.

There are other options aside from not using NLA and storing passwords - namely integrating with LDAP and using parameter tokens to pass those authenticators through to the connections.  I work in an environment with Active Directory and a lot of RDP and use this routinely, without storing passwords and without disabling NLA.

jeremylarose:
@necouchman 
Thanks.

CAS is an example that would highly benefit from this.  Users can authenticate successfully but the ${GUAC_PASSWORD} doesn't redirect.  The only way to allow users to enter a password is disabling NLA in RDP.  

Also, not all systems use the same authentication methods.  Maybe you want to share a VNC connection with multiple users without hardcoding a password in the database.   It's a major addition, but also an important addition for versatility of guacamole.

necouchman:
@jeremylarose 
You're right, there are situations where it would be beneficial, and we definitely want to do it.  Just hasn't made it to the top of the pile, yet.

By the way, with CAS and Guacamole, you can do ClearPass, and ${GUAC_PASSWORD} will be avaialble.  However, not all SSO modules support this, so the point is still well-taken - prompting would be very useful.

DerDanilo:
Hello there. Can we please get this fixed soon? This is required and since GDPR some clients are about to ditch GUA since it is a no go for the admin to know the passwords of the customers account. Plus the fact that the passwords have to be changed every 90 days on some client systems and require the admin to coordinate the communication with the customers every time.

It would be super nice to get this done ASAP. :)

mike-jumper:
@necouchman, what if we implemented this at the instruction level? With the recent addition of the `argv` instruction, allowing the protocol implementations to request that the user be prompted for additional parameters would be in line with the prompting already implemented by SSH. Might be much simpler.

necouchman:
@mike-jumper: I'm totally open to this.  I'm not in any way tied to the current implementation.

@DerDanilo: With regard to GDPR and storing passwords, if you're authenticating against the same AD directory of which your hosts are a member you can use Parameter Tokens, instead, and you don't have to store the actual values - they will be passed through.

DerDanilo:
I know that this works. But then the AD passwords are potentially exposed. Having secondary login credentials is benefitial securitywise.

necouchman:
@DerDanilo:
> I know that this works. But then the AD passwords are potentially exposed. Having secondary login credentials is benefitial securitywise.

I don't understand how using parameter tokens exposes AD passwords any more than prompting?  In fact, it seems more secure to me, because the tokens are processed completely inside the Java application on the back-end.  They are never stored in the database, and they are never passed between the browser and the REST API.  They are injected by the Java application from in-memory values stored for the current session.

With prompting, values will be passed from the browser through to the stream, either in the tunnel or the REST API.  While this should be reasonably secure if TLS is used to secure the tunnel, I would still consider it less secure than using parameter tokens.

DerDanilo:
I'm just about to setup another set of services for a customer. Does prompting work properly with an MS AD behind?

justvanbloom:
Yes it will work.


Von meinem iPhone gesendet

necouchman:
Closing out in favor of changes in #430.

PR #195, opened 2017-10-07
GUACAMOLE-408: Use Docker multi-stage build to create guacamole image

ceharris:
Since Docker CE 17.05 (edge) and 17.06 (stable), multi-stage builds can be used to create images using different base images for build versus runtime.

This allows the build itself to be performed with the official Maven image, rather than using the Tomcat image as the base and installing the necessary dependencies for performing the Guacamole build. It also obviates to the need to clean up after the build. In the second stage of the build, the official Tomcat image is used at the base for the runtime image. The artifacts built in the first stage are simply copied from the build stage into the runtime stage.

Taking advantage of another minor feature added in Docker CE 17.05, the Tomcat version and JRE details are specified using build arguments. This allows Tomcat or underlying JRE to be easily varied by the image builder. For example, minimalists may wish to build using `docker image build --build-arg TOMCAT_JRE=jre8-alpine` to get a smaller runtime image.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,5 @@
 
 .git
-**/.gitignore
 
 **/*~
 **/target/
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,29 +21,36 @@
 # Dockerfile for guacamole-client
 #
 
-# Start from Tomcat image
-FROM tomcat:8.0.20-jre8
+# Use args for Tomcat image label to allow image builder to choose alternatives
+# such as `--build-arg TOMCAT_JRE=jre8-alpine`
+#
+ARG TOMCAT_VERSION=8.0.20
+ARG TOMCAT_JRE=jre8
 
-# Environment variables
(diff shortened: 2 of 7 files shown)


mike-jumper:
> ... When building inside of the Docker container, some files with
generated HTML are being validated by RAT and failing due to the
lack of a license header. When building outside of Docker, this
problem does not occur. ...
>
> ```xml
> <exclude>src/main/resources/generated/**</exclude>
> ```

The `.gitignore` file which includes that path must no longer be getting into the Docker build context. The RAT plugin takes the contents of `.gitignore` into account when determining the default excludes. This was encountered during testing of release .tar.gz files, as well, resulting in #91.

See: https://issues.apache.org/jira/browse/GUACAMOLE-132

With that in mind, I think the same should apply here, and we should avoid duplicating excludes.

ceharris:
That makes sense. I'll adjust accordingly and back out the duplicate exclusions.

PR #196, opened 2017-10-10
GUACAMOLE-248: Support for connection templates

necouchman:
This PR adds support for for assigning a connection as a template to another connection, and thus inheriting parameters from the template connection.

The only sort of outstanding issue that needs to be discussed and dealt with is handling checkbox values.  At this point, the a checkbox is unchecked if the value is null, and checked if it is otherwise (true).    It's possible things will work fine this way; however, the ideal solution would be to recognize no value as different from false when it comes to boolean options (checkboxes) so that you can specifically set a boolean parameter to false, and the value will not be passed through from the template.

This also implements a template chooser, similar to the location chooser, which keeps the template connections organized a little bit better than just throwing them in a select box.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java
@@ -69,7 +69,7 @@ public class ConnectionModel extends ChildObjectModel {
     private boolean failoverOnly;
 
     /**
-     * The ID of the connection that this connection is a template of.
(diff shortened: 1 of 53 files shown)


mike-jumper:
Needs further discussion, but I would be strongly -1 on this.

I'm really not sure about the idea of having skeleton connection objects serve as templates for other connection objects, or the idea of connection templates in general. I'm open to being convinced otherwise, but the concept strikes me as a hack.

necouchman:
Do you have other ideas for how to ease the administrative burden of scaling Guacamole connections?  I'm open to being convinced that there are better ways to scale managing hundreds of thousands of connections in Guacamole, and open to trying other methods for implementing it, but it's clear to me from past mailing list discussions that there needs to be some way to scale installations and keep administrative time down.  Using templates or inheritance is one such method, and, since I didn't get very much feedback when I tried to start the discussions, I went ahead and took a stab at an implementation.  I'm open to suggestions.

PR #197, opened 2017-10-12
GUACAMOLE-412: Prevent exception propagation for null UserContext

necouchman:
Changes to event listeners from GUACAMOLE-364 (#184) cause modules that do not have a UserContext object to fail at login.  This is a potential fix for that issue - preventing the GuacamoleResourceNotFoundException from propagating down and instead just throwing a warning and debug message into the log file.  Not sure if this is the preferred way to go, or if there are other implications to this?

### DIFF
--- a/guacamole-ext/src/main/java/org/apache/guacamole/net/event/AuthenticationSuccessEvent.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/net/event/AuthenticationSuccessEvent.java
@@ -19,6 +19,7 @@
 
 package org.apache.guacamole.net.event;
 
+import org.apache.guacamole.net.auth.AuthenticatedUser;
 import org.apache.guacamole.net.auth.Credentials;
 import org.apache.guacamole.net.auth.UserContext;
 
@@ -32,7 +33,7 @@ import org.apache.guacamole.net.auth.UserContext;
  * is effectively <em>vetoed</em> and will be subsequently processed as though the
  * authentication failed.
  */
-public class AuthenticationSuccessEvent implements UserEvent, CredentialEvent {
+public class AuthenticationSuccessEvent implements UserEvent, CredentialEvent, AuthenticatedUserEvent {
 
     /**
      * The UserContext associated with the request that is connecting the
@@ -43,7 +44,7 @@ public class AuthenticationSuccessEvent implements UserEvent, CredentialEvent {
     /**
      * The credentials which passed authentication.
      */
-    private Credentials credentials;
+    private AuthenticatedUser authenticatedUser;
 
     /**
      * Creates a new AuthenticationSuccessEvent which represents a successful
@@ -51,11 +52,11 @@ public class AuthenticationSuccessEvent implements UserEvent, CredentialEvent {
      *
      * @param context The UserContext created as a result of successful
      *                authentication.
-     * @param credentials The credentials which passed authentication.
+     * @param authenticatedUser The user which passed authentication.
      */
-    public AuthenticationSuccessEvent(UserContext context, Credentials credentials) {
+    public AuthenticationSuccessEvent(UserContext context, AuthenticatedUser authenticatedUser) {
         this.context = context;
-        this.credentials = credentials;
+        this.authenticatedUser = authenticatedUser;
     }
 
     @Override
@@ -65,7 +66,14 @@ public class AuthenticationSuccessEvent implements UserEvent, CredentialEvent {
 
     @Override
     public Credentials getCredentials() {
-        return credentials;
+        if (authenticatedUser == null)
+            return null;
+        else
+            return authenticatedUser.getCredentials();
+    }
+
+    public AuthenticatedUser getAuthenticatedUser() {
+        return authenticatedUser;
     }
 
 }
--- a/guacamole-ext/src/main/java/org/apache/guacamole/net/event/AuthenticationSuccessEvent.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/net/event/AuthenticationSuccessEvent.java
@@ -19,6 +19,8 @@
 
 package org.apache.guacamole.net.event;
 
+import org.apache.guacamole.GuacamoleException;
+import org.apache.guacamole.net.auth.AuthenticationProvider;
 import org.apache.guacamole.net.auth.AuthenticatedUser;
 import org.apache.guacamole.net.auth.Credentials;
 import org.apache.guacamole.net.auth.UserContext;
@@ -33,13 +35,8 @@ import org.apache.guacamole.net.auth.UserContext;
  * is effectively <em>vetoed</em> and will be subsequently processed as though the
  * authentication failed.
  */
-public class AuthenticationSuccessEvent implements UserEvent, CredentialEvent, AuthenticatedUserEvent {
-
-    /**
-     * The UserContext associated with the request that is connecting the
-     * tunnel, if any.
-     */
-    private UserContext context;
+public class AuthenticationSuccessEvent implements
+        UserEvent, CredentialEvent, AuthenticatedUserEvent, AuthenticationProviderEvent {
 
     /**
      * The credentials which passed authentication.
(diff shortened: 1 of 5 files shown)


mike-jumper:
We may have more problems than simply the unexpected ramifications of an exception:

> ```java
> userContext = session.getUserContext(
>     authenticatedUser.getAuthenticationProvider().getIdentifier());
> ```

I didn't notice this angle during my review, but it doesn't make all that much sense to fire an event with the `UserContext` associated with the `AuthenticationProvider` that produced the `AuthenticatedUser`. Back when event listeners originally existed, Guacamole only supported one `AuthenticationProvider` at a time, and there was only one `UserContext`. That API doesn't make much sense now, since multiple `UserContext` objects may exist for any one session.

necouchman:
> That API doesn't make much sense now, since multiple UserContext objects may exist for any one session.

So, are you thinking that the AuthenticationSuccessEvent needs to be changed to just get the entire GuacamoleSession object?  Or some other direction?

mike-jumper:
> So, are you thinking that the AuthenticationSuccessEvent needs to be changed to just get the entire GuacamoleSession object? Or some other direction?

Not the `GuacamoleSession` object, as that's internal to the webapp and not part of the public API, but yeah - it looks like `AuthenticationSuccessEvent` (and perhaps anything from `UserEvent` onward) needs some rethinking.

Exposing `AuthenticatedUser` instead could make sense. If the event is fired once for each `AuthenticationProvider`, exposing both `AuthenticatedUser` and `UserContext` could also work.

Any thoughts, @ceharris?

necouchman:
> If the event is fired once for each `AuthenticationProvider`, exposing both `AuthenticatedUser` and `UserContext` could also work.

...keeping in mind that, if we do this, we still have to deal with null `UserContext` objects in a way that doesn't kill the authentication process.

mike-jumper:
Sort of, but in the current API, it's `AuthenticatedUser` that defines identity (and that identity may not be readily derived from simply the `Credentials`), so receiving a `Credentials` instead of `AuthenticatedUser` makes less sense now.

As for the potential lack of `UserContext`, as it's not guaranteed to be defined for a given `AuthenticationProvider`, this should simply be `null` if unavailable.

necouchman:
> ...so receiving a `Credentials` instead of `AuthenticatedUser` makes less sense now.

Okay...I can take a stab at reworking the listeners with the `AuthenticatedUser` object instead of `Credentials`, and then see if @ceharris has any feedback or suggestions.

> As for the potential lack of `UserContext`, as it's not guaranteed to be defined for a given `AuthenticationProvider`, this should simply be null if unavailable.

Agreed, but the point of this PR is that it's the attempt to retrieve the `UserContext` object that causes the exception to be thrown that kills the authentication process when `UserContext` `null`.  So, either the code above is needed to prevent that exception from being passed on down the line, or the code in the `GuacamoleSession` object that throws the exception needs to be tweaked to just `return null` instead of throwing an exception (which then has other implications for stuff that expects a 404 error when a `UserContext` is not available.  Or is there another alternative you're thinking of to handle this?

ceharris:
Including `AuthenticatedUser` in `UserEvent` makes a lot of sense. This would allow `UserEvent.getUserContext` to be implemented such it delegates to the `AuthenticationProvider` associated with `AuthenticatedUser` to get the `UserContext` instead of requiring this context object in the event constructor. Not needing the context object to construct the event would avoid the exception that is the subject of this PR.

Similarly `CredentialsEvent.getCredentials` could be implemented such that it calls `AuthenticatedUser.getCredentials`, eliminating the need for it in the event constructor as well.

In thinking about per-provider notifications, it also occurs to me that these authentication events should probably include an `AuthenticationProvider` property.  As it stands right now, a recipient of an `AuthenticationFailureEvent` doesn't seem to have an easy way to figure out which provider failed the authentication.

necouchman:
> Including AuthenticatedUser in UserEvent makes a lot of sense. This would allow UserEvent.getUserContext to be implemented such it delegates to the AuthenticationProvider associated with AuthenticatedUser to get the UserContext instead of requiring this context object in the event constructor. Not needing the context object to construct the event would avoid the exception that is the subject of this PR.
> Similarly implementing CredentialsEvent.getCredentials for AuthenticationSuccessEvent could delegate to AuthenticatedUser.getCredentials, eliminating the need for it in the event constructor as well.

So, do the most recent changes look like they address this correctly?

> In thinking about per-provider notifications, it also occurs to me that these authentication events should probably include an AuthenticationProvider property. As it stands right now, a recipient of an AuthenticationFailureEvent doesn't seem to have an easy way to figure out which provider failed the authentication.

I glanced briefly at the failure event, but didn't look real close - I'll try to tweak that one to address this, as well.

guacamole-ext/src/main/java/org/apache/guacamole/net/event/AuthenticationSuccessEvent.java
```diff
@@ -43,19 +44,19 @@
...
      * Creates a new AuthenticationSuccessEvent which represents a successful
      * authentication attempt with the given credentials.
      *
      * @param context The UserContext created as a result of successful
      *                authentication.
-     * @param credentials The credentials which passed authentication.
+     * @param authenticatedUser The user which passed authentication.
      */
-    public AuthenticationSuccessEvent(UserContext context, Credentials credentials) {
+    public AuthenticationSuccessEvent(UserContext context, AuthenticatedUser authenticatedUser) {
```
ceharris: You could drop `UserContext` in the constructor here and drop the associated field. Then you could implement `getUserContext` as:

```
public UserContext getUserContext() {
  return authenticatedUser.getAuthenticationProvider().getUserContext(authenticatedUser);
}
```

This will avoid the need to handle the exception for the missing userContext when the event object is created.

necouchman: Works for me.  Done.

necouchman:
> I glanced briefly at the failure event, but didn't look real close - I'll try to tweak that one to address this, as well.

Turns out this will take a little bit more than minor tweaks to unwind - getting the authentication provider object with what's currently there takes a little doing.  I think it could be retrieved from the UserContext object, but this requires retrieving the UserContext object (which may or may not exist), so it throws the whole ResourceNotFoundException back into the mix, this time on the failure side.  I'll take a little bit of a closer look and see what I can figure out.

guacamole-ext/src/main/java/org/apache/guacamole/net/event/AuthenticationSuccessEvent.java
mike-jumper: > `authenticatedUser.getAuthenticationProvider().getUserContext(authenticatedUser)`

I'm worried about that approach.

The `UserContext` is effectively the data side of a user's session, and typically is only created once, with that instance existing until the user logs out. If an extension implementation needed to persist some sort of state throughout the user's session, the lifecycle of that state would be tied to the `UserContext`. There's also the new `invalidate()` function to consider, which will need to be properly invoked for that short-lived `UserContext`.

This could have performance consequences down the line.

necouchman: Is there an expectation that this getUserContext() method in the event handler will get called repeatedly?

I can revert back to the original method of storing it there - but, if we do that, we have to handle the exception that happens when it is null.  I'm happy to put back my code for just logging the exception and not passing it through in this instance, if that's okay, or I'm happy to investigate other routes if those are preferred?

ceharris: Since `getUserContext` looks like an accessor method, it was an (obviously incorrect) assumption that a provider would create it once and return the same reference for a given authenticated user thereafter

necouchman: Okay, reverted to having the objects in here, but took a slightly different angle on it by having the `AuthenticationSuccessEvent` constructor handle pulling all of the objects out of the `AuthenticatedUser` object, and handled catching the `GuacamoleResourceNotFoundException` exception there, instead of in `AuthenticationService`.  Let me know if this looks okay, or if it should go back to the way I originally had it, catching the error in `AuthenticationService`, instead.  Or if there's some other way that it needs to be done?

mike-jumper:
Per [recent discussion on the mailing lists](https://lists.apache.org/thread.html/d75b09f4636789bd38ed2eff40ed27dd36faa7fbf200c701d202a4b3@%3Cdev.guacamole.apache.org%3E), the release process for 0.9.14-incubating has begun, and the issue associated with this change ([GUACAMOLE-412](https://issues.apache.org/jira/browse/GUACAMOLE-412)) is within scope. Please update the merge base of this PR to point to `staging/0.9.14-incubating`.

necouchman:
> I glanced briefly at the failure event, but didn't look real close - I'll try to tweak that one to address this, as well.

FWIW, I'm not seeing a great way to do this right now.  Maybe I'm missing something really simple, but it seems like it would take a lot of changes in a lot of places to get this info into the failure event.  I move that we table that for another time/issue, and move forward with whatever changes need to be made here to deal with this regression.  Objections?

mike-jumper:
Well, handling the regression is definitely the priority, but creating a new instance of the `UserContext` for each event isn't something we should do. There isn't a way around this which will not require API changes, either now or in a subsequent release, such that the proper data is made available at the proper time.

If we want the event API to be part of 0.9.14-incubating, I would say the only thing that makes sense is to remove `UserContext` from the auth success event entirely. I'd be glad to take a stab at the refactor for this.

If there is concern over the new API in general, then the choice would be to roll back those changes, and let them be reconsidered for a later release.

necouchman:
>  I'd be glad to take a stab at the refactor for this.

Sounds good - I won't be offended :-).

mike-jumper:
Heh. Very well.

necouchman:
> Heh. Very well.

Sounds so enthusiastic.  I'm not opposed to looking at it myself, but I might take longer than we want to wait for 0.9.14-incubating, and I'm pretty sure this fixed in there one way or the other.

> If there is concern over the new API in general, then the choice would be to roll back those changes, and let them be reconsidered for a later release.

I think the new API is fine, and this is the only error I've run into with those changes, but obviously impacts any of the extensions that don't provide `UserContext` objects, which is no longer an insignificant number.  So, I lean more toward sticking with the new API and taking care of this regression than rolling everything back - unless you'd like me to keep hacking it and/or think it's just going to cause significant delays to getting the 0.9.14-incubating release out??

mike-jumper:
> Sounds so enthusiastic.

Don't read too much into my periods. ;)

I'm about to open a PR with my changes.

necouchman:
Closing in favor of Mike's changes.

PR #198, opened 2017-10-13
GUACAMOLE-415: Add support for SQLite Authentication Module

necouchman:
This pull request adds support for the SQLite JDBC driver to the JDBC authentication module.  There are a couple of things worth noting:
- Support for SQLite in JdbcHelper did not show up until MyBatis Guice 3.9.  I've gone ahead and bumped that version up, and tested it with both SQLite and PostgreSQL, without an adverse behavior, but that's worth discussing.  I think I can probably get the driver going without the JdbcHelper by manually setting JDBC.driver and JDBC.url properties, if we don't want to rev the MyBatis Guice version - just let me know.
- There's a bug [1] in the SQLite JDBC driver that impacts parameters being inserted into prepared statements, and, as a result, I had to break the select statements for this driver into individual blocks, which resulted in some badly duplicated XML code.  It works, it's just not ideal, from a maintainability or performance perspective.

[1] [https://github.com/xerial/sqlite-jdbc/issues/277](url)


extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/pom.xml
```diff
@@ -116,7 +116,7 @@
         <dependency>
             <groupId>org.mybatis</groupId>
             <artifactId>mybatis-guice</artifactId>
-            <version>3.6</version>
+            <version>3.9</version>
```
necouchman: Needs discussion...

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlite/schema/002-create-admin-user.sql
```diff
@@ -0,0 +1,49 @@
...
+-- specific language governing permissions and limitations
+-- under the License.
+--
+
+-- Create default user "guacadmin" with password "guacadmin"
+INSERT INTO guacamole_user (username, password_hash, password_salt, password_date)
+VALUES ('guacadmin',
+    x'CA458A7D494E3BE824F5E1E175A1556C0F8EEF2C2D7DF3633BEC4A29C4411960',  -- 'guacadmin'
+    x'FE24ADC5E11E2B25288D1704ABE67A79E342ECC26064CE69C5B3177795A82264',
+    datetime('now','localtime'));
```
necouchman: This actually doesn't work, needs to be replaced with strftime().

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlite/src/main/java/org/apache/guacamole/auth/sqlite/SQLiteGuacamoleProperties.java
```diff
@@ -0,0 +1,127 @@
...
+     * The maximum number of concurrent connections to allow overall. Zero
+     * denotes unlimited.
+     */
+    public static final IntegerGuacamoleProperty
+            SQLITE_ABSOLUTE_MAX_CONNECTIONS =
+            new IntegerGuacamoleProperty() {
+
+        @Override
+        public String getName() { return "sqlite-absolute-max-connections"; }
+
```
necouchman: Does these actually apply to SQLite??

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlite/src/main/resources/org/apache/guacamole/auth/jdbc/connection/ConnectionMapper.xml
```diff
@@ -0,0 +1,253 @@
...
+        <result column="connection_weight"        property="connectionWeight"      jdbcType="INTEGER"/>
+        <result column="failover_only"            property="failoverOnly"          jdbcType="BOOLEAN"/>
+
+        <!-- Associated sharing profiles -->
+        <collection property="sharingProfileIdentifiers" resultSet="sharingProfiles" ofType="java.lang.String"
+                    column="connection_id" foreignColumn="primary_connection_id" select="selectSharingProfiles"/>
+
+    </resultMap>
+
+    <!-- Result mapper for connection objects with read permission check -->
```
necouchman: This is the result of the SQLite JDBC bug.  Yeah, it's bad.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlite/src/main/resources/org/apache/guacamole/auth/jdbc/connectiongroup/ConnectionGroupMapper.xml
```diff
@@ -0,0 +1,250 @@
...
+        <collection property="connectionGroupIdentifiers" resultSet="childConnectionGroups" ofType="java.lang.String"
+                    column="connection_group_id" foreignColumn="parent_id" select="selectChildGroups"/>
+
+        <!-- Child connections -->
+        <collection property="connectionIdentifiers" resultSet="childConnections" ofType="java.lang.String"
+                    column="connection_group_id" foreignColumn="parent_id" select="selectChildConnections"/>
+
+    </resultMap>
+
+    <!-- Result mapper for connection objects with read permission check -->
```
necouchman: Ick.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlite/src/main/java/org/apache/guacamole/auth/sqlite/SQLiteGuacamoleProperties.java
mike-jumper: Which?

necouchman: All of the connection-limiting properties.  Just not sure those actually apply to something file-based like SQLite?

mike-jumper: They're applicable. The connection-limiting properties deal with the default behavior of Guacamole connections, not database connections.

The hostname and port properties definitely wouldn't apply, though.

necouchman: Sounds good.  I think I got the rest of them removed, just wasn't sure about those.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/pom.xml
necouchman: I can probably do this without updating mybatis-guice, without the JdbcHelper support for SQLite, if this change needs to be a separate issue.  I was able to bump up both mybatis and mybatis-guice versions to the latest without seeing any adverse effects, but it probably warrants further discussion and testing.

necouchman:
> There's a bug [1] in the SQLite JDBC driver...

Unfortunately this bug hasn't received any attention from the community/maintainer(s) of the SQLite JDBC driver.  So, I may end up having to track that down a little more and submit a PR and see if it gets accepted, or we may have to cope with the ugly XML code for the time being.

PR #199, opened 2017-10-13
GUACAMOLE-416: Add translation entries for SQL Server module.

necouchman:
Minor tweak to add missing entries for the SQL Server module to the base jdbc en.json file.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/resources/translations/en.json
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/resources/translations/en.json
@@ -63,6 +63,14 @@
         "NAME" : "Shared Connections (PostgreSQL)"
     },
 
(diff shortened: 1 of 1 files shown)


necouchman:
Ping @mike-jumper and/or @ceharris 

Simple review and merge, here.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/resources/translations/en.json
```diff
@@ -63,6 +63,14 @@
         "NAME" : "Shared Connections (PostgreSQL)"
     },
 
+    "DATA_SOURCE_SQLSERVER" : {
+        "NAME" : "SQL Server"
+    },
+
+    "DATA_SOURCe_SQLSERVER_SHARED" : {
```
mike-jumper: I'm clearly late to the party, but that "e" should be uppercase. It will not match otherwise.

PR #202, opened 2017-10-23
GUACAMOLE-243: Allow LDAP referral following to be configured

necouchman:
This PR fixes the JIRA issue 243, where errors in following referrals block LDAP authentication from succeeding, by allowing referral support to be configured.

This closes #129.

### DIFF
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
@@ -257,7 +257,7 @@ public class ConfigurationService {
      *
      * @return
      *     The boolean value of whether to follow referrals
-     *     as configured in guacamole.properties
+     *     as configured in guacamole.properties.
      *
      * @throws GuacamoleException
      *     If guacamole.properties cannot be parsed.
@@ -295,7 +295,7 @@ public class ConfigurationService {
      *
      * @return
      *     The maximum number of referral hops to follow
-     *     as configured in guacamole.properties
+     *     as configured in guacamole.properties.
      *
      * @throws GuacamoleException
      *     If guacamole.properties cannot be parsed.
@@ -328,11 +328,11 @@ public class ConfigurationService {
     }
 
     /**
-     * Returns the maximum number of seconds to wait for LDAP operations
+     * Returns the maximum number of seconds to wait for LDAP operations.
      *
      * @return
      *     The maximum number of seconds to wait for LDAP operations
-     *     as configured in guacamole.properties
+     *     as configured in guacamole.properties.
      *
      * @throws GuacamoleException
      *     If guacamole.properties cannot be parsed.
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
@@ -316,6 +316,7 @@ public class ConfigurationService {
      *     The search filter that should be used when querying the
      *     LDAP server for users that are valid in Guacamole, or
      *     "(objectClass=*)" if not specified.
+     *
      * @throws GuacamoleException
      *     If guacamole.properties cannot be parsed.
      */
@@ -324,37 +325,6 @@ public class ConfigurationService {
             LDAPGuacamoleProperties.LDAP_USER_SEARCH_FILTER,
             "(objectClass=*)"
         );
-    }
-
-    /**
-     * Returns the authentication method to use during referral following.
-     *
-     * @return
-     *     The authentication method to use during referral following
-     *     as configured in guacamole.properties or as derived from
-     *     other configuration options.
-     *
-     * @throws GuacamoleException
-     *     If guacamole.properties cannot be parsed.
-     */
-    public String getReferralAuthentication() throws GuacamoleException {
-        String confMethod = environment.getProperty(
-            LDAPGuacamoleProperties.LDAP_REFERRAL_AUTHENTICATION
-        );
-
-        if (confMethod == null)
-
-            if (getSearchBindDN() != null && getSearchBindPassword() != null)
-                return "bind";
-
-            else
-                return "anonymous";
-
-        else if (confMethod.equals("bind") && (getSearchBindDN() == null || getSearchBindPassword() == null))
-            throw new GuacamoleException("Referral is set to bind with credentials, but credentials are not configured.");
-
-        return confMethod;
-
     }
 
     /**
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
@@ -252,6 +252,24 @@ public class ConfigurationService {
     }
 
     /**
+     * Returns the boolean value for whether the connection should
+     * follow referrals or not.  By default, it will not.
+     *
+     * @return
+     *     The boolean value of whether to follow referrals
+     *     as configured in guacamole.properties
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed.
+     */
+    public boolean getFollowReferrals() throws GuacamoleException {
+        return environment.getProperty(
+            LDAPGuacamoleProperties.LDAP_FOLLOW_REFERRALS,
+            false
+        );
+    }
+
+    /**
      * Returns a set of LDAPSearchConstraints to apply globally
      * to all LDAP searches.
      *
@@ -273,6 +291,23 @@ public class ConfigurationService {
     }
 
     /**
+     * Returns the maximum number of referral hops to follow.
+     *
+     * @return
+     *     The maximum number of referral hops to follow
+     *     as configured in guacamole.properties
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed.
+     */
+    public int getMaxReferralHops() throws GuacamoleException {
+        return environment.getProperty(
+            LDAPGuacamoleProperties.LDAP_MAX_REFERRAL_HOPS,
+            5
+        );
+    }
+
+    /**
      * Returns the search filter that should be used when querying the
      * LDAP server for Guacamole users.  If no filter is specified,
      * a default of "(objectClass=*)" is returned.
@@ -281,7 +316,6 @@ public class ConfigurationService {
      *     The search filter that should be used when querying the
      *     LDAP server for users that are valid in Guacamole, or
      *     "(objectClass=*)" if not specified.
-     *
      * @throws GuacamoleException
      *     If guacamole.properties cannot be parsed.
      */
@@ -289,6 +323,54 @@ public class ConfigurationService {
         return environment.getProperty(
             LDAPGuacamoleProperties.LDAP_USER_SEARCH_FILTER,
             "(objectClass=*)"
+        );
+    }
+
+    /**
+     * Returns the authentication method to use during referral following.
+     *
+     * @return
+     *     The authentication method to use during referral following
+     *     as configured in guacamole.properties or as derived from
+     *     other configuration options.
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed.
+     */
+    public String getReferralAuthentication() throws GuacamoleException {
+        String confMethod = environment.getProperty(
+            LDAPGuacamoleProperties.LDAP_REFERRAL_AUTHENTICATION
+        );
+
+        if (confMethod == null)
+
+            if (getSearchBindDN() != null && getSearchBindPassword() != null)
+                return "bind";
+
+            else
+                return "anonymous";
+
+        else if (confMethod.equals("bind") && (getSearchBindDN() == null || getSearchBindPassword() == null))
+            throw new GuacamoleException("Referral is set to bind with credentials, but credentials are not configured.");
+
+        return confMethod;
+
+    }
+
+    /**
+     * Returns the maximum number of seconds to wait for LDAP operations
+     *
+     * @return
+     *     The maximum number of seconds to wait for LDAP operations
+     *     as configured in guacamole.properties
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed.
+     */
+    public int getOperationTimeout() throws GuacamoleException {
+        return environment.getProperty(
+            LDAPGuacamoleProperties.LDAP_OPERATION_TIMEOUT,
+            30
         );
     }
 
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPConnectionService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPConnectionService.java
@@ -123,8 +123,7 @@ public class LDAPConnectionService {
 
         // If the referral auth method is set to bind, we set it using the existing
         // username and password.
-        String refAuthMethod = confService.getReferralAuthentication();
-        if (refAuthMethod != null && refAuthMethod.equals("bind"))
+        if (userDN != null && !userDN.isEmpty())
             ldapConstraints.setReferralHandler(new ReferralAuthHandler(userDN, password));
 
         // Set the maximum number of referrals we follow
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPConnectionService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPConnectionService.java
@@ -21,12 +21,14 @@ package org.apache.guacamole.auth.ldap;
 
 import com.google.inject.Inject;
 import com.novell.ldap.LDAPConnection;
+import com.novell.ldap.LDAPConstraints;
 import com.novell.ldap.LDAPException;
 import com.novell.ldap.LDAPJSSESecureSocketFactory;
 import com.novell.ldap.LDAPJSSEStartTLSFactory;
 import java.io.UnsupportedEncodingException;
 import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.GuacamoleUnsupportedException;
+import org.apache.guacamole.auth.ldap.ReferralAuthHandler;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -110,6 +112,26 @@ public class LDAPConnectionService {
 
         // Obtain appropriately-configured LDAPConnection instance
         LDAPConnection ldapConnection = createLDAPConnection();
+
+        // Configure LDAP connection constraints
+        LDAPConstraints ldapConstraints = ldapConnection.getConstraints();
+        if (ldapConstraints == null)
+          ldapConstraints = new LDAPConstraints();
+
+        // Set whether or not we follow referrals, and max hops
+        ldapConstraints.setReferralFollowing(confService.getFollowReferrals());
+        String refAuthMethod = confService.getReferralAuthentication();
+
+        if (refAuthMethod != null && refAuthMethod.equals("bind"))
+            ldapConstraints.setReferralHandler(new ReferralAuthHandler(userDN, password));
+
+        ldapConstraints.setHopLimit(confService.getMaxReferralHops());
+
+        // Set timelimit to wait for LDAP operations, converting to ms
+        ldapConstraints.setTimeLimit(confService.getOperationTimeout() * 1000);
+
+        // Apply the constraints to the connection
+        ldapConnection.setConstraints(ldapConstraints);
 
         try {
 
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPConnectionService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPConnectionService.java
@@ -118,13 +118,16 @@ public class LDAPConnectionService {
         if (ldapConstraints == null)
           ldapConstraints = new LDAPConstraints();
 
-        // Set whether or not we follow referrals, and max hops
+        // Set whether or not we follow referrals
         ldapConstraints.setReferralFollowing(confService.getFollowReferrals());
-        String refAuthMethod = confService.getReferralAuthentication();
 
+        // If the referral auth method is set to bind, we set it using the existing
+        // username and password.
+        String refAuthMethod = confService.getReferralAuthentication();
         if (refAuthMethod != null && refAuthMethod.equals("bind"))
             ldapConstraints.setReferralHandler(new ReferralAuthHandler(userDN, password));
 
+        // Set the maximum number of referrals we follow
         ldapConstraints.setHopLimit(confService.getMaxReferralHops());
 
         // Set timelimit to wait for LDAP operations, converting to ms
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPConnectionService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPConnectionService.java
@@ -121,8 +121,7 @@ public class LDAPConnectionService {
         // Set whether or not we follow referrals
         ldapConstraints.setReferralFollowing(confService.getFollowReferrals());
 
-        // If the referral auth method is set to bind, we set it using the existing
-        // username and password.
+        // Set referral authentication to use the provided credentials.
         if (userDN != null && !userDN.isEmpty())
             ldapConstraints.setReferralHandler(new ReferralAuthHandler(userDN, password));
 
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
@@ -19,6 +19,7 @@
 
 package org.apache.guacamole.auth.ldap;
 
+import org.apache.guacamole.properties.BooleanGuacamoleProperty;
 import org.apache.guacamole.properties.IntegerGuacamoleProperty;
 import org.apache.guacamole.properties.StringGuacamoleProperty;
 
@@ -171,6 +172,46 @@ public class LDAPGuacamoleProperties {
 
         @Override
         public String getName() { return "ldap-user-search-filter"; }
+
+    };
+
+    /**
+     * Whether or not we should follow referrals
+     */
+    public static final BooleanGuacamoleProperty LDAP_FOLLOW_REFERRALS = new BooleanGuacamoleProperty() {
+
+        @Override
+        public String getName() { return "ldap-follow-referrals"; }
+
+    };
+
+    /**
+     * Maximum number of referral hops to follow
+     */
+    public static final IntegerGuacamoleProperty LDAP_MAX_REFERRAL_HOPS = new IntegerGuacamoleProperty() {
+
+        @Override
+        public String getName() { return "ldap-max-referral-hops"; }
+
+    };
+
+    /**
+     * Authentication method to use to follow referrals
+     */
+    public static final StringGuacamoleProperty LDAP_REFERRAL_AUTHENTICATION = new StringGuacamoleProperty() {
+
+        @Override
+        public String getName() { return "ldap-referral-authentication"; }
+
+    };
+
+    /**
+     * Number of seconds to wait for LDAP operations to complete
+     */
+    public static final IntegerGuacamoleProperty LDAP_OPERATION_TIMEOUT = new IntegerGuacamoleProperty() {
+
+        @Override
+        public String getName() { return "ldap-operation-timeout"; }
 
     };
 
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
@@ -176,7 +176,7 @@ public class LDAPGuacamoleProperties {
     };
 
     /**
-     * Whether or not we should follow referrals
+     * Whether or not we should follow referrals.
      */
     public static final BooleanGuacamoleProperty LDAP_FOLLOW_REFERRALS = new BooleanGuacamoleProperty() {
 
@@ -186,7 +186,7 @@ public class LDAPGuacamoleProperties {
     };
 
     /**
-     * Maximum number of referral hops to follow
+     * Maximum number of referral hops to follow.
      */
     public static final IntegerGuacamoleProperty LDAP_MAX_REFERRAL_HOPS = new IntegerGuacamoleProperty() {
 
@@ -196,7 +196,7 @@ public class LDAPGuacamoleProperties {
     };
 
     /**
-     * Number of seconds to wait for LDAP operations to complete
+     * Number of seconds to wait for LDAP operations to complete.
      */
     public static final IntegerGuacamoleProperty LDAP_OPERATION_TIMEOUT = new IntegerGuacamoleProperty() {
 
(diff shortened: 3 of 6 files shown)


mike-jumper:
Per [recent discussion on the mailing lists](https://lists.apache.org/thread.html/d75b09f4636789bd38ed2eff40ed27dd36faa7fbf200c701d202a4b3@%3Cdev.guacamole.apache.org%3E), the release process for 0.9.14-incubating has begun, and the issue associated with this change ([GUACAMOLE-243](https://issues.apache.org/jira/browse/GUACAMOLE-243)) is within scope. Please update the merge base of this PR to point to `staging/0.9.14-incubating`.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPConnectionService.java
```diff
@@ -111,6 +113,29 @@ public LDAPConnection bindAs(String userDN, String password)
...
+          ldapConstraints = new LDAPConstraints();
+
+        // Set whether or not we follow referrals
+        ldapConstraints.setReferralFollowing(confService.getFollowReferrals());
+
+        // If the referral auth method is set to bind, we set it using the existing
+        // username and password.
+        String refAuthMethod = confService.getReferralAuthentication();
+        if (refAuthMethod != null && refAuthMethod.equals("bind"))
+            ldapConstraints.setReferralHandler(new ReferralAuthHandler(userDN, password));
```
mike-jumper: So, what's the idea behind configuring whether referrals auth with the same credentials?

necouchman: I believe it's the difference between using the credentials that the user logged in with vs. using the credentials specified in the guacamole.properties file, if any.  So, in some cases it'll be the difference between an anonymous bind and an authenticated bind, in other cases it'll be the difference between binding with the configured search user vs. the logged in user.

There are cases where the user who logged in may not have access to the entire tree, or even the base of the tree, so you'd want the search & referral following to happen under that user account instead of the logged in account.

At least, that's the theory, I think...whether or not I've actually implemented it that way, I probably should go back and take another look and make sure.

necouchman: Okay, I configured it to just use the search credentials.

The one situation that I thought about here that you might want the ability to configure separate referral credentials is where you know you have a situation where your base LDAP tree points to another system, and the credentials are different on the referred to system than they are on the base system.  This is somewhat of a corner case, and I think it unlikely there are just huge numbers of people running that, but it's worth keeping in mind for future discussions.  In that case, however, just having this parameter - ldap-referral-auth - wouldn't be enough - you'd also want to ability to configure a separate username and password for the referrals, so you'd end up adding a couple more parameters.

mike-jumper: Well, to clarify a bit, there are two types of searches which may occur in the LDAP auth:

* The search performed to resolve the user's DN. This is the only case where the search DN and password are used.
* The search to retrieve the users and/or connections defined within LDAP and accessible to the user that authenticated. These searches naturally **must** use the user's own DN and password, since part of the point of the LDAP auth is to allow the LDAP directory's own access control to take effect.

If users or connections cannot be retrieved via LDAP because the authenticated user cannot access them, then they shouldn't have access, referral or not.

necouchman: Yeah, I see what you're saying.  The changes here should be consistent with that - the referral authentication just uses what is passed in to the bindAs method for userDN and password, which, for the initial user search will be the search credentials, and subsequently will be the retrieved userDN and the password used for authentication.

mike-jumper: Ah, OK then.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPConnectionService.java
```diff
@@ -111,6 +113,28 @@ public LDAPConnection bindAs(String userDN, String password)
...
 
+        // Configure LDAP connection constraints
+        LDAPConstraints ldapConstraints = ldapConnection.getConstraints();
+        if (ldapConstraints == null)
+          ldapConstraints = new LDAPConstraints();
+
+        // Set whether or not we follow referrals
+        ldapConstraints.setReferralFollowing(confService.getFollowReferrals());
+
+        // If the referral auth method is set to bind, we set it using the existing
```
mike-jumper: With the removal of the auth method property, this comment is no longer correct.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ReferralAuthHandler.java
```diff
@@ -0,0 +1,97 @@
...
+import com.google.inject.Inject;
+import com.novell.ldap.LDAPAuthHandler;
+import com.novell.ldap.LDAPAuthProvider;
+import com.novell.ldap.LDAPConnection;
+import java.io.UnsupportedEncodingException;
+import org.apache.guacamole.GuacamoleException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ReferralAuthHandler implements LDAPAuthHandler {
```
mike-jumper: JavaDoc missing for this class.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ReferralAuthHandler.java
```diff
@@ -0,0 +1,97 @@
...
+            logger.error("Unexpected lack of support for UTF-8: {}", e.getMessage());
+            logger.debug("Support for UTF-8 (as required by Java spec) not found.", e);
+            throw new GuacamoleException("Could not set password due to missing support for UTF-8 encoding.");
+        }
+
+        ldapAuth = new LDAPAuthProvider(binddn, passwordBytes);
+
+    }
+
+    public ReferralAuthHandler(String dn, String password) throws GuacamoleException {
```
mike-jumper: Missing JavaDoc.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ReferralAuthHandler.java
```diff
@@ -0,0 +1,97 @@
...
+     * Service for retrieving LDAP server configuration information.
+     */
+    @Inject
+    private ConfigurationService confService;
+
+
+    public ReferralAuthHandler() throws GuacamoleException {
+        String binddn = confService.getSearchBindDN();
+        String password = confService.getSearchBindPassword();
+        byte[] passwordBytes;
```
mike-jumper: Rather than copy the entire body of the other constructor, you can invoke it here. You would do:

    this(confService.getSearchBindDN(), confService.getSearchBindPassword());

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ReferralAuthHandler.java
```diff
@@ -0,0 +1,97 @@
...
+    private final LDAPAuthProvider ldapAuth;
+
+    /**
+     * Service for retrieving LDAP server configuration information.
+     */
+    @Inject
+    private ConfigurationService confService;
+
+
+    public ReferralAuthHandler() throws GuacamoleException {
```
mike-jumper: I don't see this constructor used anywhere. Perhaps it's no longer needed?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ReferralAuthHandler.java
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * The LDAPAuthProvider object that will be set and returned to the referral handler.
+     */
+    private final LDAPAuthProvider ldapAuth;
+
+    /**
+     * Service for retrieving LDAP server configuration information.
+     */
+    @Inject
+    private ConfigurationService confService;
```
mike-jumper: Beware that, despite the `@Inject` annotation, this will not happen automatically upon construction. Guice will only perform the injection via a `Provider`, or if the various constructors are properly annotated for assisted injection, which is complex.

If the default constructor is the only place this is used, and that constructor is not called anyway, I'd suggest simply getting rid of both.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -251,8 +270,23 @@ private String getConnectionSearchFilter(String userDN,
...
-                LDAPEntry entry = userRoleGroupResults.next();
-                connectionSearchFilter.append("(seeAlso=").append(escapingService.escapeLDAPSearchFilter(entry.getDN())).append(")");
+                try {
+                    LDAPEntry entry = userRoleGroupResults.next();
+                    connectionSearchFilter.append("(seeAlso=").append(escapingService.escapeLDAPSearchFilter(entry.getDN())).append(")");
+                }
+
+                catch (LDAPReferralException e) {
+                    if (confService.getFollowReferrals()) {
+                        logger.error("Could not follow referral.", e.getMessage());
```
mike-jumper: If you want this message to be logged, you'll need to include `{}` within the string to define the location for the value to be inserted. For example:

    logger.error("Could not follow referral: {}", e.getMessage());

Same thing for the other new logging statements within these changes.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -251,8 +270,23 @@ private String getConnectionSearchFilter(String userDN,
...
+                        logger.debug("Error encountered trying to follow referral.", e);
+                        throw new GuacamoleServerException("Could not follow LDAP referral.", e);
+                    }
+                    else {
+                        logger.warn("Given a referral, but referrals are disabled.", e.getMessage());
+                        logger.debug("Got a referral, but configured to not follow them.", e);
+                        continue;
```
mike-jumper: Being the last statement in the loop, this `continue` has no effect.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
```diff
@@ -107,19 +108,37 @@ private void putAllUsers(Map<String, User> users, LDAPConnection ldapConnection,
...
+                catch (LDAPReferralException e) {
+                    if (confService.getFollowReferrals()) {
+                        logger.error("Could not follow referral.", e.getFailedReferral());
+                        logger.debug("Error encountered trying to follow referral.", e);
+                        throw new GuacamoleServerException("Could not follow LDAP referral.", e);
+                    }
+                    else {
+                        logger.warn("Given a referral, but referrals are disabled.", e.getMessage());
+                        logger.debug("Got a referral, but configured to not follow them.", e);
+                        continue;
```
mike-jumper: Being the last statement in the loop, this `continue` has no effect.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -251,8 +270,23 @@ private String getConnectionSearchFilter(String userDN,
...
+                try {
+                    LDAPEntry entry = userRoleGroupResults.next();
+                    connectionSearchFilter.append("(seeAlso=").append(escapingService.escapeLDAPSearchFilter(entry.getDN())).append(")");
+                }
+
+                catch (LDAPReferralException e) {
+                    if (confService.getFollowReferrals()) {
+                        logger.error("Could not follow referral.", e.getMessage());
+                        logger.debug("Error encountered trying to follow referral.", e);
+                        throw new GuacamoleServerException("Could not follow LDAP referral.", e);
```
mike-jumper: Recalling the changes proposed via #129, is it desirable that the entire search will abort with a failure if a referral error occurs? Or should the error be non-fatal and simply logged?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPConnectionService.java
necouchman: Corrected.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ReferralAuthHandler.java
necouchman: Added.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ReferralAuthHandler.java
necouchman: Added.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ReferralAuthHandler.java
necouchman: Removed this constructor.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ReferralAuthHandler.java
necouchman: Removed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ReferralAuthHandler.java
necouchman: Removed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
necouchman: Corrected.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
necouchman: Removed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
necouchman: Removed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
necouchman: My thinking is that if you explicitly enable referral following, an error following the referrals should cause an abort, and if you disable it, you can log it and ignore it.  I'm not opposed to a non-fatal logging, here, though, if that makes more sense.  I don't know off the top of my head what the generally accepted behavior is for other implementations.

In the case of the issue with Active Directory, disabling referral following completely allows you to query it via LDAP correctly - having it disabled doesn't cut off expected results or anything like that.

necouchman: @mike-jumper Any further thoughts on this, or opinions on which way it should go?  I'm not sure one way or the other - I do tend toward not wanting to break things if a single operation fails; however, I also think that if you're explicitly enabling it and it fails, you want a very clear indication that something is not working.  I see both sides, so I'm good going either way - it's easy enough to change down the road, and will get tweaked, anyway, when we switch LDAP APIs.

necouchman:
@mike-jumper Aside from the one discussion point on logging vs. failing, any other issues with this?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ReferralAuthHandler.java
```diff
@@ -0,0 +1,80 @@
...
+            if (password != null)
+                passwordBytes = password.getBytes("UTF-8");
+            else
+                passwordBytes = null;
+
+        }   
+        catch (UnsupportedEncodingException e) {
+            logger.error("Unexpected lack of support for UTF-8: {}", e.getMessage());
+            logger.debug("Support for UTF-8 (as required by Java spec) not found.", e); 
+            throw new GuacamoleException("Could not set password due to missing UTF-8 support.");
```
mike-jumper: In general:

1. Rather than throw the base `GuacamoleException`, we should be throwing an appropriate subclass. `GuacamoleServerException` would be a reasonable choice in this case.
2. For any exception caused by another, [the constructor which includes the causing exception](http://guacamole.incubator.apache.org/doc/guacamole-common/org/apache/guacamole/GuacamoleServerException.html#GuacamoleServerException-java.lang.String-java.lang.Throwable-) should be used (rather than `new GuacamoleServerException("message")`, use `new GuacamoleServerException("message", e)`).

That said, since UTF-8 support is guaranteed by Java, there's no need for checked exceptions here. Something like `UnsupportedOperationException` might be more appropriate. For example:

https://github.com/apache/incubator-guacamole-client/blob/0611fe8fff694dff7e3cb6a62918f6c90668728c/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/form/CASTicketField.java#L80-L83

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ReferralAuthHandler.java
```diff
@@ -0,0 +1,80 @@
...
+     */
+    private final LDAPAuthProvider ldapAuth;
+
+    /**
+     * Creates a ReferralAuthHandler object to handle authentication when
+     * following referrals in a LDAP connection, using the provided dn and
+     * password.
+     *
+     * @throws GuacamoleException
+     *     If exceptions are caught while converting the password from a string
```
mike-jumper: "If exceptions are caught" doesn't really provide much information, and "converting into a byte array" is an internal concern. This may be a moot point if switching over to `UnsupportedOperationException` (see below), as the exception need not be declared in that case, but otherwise perhaps there is a way to describe the cases where exceptions will be thrown which does not reference internal implementation details?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -129,62 +130,79 @@
...
-                String name = cn.getStringValue();
-                Connection connection = new SimpleConnection(name, name, config);
-                connection.setParentIdentifier(LDAPAuthenticationProvider.ROOT_CONNECTION_GROUP);
-                connections.put(name, connection);
+                }
+
+                // Deal with issues following LDAP referrals
+                catch (LDAPReferralException e) {
+                    if (confService.getFollowReferrals()) {
+                        logger.error("Could not follow referral.", e.getFailedReferral());
```
mike-jumper: If the intent is to include `e.getFailedReferral()` within the logged message, then the `{}` string must be present in the message at the location the value should be inserted.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -129,62 +130,79 @@
...
+                        logger.debug("Error encountered trying to follow referral.", e);
+                        throw new GuacamoleServerException("Could not follow LDAP referral.", e);
+                    }
+                    else {
+                        logger.warn("Given a referral, but referrals are disabled.", e.getMessage());
```
mike-jumper: As above, if the intent is to include `e.getMessage()` within the logged message, then the `{}` string must be present in the message at the location the value should be inserted.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -251,8 +269,22 @@ private String getConnectionSearchFilter(String userDN,
...
+                }
+
+                catch (LDAPReferralException e) {
+                    if (confService.getFollowReferrals()) {
+                        logger.error("Could not follow referral: {}", e.getFailedReferral());
+                        logger.debug("Error encountered trying to follow referral.", e);
+                        throw new GuacamoleServerException("Could not follow LDAP referral.", e);
+                    }
+                    else {
+                        logger.warn("Given a referral, but referrals are disabled.", e.getMessage());
```
mike-jumper: Missing `{}` in log message.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
```diff
@@ -107,19 +108,36 @@ private void putAllUsers(Map<String, User> users, LDAPConnection ldapConnection,
...
-                    logger.warn("Possibly ambiguous user account: \"{}\".", identifier);
+                // Deal with errors trying to follow referrals
+                catch (LDAPReferralException e) {
+                    if (confService.getFollowReferrals()) {
+                        logger.error("Could not follow referral: {}", e.getFailedReferral());
+                        logger.debug("Error encountered trying to follow referral.", e);
+                        throw new GuacamoleServerException("Could not follow LDAP referral.", e);
+                    }
+                    else {
+                        logger.warn("Given a referral, but referrals are disabled.", e.getMessage());
```
mike-jumper: Missing `{}` in log message.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
```diff
@@ -267,8 +285,23 @@ private String generateLDAPQuery(String username)
...
-                userDNs.add(entry.getDN());
+                try {
+                    LDAPEntry entry = results.next();
+                    userDNs.add(entry.getDN());
+                }
+          
+                // Deal with errors following referrals
+                catch (LDAPReferralException e) {
+                    if (confService.getFollowReferrals()) {
+                        logger.error("Error trying to follow a referral.", e.getMessage());
```
mike-jumper: Missing `{}` in log message.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
```diff
@@ -267,8 +285,23 @@ private String generateLDAPQuery(String username)
...
+                        logger.debug("Encountered an error trying to follow a referral.", e);
+                        throw new GuacamoleServerException("Failed while trying to follow referrals.", e);
+                    }
+                    else {
+                        logger.warn("Given a referral, not following it.", e.getMessage());
```
mike-jumper: Missing `{}` in log message.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
```diff
@@ -252,6 +252,24 @@ private DereferenceAliasesMode getDereferenceAliases() throws GuacamoleException
     }
 
     /**
+     * Returns the boolean value for whether the connection should
+     * follow referrals or not.  By default, it will not.
+     *
+     * @return
+     *     The boolean value of whether to follow referrals
+     *     as configured in guacamole.properties
```
mike-jumper: JavaDoc formatting: missing period.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
```diff
@@ -273,6 +291,23 @@ public LDAPSearchConstraints getLDAPSearchConstraints() throws GuacamoleExceptio
     }
 
     /**
+     * Returns the maximum number of referral hops to follow.
+     *
+     * @return
+     *     The maximum number of referral hops to follow
+     *     as configured in guacamole.properties
```
mike-jumper: JavaDoc formatting: missing period.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
```diff
@@ -292,4 +327,21 @@ public String getUserSearchFilter() throws GuacamoleException {
         );
     }
 
+    /**
+     * Returns the maximum number of seconds to wait for LDAP operations
```
mike-jumper: JavaDoc formatting: missing period.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
```diff
@@ -292,4 +327,21 @@ public String getUserSearchFilter() throws GuacamoleException {
...
+     *
+     * @return
+     *     The maximum number of seconds to wait for LDAP operations
+     *     as configured in guacamole.properties
```
mike-jumper: JavaDoc formatting: missing period.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
```diff
@@ -174,4 +175,34 @@ private LDAPGuacamoleProperties() {}
 
     };
 
+    /**
+     * Whether or not we should follow referrals
```
mike-jumper: JavaDoc formatting: missing period.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
```diff
@@ -174,4 +175,34 @@ private LDAPGuacamoleProperties() {}
...
+     */
+    public static final BooleanGuacamoleProperty LDAP_FOLLOW_REFERRALS = new BooleanGuacamoleProperty() {
+
+        @Override
+        public String getName() { return "ldap-follow-referrals"; }
+
+    };
+
+    /**
+     * Maximum number of referral hops to follow
```
mike-jumper: JavaDoc formatting: missing period.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
```diff
@@ -174,4 +175,34 @@ private LDAPGuacamoleProperties() {}
...
+     */
+    public static final IntegerGuacamoleProperty LDAP_MAX_REFERRAL_HOPS = new IntegerGuacamoleProperty() {
+
+        @Override
+        public String getName() { return "ldap-max-referral-hops"; }
+
+    };
+
+    /**
+     * Number of seconds to wait for LDAP operations to complete
```
mike-jumper: JavaDoc formatting: missing period.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
necouchman: Fixed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
necouchman: Fixed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java
necouchman: Fixed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
necouchman: Fixed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
necouchman: Fixed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
necouchman: Fixed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ReferralAuthHandler.java
necouchman: Changed as suggested.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ReferralAuthHandler.java
necouchman: Removed since using `UnsupportedOperationException`.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
necouchman: Fixed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
necouchman: Fixed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
necouchman: Fixed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
necouchman: Fixed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
necouchman: Fixed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
necouchman: Fixed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
necouchman: Fixed.

necouchman:
Okay, updates done as suggested.

PR #204, opened 2017-10-26
GUACAMOLE-234: Migration from JLDAP to Apache LDAP API

necouchman:
This PR migrates the guacamole-auth-ldap extension from the legacy Novell JLDAP API to the newer Apache Directory LDAP API.  Wasn't too bad, but there are some semi-significant changes, here, so the code needs careful review and testing.  I've done some initial testing where I can, but I do not currently have the ability to extend the LDAP schema on the systems I work on, so I cannot try putting connections in LDAP.

Also, this requires that the guacamole-auth-ldap extension be updated to Java 1.7 as a required (vs. 1.6).  Probably not a big deal, and I think we should consider bumping it up across the board, but that's a discussion for the dev list.


necouchman:
Also worth noting, with relation to GUACAMOLE-234, regarding referral following, that this implementation pretty much silently ignores referrals, only processing search results that are actually entries.  So, it isn't subject to the "bug" in GUACAMOLE-234 (AD works fine with this implementation); however, I still consider the ideal implementation one where you can configure that behavior and not just silently ignore referrals.

necouchman:
Well, I may have configured support for LDAP referrals, but I don't currently have a LDAP tree with any referrals in it that I can actually test against, so I'm not certain about that.  Anyway, I've added the code as I believe it needs to be, I'm just not sure if anything else is required (any manual traversal of the referrals) or if it does it automatically with what I've added.

necouchman:
Rebased to current master...

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
```diff
@@ -176,32 +176,22 @@ private LDAPGuacamoleProperties() {}
...
-
-        @Override
-        public String getName() { return "ldap-follow-referrals"; }
-
-    };
-
-    /**
-     * Maximum number of referral hops to follow.
-     */
-    public static final IntegerGuacamoleProperty LDAP_MAX_REFERRAL_HOPS = new IntegerGuacamoleProperty() {
```
mike-jumper: Is this not supported by the Apache Directory LDAP API?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
```diff
@@ -176,32 +176,22 @@ private LDAPGuacamoleProperties() {}
...
+        public String getName() { return "ldap-operation-timeout"; }
 
     };
 
     /**
-     * Number of seconds to wait for LDAP operations to complete.
+     * Whether or not to follow referrals.
      */
-    public static final IntegerGuacamoleProperty LDAP_OPERATION_TIMEOUT = new IntegerGuacamoleProperty() {
+    public static final BooleanGuacamoleProperty LDAP_FOLLOW_REFERRALS = new BooleanGuacamoleProperty() {
```
mike-jumper: I see that `LDAP_FOLLOW_REFERRALS` and `LDAP_OPERATION_TIMEOUT` have switched positions. Any reason for this?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -113,56 +123,65 @@
...
-                    if (parameterAttribute != null) {
+                // Get parameters, if any
+                Attribute parameterAttribute = entry.get("guacConfigParameter");
+                if (parameterAttribute != null) {
 
-                        // For each parameter
-                        Enumeration<?> parameters = parameterAttribute.getStringValues();
-                        while (parameters.hasMoreElements()) {
+                    // For each parameter
+                    Iterator parameters = parameterAttribute.iterator();
```
mike-jumper: Why not `Iterator<String>`? Or does the new API not use generics?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -189,30 +208,24 @@
                     connection.setParentIdentifier(LDAPAuthenticationProvider.ROOT_CONNECTION_GROUP);
                     connections.put(name, connection);
 
-                }
-
-                // Deal with issues following LDAP referrals
-                catch (LDAPReferralException e) {
```
mike-jumper: Does the new LDAP API not throw exceptions for referrals?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -124,10 +126,15 @@ private String getUserBindDN(String username)
                 }
 
                 // Return the single possible DN
-                return userDNs.get(0);
+                return new Dn(userDNs.get(0));
```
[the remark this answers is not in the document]
necouchman: Good idea - let me see if there's any reason not to do this.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
```diff
@@ -176,32 +176,22 @@ private LDAPGuacamoleProperties() {}
     };
 
     /**
-     * Whether or not we should follow referrals.
+     * A time limit on LDAP operations.
```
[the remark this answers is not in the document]
necouchman: Not sure...will revert this.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
necouchman: I think it was a side-effect of rebasing and the LDAP Referral work that got pushed in ahead of that.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
necouchman: Not sure why it got removed and did not get added back.  Added it back and cleaned up some of the other error messages.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
necouchman: Not that I can tell - let me know if you can find something on it, but I'm unable to find any details on how to do it.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
necouchman: Added back.

necouchman:
Okay, actually, this needs quite a bit of re-working after the rebase over the LDAP referral changes, so let me clean it up a bit and I'll reply back when it's ready for review, again.

necouchman:
@mike-jumper Okay, that actually wasn't too bad to clean up the rebase changes, so it should be ready for review, again.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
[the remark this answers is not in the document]
necouchman: Done - pretty simple change and no apparent downsides.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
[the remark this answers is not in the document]
necouchman: Reverted.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
necouchman: Reverted.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
necouchman: Yeah, I think the API does not implement any method that allows for retrieving all of the string values at one time, so instead you have to get the iterator and got through and retrieve each value.  Let me know if you see something that makes you think there's an alternate way to do this.

mike-jumper: I don't see any problem with using an iterator - the issue here is that the type parameter for `Iterator` is not being used, and thus we are later having to typecast to `String`. Checking the API for [`Attribute`](http://directory.apache.org/api/gen-docs/latest/apidocs/org/apache/directory/api/ldap/model/entry/Attribute.html), the `iterator()` function returns an `Iterator<Value<?>>`, with `Value<?>` providing a `getString()`. I'm not sure why typecasting to `String` would be working, but I think we should be using proper Java generics.

mike-jumper: By the way, for anything implementing `Iterable`, you can use a special `for` syntax to iterate through its contents. In this case, with `Iterable<Value<?>>`, you can:

    for (Value<?> value : parameterAttribute) {
        ...
    }

There's generally no need to obtain the iterator itself unless you're going to be altering the contents of the iterated object through invoking `iterator.remove()`.

necouchman: Fixed, x2.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
```diff
@@ -185,16 +185,6 @@ private LDAPGuacamoleProperties() {}
 
     };
 
-    /**
-     * Maximum number of referral hops to follow.
-     */
-    public static final IntegerGuacamoleProperty LDAP_MAX_REFERRAL_HOPS = new IntegerGuacamoleProperty() {
-
-        @Override
-        public String getName() { return "ldap-max-referral-hops"; }
```
mike-jumper: Is this not supported by the Apache Directory API?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
```diff
@@ -328,20 +273,40 @@ public String getUserSearchFilter() throws GuacamoleException {
...
+     * search operations follow referrals in the tree.
+     *
+     * @return
+     *     True if LDAP referrals will be followed; false if not.
+     *     Default is false.
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed.
+     */
+    public Boolean getFollowReferrals() throws GuacamoleException {
```
mike-jumper: The comment documenting this function seems improved (more detailed), but why has it been reordered relative to the other functions here?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
necouchman: See my comment above - as far as I can tell, no, I have not found a place in the Apache Directory API that lets you control number of referral hops.  I've looked a couple more times to make sure and am not seeing it...

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
necouchman: I don't think there's any particular reason for reordering it - I think, as with some of the other stuff you mentioned earlier, the reordering happened when rebasing these changes onto the changes for -243 (following referrals).  Let me see if I can fix this one...

necouchman: Fixed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
mike-jumper: Hm... OK. Well, assuming that this is impossible to replicate with the Apache Directory API, then I have no issue with its removal. Who am I to argue with reality. ;)

It's fine to remove support for a property, but we'll need to follow normal deprecation procedures. Ideally, this would mean allowing the property to continue functioning as normal, but failing that we at least need to warn in the logs when the property is used, notifying that the property is deprecated and no longer has any effect.

For example, after `auth-provider` and `lib-directory` were deprecated:

https://github.com/apache/guacamole-client/blob/027fa1d85e72352e3b9c634ae48ec5c0315b9672/guacamole/src/main/java/org/glyptodon/guacamole/net/basic/extension/ExtensionModule.java#L162-L165

Or with the deprecation of the `mysql-disallow-*` properties:

https://github.com/apache/guacamole-client/blob/162f4db386c3a612e8a1189e92b7f789d51945d8/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/net/sourceforge/guacamole/net/auth/mysql/MySQLEnvironment.java#L123-L127

necouchman: > Hm... OK. Well, assuming that this is impossible to replicate with the Apache Directory API, then I have no issue with its removal. Who am I to argue with reality. ;)

Yeah...it is a shame, but I do think the benefits of moving to the Apache API outweigh keeping this one feature.

I do need to do a little more research and just make sure there's not something I'm missing.  I'm also wondering how hard it would be and if it would be worth it to try to implement some sort of form of this inside the Guacamole LDAP extension.  It seems like maybe in the sections that process the search results there might be a way to keep track of how many times a referral has occurred and drop out if it goes beyond the limit.  But, it may be too complicated to be worth implementing.

necouchman: I wonder if the `getReferralLength()` method could be used to check this?  It's a little hard to tell because the documentation on the Apache API is still in its early stages, and it's hard to tell what `referralLength` means in this context:

http://directory.apache.org/api/gen-docs/latest/apidocs/org/apache/directory/api/ldap/model/message/ReferralImpl.html#getReferralLength--

Thoughts??

mike-jumper: Yeah, the documentation for `getReferralLength()` doesn't really tell us anything. Looking through the underlying implementation, the value returned is manually populated through some external mechanism via `setReferralLength()`:

https://github.com/apache/directory-ldap-api/blob/9f888e56bcc65c6a5efe7a49bf7f59d45c9fc935/ldap/codec/core/src/main/java/org/apache/directory/api/ldap/codec/api/LdapEncoder.java#L345-L353

The above block of code gradually increments a length counter taking bytes into account:

    referralLength += 1 + TLV.getNbBytes( ldapUrlBytes.length ) + ldapUrlBytes.length;

Whatever referral length is supposed to be here, it seems to be more related to the overall byte length of some packet, not the depth of the chain of referrals.

mike-jumper: > It seems like maybe in the sections that process the search results there might be a way to keep track of how many times a referral has occurred and drop out if it goes beyond the limit.

This seems important functionality to have. As LDAP search utilities and the old JLDAP API both provide a means of defining the maximum number of hops (and since unbounded following of pointers is generally dangerous behavior), I think we should be absolutely certain that this is impossible before removing the functionality.

In fact, assuming that the functionality is indeed critical, and assuming that it's impossible to implement with the Apache Directory API, it might instead be best to:

1. Keep the property around, but warn that the underlying LDAP implementation does not currently implement it.
2. Open an issue with the Apache Directory project for implementing such a limit.

The Apache Directory mailing lists may be a good resource if we can't figure out what we need from the docs alone.

necouchman: > Yeah, the documentation for getReferralLength() doesn't really tell us anything. Looking through the underlying implementation, the value returned is manually populated through some external mechanism via setReferralLength():

Yeah, I looked at that after I sent that response and came to the same conclusion.  So, that's not going to work.

> This seems important functionality to have.

Yeah, I'd have to agree.  Seems like you could get yourself into a situation that could end badly without it.

> The Apache Directory mailing lists may be a good resource if we can't figure out what we need from the docs alone.

Already signed up for the api list and posted a message, there, so we'll see if I get a response.  It isn't an overly-active list, so not sure.

I'm a little afraid that, if an obvious solution doesn't come along, I'm going to have to implement the entire result processing functionality in the extension - which might be worth it, and might end up being necessary, anyway, to deal with paged results, but I really like leaving the implementation to the API and not having to do it myself.

necouchman: > The Apache Directory mailing lists may be a good resource if we can't figure out what we need from the docs alone.

Well, this particular issue ends up taking a new direction, since, apparently the Apache LDAP API does not automatically follow referrals, and that has to be manually coded.  So, guess I get to manually code that - and while I'm at it, keep track of/check referral depth.

necouchman:
@mike-jumper This is low on the priority list, especially given the pending 1.0.0 release and stuff that needs to be reviewed and fixed up for that, but I've cleaned this up some and taken a stab at implementing the necessary stuff to make referral following work.  I've tested it with ActiveDirectory, which notoriously contains referrals, and it seems to work.  When you get a chance to review, again, let me know what needs to be fixed.

necouchman:
Rebased on current master and reorganized the configuration into its own package.

necouchman:
Closing this in favor of #345, which is a fresh attempt after pulling in all of the recent changes.

PR #206, opened 2017-10-28
GUACAMOLE-362: Fix resource leak in FileInputStream

necouchman:
Quick attempt at fixing a resource leak that Coverity identified when reading the private key file.  I think this is still in-scope for 0.9.14-incubating...

### DIFF
--- a/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/PrivateKeyGuacamoleProperty.java
+++ b/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/PrivateKeyGuacamoleProperty.java
@@ -60,6 +60,7 @@ public abstract class PrivateKeyGuacamoleProperty implements GuacamoleProperty<P
             for (int readBytes; (readBytes = keyStreamIn.read(keyBuffer)) != -1;)
                 keyStreamOut.write(keyBuffer, 0, readBytes);
 
+            keyStreamIn.close();
             final byte[] keyBytes = keyStreamOut.toByteArray();
 
             // Set up decryption infrastructure
--- a/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/PrivateKeyGuacamoleProperty.java
+++ b/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/PrivateKeyGuacamoleProperty.java
@@ -81,7 +81,8 @@ public abstract class PrivateKeyGuacamoleProperty implements GuacamoleProperty<P
                 throw new GuacamoleServerException("Key is not in expected PKCS8 encoding.", e);
             }
             finally {
-                keyStreamIn.close();
+                if (keyStreamIn != null)
+                    keyStreamIn.close();
             }
         }
         catch (IOException e) {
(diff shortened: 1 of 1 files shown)


mike-jumper:
>  I think this is still in-scope for 0.9.14-incubating...

Yes, definitely.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/PrivateKeyGuacamoleProperty.java
```diff
@@ -60,6 +60,7 @@ public PrivateKey parseValue(String value) throws GuacamoleServerException {
             for (int readBytes; (readBytes = keyStreamIn.read(keyBuffer)) != -1;)
                 keyStreamOut.write(keyBuffer, 0, readBytes);
 
+            keyStreamIn.close();
```
mike-jumper: If `write()` throws an exception, then this line will never be reached. You'll need something like:

```
try {
    // write() loop here
}
finally {
    keyStreamIn.close();
}

Comment (necouchman) [reply extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/PrivateKeyGuacamoleProperty.java]: Well, that was fun.  Nested `try{}` blocks to correctly capture all of the possible exceptions.  I *think* this works out correctly...though it seems like you could still run into issues if you hit the IOException during the file close, but I can't really see any way to get around that.

File: extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/PrivateKeyGuacamoleProperty.java [hunk 2]
```diff
@@ -49,38 +49,44 @@ public PrivateKey parseValue(String value) throws GuacamoleServerException {
...
+                throw new GuacamoleServerException("Could not find the specified key file.", e);
+            }
+            catch (NoSuchAlgorithmException e) {
+                throw new GuacamoleServerException("RSA algorithm is not available.", e);
+            }
+            catch (InvalidKeySpecException e) {
+                throw new GuacamoleServerException("Key is not in expected PKCS8 encoding.", e);
+            }
+            finally {
+                keyStreamIn.close();
```
mike-jumper: This will throw a `NullPointerException` if an exception occurs during the creation of `keyStreamIn`

necouchman: Oops...had it in there, originally, and it got removed somewhere along the way.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/PrivateKeyGuacamoleProperty.java
```diff
@@ -49,38 +49,45 @@ public PrivateKey parseValue(String value) throws GuacamoleServerException {
         if (value == null || value.isEmpty())
             return null;
 
+        FileInputStream keyStreamIn = null;
+
         try {
+            try {
```
mike-jumper: Rather than nest `try` blocks and initialize with `null`, you could always just move the handling for `FileNotFoundException` up:

    FileInputStream keyStreamIn;
    try {
        keyStreamIn = new FileInputStream(value);
    }
    catch (FileNotFoundException e) {
        throw new GuacamoleServerException("Could not find the specified key file.", e);
    }

Since `keyStreamIn` would then be guaranteed to be non-null, complexity of the rest would be reduced, and all you would need is a simple `close()` within a `finally` at the end, like you had before.

necouchman: So, the problem I'm running into is that the `close()` operation can actually throw an `IOException`, which is why I have to nest the `try{}` statements.  Apparently with Java 7+ it gets a little easier, but 6 is a little onerous:

https://stackoverflow.com/questions/156508/closing-a-java-fileinputstream

mike-jumper: Ah, indeed.

PR #208, opened 2017-11-07
GUACAMOLE-431: Fix Tunnel handling of CLOSED state

lightpriest:
When the tunnel is closed and another `connect()` call is made, it stops
notifying about state changes (`onstatechanges`) if the connection continues
to fail.

Another issue is that the `Tunnel.onerror()` handler is never called the
second time, since it thinks the tunnel is already closed.

The fix is to set the state to `CONNECTING` when calling `connect()`.
I also moved the `setState()` function to the base Tunnel class and used
that whenever a state needs to change in any tunnel implementation, as this
seemed appropriate.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Tunnel.js
+++ b/guacamole-common-js/src/main/webapp/modules/Tunnel.js
@@ -55,6 +55,25 @@ Guacamole.Tunnel = function() {
     this.sendMessage = function(elements) {};
 
     /**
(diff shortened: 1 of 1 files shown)


necouchman:
This looks okay to me.  @mike-jumper @jmuehlner Any comments?

jmuehlner:
Everything looks good to me. 👍

mike-jumper:
Looks good to me as well.

PR #209, opened 2017-11-09
GUACAMOLE-161: Add missing press and release CapsLock events on Mac O…

flangelo:
…S. Add CapsLock to the no_repeat key list.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Keyboard.js
+++ b/guacamole-common-js/src/main/webapp/modules/Keyboard.js
@@ -515,6 +515,7 @@ Guacamole.Keyboard = function(element) {
         0xFFE2: true, // Right shift
         0xFFE3: true, // Left ctrl 
         0xFFE4: true, // Right ctrl 
+        0xFFE5: true, // Caps Lock
         0xFFE7: true, // Left meta 
         0xFFE8: true, // Right meta 
         0xFFE9: true, // Left alt


mike-jumper:
Adding Caps Lock to the `no_repeat` list is definitely a well-needed change. I'm iffy on all the browser sniffing, though. Is there no other way to achieve this?

flangelo:
I do not see any harm in removing the browser sniffing; it would normalize CapsLock behavior should other platforms/browsers behave the same.

mike-jumper:
Admittedly, this wouldn't be the first time we had to add Mac-specific browser sniffing to work around a quirk. Due to old [GUAC-391](https://glyptodon.org/jira/browse/GUAC-391) via https://github.com/apache/incubator-guacamole-client/commit/b6769e4ba0d33fbcd6f4e8673078661636632403, from the pre-Apache days:

https://github.com/apache/incubator-guacamole-client/blob/e2c8594b3a0fa07ec766865e80211f29021e88d8/guacamole-common-js/src/main/webapp/modules/Keyboard.js#L187-L189

Overall, it seems the currently-identified quirks on Mac which are detectable only through browser-sniffing:

1. Alt keydown must not be `preventDefault()`-ed (or typing of certain symbols becomes impossible)
2. Caps Lock key events are unreliable (there may not be a corresponding keyup/keydown for every keydown/keyup)

If we're now handling more than one Mac quirk, it might be time to unify the checks, and allow the code handling those quirks to be a bit more semantically-driven (`if (capsLockUnreliable) { ... }` rather than `if (isAMac) { ... }`).

I'd also be OK with simply assuming Caps Lock is unreliable, always releasing the key after keydown, and always pressing it prior to keyup.

guacamole-common-js/src/main/webapp/modules/Keyboard.js
```diff
@@ -984,8 +1003,15 @@ Guacamole.Keyboard = function(element) {
             // Release specific key if known
             var keysym = first.keysym;
             if (keysym) {
+
+                // Press CapsLock first if Keyup event is alone.
+                // https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent#Special_cases
+                if (keysym === 0xFFE5)
+                    guac_keyboard.press(keysym);
```
mike-jumper: Given the following events, happening in order:

1. `keydown` Caps Lock
2. `keyup` Caps Lock

due to a single key press, would this then result in the following series of Guacamole keyboard events:

1. `keydown` Caps Lock (due to original `keydown`)
2. `keyup` Caps Lock (due to unreliable `keydown`)
3. `keydown` Caps Lock (due to unreliable `keyup`)
4. `keyup` Caps Lock (due to original `keyup`)

effectively doubling things on the remote end?

flangelo: You are correct; looking for the Keydown and Keyup events have no effect:
```
3.key,5.65509,1.1;	18	15:08:08.502
3.key,5.65509,1.0;	18	15:08:08.503
3.key,5.65509,1.1;	18	15:08:08.513
3.key,5.65509,1.0;	18	15:08:08.513
```
This leaves relying only on assuming Mac OS is quirky; which may be true until the behavior is 
fixed in the browser(s) and we'd have to handle that later.. Are we OK with that though.

mike-jumper: Generally, yes, that seems the only safe option - always releasing after handling keydown for Caps Lock, and then leaving the keyup behavior unmodified. There's no need to inspect the keyboard object's event queue to determine whether a keyup event is already present, since if so that event will be handled immediately after the keydown handling is complete, and the keyup event will simply be ignored if the corresponding key has already been marked as released.

Alternatively, if you want to browser sniff the quirk, I think that would be fine as well given the docs you've cited, but we should look to avoid repeating the sniff and to make the representation of the quirk(s) more semantic. From the iterations so far, it looks like this may even be as simple as just adding a call to `release()` if Caps Lock is known to be unreliable based on the platform test.

heyan01:
I want to ask a question : how to change the canvas height, and where ? I add a header in page, lead to the canvas cannot show completely。

necouchman:
@heyan01: Please don't ask the question, here, please ask on either the user or dev mailing list:

mike-jumper:
@flangelo - if you're OK with narrowing the scope of this PR to just adding Caps Lock to the no-repeat list, I've been looking into this further as I work on debugging the behavior of deadkeys across platforms, and I think I have a working set of changes which partly abstracts away the detection and representation of  all these keyboard quirks: #227

I think these changes should address the issue at hand, as well as other Mac- and iOS-specific issues with key repeat, while providing a framework for cleanly handling keyboard behavior quirks which can only be detected through browser sniffing.

flangelo:
@mike-jumper - sure, I can do that.

necouchman:
Ping @flangelo - any progress in updating this with the narrower scope?

mike-jumper:
Not sure when this changed, but it looks like this has been narrowed already. I now see only one commit, and the commit deals only with adding caps lock to the no-repeat list.

PR #210, opened 2017-11-16
GUACAMOLE-437: Allow to send custom headers with the Guacamole tunnel requests

nacx:
https://issues.apache.org/jira/browse/GUACAMOLE-437

When establishing a Guacamole tunnel to an endpoint that requires (for example) authentication, it is often needed to send custom headers. The existing `withCredentials` option works well for session cookies in cross-domain environments, but when there are no sessions or authentication relies on headers, there is no way to have an authenticated tunnel request.

This PR adds an optional variable to the `HTTPTunnel` so users can configure a set of custom headers to be sent on every tunnel request. That can be used to configure additional headers for authentication, distributed tracing, and other needs. Something like:

```javascript
var tunnel = new Guacamole.HTTPTunnel(endpoint, true);
tunnel.extraHeaders = {
    "X-Auth-Token": "Foo",
    "X-Auth-Signature": "Bar",
    "X-Trace-Id": "uuid"
};
var guac = new Guacamole.Client(tunnel);
````

I've also preferred not to add the extra headers to the tunnel constructor to keep backward compatibility.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Tunnel.js
+++ b/guacamole-common-js/src/main/webapp/modules/Tunnel.js
@@ -182,8 +182,13 @@ Guacamole.Tunnel.State = {
  *     Whether tunnel requests will be cross-domain, and thus must use CORS
  *     mechanisms and headers. By default, it is assumed that tunnel requests
  *     will be made to the same domain.
+ *
+ * @param {Object} [extraTunnelHeaders={}]
+ *     Key value pairs containing the header names and values of any additional
+ *     headers to be sent in tunnel requests. By default, no extra headers will
+ *     be added.
  */
-Guacamole.HTTPTunnel = function(tunnelURL, crossDomain) {
+Guacamole.HTTPTunnel = function(tunnelURL, crossDomain, extraTunnelHeaders) {
 
     /**
      * Reference to this HTTP tunnel.
@@ -213,6 +218,29 @@ Guacamole.HTTPTunnel = function(tunnelURL, crossDomain) {
      * @private
      */
     var receive_timeout = null;
+
+    /**
+     * Additional headers to be sent in tunnel requests. This dictionary can be
+     * populated with key/value header pairs to pass information such as authentication
+     * tokens, etc.
+     *
+     * @private
+     */
+    var extraHeaders = extraTunnelHeaders || {}
+
+    /**
+     * Adds the configured additional headers to the given request.
+     *
+     * @params {XMLHttpRequest} request
+     *     The request where the configured extra headers will be added.
+     *
+     * @private
+     */
+    function addExtraHeaders(request) {
+        for (var name in tunnel.extraHeaders) {
+            request.setRequestHeader(name, tunnel.extraHeaders[name]);
+        }
+    }
 
     /**
      * Initiates a timeout which, if data is not received, causes the tunnel
@@ -322,6 +350,7 @@ Guacamole.HTTPTunnel = function(tunnelURL, crossDomain) {
             var message_xmlhttprequest = new XMLHttpRequest();
             message_xmlhttprequest.open("POST", TUNNEL_WRITE + tunnel.uuid);
             message_xmlhttprequest.withCredentials = withCredentials;
+            addExtraHeaders(message_xmlhttprequest);
             message_xmlhttprequest.setRequestHeader("Content-type", "application/octet-stream");
 
             // Once response received, send next queued event.
@@ -553,6 +582,7 @@ Guacamole.HTTPTunnel = function(tunnelURL, crossDomain) {
         var xmlhttprequest = new XMLHttpRequest();
         xmlhttprequest.open("GET", TUNNEL_READ + tunnel.uuid + ":" + (request_id++));
         xmlhttprequest.withCredentials = withCredentials;
+        addExtraHeaders(xmlhttprequest);
         xmlhttprequest.send(null);
 
         return xmlhttprequest;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by necouchman** [comment]
To start, you need to open a JIRA issue for this and tag both this pull request and the commits with the JIRA issue (GUACAMOLE-XXX).

File: guacamole-common-js/src/main/webapp/modules/Tunnel.js [hunk 1]
```diff
@@ -215,6 +215,24 @@ Guacamole.HTTPTunnel = function(tunnelURL, crossDomain) {
     var receive_timeout = null;
 
     /**
+     * Additional headers to be sent in tunnel requests. This dictionary can be
+     * populated with key/value header pairs to pass information such as authentication
+     * tokens, etc.
+     */
+    var extraHeaders = {};
```
mike-jumper: 1. Declaring this via `var` produces a private variable, accessible only within the scope of this constructor. The `add_extra_headers()` function you've written will still work, but it's not accessing the `extraHeaders` declared here. If using this approach, `extraHeaders` will need to be defined as:

       this.extraHeaders = {};

2. Why not an optional parameter to the constructor, similar to `crossDomain`, rather than a public property?

guacamole-common-js/src/main/webapp/modules/Tunnel.js
```diff
@@ -215,6 +215,24 @@ Guacamole.HTTPTunnel = function(tunnelURL, crossDomain) {
...
+
+    /**
+     * Adds the configured additional headerss to the given request.
+     *
+     * @private
+     */
+    function add_extra_headers(request) {
```
mike-jumper: Though you'll find much old JavaScript code here which uses the lowercase_and_underscores style for functions, variables, etc., new code should use headlessCamelCase:

http://guacamole.apache.org/guac-style/#naming

Please also be sure to document the parameters of the function, even for functions which are private.

guacamole-common-js/src/main/webapp/modules/Tunnel.js
```diff
@@ -215,6 +215,24 @@ Guacamole.HTTPTunnel = function(tunnelURL, crossDomain) {
...
+        for (name in tunnel.extraHeaders) {
```
mike-jumper: `name` here is implicitly declared, and thus [will actually be on the global scope](http://blog.niftysnippets.org/2008/03/horror-of-implicit-globals.html). Variables need to be explicitly declared to avoid such scary defaults. You'll need something like:

    for (var name in tunnel.extraHeaders)

to properly restrict `name` to the scope of its function.

mike-jumper:
Though useful in the case of the HTTP tunnel, anyone that must rely on this feature will implicitly be unable to use WebSocket, since there is no mechanism to add such headers via JavaScript's WebSocket API.

This is not necessarily a sign that this shouldn't be done, but something worth thinking about. Any solution which must leverage this feature will have inherently limited performance.

guacamole-common-js/src/main/webapp/modules/Tunnel.js
mike-jumper: I just saw you mentioned:

> I've also preferred not to add the extra headers to the tunnel constructor to keep backward compatibility.

JavaScript will assign the value `undefined` to any parameter omitted from function calls, including constructors. As long as you ensure the behavior of the HTTP tunnel is identical to past releases in the case the parameter is omitted, this can be done without breaking compatibility.

nacx:
Thanks for the review and the javascript hints. I am from the Java world and trying to get used to it :)
I've applied the suggested fixes and moved the configuration of the headers to the constructor.

>Though useful in the case of the HTTP tunnel, anyone that must rely on this feature will implicitly be unable to use WebSocket, since there is no mechanism to add such headers via JavaScript's WebSocket API.

Yes, and that's a limitation on the WebSocket API itself. Whilst it can be used for Basic auth, by setting the credentials in the websocket URL, it does not support headers or an other auth schema. It is not a Guacamole limitation but a WebSocket API one, so I'd say it is fair to assume users connecting to a websocket endpoint will be already aware of that.

nacx:
Once the PR is good to merge I'll squash the commits into a single one and properly tag the commit with the JIRA id for a clean merge.

guacamole-common-js/src/main/webapp/modules/Tunnel.js
```diff
@@ -214,6 +219,29 @@ Guacamole.HTTPTunnel = function(tunnelURL, crossDomain) {
...
+     * populated with key/value header pairs to pass information such as authentication
+     * tokens, etc.
+     *
+     * @private
+     */
+    var extraHeaders = extraTunnelHeaders || {}
+
+    /**
+     * Adds the configured additional headerss to the given request.
+     *
```
mike-jumper: `request` is not an optional parameter of this function, and thus must not be within brackets here.

guacamole-common-js/src/main/webapp/modules/Tunnel.js
```diff
@@ -214,6 +219,29 @@ Guacamole.HTTPTunnel = function(tunnelURL, crossDomain) {
...
+    /**
+     * Additional headers to be sent in tunnel requests. This dictionary can be
+     * populated with key/value header pairs to pass information such as authentication
+     * tokens, etc.
+     *
+     * @private
+     */
+    var extraHeaders = extraTunnelHeaders || {}
+
+    /**
```
mike-jumper: "headerss"

mike-jumper:
Looks generally good, however:

> Once the PR is good to merge I'll squash the commits into a single one and properly tag the commit with the JIRA id for a clean merge.

If you wish to squash things, that's definitely your choice, but we don't have a requirement for this. As long as commits cover logical changes and properly tagged with the JIRA issue, things are good.

That said, please don't put off intended changes until after review passes. That would defeat the purpose of review, and could result in your changes being lost during merge of the approved changes.

nacx:
Fixed.

I tend to squash the commits to avoid merging meaningless commits like *review fixes, etc*, after the PR is approved, but that's entirely up to the reviewers. If you want me to squash the commits before merging, just let me know. If you want to directly merge as-is or squash them yourselves, please, do it 😄 

Thanks for having a look at the PR so quickly!

necouchman:
> If you want me to squash the commits before merging...

Squashing the commits is up to you, but you should go ahead and fix the commit messages, now, prepending the JIRA issue.

nacx:
Done

PR #211, opened 2017-11-20
GUACAMOLE-437: Fix extraHeaders scope and StaticHTTPTunnel

nacx:
After the changes in https://github.com/apache/incubator-guacamole-client/pull/210, the `extraHeaders` variable was not visible to the tunnel object in the scope of the add headers function.

This PR fixes that by letting the function receive the headers to add as a parameter (which also makes more sense), and fixes the `StaticHTTPTunnel` that was using the same add headers approach without having it defined.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Tunnel.js
+++ b/guacamole-common-js/src/main/webapp/modules/Tunnel.js
@@ -1091,8 +1091,13 @@ Guacamole.ChainedTunnel.prototype = new Guacamole.Tunnel();
  *     Whether tunnel requests will be cross-domain, and thus must use CORS
  *     mechanisms and headers. By default, it is assumed that tunnel requests
  *     will be made to the same domain.
+ *
+ * @param {Object} [extraTunnelHeaders={}]
+ *     Key value pairs containing the header names and values of any additional
+ *     headers to be sent in tunnel requests. By default, no extra headers will
+ *     be added.
  */
-Guacamole.StaticHTTPTunnel = function StaticHTTPTunnel(url, crossDomain) {
+Guacamole.StaticHTTPTunnel = function StaticHTTPTunnel(url, crossDomain, extraTunnelHeaders) {
 
     /**
      * Reference to this Guacamole.StaticHTTPTunnel.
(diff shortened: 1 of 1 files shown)


mike-jumper:
Yep, I have a few - just a moment.

guacamole-common-js/src/main/webapp/modules/Tunnel.js
```diff
@@ -226,19 +226,21 @@ Guacamole.HTTPTunnel = function(tunnelURL, crossDomain, extraTunnelHeaders) {
...
-    var extraHeaders = extraTunnelHeaders || {}
+    var extraHeaders = extraTunnelHeaders || {};
 
     /**
      * Adds the configured additional headers to the given request.
      *
-     * @params {XMLHttpRequest} request
+     * @param {XMLHttpRequest} request
      *     The request where the configured extra headers will be added.
+     * @param {Object} headers
```
mike-jumper: Please leave a blank line separating different `@param` (see other recent JSDoc'd or JavaDoc'd functions).

guacamole-common-js/src/main/webapp/modules/Tunnel.js
```diff
@@ -1088,8 +1090,12 @@ Guacamole.ChainedTunnel.prototype = new Guacamole.Tunnel();
  *     Whether tunnel requests will be cross-domain, and thus must use CORS
  *     mechanisms and headers. By default, it is assumed that tunnel requests
  *     will be made to the same domain.
+ * @param {Object} [extraTunnelHeaders={}]
```
mike-jumper: As above, please leave a blank line between different `@param`.

guacamole-common-js/src/main/webapp/modules/Tunnel.js
```diff
@@ -1107,6 +1113,31 @@ Guacamole.StaticHTTPTunnel = function StaticHTTPTunnel(url, crossDomain) {
...
+     * @private
+     */
+    var extraHeaders = extraTunnelHeaders || {};
+
+    /**
+     * Adds the configured additional headers to the given request.
+     *
+     * @param {XMLHttpRequest} request
+     *     The request where the configured extra headers will be added.
+     * @param {Object} headers
```
mike-jumper: Same here - blank line, plase.

mike-jumper:
OK - spoke too soon. The main concern I had was that the CORS parameter wasn't added to `StaticHTTPTunnel` (and thus the prototype of the constructor would end up diverging from convention), but it's there, and I misread.

All I have are JSDoc style nitpicks.

mike-jumper:
@nacx - here, I think, is a case where it definitely makes sense to not squash everything into one commit. Fixing the scope of `extraHeaders` and adding the same functionality to `StaticHTTPTunnel` are not the same logical change, but two distinct changes, and don't belong in the same commit. The changes, in this case, may be safe enough, but as general practice is concerned combining such things makes `git bisect`, `git blame`, etc. problematic.

nacx:
Applied review fixes and split the changes into two commits, as suggested.

mike-jumper:
OK - things look good. I will say that "Fix extraHeaders in the StaticHTTPTunnel" is not really correct, since you're adding the support, not fixing it. The intent is clear enough, though, and I'm fine either way.

nacx:
Well, in the previous PR I partially (badly) introduced it. I added the call to the `addHeaders` method to the StaticHTTPTunnel... without having defined it. A mess. The second commit in this PR fixes that, but I'll amend the message if needed.

necouchman:
Merged.

PR #214, opened 2017-12-06
GUACAMOLE-423: Bump version numbers for all modified components to 0.9.14.

mike-jumper:
Absolutely everything has been modified for 0.9.14, with the slight exception of guacamole-common which is modified only because of the -incubator rename.

### DIFF
--- a/doc/guacamole-example/pom.xml
+++ b/doc/guacamole-example/pom.xml
@@ -26,7 +26,7 @@
     <groupId>org.apache.guacamole</groupId>
     <artifactId>guacamole-example</artifactId>
     <packaging>war</packaging>
-    <version>0.9.13-incubating</version>
+    <version>0.9.14</version>
     <name>guacamole-example</name>
     <url>http://guacamole.apache.org/</url>
 
@@ -106,7 +106,7 @@
         <dependency>
             <groupId>org.apache.guacamole</groupId>
             <artifactId>guacamole-common</artifactId>
-            <version>0.9.13-incubating</version>
+            <version>0.9.14</version>
             <scope>compile</scope>
         </dependency>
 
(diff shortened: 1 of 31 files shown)


necouchman:
> Absolutely everything has been modified for 0.9.14, with the slight exception of guacamole-common which is modified only because of the -incubator rename.

So, we're going ahead with change for -common to 0.9.14 because of the -incubator rename?  Just want to make sure that's what you intended.

mike-jumper:
> So, we're going ahead with change for -common to 0.9.14 because of the -incubator rename? Just want to make sure that's what you intended.

Ultimately, yes, but the reason for bumping -common here is the same as the other components: it was modified. The scope of those modifications was small, in this case just updating metadata in the `pom.xml` for the -incubator rename, but it has been modified nonetheless.

> So, the only thing that I see that looks a little bit off is that the guacamole-docker/README.md file still references 0.9.10-incubating trees for the SQL scripts used to load the schema. I would guess we want to tweak that? Not sure if it belongs in this commit or another one.

You're right that this should be updated. It's unfortunate that doing so will temporarily break the link the `README.md`, at least until the release is actually out.

What about linking to http://guacamole.apache.org/releases/, and rewording to avoid having to bump this going forward? Something like: "Alternatively, you can use the SQL scripts included with the guacamole-auth-jdbc extension from [the corresponding release](http://guacamole.apache.org/releases/)."

necouchman:
> What about linking to http://guacamole.apache.org/releases/, and rewording to avoid having to bump this going forward? Something like: "Alternatively, you can use the SQL scripts included with the guacamole-auth-jdbc extension from the corresponding release."

Fine with me.  You want to make that change in this PR, or do a separate one for that?

mike-jumper:
Definitely a relevant change - I'll make the commit here momentarily.

PR #215, opened 2017-12-07
GUACAMOLE-457: include CAS login URI when requesting auth ticket

ceharris:
When attempting to use the CAS extension with a protocol-compliant CAS server, I discovered that when requesting an authentication ticket the URL used by the extension is missing the `/login` URI path defined in the spec.

This PR adds the necessary additional path segment before appending the service name parameter.

### DIFF
--- a/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/form/CASTicketField.java
+++ b/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/form/CASTicketField.java
@@ -41,6 +41,12 @@ public class CASTicketField extends Field {
     public static final String PARAMETER_NAME = "ticket";
 
     /**
+     * The standard URI name for the CAS login resource.
+     */
+    private static final String CAS_LOGIN_URI = "login";
+
+
+    /**
      * The full URI which the field should link to.
      */
     private final String authorizationURI;
@@ -57,11 +63,6 @@ public class CASTicketField extends Field {
      *     The full URL of the endpoint accepting CAS authentication
      *     requests.
      *
-     * @param clientID
-     *     The ID of the CAS client. This is normally determined ahead of
-     *     time by the CAS service through some manual credential request
-     *     procedure.
-     *
      * @param redirectURI
      *     The URI that the CAS service should redirect to upon successful
      *     authentication.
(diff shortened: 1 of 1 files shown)


necouchman:
So, a couple of things about this:
* In my experience, this isn't required - I use Guacamole against CAS 4.2.x and 5.0.x, and sending it the location without the /login results in it automatically forwarding over to the /login location.  I believe someone else posted to the list with the issue that they were using a "CAS-compatible" server that did not behave this way, but the official Apereo one seems to work without having to put /login on the URL.
* In at least some versions of the CAS server, the login endpoint can be configured with the following property:
`cas.securityContext.casProcessingFilterEntryPoint.loginUrl`

Hard-coding /login in the code will break Guacamole's ability to log in for anyone who customizes their URLs.  I think the number of people who customize their CAS server to a different login endpoint is probably pretty low, but it's probably equally likely that the number of people running a CAS server that does not redirect login requests transparently is probably equally low.  The CAS protocol may actually define /login as the endpoint for the login, so maybe it's worth going with the protocol, but, then, if the Apereo server lets you customize it, that seems a little strange.

I'm not dead set against making this change, but I think it's worth at least noting those issues and having at least a small discussion on the issue.

ceharris:
The Shib server, which is at least as broadly used as Apereo does not work with the extension as shipped. On the other hand, the Guacamole extension will successfully interoperate with an Apereo server with this change.

Protocols exist for a reason -- to ensure interoperable communications. The Postel principle suggests _be conservative in what you do, be liberal in what you accept from others_. That the Apereo implementation of the server protocol allows authentication to work despite the absence of the `/login` URI is an example of the latter. But for the Guacamole extension to depend on that behavior is the antithesis of the former.

I've written several CAS client implementations over the last ten years or so. I contributed to the original Yale CAS client implementation, wrote two alternatives, and I still support a [CAS extension for Wildfly](https://github.com/soulwing/cas-extension). In every case we ran across users who insisted that they needed to modify aspects of the protocol (particularly the URIs, but also XML element names, namespaces, etc) to suit local circumstances. It invariably ended badly. 

It's fine to try to support those who choose to run non-standard CAS, but to do so at the expense of supporting a server that correctly implements the documented protocol is indefensible.

necouchman:
> It's fine to try to support those who choose to run non-standard CAS, but to do so at the expense of supporting a server that correctly implements the documented protocol is indefensible.

Whoa, okay :-).  You've got me on board with the change.  Most of the other people I've encountered have been of the ilk of "Well, I tried xyz once and could not get it working."  Clearly you've got substantial experience with this, and I will gladly defer to your expertise and experience.

I'll review the PR more thoroughly as soon as I get a chance.

PR #217, opened 2017-12-12
GUACAMOLE-394: Add interface for browsing user access history.

mike-jumper:
Analogous to the connection history REST endpoint, these changes add a new user history endpoint which lists the start and end times of past user sessions. The user management interface is also updated, exposing the time that each user was last active (if known):

guac-user-last-active

### DIFF
--- a/guacamole/src/main/webapp/app/rest/types/Connection.js
+++ b/guacamole/src/main/webapp/app/rest/types/Connection.js
@@ -104,6 +104,15 @@ angular.module('rest').factory('Connection', [function defineConnection() {
          */
         this.sharingProfiles = template.sharingProfiles;
 
+        /**
+         * The time that this connection was last used, in seconds since
+         * 1970-01-01 00:00:00 UTC. If this information is unknown or
+         * unavailable, this will be null.
+         *
+         * @type Number
+         */
+        this.lastActive = template.lastActive;
+
     };
 
     return Connection;
--- a/guacamole/src/main/webapp/app/rest/types/Connection.js
+++ b/guacamole/src/main/webapp/app/rest/types/Connection.js
@@ -105,7 +105,7 @@ angular.module('rest').factory('Connection', [function defineConnection() {
         this.sharingProfiles = template.sharingProfiles;
 
         /**
-         * The time that this connection was last used, in seconds since
+         * The time that this connection was last used, in milliseconds since
          * 1970-01-01 00:00:00 UTC. If this information is unknown or
          * unavailable, this will be null.
          *
--- a/guacamole/src/main/webapp/translations/it.json
+++ b/guacamole/src/main/webapp/translations/it.json
@@ -543,9 +543,13 @@
 
         "FIELD_PLACEHOLDER_FILTER" : "@:APP.FIELD_PLACEHOLDER_FILTER",
 
+        "FORMAT_DATE" : "@:APP.FORMAT_DATE_TIME_PRECISE",
+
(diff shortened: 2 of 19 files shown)


guacamole/src/main/webapp/app/rest/types/Connection.js
```diff
@@ -104,6 +104,15 @@ angular.module('rest').factory('Connection', [function defineConnection() {
          */
         this.sharingProfiles = template.sharingProfiles;
 
+        /**
+         * The time that this connection was last used, in seconds since
+         * 1970-01-01 00:00:00 UTC. If this information is unknown or
+         * unavailable, this will be null.
+         *
+         * @type Number
```
necouchman: Based on the API code above on the servlet side, it looks like this parameter is delivered by Java as a Date/Time, but is a number of seconds (epoch) here.  Does the API deliver it as a number of seconds, or is this translation done automatically?

guacamole/src/main/webapp/translations/it.json
```diff
@@ -543,9 +543,13 @@
...
         "FIELD_PLACEHOLDER_FILTER" : "@:APP.FIELD_PLACEHOLDER_FILTER",
 
+        "FORMAT_DATE" : "@:APP.FORMAT_DATE_TIME_PRECISE",
+
         "HELP_USERS" : "Click or tap on a user below to manage that user. Depending on your access level, users can be added and deleted, and their passwords can be changed.",
 
-        "SECTION_HEADER_USERS"       : "Utenti"
+        "SECTION_HEADER_USERS"       : "Utenti",
+
+        "TABLE_HEADER_USERNAME" : "Username"
```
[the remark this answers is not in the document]
mike-jumper: I have no idea, personally. In this case, I copied the value from elsewhere in `it.json`:

https://github.com/apache/guacamole-client/blob/649fd8c036861014a6064f4af3e05f309cd92973/guacamole/src/main/webapp/translations/it.json#L191-L193

guacamole/src/main/webapp/app/rest/types/Connection.js
necouchman: @mike-jumper Any insight on this one?

guacamole/src/main/webapp/translations/it.json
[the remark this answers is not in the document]
necouchman: Okay, we'll go with that for now.

guacamole/src/main/webapp/app/rest/types/Connection.js
mike-jumper: Whoops - didn't see this comment. Yes, the Java `Date` is automatically converted to a timestamp in seconds by Jersey+Jackson when the object is serialized to JSON.

mike-jumper: Actually, I may be wrong here ... all the documentation I'm able to find thus far indicates that Jackson serializes to milliseconds, not seconds. @jmuehlner - do you recall offhand?

necouchman: Okay...I'll hold off on the merge until we nail that down.

jmuehlner: It looks like milliseconds, yeah. The documentation indicates so, as does inspection of existing API responses.

jmuehlner:
Aside from the seconds/milliseconds thing, everything looks fine to me.

mike-jumper:
OK - all milliseconds now.

PR #218, opened 2017-12-12
GUACAMOLE-363: Correct DOS-style newlines within SQL Server schema before it's too late.

mike-jumper:
While working on updating the SQL Server schema with respect to the user history changes, I ran into trouble due to the line endings of the SQL Server schema files using DOS-style newlines rather than UNIX-style like the rest of the codebase.

This change does nothing more than `dos2unix` the SQL Server schema files, before things get too baked into history and such a change becomes impractical.

Adding a `w=1` query parameter to the GitHub URL for the comparison/diff should to ignore whitespace should show zero other changes (see https://github.com/blog/967-github-secrets).


mike-jumper:
Yep, SSMS is fine with either style of line endings.

Nothing broke per se. It was just the diffs that looked extremely odd, with git including a `^M` at the end of lines which used the DOS-style endings. Since maintaining source code with mixed line endings can get hairy, I figured we should correct this before things get too far.

necouchman:
Cool.  As long as that still works, I'm fine with the change and will get it merged here in a little while.

PR #221, opened 2017-12-14
GUACAMOLE-464: configuration properties from OS environment

ceharris:
Wanted to see how others feel about this feature before spending more time on it. If the concept is agreeable, there's some work to do in updating documentation. Additionally, the start.sh script for the docker container image should probably be revised to take advantage of this feature.

Might also want to consider whether this feature must be enabled by the presence of another environment variable. This would prevent unexpected configuration of extension properties via the environment, with minor additional nuisance factor for using the feature.

### DIFF
--- a/guacamole-ext/src/main/java/org/apache/guacamole/environment/LocalEnvironment.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/environment/LocalEnvironment.java
@@ -78,11 +78,11 @@ public class LocalEnvironment implements Environment {
      * A property that determines whether environment variables are evaluated
      * to override properties specified in guacamole.properties.
      */
-    private static final BooleanGuacamoleProperty ENVIRONMENT_PROPERTY_OVERRIDE =
+    private static final BooleanGuacamoleProperty ENABLE_ENVIRONMENT_PROPERTIES =
         new BooleanGuacamoleProperty() {
             @Override
             public String getName() {
-                return "enable-environment-overrides";
+                return "enable-environment-properties";
             }
         };
 
@@ -104,7 +104,7 @@ public class LocalEnvironment implements Environment {
     /**
      * Flag indicating whether environment variables can override properties.
      */
-    private final boolean environmentPropertyOverride;
+    private final boolean environmentPropertiesEnabled;
 
     /**
      * The Jackson parser for parsing JSON files.
@@ -161,7 +161,7 @@ public class LocalEnvironment implements Environment {
         availableProtocols = readProtocols();
 
         // Should environment variables override configuration properties?
-        environmentPropertyOverride = propertyOverrideEnabled(properties);
+        environmentPropertiesEnabled = environmentPropertiesEnabled(properties);
     }
 
     /**
@@ -321,7 +321,7 @@ public class LocalEnvironment implements Environment {
     }
 
     /**
-     * Checks for the presence of the {@link #ENVIRONMENT_PROPERTY_OVERRIDE}
+     * Checks for the presence of the {@link #ENABLE_ENVIRONMENT_PROPERTIES}
      * property in the given properties collection.
      *
      * @param properties
@@ -335,13 +335,13 @@ public class LocalEnvironment implements Environment {
      *                            cannot be successfully parsed as a Boolean
      *
      */
-    private static boolean propertyOverrideEnabled(Properties properties)
+    private static boolean environmentPropertiesEnabled(Properties properties)
             throws GuacamoleException {
 
-        final Boolean enableOverrides = ENVIRONMENT_PROPERTY_OVERRIDE.parseValue(
-                properties.getProperty(ENVIRONMENT_PROPERTY_OVERRIDE.getName()));
+        final Boolean enabled = ENABLE_ENVIRONMENT_PROPERTIES.parseValue(
+                properties.getProperty(ENABLE_ENVIRONMENT_PROPERTIES.getName()));
 
-        return enableOverrides != null && enableOverrides;
+        return enabled != null && enabled;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


necouchman:
> Wanted to see how others feel about this feature before spending more time on it. If the concept is agreeable, there's some work to do in updating documentation. Additionally, the start.sh script for the docker container image should probably be revised to take advantage of this feature.

I think it's a good idea, and certainly seems like it would make Docker, in particular, a little easier to configure.

> Might also want to consider whether this feature must be enabled by the presence of another environment variable. This would prevent unexpected configuration of extension properties via the environment, with minor additional nuisance factor for using the feature.

I'd say there probably should be some way to explicitly enable it, either by another environment variable or maybe even a servlet container system property (e.g. TOMCAT_HOME/conf/catalina.properties).  I think the slight nuisance is worth avoiding confusion and maybe even some possible security concerns with it being enabled without the operator being aware.

mike-jumper:
It's an interesting idea. I agree that it might make configuration easier in certain environments, particularly Docker.

> I'd say there probably should be some way to explicitly enable it either by another environment variable or maybe even a servlet container system property

As manipulation of environment variables can sometimes be difficult to prevent, my $0.02 would be that there should a non-environment-variable mechanism for disabling this behavior.

Either way, I would also suggest that the logic of the current implementation be inverted. The property values within `guacamole.properties` should trump all, with environment variables being used only as a last resort.

ceharris:
Having guacamole.properties take precedence over the environment would interfere with a fairly common devops practice: the devops the team produces a local container image from the official image that lays in the local defaults via properties files that are backed in the image. Environment variables are then used to override some of those defaults for particular deployments/environments/situations.

I guess if we're going to have a non-environment mechanism for controlling the behavior, that mechanism could be used to specify which sources are evaluated and in what order. Perhaps that's getting too complicated?

necouchman:
> I guess if we're going to have a non-environment mechanism for controlling the behavior, that mechanism could be used to specify which sources are evaluated and in what order. Perhaps that's getting too complicated?

It seems like there could be a system property, something put in catalina.properties or the like, that could either change the order or at least tell Guacamole to allow environment variables to override guacamole.properties.  Maybe it should be some sort of tri-state parameter:

guacamole.use_env_variables = yes/prefer/no

Where yes is the behavior Mike suggested (guacamole.properties peferred, then environment variables), prefer allows the environment variables to take precedence over the guacamole.properties file, and no disables environment variable processing and is the default.

mike-jumper:
> It seems like there could be a system property, something put in catalina.properties or the like

Why not have the property within `guacamole.properties`?

> I guess if we're going to have a non-environment mechanism for controlling the behavior, that mechanism could be used to specify which sources are evaluated and in what order. Perhaps that's getting too complicated?

This and the tri-state parameter are probably unnecessarily complex, yeah. If it really is the generally-accepted behavior that environment variables should override the contents of the config file, with the config file mainly serving as a set a defaults, I would be OK with:

1. A simple boolean property within `guacamole.properties` which controls whether environment variables are used for property values.
2. The default behavior being to not use environment variables for property values.

ceharris:
I agree with the idea that sometimes it's difficult to control the (inherited) environment when running outside of containers, and that simply having an environment variable that enables the evaluation of other environment variables might violate the principle of least astonishment.

> Why not have the property within guacamole.properties?

I think that makes sense. Most of the valid use cases are going to involve putting something into the container filesystem (e.g. extensions, lib jars) anyway, either through a custom image, or by bind mounting a location of the host filesystem into the container. So having a simple boolean property to enable environment evaluation seems like a pretty workable solution.

ceharris:
Any suggestions/opinions on the name for the boolean property that will enable evaluation of environment variables as properties?

Just some ideas to seed some discussion... _zero_ personal investment in any of these:

* `allow-properties-from-environment`
* `enable-environment-properties`
* `evaluate-environment-as-properties`

ceharris:
This has been revised to use the boolean property `environment-property-override` to determine whether environment variables should be used to override/supplement properties that can be set in `guacamole.properties`.

Would still welcome feedback and/or other suggestions for the property name.

On a side note, the constructor in this class is a bit of a mess. There is a block of code that loads the properties collection that probably be should be factored out as a method. And some of the methods being called in the constructor rely on field state set by previously executed code in the constructor. These dependencies should probably be made explicit by passing the required state as method parameters and making the methods called by the constructor static so there can be no unintentional/non-obvious field initialization order dependencies.

I resisted the temptation to clean this up, because it's not what this PR is about. Is the appropriate approach to create a JIRA issue and corresponding PR for refactoring of this sort?

mike-jumper:
> This has been revised to use the boolean property `environment-property-override` to determine whether environment variables should be used to override/supplement properties that can be set in `guacamole.properties`.

My personal preference would be `enable-environment-properties` or `evaluate-environment-properties`, as the fact that the environment will now be interpreted for property values seems more obvious.

`environment-property-override` seems OK, but is a bit less clear on what a true/false value would mean.

> On a side note, the constructor in this class is a bit of a mess. ... I resisted the temptation to clean this up, because it's not what this PR is about. Is the appropriate approach to create a JIRA issue and corresponding PR for refactoring of this sort?

Yep.

I've been considering similar for cleaning up some of the oldest parts of guacamole-common-js. Most newer code is quite nice, but some of the older stuff is varying degrees of beastly.

necouchman:
I vote for `enable-environment-properties`.

guacamole-ext/src/main/java/org/apache/guacamole/environment/LocalEnvironment.java
```diff
@@ -27,6 +27,8 @@
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
+
+import org.apache.guacamole.properties.BooleanGuacamoleProperty;
```
necouchman: Minor style detail - in general I think we try to keep the import statements groups together and in mostly alphabetical order, and we don't tend to put spaces in between different groups of imports.  I don't have a preference one way or the other, but, for consistency's sake, seems like we should stick to that, here.

ceharris: Still trying to tune my style setup for this project. Thanks for catching that.

mike-jumper:
Looks good to me.

PR #222, opened 2018-01-02
GUACAMOLE-237: Move to new getUserMedia method.

necouchman:
Pretty simple change here to prefer the newer getUserMedia() method.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/AudioRecorder.js
+++ b/guacamole-common-js/src/main/webapp/modules/AudioRecorder.js
@@ -182,15 +182,18 @@ Guacamole.RawAudioRecorder = function RawAudioRecorder(stream, mimetype) {
     var context = Guacamole.AudioContextFactory.getAudioContext();
 
     /**
-     * A function which directly invokes the browser's implementation of
-     * navigator.mediaDevices.getUserMedia() with all provided parameters.
-     *
-     * @type Function
+     * Some browsers do not implement navigator.mediaDevices - this
+     * shims in this functionality to ensure code compatibility.
      */
-
     if (!navigator.mediaDevices)
         navigator.mediaDevices = {};
 
+    /**
+     * Browsers that either do not implement navigator.mediaDevices
+     * at all or do not implement it completely need the getUserMedia
+     * method defined.  This shims in this function by detecting
+     * one of the supported legacy methods.
+     */
     if (!navigator.mediaDevices.getUserMedia)
         navigator.mediaDevices.getUserMedia = (navigator.getUserMedia
                 || navigator.webkitGetUserMedia
--- a/guacamole-common-js/src/main/webapp/modules/AudioRecorder.js
+++ b/guacamole-common-js/src/main/webapp/modules/AudioRecorder.js
@@ -181,19 +181,15 @@ Guacamole.RawAudioRecorder = function RawAudioRecorder(stream, mimetype) {
      */
     var context = Guacamole.AudioContextFactory.getAudioContext();
 
-    /**
-     * Some browsers do not implement navigator.mediaDevices - this
-     * shims in this functionality to ensure code compatibility.
-     */
+    // Some browsers do not implement navigator.mediaDevices - this
+    // shims in this functionality to ensure code compatibility.
     if (!navigator.mediaDevices)
         navigator.mediaDevices = {};
 
-    /**
-     * Browsers that either do not implement navigator.mediaDevices
-     * at all or do not implement it completely need the getUserMedia
-     * method defined.  This shims in this function by detecting
-     * one of the supported legacy methods.
-     */
+    // Browsers that either do not implement navigator.mediaDevices
+    // at all or do not implement it completely need the getUserMedia
+    // method defined.  This shims in this function by detecting
+    // one of the supported legacy methods.
     if (!navigator.mediaDevices.getUserMedia)
         navigator.mediaDevices.getUserMedia = (navigator.getUserMedia
                 || navigator.webkitGetUserMedia
--- a/guacamole-common-js/src/main/webapp/modules/AudioRecorder.js
+++ b/guacamole-common-js/src/main/webapp/modules/AudioRecorder.js
@@ -187,11 +187,15 @@ Guacamole.RawAudioRecorder = function RawAudioRecorder(stream, mimetype) {
      *
      * @type Function
      */
-    var getUserMedia = (navigator.mediaDevices.getUserMedia
-            || navigator.getUserMedia
-            || navigator.webkitGetUserMedia
-            || navigator.mozGetUserMedia
-            || navigator.msGetUserMedia).bind(navigator);
+
+    if (navigator.mediaDevices === undefined)
+        navigator.mediaDevices = {};
+
+    if (navigator.mediaDevices.getUserMedia === undefined)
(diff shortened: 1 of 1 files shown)


guacamole-common-js/src/main/webapp/modules/AudioRecorder.js
```diff
@@ -183,11 +183,12 @@ Guacamole.RawAudioRecorder = function RawAudioRecorder(stream, mimetype) {
 
     /**
      * A function which directly invokes the browser's implementation of
-     * navigator.getUserMedia() with all provided parameters.
+     * navigator.mediaDevices.getUserMedia() with all provided parameters.
      *
      * @type Function
      */
-    var getUserMedia = (navigator.getUserMedia
+    var getUserMedia = (navigator.mediaDevices.getUserMedia
```
mike-jumper: I'm not sure this will work:

1. If `navigator.mediaDevices` is not defined (older browsers), then the fallback to `navigator.getUserMedia`, etc. will not take effect as the attempt to access `navigator.mediaDevices.getUserMedia` will fail. The example at https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia gets around this by ensuring `navigator.mediaDevices` is set to `{}` if not otherwise defined, but I'm sure there are other and/or better approaches.
2. The call to `(...).bind(navigator)` creates a function which automatically uses `navigator` for `this`. This is correct for functions like `navigator.getUserMedia()`, but not for `navigator.mediaDevices.getUserMedia()`. I have not checked, but I expect that `navigator.mediaDevices.getUserMedia()` would need `this` to be `navigator.mediaDevices`.

Rather than point directly to `getUserMedia()`, we could instead point to `mediaDevices`, invoking the function later on as `mediaDevices.getUserMedia()`, defining a `mediaDevices` object with a `getUserMedia` property if not already provided on `navigator`. That might solve things cleanly, and the `bind()` would only be necessary in the `navigator` fallback case.

mike-jumper: In fact ... it might be even better to simply:

1. Test whether `mediaDevices` is defined on `navigator`.
2. If not, define a shim `mediaDevices` object on `navigator` having a single `getUserMedia` property pointing to the function we used to reference here.
3. Use `navigator.mediaDevices.getUserMedia()` directly where we previously used the locally-defined `getUserMedia()`.

necouchman: Okay...I'll take a stab at that...

necouchman:
How does this look?  Is this the right place for it, or should this check/initialization move elsewhere?

As far as I can tell there is just the one call to getUserMedia, so I'm thinking this should work.  Tested it out and I don't get any errors in my recent version of Chrome.

guacamole-common-js/src/main/webapp/modules/AudioRecorder.js
```diff
@@ -183,14 +183,19 @@ Guacamole.RawAudioRecorder = function RawAudioRecorder(stream, mimetype) {
...
+     * navigator.mediaDevices.getUserMedia() with all provided parameters.
      *
      * @type Function
      */
-    var getUserMedia = (navigator.getUserMedia
-            || navigator.webkitGetUserMedia
-            || navigator.mozGetUserMedia
-            || navigator.msGetUserMedia).bind(navigator);
+
+    if (navigator.mediaDevices === undefined)
```
mike-jumper: Probably just want to simply:

```
if (!navigator.mediaDevices)
```

guacamole-common-js/src/main/webapp/modules/AudioRecorder.js
```diff
@@ -183,14 +183,19 @@ Guacamole.RawAudioRecorder = function RawAudioRecorder(stream, mimetype) {
...
+        navigator.mediaDevices = {};
+
+    if (navigator.mediaDevices.getUserMedia === undefined)
```
mike-jumper: Since this will only be valid if `navigator.mediaDevices` was not defined in the first place (we'd need to `bind()` to `navigator` otherwise). I'd suggest:

```javascript
navigator.mediaDevices = {

    /**
     * That comment from before.
     */
    getUserMedia: (...all those tests...).bind(navigator)

};
```

necouchman: So, the reason I went this route is based on one of the examples on that MDN page ([1]).  The comment in the code at the bottom of the page says:

    // Some browsers partially implement mediaDevices. We can't just assign an object
    // with getUserMedia as it would overwrite existing properties.
    // Here, we will just add the getUserMedia property if it's missing.

Thus, I was going for the same level of compatibility as defined in that code sample.

guacamole-common-js/src/main/webapp/modules/AudioRecorder.js
```diff
@@ -183,14 +183,19 @@ Guacamole.RawAudioRecorder = function RawAudioRecorder(stream, mimetype) {
 
     /**
      * A function which directly invokes the browser's implementation of
```
mike-jumper: This JSDoc comment is no longer valid now that `var getUserMedia` is gone.

guacamole-common-js/src/main/webapp/modules/AudioRecorder.js
```diff
@@ -183,14 +183,19 @@ Guacamole.RawAudioRecorder = function RawAudioRecorder(stream, mimetype) {
...
+     * navigator.mediaDevices.getUserMedia() with all provided parameters.
      *
      * @type Function
      */
-    var getUserMedia = (navigator.getUserMedia
-            || navigator.webkitGetUserMedia
-            || navigator.mozGetUserMedia
-            || navigator.msGetUserMedia).bind(navigator);
+
+    if (!navigator.mediaDevices)
```
mike-jumper: With the old JSDoc comment no longer being valid, there should be some comments here describing what's going on.

guacamole-common-js/src/main/webapp/modules/AudioRecorder.js
mike-jumper: Ah, OK. That makes good sense.

guacamole-common-js/src/main/webapp/modules/AudioRecorder.js
necouchman: Removed.

guacamole-common-js/src/main/webapp/modules/AudioRecorder.js
necouchman: Added comments.

mike-jumper: Looks much better, however these new comments are JSDoc comments. JSDoc comments are used for declarations only, and then to document a function, class, variable, etc.:

```javascript
/**
 * This is correct.
 *
 * @private
 */
var foo;
```

They cannot / should not be used for general code which is not a declaration:

```javascript
/**
 * This is not correct.
 */
if (foo) {
    bar();
}
```

Normal `//` comments are the correct thing to use here:

```javascript
// This is correct
if (foo) {
    bar();
}
```

necouchman: Fixed.

PR #224, opened 2018-01-03
GUACAMOLE-113: Implement hotkey for sending Ctrl-Alt-Del

necouchman:
This change implements sending Ctrl-Alt-Del to the remote system by capturing Ctrl-Alt-End, which is a pretty standard hotkey among RDP clients for passing through the three-finger-salute to remote systems.

### DIFF
--- a/guacamole/src/main/webapp/app/client/controllers/clientController.js
+++ b/guacamole/src/main/webapp/app/client/controllers/clientController.js
@@ -77,12 +77,9 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
         MENU_KEYS   = angular.extend({}, SHIFT_KEYS, ALT_KEYS, CTRL_KEYS);
 
     /**
-     * Keys needed to support the Ctrl-Alt-End hotkey for sending
-     * Ctrl-Alt-Delete.
+     * Keysym for sending the DELETE key when the Ctrl-Alt-End hotkey
+     * combo is pressed.
      */
-    CAD_KEYS    = angular.extend({}, ALT_KEYS, CTRL_KEYS, END_KEYS);
-    var CTRL_KEY    = 0xFFE3;
-    var ALT_KEY     = 0xFFE9;
     var DEL_KEY     = 0xFFFF;
 
     /**
--- a/guacamole/src/main/webapp/app/client/controllers/clientController.js
+++ b/guacamole/src/main/webapp/app/client/controllers/clientController.js
@@ -278,6 +278,7 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
      * Map of all substituted key presses.  If one key is pressed in place of another
      * the value of the substituted key is stored in an object with the keysym of
      * the original key.
+     *
      * @type Object.<Number, Number>
      */
     var substituteKeysPressed = {};
@@ -505,20 +506,17 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
 
     };
 
-    /**
-     * Track pressed keys, opening the Guacamole menu after Ctrl+Alt+Shift, or
-     * send Ctrl-Alt-Delete when Ctrl-Alt-End is pressed.
-     */
+    // Track pressed keys, opening the Guacamole menu after Ctrl+Alt+Shift, or
+    // send Ctrl-Alt-Delete when Ctrl-Alt-End is pressed.
     $scope.$on('guacKeydown', function keydownListener(event, keysym, keyboard) {
 
         // Record key as pressed
         keysCurrentlyPressed[keysym] = true;   
         
         var currentKeysPressedKeys = Object.keys(keysCurrentlyPressed);
-        /**
-         * If only menu keys are pressed, and we have one keysym from each group,
-         * and one of the keys is being released, show the menu. 
-         */
+
+        // If only menu keys are pressed, and we have one keysym from each group,
+        // and one of the keys is being released, show the menu. 
         if (checkMenuModeActive()) {
             
             // Check that there is a key pressed for each of the required key classes
@@ -541,10 +539,8 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
             }
         }
 
-        /**
-         * If only Ctrl-Alt-End is pressed, and we have a one keysym from each
-         * group, and one key is being released, send Ctrl-Alt-Delete.
-         */
+        // If only Ctrl-Alt-End is pressed, and we have a one keysym from each
+        // group, and one key is being released, send Ctrl-Alt-Delete.
         if (END_KEYS[keysym] &&
             !_.isEmpty(_.pick(ALT_KEYS, currentKeysPressedKeys)) &&
             !_.isEmpty(_.pick(CTRL_KEYS, currentKeysPressedKeys))
--- a/guacamole/src/main/webapp/app/client/controllers/clientController.js
+++ b/guacamole/src/main/webapp/app/client/controllers/clientController.js
@@ -72,7 +72,9 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
         ALT_KEYS    = {0xFFE9 : true, 0xFFEA : true, 0xFE03 : true,
                        0xFFE7 : true, 0xFFE8 : true},
         CTRL_KEYS   = {0xFFE3 : true, 0xFFE4 : true},
+        END_KEYS    = {0xFF57 : true, 0xFFB1 : true},
         MENU_KEYS   = angular.extend({}, SHIFT_KEYS, ALT_KEYS, CTRL_KEYS);
+        CAD_KEYS    = angular.extend({}, ALT_KEYS, CTRL_KEYS, END_KEYS);
 
     /**
      * All client error codes handled and passed off for translation. Any error
@@ -288,6 +290,16 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
         return true;
     }
 
+    function checkCADHotkeyActive() {
+        for(var keysym in keysCurrentlyPressed) {
+            if(!CAD_KEYS[keysym]) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+
     // Hide menu when the user swipes from the right
     $scope.menuDrag = function menuDrag(inProgress, startX, startY, currentX, currentY, deltaX, deltaY) {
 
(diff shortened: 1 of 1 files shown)


necouchman:
> It would make more sense to just check within keydown/keyup:
> Is the key being pressed End?
> If so, are Ctrl + Alt currently being held?
> If so, send Delete instead.

Ah, okay, so, something (pseudo-code) like:

    if (keysym is in END_KEYS && CTRL and ALT are pressed) {
        event.preventDefault();
        keydown DELETE
    }

and then the matching one for keyup?

necouchman:
Okay, so, refactored a little based on what I think you were suggesting.  It seems to work - the only thing I'm concerned about with this code is the order in which keys are released - that is, if the Ctrl or Alt keyup events occur prior to the End keyup event, then it seems like the "Delete" key could get stuck down, which doesn't seem like a good thing.  I haven't been able to cause that to happen, just seems like a possibility.

mike-jumper:
> ... if the Ctrl or Alt keyup events occur prior to the End keyup event, then it seems like the "Delete" key could get stuck down, ...

That's a good point.

Perhaps `Guacamole.Keyboard` itself should be expanded to allow:

* Testing whether a key is currently pressed/released according to `Guacamole.Keyboard`'s own tracking
* Overriding the identity of a pressed key, such that the identity of that key is properly handled on keyup

necouchman:
> Perhaps Guacamole.Keyboard itself should be expanded to allow:

Okay, so that would mean these changes get moved to the guacamole-common-js area, right?  Is that really what we want - defining this sort of hotkey at the JS API level?  Or did you have something else in mind - adding a more abstract method of overriding the keys in the common-js area, but still actually defining the hotkey in the client controller?

mike-jumper:
The latter - providing some abstract means of overriding `Guacamole.Keyboard`'s interpretation of the keys pressed, such that shortcuts are kept within application logic while proper tracking and release of keys can be left up to `Guacamole.Keyboard`.

necouchman:
So, was not trying to ignore your advice, but was looking at `Keyboard.js` and didn't really see any logic in there currently for keeping track of what keys are actually pressed, outside of the last pressed one.  So, I went a little bit different route and added a `substituteKeysPressed` object to the `clientController.js` file to keep track of things that way.  Let me know if that looks okay, or if you think there's a better way in `Keyboard.js` to do this tracking.

mike-jumper:
> ... didn't really see any logic in there currently for keeping track of what keys are actually pressed, outside of the last pressed one.

Ah - I can help clarify here. The `Guacamole.Keyboard` object tracks the current expected state of all keys within its `pressed` property:

https://github.com/apache/guacamole-client/blob/00d464b51181d04e123472050833aa265dfdfee2/guacamole-common-js/src/main/webapp/modules/Keyboard.js#L531-L536

This property is publicly accessible (for checking key status) and is also used internally by `Guacamole.Keyboard` to drive the `reset()` function:

https://github.com/apache/guacamole-client/blob/00d464b51181d04e123472050833aa265dfdfee2/guacamole-common-js/src/main/webapp/modules/Keyboard.js#L787-L800

> I went a little bit different route and added a `substituteKeysPressed` object to the `clientController.js` file to keep track of things that way. Let me know if that looks okay, or if you think there's a better way in `Keyboard.js` to do this tracking.

I'll take a look. In abstract, this sounds fine, but could lead to some complexity (or, at least, redundancy) in the handling of key release, particularly when due to `reset()`. If `Guacamole.Keyboard` sees a keyup for a key that isn't currently pressed, or if the Guacamole webapp loses focus, `reset()` is invoked to ensure remote key state is sane. If keydown is being reinterpreted outside `Guacamole.Keyboard`, then the keyups from `reset()` will not match.

necouchman:
> This property is publicly accessible (for checking key status) and is also used internally by Guacamole.Keyboard to drive the reset() function:

Okay, cool, I'll take a look at that.

> I'll take a look. In abstract, this sounds fine, but could lead to some complexity (or, at least, redundancy) in the handling of key release, particularly when due to reset(). If Guacamole.Keyboard sees a keyup for a key that isn't currently pressed, or if the Guacamole webapp loses focus, reset() is invoked to ensure remote key state is sane. If keydown is being reinterpreted outside Guacamole.Keyboard, then the keyups from reset() will not match.

I believe I have implemented this in a way that is safe in that regard, because the `clientController.js` file still uses the `guacSyntheticKeyUp/Down` calls to press/release the substituted key, so the actual key that is being pressed (and therefore needs to be released/reset) should be consistent when passed to the `Guacamole.Keyboard` object.  I think.  Maybe.  :-)

mike-jumper:
That sounds pretty reasonable.

guacamole/src/main/webapp/app/client/controllers/clientController.js
```diff
@@ -64,16 +65,23 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
...
-    /*
+    /**
      * In order to open the guacamole menu, we need to hit ctrl-alt-shift. There are
      * several possible keysysms for each key.
      */
     var SHIFT_KEYS  = {0xFFE1 : true, 0xFFE2 : true},
         ALT_KEYS    = {0xFFE9 : true, 0xFFEA : true, 0xFE03 : true,
                        0xFFE7 : true, 0xFFE8 : true},
         CTRL_KEYS   = {0xFFE3 : true, 0xFFE4 : true},
+        END_KEYS    = {0xFF57 : true, 0xFFB1 : true},
```
mike-jumper: As the comment for this set of declarations deals specifically with Ctrl+Alt+Shift, I suggest not including this declaration here, but rather separating it like you did with `DEL_KEY`.

guacamole/src/main/webapp/app/client/controllers/clientController.js
```diff
@@ -522,21 +539,44 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
...
+                $rootScope.$broadcast('guacSyntheticKeydown', DEL_KEY);
+        }
+
     });
 
     // Update pressed keys as they are released, synchronizing the clipboard
     // with any data that appears to have come from those key presses
     $scope.$on('guacKeyup', function keyupListener(event, keysym, keyboard) {
 
+        var currentKeysPressedKeys = Object.keys(keysCurrentlyPressed);
```
mike-jumper: This variable doesn't seem to be used here.

guacamole/src/main/webapp/app/client/controllers/clientController.js
```diff
@@ -522,21 +539,44 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
             }
         }
 
+        // If only Ctrl-Alt-End is pressed, and we have a one keysym from each
+        // group, and one key is being released, send Ctrl-Alt-Delete.
```
mike-jumper: Is that "... and one key is being released ..." aspect correct?

guacamole/src/main/webapp/app/client/controllers/clientController.js
```diff
@@ -64,16 +65,23 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
...
                        0xFFE7 : true, 0xFFE8 : true},
         CTRL_KEYS   = {0xFFE3 : true, 0xFFE4 : true},
+        END_KEYS    = {0xFF57 : true, 0xFFB1 : true},
         MENU_KEYS   = angular.extend({}, SHIFT_KEYS, ALT_KEYS, CTRL_KEYS);
 
+    /**
+     * Keysym for sending the DELETE key when the Ctrl-Alt-End hotkey
+     * combo is pressed.
+     */
+    var DEL_KEY     = 0xFFFF;
```
mike-jumper: Need an `@type` tag. I also suggest just separating `DEL_KEY` and `=` with a single space, seeing as it's not being aligned with anything any more.

guacamole/src/main/webapp/app/client/controllers/clientController.js
```diff
@@ -522,21 +539,44 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
...
+        // If only Ctrl-Alt-End is pressed, and we have a one keysym from each
+        // group, and one key is being released, send Ctrl-Alt-Delete.
+        if (END_KEYS[keysym] &&
+            !_.isEmpty(_.pick(ALT_KEYS, currentKeysPressedKeys)) &&
+            !_.isEmpty(_.pick(CTRL_KEYS, currentKeysPressedKeys))
+        ) {
+                event.preventDefault();
+                delete keysCurrentlyPressed[keysym];
+                substituteKeysPressed[keysym] = DEL_KEY;
+                $rootScope.$broadcast('guacSyntheticKeydown', DEL_KEY);
```
mike-jumper: Out of curiosity, have you tried just using `$scope.$broadcast()`? There may not be a need to use `$rootScope`.

guacamole/src/main/webapp/app/client/controllers/clientController.js
necouchman: Oops, removed.

guacamole/src/main/webapp/app/client/controllers/clientController.js
necouchman: Have not - will check and see what happens.

guacamole/src/main/webapp/app/client/controllers/clientController.js
necouchman: Yeah, this comment needed a little cleaning.

guacamole/src/main/webapp/app/client/controllers/clientController.js
necouchman: Fixed.

guacamole/src/main/webapp/app/client/controllers/clientController.js
necouchman: Fixed.

guacamole/src/main/webapp/app/client/controllers/clientController.js
necouchman: Seems fine, change committed.

necouchman:
Let me know if there's anything else that needs tweaking.

guacamole/src/main/webapp/app/client/controllers/clientController.js
```diff
@@ -522,6 +545,18 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
             }
         }
 
+        // If one of the End keys is pressed, and we have a one keysym from each
+        // of Ctrl and Alt groups, send Ctrl-Alt-Delete.
+        if (END_KEYS[keysym] &&
+            !_.isEmpty(_.pick(ALT_KEYS, currentKeysPressedKeys)) &&
+            !_.isEmpty(_.pick(CTRL_KEYS, currentKeysPressedKeys))
+        ) {
+                event.preventDefault();
```
mike-jumper: Why is the body of this `if` indented by two levels?

necouchman: Ummm....variety is the spice of code?

Let me fix that real quick...

PR #225, opened 2018-01-14
GUACAMOLE-96: Allow extensions to decorate each other's objects.

mike-jumper:
Part of the core requirements of implementing the TOTP authentication extension is the ability to add and manage TOTP-specific attributes on the `User` objects of other extensions (to store each user's unique, secret key). To achieve this, this change adds the fundamental ability for extensions to modify each other's objects through adding two new functions to `UserContext`:

1. `decorate()` - allows a `UserContext` from another `AuthenticationProvider` to be wrapped by the current `AuthenticationProvider`
2. `redecorate()` - allows a previously-decorated `UserContext` to be updated in case the underlying `UserContext` has been updated.

See also: https://en.wikipedia.org/wiki/Decorator_pattern

This is admittedly a very large change, but it is only part 1 of 3 as far as TOTP is concerned. Following this is the other core requirement of allowing extensions to voluntarily store each other's data (and modifying the JDBC auth to do so), and finally the TOTP implementation itself.

I expect the documentation surrounding this new API is in need of particular review attention, as describing `decorate()`, `redecorate()`, and related functions proved difficult.

The core decorate/redecorate API
--------------------------------------------------

The `decorate()` function is analogous to `getUserContext()` and is invoked only during the initial authentication process for a user. For a `UserContext` retrieved via `getUserContext()` from a particular `AuthenticationProvider`, `decorate()` is invoked to decorate that `UserContext` for all **other** `AuthenticationProvider`s.

The `redecorate()` function is analogous to `updateUserContext() and is invoked only for new requests related to an existing session. If an `AuthenticationProvider` previously decorated a `UserContext` via `decorate()` or `redecorate()`, it is given that `UserContext` again as well as the `UserContext` returned by `updateUserContext() to allow the `AuthenticationProvider` to clean up its old object and decorate the new object.

Simplified decorate/undecorate API
--------------------------------------------------

To make things easier, I've also added several abstract classes for wrapping objects, as well as abstract `Directory` implementations which simplify decoration semantics. A typical pair of `decorate()` and `redecorate()` implementations would be something like:

```java
@Override
public UserContext decorate(UserContext context,
        AuthenticatedUser authenticatedUser, Credentials credentials)
        throws GuacamoleException {

    // Decorate the UserContext such that all Connection objects are wrapped
    return new DelegatingUserContext(context) {

        @Override
        public Directory<Connection> getConnectionDirectory()
                throws GuacamoleException {
            return new DecoratingDirectory<Connection>(super.getConnectionDirectory()) {

                @Override
                protected Connection decorate(Connection object)
                        throws GuacamoleException {

                    // Return a wrapped version of the original object which
                    // adds functionality of some kind
                    return new MyConnection(object);

                }

                @Override
                protected Connection undecorate(Connection object)
                        throws GuacamoleException {

                    // Return the original object
                    return ((MyConnection) object).getWrappedConnection();

                }

            };
        }

    };

}

@Override
public UserContext redecorate(UserContext decorated, UserContext context,
        AuthenticatedUser authenticatedUser, Credentials credentials)
        throws GuacamoleException {

    // No need to clean up / update the previously-decorated UserContext
    return decorate(context, authenticatedUser, credentials);

}
```


guacamole-ext/src/main/java/org/apache/guacamole/net/auth/DecoratingDirectory.java
```diff
@@ -0,0 +1,133 @@
...
+            throws GuacamoleException;
+
+    @Override
+    public ObjectType get(String identifier) throws GuacamoleException {
+
+        // Decorate only if object exists
+        ObjectType object = super.get(identifier);
+        if (object == null)
+            return object;
+
```
necouchman: This is more a curiosity question for me than anything, but is there some reason you chose this implementation rather than, for example:

    if (object != null)
        return decorate(object);

    return null;

?  I guess I'm a little fuzzy on why you'd return the object if there's nothing there, anyway, instead of flipping the logic?

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleAuthenticationProvider.java
```diff
@@ -260,6 +260,23 @@ public UserContext updateUserContext(UserContext context,
...
+        // Simply return the given context, decorating nothing
+        return context;
+
+    }
+
+    @Override
+    public UserContext redecorate(UserContext decorated, UserContext context,
+            AuthenticatedUser authenticatedUser, Credentials credentials)
+            throws GuacamoleException {
+        return decorate(context, authenticatedUser, credentials);
```
necouchman: Any particular reason in this instance you chose to pass up to the `decorate()` method rather than just returning the context?  In the extension modules you just `return context`, instead - why the difference?

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/DecoratingDirectory.java
mike-jumper: Yeah, that's a fair point. I'll change this to `return null`.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleAuthenticationProvider.java
mike-jumper: Mainly because this is part of the public API and not the internal workings of an extension. I chose to do things this way so that the class remains compliant with the expectations of `decorate()` / `redecorate()` if `decorate()` is overridden while `redecorate()` is not.

necouchman: :+1:

PR #228, opened 2018-01-15
GUACAMOLE-352: Work around lack of key events for dead keys.

mike-jumper:
Some platforms (but not all) will not send traditional keydown/keypress/keyup events for [dead keys](https://en.wikipedia.org/wiki/Dead_key), instead sending input or composition events, often requiring that an input field be focused for such events to be dispatched at all.

These changes modify `Guacamole.Keyboard` such that it additionally hooks into the "input" and "compositionend" events, thus producing key events for characters typed through dead keys even when traditional key events are not dispatched.

Because an input field often must be focused for these events to work, these changes also add a new object, `Guacamole.InputSink`, which acts as an automatically-focused input field that takes the focus whenever the focus is not on some other input field.

Additional and backwards-compatible modifications to `Guacamole.Keyboard` allow it to listen to the keyboard/input/composition events of multiple objects, such as both `document` (for intercepting all key events) and the element exposed by `Guacamole.InputSink` (for receiving events which are only dispatched if an input field is focused).

This all has the happy side effect of allowing [IMEs](https://en.wikipedia.org/wiki/Input_method) to function without using Guacamole's "text input" mechanism.


jmuehlner:
The code looks good to me, though there does appear to be a merge conflict.

Assuming this has been tested well, looks good to me!

mike-jumper:
I'll rebase to resolve those conflicts.

mike-jumper:
OK - rebased.

necouchman:
So, I've managed to reproduce a slightly odd behavior here a couple of times.  Here's what I'm seeing:
- Open a SSH terminal
- Hold down Shift + T, allowing it to type several upper-case T characters.
- Release Shift *before* releasing T

At this point, I can no longer use the Ctrl-Alt-Shift hot-key to produce the hidden Guacamole menu.  I can type other characters, including using "Shift" to modify other keys.  However, I cannot use the Ctrl-Alt-Shift hot-key, again, until I press Shift + T once again - almost like the Shift key state is somehow being tied to the combination of Shift + T, and the Shift state is not released properly when it is released, first - at least, not release "enough" in order to be used with the hot-key.

I'm using Chrome 60-something on OpenSuSE Leap 42.3 with XFCE.

necouchman:
> So, I've managed to reproduce a slightly odd behavior here a couple of times. Here's what I'm seeing:

...but, I've managed to reproduce this behavior in the current master branch *without* this PR applied, so doesn't look like it's specifically related to this.  Will go ahead with the merge.

mike-jumper:
On Jan 16, 2018 10:39, "necouchman" <notifications@github.com> wrote:

So, I've managed to reproduce a slightly odd behavior here a couple of
times. Here's what I'm seeing:

...but, I've managed to reproduce this behavior in the current master
branch *without* this PR applied, so doesn't look like it's specifically
related to this. Will go ahead with the merge.


I bet this behavior is related to the new automatic synchronization of
modifiers.

mike-jumper:
>>
>> So, I've managed to reproduce a slightly odd behavior here a couple of times. Here's what I'm seeing:
>>
>
> ...but, I've managed to reproduce this behavior in the current master branch without this PR applied, so doesn't look like it's specifically related to this. Will go ahead with the merge.
>

@necouchman, regarding this behavior, I suspect this is a regression due to my recent changes. I'll open an issue in JIRA and try to reproduce this tonight.

Buuuut ... as the proud owner of the environment where this problem occurs, if I'm unable to reproduce the problem, you might need to do a `git bisect` to confirm whether this is the case.

necouchman:
:+1:

PR #229, opened 2018-01-18
GUACAMOLE-161: Do not rely on receiving keyup for Caps Lock on Mac.

mike-jumper:
As noted by @flangelo in #209, Caps Lock will only result in keydown events on Mac OS X (keyup events are not dispatched). This change automatically releases Caps Lock immediately after it's pressed if the platform appears to be Mac.


mike-jumper:
Not completely - this PR does not change the set of keys for which Guacamole does automatic key repeat, which is part of #209 and definitely necessary.

It does, however, take the place of the changes within #209 which relate to detecting whether a Mac is in use and altering behavior of keyup for Caps Lock.

necouchman:
Okay - I'll go ahead and merge this.

PR #231, opened 2018-01-29
GUACAMOLE-434: Add parameter value and translation string for UK English keyboard layout.

mike-jumper:


guacamole-ext/src/main/resources/org/apache/guacamole/protocols/rdp.json
```diff
@@ -92,6 +92,7 @@
                     "options" : [
                         "",
                         "de-de-qwertz",
+                        "en-uk-qwerty",
```
necouchman: Just noticed - the manual documentation you added had this as "en-gb-qwerty", but here it is en-uk-qwerty.  Should it be different?  Or should they be the same?

mike-jumper: Whoops - good catch. The correct value is "en-gb-qwerty":

https://github.com/apache/guacamole-server/blob/493fa4df165b99357184f31db735aa701b50f7e4/src/protocols/rdp/keymaps/en_gb_qwerty.keymap#L21

Will fix.

mike-jumper: There. That should do it.

PR #232, opened 2018-01-29
GUACAMOLE-273: Add parameter value and translation string for Portuguese Brazilian keyboard layout.

mike-jumper:


mike-jumper:
OK - rebased against master.

PR #241, opened 2018-02-03
GUACAMOLE-197: Specify Charset when creating String from byte array.

necouchman:
This is a quick fix for the issue identified by Coverity in assuming the default encoding.


mike-jumper:
So, from your comment on JIRA:

> ... the client MUST NOT interpret the attribute locally ...

By decoding the binary value, whether using the platform default encoding or UTF-8, that's exactly what we're doing. To transform the opaque binary value into a string such that we aren't interpreting it, we need to encode it using a mechanism which does so, like base64 or hex.

necouchman:
Okay, new PR coming soon.

PR #245, opened 2018-02-04
GUACAMOLE-300: Support posixGroup in LDAP Authentication and Group-ba…

steffen-moser:
…sed Session Admission

The change will extend the user lookup code by the ability to search not only through the groupOfNames but also through the posixGroup scheme. The piece of code seems to work with both schemes in my tests successfully. 

I am not sure if there are any pitfalls when just combining the possible results. Maybe introducing a configuration flag to choose whether searching posixGroup or groupOfNames would be a better approach.

### DIFF
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
@@ -107,7 +107,7 @@ public class ConnectionService {
 
             // Get the search filter for finding connections accessible by the
             // current user
-            String connectionSearchFilter = getConnectionSearchFilter(userDN, ldapConnection);
+            String connectionSearchFilter = getConnectionSearchFilter(userDN, user, ldapConnection);
 
             // Find all Guacamole connections for the given user by
             // looking for direct membership in the guacConfigGroup
@@ -237,7 +237,7 @@ public class ConnectionService {
      * @throws GuacamoleException
      *     If an error occurs retrieving the group base DN.
      */
-    private String getConnectionSearchFilter(String userDN,
+    private String getConnectionSearchFilter(String userDN, AuthenticatedUser user,
             LDAPConnection ldapConnection)
             throws LDAPException, GuacamoleException {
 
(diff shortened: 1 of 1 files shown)


necouchman:
> Maybe introducing a configuration flag to choose...

I'm perfectly good with the changes you suggested here, but I would vote for making it configurable, as that would improve LDAP performance on searches, especially for larger trees.

mike-jumper:
My only concerns would be with respect to the `guacConfigGroup` schema:

https://github.com/apache/guacamole-client/blob/6db25395698632d93e766e73ec73437eb262b5a3/extensions/guacamole-auth-ldap/schema/guacConfigGroup.schema

The `guacConfigGroup` is defined as extending `groupOfNames`, which is a standard object class which defines and requires the `member` attribute (see [RFC 2519](https://tools.ietf.org/html/rfc4519#page-22)). Adding a `memberUid` attribute to a `guacConfigGroup` technically doesn't follow the schema definition of the object class itself, and though it does follow the schema of the object class referenced in this PR and JIRA (`posixGroup`, defined by [RFC 2307](https://tools.ietf.org/html/rfc2307)), I'm on unclear on:

1. Why does `guacConfigGroup`, a connection configuration mechanism specific to Guacamole, need to support `memberUid`? Why use `member`, as defined by `groupOfNames`?
2. If this is something that is desirable, how does it affect our defined LDAP schema?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -237,7 +237,7 @@
      * @throws GuacamoleException
```
necouchman: New parameter needs to be added to the documentation.

mike-jumper:
Yup, that's correct. Never mind - I misread the query.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -257,7 +257,7 @@ private String getConnectionSearchFilter(String userDN,
             LDAPSearchResults userRoleGroupResults = ldapConnection.search(
                 groupBaseDN,
                 LDAPConnection.SCOPE_SUB,
-                "(&(!(objectClass=guacConfigGroup))(member=" + escapingService.escapeLDAPSearchFilter(userDN) + "))",
+                "(&(!(objectClass=guacConfigGroup))(|(member=" + escapingService.escapeLDAPSearchFilter(userDN) + ")(memberUid=" + user.getCredentials().getUsername() + ")))",
```
mike-jumper: The username value here will need to be escaped for inclusion in a filter.

necouchman:
@steffen-moser Any chance you can work on this a bit more to get it ready for merge?  I'd like to set the final list of issues for inclusion in the version 1.0.0 release, and this looks like a good candidate to me if it's something you can spend a little more time on?

necouchman:
Ping @steffen-moser - see last note, would like to include this in 1.0.0...

necouchman:
@steffen-moser Merge base needs to be changed to staging/1.0.0 to be included in 1.0.0 release.

necouchman:
@steffen-moser Ping...

necouchman:
Closed by #450

PR #247, opened 2018-02-05
GUACAMOLE-96: Add support for TOTP as an additional authentication factor.

mike-jumper:
The last in a series of changes building the foundation for TOTP support (see #225 and #233), this change adds actual support for TOTP authentication. It builds off the reference implementation provided by the IETF, and supports enrollment via QR codes or manual entry of parameters into the authentication device.

Configurable values are given defaults compatible with Google Authenticator, thus the extension will work with Google Authenticator and similar solutions out-of-the-box, as well as others with minor additions to `guacamole.properties`.


extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/user/CodeUsageTrackingService.java
```diff
@@ -0,0 +1,264 @@
...
+     *     time a code should be marked as invalid cannot be read from
+     *     guacamole.properties.
+     */
+    public boolean useCode(String username, String code)
+            throws GuacamoleException {
+
+        // Repeatedly attempt to use the given code until an explicit success
+        // or failure has occurred
+        UsedCode usedCode = new UsedCode(username, code);
+        for (;;) {
```
necouchman: Any reason this is a `for` loop and not `while (true)`?

extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/user/CodeUsageTrackingService.java
```diff
@@ -0,0 +1,264 @@
...
+         */
+        public String getCode() {
+            return code;
+        }
+
+        @Override
+        public int hashCode() {
+            int hash = 7;
+            hash = 79 * hash + this.username.hashCode();
+            hash = 79 * hash + this.code.hashCode();
```
necouchman: Maybe this is part of the IETF reference implementation, but is there some significance to the values 7 and 79, here?  Is it worth making these constants, or is there a particular reason they should not be made constant?

mike-jumper: > Maybe this is part of the IETF reference implementation, but is there some significance to the values 7 and 79, here?

Nope, it's not specific to the IETF implementation. It's a generic hash function for the values within this class. The primes 7 and 79 were arbitrarily chosen and have no meaning other than to generally distribute the values well.

extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/user/CodeUsageTrackingService.java
mike-jumper: Only that it's what I'm used to. If you think `while (true)` is more clear, I'm happy to switch things.

mike-jumper:
> Also, does it make sense to have some sort of "totp-required" field that can be enabled to force the use of the TOTP module? Right now it looks like if a user doesn't have a key authentication succeeds ...

At the moment, if a user does not have a key, they are forced to enroll before being allowed to authenticate. If enrollment is impossible, or if the data associated with their account is invalid, then authentication is allowed through without TOTP.

> ... I can also situations where administrators would want to force enrollment and/or usage of 2FA. Thoughts?

That would probably be the most likely use case for any 2FA, including TOTP. We could simply fail the authentication attempt in all cases, always requesting a TOTP code, with the exception of initial enrollment.

extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/user/CodeUsageTrackingService.java
necouchman: Nah, doesn't matter much to me, just curious.

extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/user/CodeUsageTrackingService.java
necouchman: ok

necouchman:
> At the moment, if a user does not have a key, they are forced to enroll before being allowed to authenticate.

Okay, I would agree that this is, in general, the most desirable behavior.

> If enrollment is impossible, or if the data associated with their account is invalid, then authentication is allowed through without TOTP.

I already merged this (oops), but it might be worth considering making this configurable.  I would think that you'd want the default behavior to be that authentication fails if user data is invalid or enrollment is impossible, but that the administrator could turn that off (maybe even across the board, encompassing scenarios where users choose not to enroll) if desired.  Just something to think about, maybe to add/modify in the future.  For now it's merged and can be used as you've designed it.

mike-jumper:
No worries. I'll look into switching the logic around such that it fails the authentication attempt for all cases except enrollment and correct TOTP. Should be a pretty simply change.

mike-jumper:
@docelmo, if you have a question regarding Guacamole, please use the user@guacamole.apache.org mailing list. This pull request has been merged and is closed.

docelmo:
Sorry Mike I was referring to your extensions.  I would love to get 2FA setup where I can use OTP and AD LDAP integration for my users.  I will hit up the mailing list.

necouchman:
@bmullan @docelmo Probably best to take this conversation to the mailing list:

A closed Pull Request is probably not the best place for this conversation.

PR #250, opened 2018-02-07
GUACAMOLE-480: Read user attributes only if actually defined.

mike-jumper:


jmuehlner:
Under what situations do we expect that the attributes should be non-present? Is there a semantic difference between an empty attributes map and a null value?

I'm not sure this feels like the right way to fix this issue.

mike-jumper:
Looking into this more deeply, I agree. The documentation for guacamole-ext does not allow `getAttributes()` to return `null`, and based on the way the REST API works, the object should always have an attributes map, even if it's empty.

I am double-checking the other parts of the path between the REST API and the extensions, but it looks like this may well be at best a bug in the extension which was in use at the time, not a bug in Guacamole.

PR #251, opened 2018-02-07
GUACAMOLE-471: Update copyright year in NOTICE to 2018.

mike-jumper:


mike-jumper:
Oops. They all need to be updated, yeah.

mike-jumper:
There. That should be all of them.

PR #252, opened 2018-02-08
GUACAMOLE-504: Add getHttpStatusCode() method to GuacamoleException class

necouchman:
Adds the getHttpStatusCode() method to the GuacamoleException class, which will allow for exceptions to explicitly override the HTTP status code as required.

### DIFF
--- a/guacamole-common/src/main/java/org/apache/guacamole/GuacamoleException.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/GuacamoleException.java
@@ -80,5 +80,17 @@ public class GuacamoleException extends Exception {
     public int getHttpStatusCode() {
         return getStatus().getHttpStatusCode();
     }
+
+    /**
+     * Returns the most applicable WebSocket status code that can be
+     * associated with this exception.
+     *
+     * @return
+     *     An integer representing the most applicable WebSocket status
+     *     code associated with this exception.
+     */
+    public int getWebSocketCode() {
+        return getStatus().getWebSocketCode();
+    }
     
 }
--- a/guacamole-common/src/main/java/org/apache/guacamole/GuacamoleException.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/GuacamoleException.java
@@ -70,12 +70,14 @@ public class GuacamoleException extends Exception {
     }
 
     /**
-     * Returns the numeric HTTP status code associated with this exception.
+     * Returns the most applicable HTTP status code that can be associated
+     * with this exception.
      *
      * @return
-     *     The numeric HTTP status code associated with this exception.
+     *     An integer representing the most applicable HTTP status code
+     *     associated with this exception.
      */
-    public Integer getHttpStatusCode() {
+    public int getHttpStatusCode() {
         return getStatus().getHttpStatusCode();
     }
     
--- a/guacamole-common/src/main/java/org/apache/guacamole/GuacamoleException.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/GuacamoleException.java
@@ -68,5 +68,15 @@ public class GuacamoleException extends Exception {
     public GuacamoleStatus getStatus() {
         return GuacamoleStatus.SERVER_ERROR;
     }
+
+    /**
+     * Returns the numeric HTTP status code associated with this exception.
+     *
+     * @return
+     *     The numeric HTTP status code associated with this exception.
+     */
+    public Integer getHttpStatusCode() {
+        return getStatus().getHttpStatusCode();
+    }
     
 }
--- a/guacamole-common/src/main/java/org/apache/guacamole/servlet/GuacamoleHTTPTunnelServlet.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/servlet/GuacamoleHTTPTunnelServlet.java
@@ -149,26 +149,23 @@ public abstract class GuacamoleHTTPTunnelServlet extends HttpServlet {
      * @param response
      *     The HTTP response to use to send the error.
      *
-     * @param guacStatus
-     *     The status to send
-     *
-     * @param message
-     *     A human-readable message that can be presented to the user.
+     * @param guacamoleException
+     *     The exception that caused this error.
      *
      * @throws ServletException
      *     If an error prevents sending of the error code.
      */
     protected void sendError(HttpServletResponse response,
-            GuacamoleStatus guacStatus, String message)
+            GuacamoleException guacamoleException)
             throws ServletException {
 
         try {
 
             // If response not committed, send error code and message
             if (!response.isCommitted()) {
-                response.addHeader("Guacamole-Status-Code", Integer.toString(guacStatus.getGuacamoleStatusCode()));
-                response.addHeader("Guacamole-Error-Message", message);
-                response.sendError(guacStatus.getHttpStatusCode());
+                response.addHeader("Guacamole-Status-Code", Integer.toString(guacamoleException.getStatus().getGuacamoleStatusCode()));
+                response.addHeader("Guacamole-Error-Message", guacamoleException.getMessage());
+                response.sendError(guacamoleException.getHttpStatusCode());
             }
 
         }
@@ -258,12 +255,12 @@ public abstract class GuacamoleHTTPTunnelServlet extends HttpServlet {
         // HTTP response, logging each error appropriately.
         catch (GuacamoleClientException e) {
             logger.warn("HTTP tunnel request rejected: {}", e.getMessage());
-            sendError(response, e.getStatus(), e.getMessage());
+            sendError(response, e);
         }
         catch (GuacamoleException e) {
             logger.error("HTTP tunnel request failed: {}", e.getMessage());
             logger.debug("Internal error in HTTP tunnel.", e);
-            sendError(response, e.getStatus(), "Internal server error.");
+            sendError(response, e);
         }
 
     }
--- a/guacamole-common/src/main/java/org/apache/guacamole/servlet/GuacamoleHTTPTunnelServlet.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/servlet/GuacamoleHTTPTunnelServlet.java
@@ -149,23 +149,30 @@ public abstract class GuacamoleHTTPTunnelServlet extends HttpServlet {
      * @param response
      *     The HTTP response to use to send the error.
      *
-     * @param guacamoleException
-     *     The exception that caused this error.
+     * @param guacamoleStatusCode
+     *     The GuacamoleStatus code to send.
+     *
+     * @param guacamoleHttpCode
+     *     The numeric HTTP code to send.
+     *
+     * @param message
+     *     The human-readable error message to send.
      *
      * @throws ServletException
      *     If an error prevents sending of the error code.
      */
-    protected void sendError(HttpServletResponse response,
-            GuacamoleException guacamoleException)
+    protected void sendError(HttpServletResponse response, int guacamoleStatusCode,
+            int guacamoleHttpCode, String message)
             throws ServletException {
+
 
         try {
 
             // If response not committed, send error code and message
             if (!response.isCommitted()) {
-                response.addHeader("Guacamole-Status-Code", Integer.toString(guacamoleException.getStatus().getGuacamoleStatusCode()));
-                response.addHeader("Guacamole-Error-Message", guacamoleException.getMessage());
-                response.sendError(guacamoleException.getHttpStatusCode());
+                response.addHeader("Guacamole-Status-Code", Integer.toString(guacamoleStatusCode));
+                response.addHeader("Guacamole-Error-Message", message);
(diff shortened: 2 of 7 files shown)


guacamole-common/src/main/java/org/apache/guacamole/GuacamoleException.java
```diff
@@ -68,5 +68,15 @@ public GuacamoleException(Throwable cause) {
     public GuacamoleStatus getStatus() {
         return GuacamoleStatus.SERVER_ERROR;
     }
+
+    /**
+     * Returns the numeric HTTP status code associated with this exception.
```
mike-jumper: I think the semantics of the value returned here need to be more clearly defined. With `GuacamoleStatus`, for example, `getHttpStatusCode()` is defined as returning "the most applicable HTTP error code". In this case, we should be clear that the HTTP status code should be the nearest equivalent to the status represented by this exception.

guacamole-common/src/main/java/org/apache/guacamole/GuacamoleException.java
```diff
@@ -68,5 +68,15 @@ public GuacamoleException(Throwable cause) {
...
+     *
+     * @return
+     *     The numeric HTTP status code associated with this exception.
+     */
+    public Integer getHttpStatusCode() {
```
mike-jumper: This should a primitive `int`, not `Integer`.

mike-jumper:
Should there also be a `getWebSocketCode()`?

mike-jumper:
I don't see the corresponding changes to the REST API internals leveraging this new function. Should those be here, too?

guacamole-common/src/main/java/org/apache/guacamole/GuacamoleException.java
necouchman: Reworded it a bit.

guacamole-common/src/main/java/org/apache/guacamole/GuacamoleException.java
necouchman: Changed.

necouchman:
> I don't see the corresponding changes to the REST API internals leveraging this new function. Should those be here, too?

Yep - I've implemented those, now, so it should actually be effective :-).  I also changed the `sendError` method in the `HTTPTunnelServlet` class such that it uses the getHttpStatusCode() method directly from the exception rather than from the getStatus() call.

> Should there also be a getWebSocketCode()?

I went ahead and implemented this method; however, I'm not sure how worthwhile this is, since the corresponding changes to the `closeConnection` method in the various WebSocket tunnel classes aren't quite as easy to update.  There are two reasons for this:
* Unlike the REST API and the HTTP Tunnel, the `closeConnection()` method in the WebSocket tunnel classes is not only used for exceptions, it's also used for successful/normal closure, so this method cannot be quite as easily converted to use the `getWebSocketCode()` method.  For example:

https://github.com/apache/guacamole-client/blob/49c11bbdadf976409d84cb17a1884947fcb4d126/guacamole-common/src/main/java/org/apache/guacamole/websocket/GuacamoleWebSocketTunnelEndpoint.java#L188

* Several of the calls to `closeConnection()` include direct references to constants rather than by using `getStatus()` to pull the status code.  For example:

https://github.com/apache/guacamole-client/blob/49c11bbdadf976409d84cb17a1884947fcb4d126/guacamole-common/src/main/java/org/apache/guacamole/websocket/GuacamoleWebSocketTunnelEndpoint.java#L112

Thoughts on the Websocket side of things?  Is it worth re-tooling the WebSocket stuff to have a similar `sendError()` method that deals exclusively with exceptions?

guacamole-common/src/main/java/org/apache/guacamole/servlet/GuacamoleHTTPTunnelServlet.java
```diff
@@ -149,26 +149,23 @@ protected void doPost(HttpServletRequest request, HttpServletResponse response)
...
 
         try {
 
             // If response not committed, send error code and message
             if (!response.isCommitted()) {
-                response.addHeader("Guacamole-Status-Code", Integer.toString(guacStatus.getGuacamoleStatusCode()));
-                response.addHeader("Guacamole-Error-Message", message);
-                response.sendError(guacStatus.getHttpStatusCode());
+                response.addHeader("Guacamole-Status-Code", Integer.toString(guacamoleException.getStatus().getGuacamoleStatusCode()));
+                response.addHeader("Guacamole-Error-Message", guacamoleException.getMessage());
```
mike-jumper: The exception message should only be exposed via the `Guacamole-Error-Message` header when the exception itself deals with a client-side issue (any subclass of `GuacamoleClientException`). For other exceptions, the message should be assumed to contain internal information, and should not be forwarded along to the client.

This is the reason for the difference in handling below, where "Internal server error" is explicitly substituted for the exception message if the exception is not a `GuacamoleClientException`.

mike-jumper:
> I went ahead and implemented this method; however, I'm not sure how worthwhile this is, since the corresponding changes to the closeConnection method in the various WebSocket tunnel classes aren't quite as easy to update. There are two reasons for this:
>
> * Unlike the REST API and the HTTP Tunnel, the closeConnection() method in the WebSocket tunnel classes is not only used for exceptions, it's also used for successful/normal closure, so this method cannot be quite as easily converted to use the getWebSocketCode() method.
> * Several of the calls to closeConnection() include direct references to constants rather than by using getStatus() to pull the status code.
>

While it can't be modified to accept an exception as a parameter as the sole implementation of `closeConnection()`, it should still be possible to simply grab the error code from the exception where such use is applicable. Not all usages of that function need to be driven by exceptions.

guacamole-common/src/main/java/org/apache/guacamole/servlet/GuacamoleHTTPTunnelServlet.java
necouchman: Okay, I think I've reworked this in a way that avoids duplicating too much code but successfully hides internals we don't want to reveal.

necouchman:
> While it can't be modified to accept an exception as a parameter as the sole implementation of closeConnection(), it should still be possible to simply grab the error code from the exception where such use is applicable. Not all usages of that function need to be driven by exceptions.

I shall give it a shot.

necouchman:
> While it can't be modified to accept an exception as a parameter as the sole implementation of closeConnection(), it should still be possible to simply grab the error code from the exception where such use is applicable. Not all usages of that function need to be driven by exceptions.

Okay, I took a shot at an implementation of this.  I'm not sure it's the most elegant, most correct, etc., but I was going for the fewest lines of code added and least duplication of logic.  If there's another route that I should take - overloading or something like that - let me know.

guacamole-common/src/main/java/org/apache/guacamole/servlet/GuacamoleHTTPTunnelServlet.java
```diff
@@ -256,14 +259,18 @@ else if(query.startsWith(WRITE_PREFIX))
...
         // HTTP response, logging each error appropriately.
-        catch (GuacamoleClientException e) {
-            logger.warn("HTTP tunnel request rejected: {}", e.getMessage());
-            sendError(response, e.getStatus(), e.getMessage());
-        }
         catch (GuacamoleException e) {
-            logger.error("HTTP tunnel request failed: {}", e.getMessage());
-            logger.debug("Internal error in HTTP tunnel.", e);
-            sendError(response, e.getStatus(), "Internal server error.");
+            if (e instanceof GuacamoleClientException) {
```
mike-jumper: There's no need to manually check `instanceof` - this is exactly what `catch` is for.

guacamole-common/src/main/java/org/apache/guacamole/websocket/GuacamoleWebSocketTunnelEndpoint.java
```diff
@@ -71,11 +71,15 @@
...
      * @param guac_status The status to send.
+     * @param webSocketCode The numeric WebSocket status to send.
      */
-    private void closeConnection(Session session, GuacamoleStatus guac_status) {
+    private void closeConnection(Session session, GuacamoleStatus guac_status,
+            Integer webSocketCode) {
 
         try {
-            CloseCode code = CloseReason.CloseCodes.getCloseCode(guac_status.getWebSocketCode());
+            if (webSocketCode == null)
```
mike-jumper: Why not the same approach that you used with the HTTP status code handling? (Separate parameters for each type of status code rather than passing `GuacamoleStatus`)

It's generally bad practice to rely on wrapper classes like `Integer` when not necessary, and I think this is one of those cases where we really should be looking for a way to use a primitive `int`.

guacamole-common/src/main/java/org/apache/guacamole/websocket/GuacamoleWebSocketTunnelEndpoint.java
```diff
@@ -109,15 +113,15 @@ public void onOpen(final Session session, EndpointConfig config) {
             // Get tunnel
             tunnel = createTunnel(session, config);
             if (tunnel == null) {
-                closeConnection(session, GuacamoleStatus.RESOURCE_NOT_FOUND);
+                closeConnection(session, GuacamoleStatus.RESOURCE_NOT_FOUND, null);
```
mike-jumper: Assuming that we do go the status code / WebSocket code route (ie: same as the HTTP case), the DRY approach here would be to overload `closeConnection()` with a variant which accepts only `GuacamoleStatus`, pulling the codes it needs from there.

guacamole-common/src/main/java/org/apache/guacamole/servlet/GuacamoleHTTPTunnelServlet.java
necouchman: Okay - I wasn't sure what the more accepted/elegant approach was.  Will rework it that way.

guacamole-common/src/main/java/org/apache/guacamole/websocket/GuacamoleWebSocketTunnelEndpoint.java
necouchman: Ok.

guacamole-common/src/main/java/org/apache/guacamole/websocket/GuacamoleWebSocketTunnelEndpoint.java
necouchman: Sounds good.

guacamole-common/src/main/java/org/apache/guacamole/servlet/GuacamoleHTTPTunnelServlet.java
necouchman: Fixed.

guacamole-common/src/main/java/org/apache/guacamole/websocket/GuacamoleWebSocketTunnelEndpoint.java
necouchman: Switched to `int` and the same approach - passing both parameters as an `int` and not passing `GuacamoleStatus` at all.

guacamole-common/src/main/java/org/apache/guacamole/websocket/GuacamoleWebSocketTunnelEndpoint.java
necouchman: I re-implemented this with just `int` and no overloading, since the only reason to pass the `GuacamoleStatus` parameter is to access the `getGuacamoleStatusCode()` method.  Let me know how this looks - it doesn't feel very elegant, but maybe better than `Integer` and `null` route.

necouchman: Okay, dense moment is passed - I know what to do, now...hang on, another change coming.

necouchman:
Okay, cleaned it up a bit more - hopefully good, now.

guacamole-common/src/main/java/org/apache/guacamole/websocket/GuacamoleWebSocketTunnelEndpoint.java
```diff
@@ -85,6 +92,21 @@ private void closeConnection(Session session, GuacamoleStatus guac_status) {
...
     }
 
+    /**
+     * Sends the given Guacaomle Status and closes the given
+     * connection.
+     *
+     * @param session
+     *     The outbound WebSocket connection to close.
+     *
+     * @param guac_status
```
mike-jumper: Being new/touched code, this should be updated to match current code style (`guacStatus` not `guac_status`). The old `guac_status` parameter is a holdover from the days where I used C style for everything.

necouchman: Fixed.

necouchman:
Cleaned that up, should be good, now.

PR #253, opened 2018-02-10
GUACAMOLE-304: Expand connection link to encompass entire hover area.

necouchman:
This PR expands the <a> tag for the connections on the home page such that it can be clicked anywhere that the hover region exists.

### DIFF
--- a/guacamole/src/main/webapp/app/home/styles/home.css
+++ b/guacamole/src/main/webapp/app/home/styles/home.css
@@ -50,3 +50,7 @@ div.recent-connections div.connection {
     max-width: 75%;
     overflow: hidden;
 }
+
+a.home-connection {
+    display: block;
+}


manolan1:
This changed a.home-connection to block. Given that anchor elements are inline by default, wouldn't it have been better to change it to inline-block? I ask because it broke one of my extensions, but in general, I think the principle of least change should apply. (Obviously, I can override in the extension)

mike-jumper:
@manolan1 - this is a pull request which is closed, having been merged about a year ago, for an issue which has been resolved and released. If you have questions on these changes, please use the mailing list.

> ... wouldn't it have been better to change it to inline-block?

No. The issue that this PR addressed was the size of the link not actually matching the highlighted region, leading to confusion:

https://issues.apache.org/jira/browse/GUACAMOLE-304

`inline-block` would not have produced the same behavior as `block` in that respect.

> I ask because it broke one of my extensions ...

If you would like to propose a change, feel free to open a new JIRA issue and pull request with those changes. If there is a better way to achieve this which also has a lower potential of breaking third-party styling, I see no reason why that would not be merged.

Be sure to describe what specifically broke.

PR #254, opened 2018-02-16
GUACAMOLE-103: Implement SAML Authentication Extension

necouchman:
Initial cut at a SAML authentication extension.  Plenty of room for improvement, I'm sure - I welcome all comments/suggestions/changes.  I'll kick off my own review, here, shortly.


extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/AuthenticationProviderService.java
```diff
@@ -0,0 +1,202 @@
...
+                    logger.debug("Received XPathExpressionException while processing SAML response.", e);
+                    throw new GuacamoleInvalidCredentialsException("Error during SAML login.",
+                                CredentialsInfo.USERNAME_PASSWORD);
+                }
+                catch (Exception e) {
+                    logger.warn("Exception while getting name from SAML response: {}", e.getMessage());
+                    logger.debug("Received Exception while retrieving name from SAML response.", e);
+                    throw new GuacamoleInvalidCredentialsException("Error during SAML login.",
+                                CredentialsInfo.USERNAME_PASSWORD);
+                }
```
necouchman: Wow, this is nasty.  Unfortunately the SAML client doesn't currently capture exceptions it generates, so we have to do it, here.

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/SAMLAuthenticationProviderResource.java
```diff
@@ -0,0 +1,76 @@
...
+    @POST
+    @Path("callback")
+    public Response processSamlResponse(@FormParam("SAMLResponse") String samlResponse)
+            throws GuacamoleException {
+
+        String guacBase = confService.getCallbackUrl().toString();
+        try {
+            Response redirectHome = Response.seeOther(
+                new URI(guacBase + "?SAMLResponse=" + Util.urlEncoder(samlResponse))).build();
+            return redirectHome;
```
necouchman: This works okay; however, this redirect tends to cause issues with header size in my testing - I'm using the CAS SAML IdP, and I had to bump up the maxHttpHeaderSize on Tomcat and then the proxy_buffer_size and proxy_busy_buffers_size parameter in Nginx to get this to work.  It does work reliably, but not sure if this is the best way to go, or if I should try to process the SAML response directly in here?

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/conf/ConfigurationService.java
```diff
@@ -0,0 +1,223 @@
...
+public class ConfigurationService {
+
+    /**
+     * The file containing the XML Metadata associated with the SAML IdP.
+     */
+    private static final FileGuacamoleProperty SAML_IDP_METADATA =
+            new FileGuacamoleProperty() {
+
+        @Override
+        public String getName() { return "saml-idp-metadata"; }
```
necouchman: Unfortunately this option has no effect right now, because the OneLogin SAML client has not released the version that implements pulling settings from the IdP Metadata.  Hopefully it'll be out, soon, but the lead developer hasn't been very responsive.

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/conf/ConfigurationService.java
```diff
@@ -0,0 +1,223 @@
...
+     *     of "Apache Guacamole" if not specified.
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed, or if the client ID
+     *     property is missing.
+     */
+    private String getEntityId() throws GuacamoleException {
+        return environment.getProperty(
+            SAML_ENTITY_ID,
+            "Apache Guacamole"
```
necouchman: This needs to be changed - it looks like the Entity ID should actually be a URL of some sort.

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/conf/ConfigurationService.java
```diff
@@ -0,0 +1,223 @@
...
+     */
+    public Saml2Settings getSamlSettings() throws GuacamoleException {
+
+        // Initialize and configure SAML client.
+        Map<String, Object> samlMap = new HashMap<String, Object>();
+        samlMap.put("onelogin.saml2.sp.entityid", getEntityId());
+        samlMap.put("onelogin.saml2.sp.assertion_consumer_service.url", getCallbackUrl() + "/api/ext/saml/callback");
+        samlMap.put("onelogin.saml2.idp.entityid", getIdpUrl());
+        samlMap.put("onelogin.saml2.idp.single_sign_on_service.url", getIdpUrl());
+        samlMap.put("onelogin.saml2.idp.single_sign_on_sevice.binding", "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect");
```
necouchman: Not sure if I can use constants here??

guacamole-ext/src/main/java/org/apache/guacamole/properties/UrlGuacamoleProperty.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.net.MalformedURLException;
+import java.net.URL;
+import org.apache.guacamole.GuacamoleException;
+import org.apache.guacamole.GuacamoleServerException;
+
+/**
+ * A GuacamoleProperty whose value is a URL.
+ */
+public abstract class UrlGuacamoleProperty implements GuacamoleProperty<URL> {
```
necouchman: I went ahead and put this as a `guacamole-ext` property, as it seems like this might be useful outside of just this extension.

extensions/guacamole-auth-saml/src/licenses/LICENSE
```diff
@@ -0,0 +1,257 @@
...
+JSR-330 / Dependency Injection for Java (http://code.google.com/p/atinject/)
+----------------------------------------------------------------------------
+
+    Version: 1
+    From: 'JSR-330 Expert Group' (https://jcp.org/en/jsr/detail?id=330)
+    License(s):
+        Apache v2.0 (bundled/javax.inject-1/LICENSE-2.0.txt)
+
+
+Coveo Dead Simple SAML 2.0 Client (https://github.com/coveo/saml-client)
```
necouchman: Oops, wrong client...

extensions/guacamole-auth-saml/src/licenses/bundled/saml-client-1.2.0/LICENSE
```diff
@@ -0,0 +1,20 @@
...
+    The above copyright notice and this permission notice shall be
+    included in all copies or substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+    EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
+    OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+    NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+    HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+    IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR
+    IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
```
necouchman: Wrong SAML client.

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/conf/ConfigurationService.java
necouchman: This is now implemented - OneLogin released version 2.3.0 of their Java SAML client, which includes this capability.

saydulk:
@necouchman how i can get Guacamole working with CAS for SAML authentication

necouchman:
@saydulk http://guacamole.apache.org/support/

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/AuthenticationProviderService.java
necouchman: Will just keep this as-is for now - better at this point to capture the exceptions with specific error messages than to lump them all together.

Saphirim:
Hello,

out of interest: what are the latest plans to integrate SAML into a official release of guacamole?
We would evaluate guacamole for enterpriese usage, but SSO would be a really relevant feature here.

Thx & Regards,

Michael

tuxcrafter:
I would like to donate some money to speed up the SAML SSO and SLO support for guacamole. I am waiting to upgrade my pam auth with SLO and SSO. I am using an ipsilon SAML IdP and I would like to test the new extension when ready. If you need a test infrastructure you can contact me and I will share some accounts and an IdP and SP setup.

necouchman:
@tuxcrafter: You're welcome to go ahead and try the extension, now.  It does not support SLO, yet, but should work with version 1.0.0 for logging in.

tuxcrafter:
> @tuxcrafter: You're welcome to go ahead and try the extension, now. It does not support SLO, yet, but should work with version 1.0.0 for logging in.

@necouchman We compiled and install the saml extension but I can not figure out how to configure the saml settings.

I probably need to configure the following settings somewhere: 

idp-entityId  (metadata)
idp-singleLogoutService.url
idp-singleSignOnService.url
idp-x509cert
sp-privateKey
sp-x509cert

Then it would be good to have a mapping options for the username and groups.

LeBrad:
@necouchman Could we get a sample configuration file for the saml extension? It would enable us to start hacking on our saml/guacamole setup.

necouchman:
@LeBrad @tuxcrafter 
It's been a little while since I did any testing on this, but this is what I have in my guacamole.properties file for configuring the extension from when I wrote it a while back:

```
saml-idp-metadata: /etc/guacamole/saml-idp-metadata.xml
saml-entity-id: https://1.2.3.4/guacamole/api/tokens
saml-idp-url: https://4.3.2.1/cas/idp/profile/SAML2/Redirect/SSO
saml-callback-url: https://1.2.3.4/guacamole
saml-logout-url: https://4.3.2.1/cas/logout
saml-keystore-file: /etc/guacamole/saml-keystore
saml-keystore-password: secret
saml-pkey-password: secret
```
Note that if you have an XML metadata file from your SAML IDP, you should not need the rest of those details - all of those configuration items should be in the XML file.  Anyway, test away, feel free to let me know if you run into any issues with it.  I do need to rebase it against the latest `master` repo and redo some of the testing and make sure it's still good, but not sure when I'll get to that.

LeBrad:
Thanks for the properties.

1. Where do we place the certificate for the IdP?
2. Where do we place the certificates for the guacamole saml server, public and private?
3. Where do we get the metadata of the guacamole saml server?
4. How's the single log-out support coming?

Here's my config at the moment:
```
saml-idp-metadata: /etc/guacamole/saml-idp-metadata.xml
saml-entity-id: https://guacamole.example/guacamole/api/tokens                      
saml-idp-url: https://ipsilon03.mctest.nl/idp/saml2/SSO/Redirect                           
saml-callback-url: https://guacamole.example/guacamole                              
saml-logout-url: https://ipsilon03.mctest.nl/idp/logout
```

necouchman:
Here are the answers...I think...
1. Where the certificate is placed will depend on whether you're configuring it with a Metadata file or with individual parameters.  If you have a metadata file, all of the certificates should be in the metadata file.  If you configure with individual parameters, then it'll be in the keystore file.
2. Would imagine this is the same as item 1.
3. The metadata file is generated by the IdP, so where you get it will depend on the IdP.  It's been a while since I did it with the CAS system, but I think there was a URL that was accessed that triggered the generation onto the filesystem.  Other IdPs probably have varying methods of generating it - I'd suggest consulting the documentation for the IdP.  I also found a few online metadata generators that will do most of the work for you if the IdP doesn't do it.
4. Ummmmmmmmm...slow.  Haven't had much time to work on this, sadly - what's even sadder (more sad?) is that I haven't had much time to work on anything Guacamole-related lately :frowning_face:

Tyler-:
Hi,  I pulled this, and was attempting to test the SAML auth extension.

everything seems to be building / deploying fine.  but the issue I have is with setting the configuration.

I specified the idp xml file : 

saml-idp-metadata: /etc/guacamole/sample-saml.xml

I'm running with tomcat 7,  and regardless of how I specify the saml-idp-metadata, or where I put the .xml file,  i always get the same error : 

ERROR c.o.saml2.settings.IdPMetadataParser - XML file'/etc/guacamole/sample-saml.xml' cannot be loaded.XML file '/etc/guacamole/sample-saml.xml' not found in the classpath


I've tried specifying it in relative / absolute paths

I've tried putting it in /etc/guacamole, $TOMCAT_HOME, $TOMCAT_BASE, $TOMCAT_HOME/lib, $TOMCAT_BASE/lib,  the resources folder in the source code to get into the .jar extension.  I added /etc/guacamole to the catalina properties common.loader, and shared.loader

if anyone has any ideas, I'd love to hear them.

thanks

spaceone:
Adding myself to the CC list.

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/SAMLAuthenticationProviderResource.java
necouchman: Okay - I changed this around and tried another approach.  I haven't tested, yet, to make sure it works, but it should at least take care of the need to bump up header sizes.

necouchman:
Okay, quite a bit of refactoring done, here, and I've tested successfully against the CAS implementation of SAML.

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/SAMLAuthenticationProviderResource.java
```diff
@@ -0,0 +1,156 @@
...
+        catch (IOException
+                | NoSuchAlgorithmException
+                | ParserConfigurationException
+                | SAXException
+                | SettingsException
+                | URISyntaxException
+                | ValidationError
+                | XPathExpressionException e) {
+            throw new GuacamoleServerException(e);
+        }
```
mike-jumper, line 129: The `GuacamoleServerException` thrown here should have an associated message describing the nature of the failure in context, ideally coupled with the usual logging. I'm guessing that part of the reason there isn't a message here is the breadth of this multicatch, which may need to be slit up a bit to allow useful log messages to be provided in the case of failures.

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/SAMLResponseMap.java
```diff
@@ -0,0 +1,80 @@
...
+     * provided hash, or null if no such object exists.
+     * 
+     * @param hash
+     *     The SHA-256 hash of the SamlResponse.
+     * 
+     * @return 
+     *     The SamlResponse object matching the hash provided.
+     */
+    protected SamlResponse getSamlResponse(String hash) {
+        return samlResponseMap.remove(hash);
```
mike-jumper, line 51: What if `getSamlResponse()` is not invoked for a particular hash, and the size of `samlResponseMap` builds? Is there an overall timeout that should be enforced?

extensions/guacamole-auth-saml/pom.xml
```diff
@@ -0,0 +1,260 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0
+                        http://maven.apache.org/maven-v4_0_0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>org.apache.guacamole</groupId>
+    <artifactId>guacamole-auth-saml</artifactId>
+    <packaging>jar</packaging>
+    <version>1.0.0</version>
```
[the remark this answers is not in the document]
mike-jumper: Whoops - looks like this has already been updated.

extensions/guacamole-auth-saml/src/main/resources/controllers/samlController.js
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Controller for the "GUAC_SAML_REDIRECT" field, which redirects
+ * to the provided URI.
+ */
+angular.module('guacSAML').controller('guacSAMLController', ['$scope','$window',
+    function guacSAMLController($scope,$window) {
+
+    $window.location.href = $scope.field.samlRedirect;
+
+}]);
```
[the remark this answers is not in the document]
necouchman: Yeah, that sounds like a good idea, I'll take a stab at consolidating those.

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/AuthenticationProviderService.java
```diff
@@ -0,0 +1,208 @@
...
+                catch (XPathExpressionException e) {
+                    logger.warn("Problem with XML parsing response: {}", e.getMessage());
+                    logger.debug("Received XPathExpressionException while processing SAML response.", e);
+                    throw new GuacamoleInvalidCredentialsException("Error during SAML login.",
+                                CredentialsInfo.USERNAME_PASSWORD);
+                }
+                catch (Exception e) {
```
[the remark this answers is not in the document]
necouchman: Uh, yeah...the Exception handling in this particular SAML library isn't very uniform - kind of all over the place.  Hence this plus the huge multi-catch you noted below.  I actually put in a Pull Request for that project to rework the handling a bit and come up with a common `SamlException` class, but they haven't done anything with it, yet.

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/SAMLAuthenticationProviderResource.java
necouchman: Yeah, same with catching the high-level `Exception` above, there are so many possible errors here that I just lumped them all together.  I can try to break it out a bit and at least come up with generic log messages for each of the broad exceptions.

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/SAMLResponseMap.java
necouchman: Yeah, that's probably worthwhile to implement.  Are you thinking implementing a `Runnable` and periodically cleaning the map?

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/conf/ConfigurationService.java
```diff
@@ -0,0 +1,236 @@
...
+        else {
+            samlMap = new HashMap<>();
+            samlMap.put("onelogin.saml2.sp.entityid", getEntityId().toString());
+            samlMap.put("onelogin.saml2.sp.assertion_consumer_service.url",
+                    getCallbackUrl().toString() + "/api/ext/saml/callback");
+            samlMap.put("onelogin.saml2.idp.entityid", getIdpUrl().toString());
+            samlMap.put("onelogin.saml2.idp.single_sign_on_service.url",
+                    getIdpUrl().toString());
+            samlMap.put("onelogin.saml2.idp.single_sign_on_sevice.binding",
+                    "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect");
```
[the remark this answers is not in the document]
necouchman: Ah, yes, makes sense to use those.

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/SAMLAuthenticationProviderResource.java
necouchman: Hmmm...I think this code may also have been reviewed from a previous version, because the current code already breaks that out into individual `catch()` statements.

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/conf/ConfigurationService.java
```diff
@@ -0,0 +1,236 @@
...
+        }
+        
+        SettingsBuilder samlBuilder = new SettingsBuilder();
+        Saml2Settings samlSettings = samlBuilder.fromValues(samlMap).build();
+        samlSettings.setDebug(true);
```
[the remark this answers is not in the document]
necouchman: Probably leftover from my testing.  Probably worth making this configurable...

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/SAMLResponseMap.java
mike-jumper: Probably, yeah - I'd think the usual pattern of a `ScheduledThreadPoolExecutor`.

extensions/guacamole-auth-saml/src/main/resources/controllers/samlController.js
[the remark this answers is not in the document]
necouchman: Okay, I've opened a separate PR (#530) for the common redirect field implementation.  If/when that is merged I'll pull those changes out of this one and rebase this one on top, and then it should be ready for the final tweaks.

necouchman:
Well, that was fun. Should all be rebased on the current `staging/1.2.0` branch with the new field.  Going to do some testing and review of the code to make sure it's all good, because that rebase was _no fun_.  Lesson learned.

mike-jumper:
What happened that made this rebase particularly hairy?

necouchman:
> What happened that made this rebase particularly hairy?

The changes to implement the common `RedirectField` class were intertwined at various stages along the way throughout the rest of the changes.  In retrospect, I should have developed that separately from the other changes and then rebased.

Also, this has been going on for 18 months or so, now, I think - maybe 2 years at this point - so there were lots of version updates and license changes along the way.  Extracting the `RedirectField` and squashing down the other changes was interesting :-).

mike-jumper:
Code-Refactoring

necouchman:
Yeah, pretty much.

extensions/guacamole-auth-saml/src/licenses/LICENSE
```diff
@@ -0,0 +1,257 @@
...
+   limitations under the License.
+
+
+==============================================================================
+
+APACHE GUACAMOLE SUBCOMPONENTS
+
+Apache Guacamole includes a number of subcomponents with separate copyright
+notices and license terms. Your use of these subcomponents is subject to the
+terms and conditions of the following licenses.
```
mike-jumper, line 211: I think some licenses are missing here and within `src/licenses/bundled/`. Checking with Maven:

```
[mjumper@dev-mjumper guacamole-auth-saml]$ mvn dependency:tree -Dscope=runtime
[INFO] Scanning for projects...
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] Building guacamole-auth-saml 1.2.0
[INFO] ------------------------------------------------------------------------
[INFO] 
[INFO] --- maven-dependency-plugin:2.10:tree (default-cli) @ guacamole-auth-saml ---
[INFO] org.apache.guacamole:guacamole-auth-saml:jar:1.2.0
[INFO] +- com.google.inject:guice:jar:3.0:compile
[INFO] |  +- javax.inject:javax.inject:jar:1:compile
[INFO] |  \- aopalliance:aopalliance:jar:1.0:compile
[INFO] +- com.google.inject.extensions:guice-multibindings:jar:3.0:compile
[INFO] \- com.onelogin:java-saml:jar:2.5.0:compile
[INFO]    +- com.onelogin:java-saml-core:jar:2.5.0:compile
[INFO]    +- joda-time:joda-time:jar:2.10.3:compile
[INFO]    +- org.apache.commons:commons-lang3:jar:3.4:compile
[INFO]    +- org.apache.santuario:xmlsec:jar:2.1.4:compile
[INFO]    |  \- com.fasterxml.woodstox:woodstox-core:jar:5.0.3:compile
[INFO]    |     \- org.codehaus.woodstox:stax2-api:jar:3.1.4:compile
[INFO]    \- commons-codec:commons-codec:jar:1.12:compile
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 3.839 s
[INFO] Finished at: 2020-06-23T14:00:00-07:00
[INFO] Final Memory: 14M/174M
[INFO] ------------------------------------------------------------------------
[mjumper@dev-mjumper guacamole-auth-saml]$ 
```

So, looks like Apache Commons Codec, Apache Commons Lang, Apache Santuario, Joda-Time, and Woodstox?

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/SAMLResponseMap.java
```diff
@@ -0,0 +1,123 @@
...
+        public void run() {
+
+            // Loop through responses in map and remove ones that are no longer valid.
+            for (Entry<String, SamlResponse> entry : samlResponseMap.entrySet()) {
+                try {
+                    entry.getValue().validateTimestamps();
+                }
+                catch (ValidationError e) {
+                    samlResponseMap.remove(entry.getKey());
+                }
```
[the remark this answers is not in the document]
necouchman: Ick.  Apparently `validateTimestamps()` never returns false - it either returns true, or throws a `ValidationException`.

extensions/guacamole-auth-saml/src/main/resources/translations/en.json
```diff
@@ -0,0 +1,12 @@
+{
+
+    "DATA_SOURCE_SAML" : {
+        "NAME" : "SAML Authentication Extension"
+    },
+
+    "LOGIN" : {
+        "FIELD_HEADER_SAML"     : "",
+        "INFO_SAML_REDIRECT_PENDING" : "Please wait, redirecting for SAML authentication..."
```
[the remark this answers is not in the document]
necouchman: I'm good with more generic language.

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/SAMLResponseMap.java
```diff
@@ -0,0 +1,123 @@
...
+        @Override
+        public void run() {
+
+            // Loop through responses in map and remove ones that are no longer valid.
+            for (Entry<String, SamlResponse> entry : samlResponseMap.entrySet()) {
+                try {
+                    entry.getValue().validateTimestamps();
+                }
+                catch (ValidationError e) {
+                    samlResponseMap.remove(entry.getKey());
```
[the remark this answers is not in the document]
necouchman: Okay, I _think_ I've done this correctly with values per your suggestion - tried to keep it as simple as possible.  Thanks for the info on the behavior of removing from `values()` - that was helpful.

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/SAMLResponseMap.java
```diff
@@ -0,0 +1,123 @@
...
+     * SAML responses.
+     */
+    private final ConcurrentMap<String, SamlResponse> samlResponseMap =
+        new ConcurrentHashMap<>();
+    
+    /**
+     * Executor service which runs the periodic cleanup task
+     */
+    private final ScheduledExecutorService executor =
+            Executors.newScheduledThreadPool(1);
```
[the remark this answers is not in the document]
necouchman: I believe I have implemented this as needed.

extensions/guacamole-auth-saml/src/licenses/LICENSE
necouchman: Okay, I think I got everything added properly - both to overall license file and the bundled directory.

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/AuthenticationProviderService.java
```diff
@@ -0,0 +1,289 @@
...
+     *     tokens.
+     */
+    private Map<String, String> parseTokens(Map<String,
+            List<String>> attributes) {
+        
+        Map<String, String> tokens = new HashMap<>();
+        for (Entry<String, List<String>> entry : attributes.entrySet()) {
+            
+            List<String> values = entry.getValue();
+            tokens.put(entry.getKey(), values.get(0));
```
mike-jumper, line 257: Is it OK that these tokens will have identical names to their attributes within SAML? Or should the names be canonicalized, prefixed, etc. with http://guacamole.apache.org/doc/guacamole-ext/org/apache/guacamole/token/TokenName.html#canonicalize-java.lang.String-java.lang.String-?

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/AuthenticationProviderService.java
```diff
@@ -0,0 +1,289 @@
...
+            if (responseHash != null) {
+
+                try {
+
+                    // Generate the response object
+                    if (!samlResponseMap.hasSamlResponse(responseHash)) {
+                        logger.warn("SAML response was not found.");
+                        logger.debug("SAML response hash {} not fonud in response map.", responseHash);
+                        throw new GuacamoleInvalidCredentialsException("Provided response was not found.",
+                                CredentialsInfo.USERNAME_PASSWORD);
```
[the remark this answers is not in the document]
necouchman: I'm fine with a generic `GuacamoleServerException`, instead.

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/AuthenticationProviderService.java
necouchman: They should definitely be canonicalized with a standard prefix.

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/SAMLResponseMap.java
```diff
@@ -0,0 +1,133 @@
...
+            // Loop through responses in map and remove ones that are no longer valid.
+            Collection<SamlResponse> samlResponses = samlResponseMap.values();
+            for (SamlResponse value : samlResponses) {
+                try {
+                    value.validateTimestamps();
+                }
+                catch (ValidationError e) {
+                    samlResponses.remove(value);
+                }
+            }
```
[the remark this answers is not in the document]
necouchman: Okay, I think I have this done correctly, now.

necouchman:
One additional thing I'm seeing in the SAML client - when the IdP redirects back to the Guacamole Client and the login completes successfully, the `responseHash` URL parameter is still on the URL.  As long as the user navigates to another page this isn't a problem, as the navigation removes the parameter; however, if the user reloads the page or tries to go to logout, an error is received because it tries to process the `responseHash` parameter and cannot find it in the `SAMLResponseMap`.

Is there something I should do (similar to the OpenID extension, maybe?) to remove this after a successful login?

mike-jumper:
I don't think the OpenID extension removes this, either, but it handles the presence of an invalid token identically to the total absence of a token. You could do the same - log at the info or debug level that an invalid hash was received and then move forward requesting a new SAML response?

necouchman:
Ah, okay, yes - I've tweaked it so that it should behave that way - I'll verify when I get a moment to recompile and test.

necouchman:
Okay, redirect appears to be working okay, now.  One more issue I need to fix, though - when you try to specify the `saml-idp-metadata` option with a metadata file it throws an error:
```
ERROR c.o.saml2.settings.IdPMetadataParser - XML file'/etc/guacamole/saml-idp-metadata.xml' cannot be loaded.XML file '/etc/guacamole/saml-idp-metadata.xml' not found in the classpath
```
I guess whatever method I've used to load the file doesn't accept absolutely paths, or requires a URI or something like that.

necouchman:
Okay, I switched the `saml-idp-metadata` parameter to be a `URIGuacamoleProperty` instead of `FileGuacamoleProperty` and this seems to have done the trick.  It makes configuration a little bit nuanced, but actually provides the flexibility to have it pull metadata from either a local file (via a `file://` URI) or a remote URI.

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/conf/ConfigurationService.java
```diff
@@ -0,0 +1,351 @@
...
+
+    /**
+     * The URI of the file containing the XML Metadata associated with the
+     * SAML IdP.
+     */
+    private static final URIGuacamoleProperty SAML_IDP_METADATA =
+            new URIGuacamoleProperty() {
+
+        @Override
+        public String getName() { return "saml-idp-metadata"; }
```
mike-jumper, line 51: As this is now going to be a URL, I recommend renaming to `saml-idp-metadata-url`.

necouchman: Fine with me.

necouchman:
I've tested successfully with the CAS SAML provider, both with a metadata file (using a file:// URI) and with manually specifying the parameter for accessing the server.  If I can find another SAML provider or two to test with I will try to get that done, but I don't have ready access to any aside from CAS.

puneetshukla-siemens:
@necouchman Hello can you provide information how to provide connection to user logged in with saml authentication?

necouchman:
@puneetshukla-siemens: This is a closed pull request, and is not intended to be a support forum.  Please subscribe to the mailing list and ask your question, there:

JPedroGon:
I tried with OKTA pulling the metadata from a file. Problem I get is an endless loop between OKTA and the url callback. Somehow  after OKTA authenticates successfully, Guacamole is not capturing that.

necouchman:
@JPedroGon: Please post your questions to the mailing list.  This is a closed/merged pull request and is not designed to be a support forum.

PR #255, opened 2018-02-19
GUACAMOLE-508: Take user to home page when Settings pages are available.

necouchman:
Took a stab at an implementation of the improvement for 508, taking user to home screen when Settings pages are available.  It's worth noting that the "Preferences" page for users is pretty much always available, so this only does it if there is more than one Settings page available.

Not sure it's the most efficient route...open to suggestions.

### DIFF
--- a/guacamole/src/main/webapp/app/navigation/services/userPageService.js
+++ b/guacamole/src/main/webapp/app/navigation/services/userPageService.js
@@ -60,9 +60,13 @@ angular.module('navigation').factory('userPageService', ['$injector',
      * @returns {PageDefinition}
      *     The user's home page.
      */
-    var generateHomePage = function generateHomePage(rootGroups) {
+    var generateHomePage = function generateHomePage(rootGroups,permissions) {
 
         var homePage = null;
+        var settingsPages = generateSettingsPages(permissions);
+
+        if (settingsPages.length > 1)
+            return SYSTEM_HOME_PAGE;
 
         // Determine whether a connection or balancing group should serve as
         // the home page
@@ -140,13 +144,20 @@ angular.module('navigation').factory('userPageService', ['$injector',
         var deferred = $q.defer();
 
         // Resolve promise using home page derived from root connection groups
-        dataSourceService.apply(
+        var getRootGroups = dataSourceService.apply(
             connectionGroupService.getConnectionGroupTree,
             authenticationService.getAvailableDataSources(),
             ConnectionGroup.ROOT_IDENTIFIER
-        )
-        .then(function rootConnectionGroupsRetrieved(rootGroups) {
-            deferred.resolve(generateHomePage(rootGroups));
+        );
+        var getPermissionSets = dataSourceService.apply(
+            permissionService.getPermissions,
+            authenticationService.getAvailableDataSources(),
+            authenticationService.getCurrentUsername()
+        );
+
+        $q.all([getRootGroups,getPermissionSets])
+        .then(function rootConnectionGroupsPermissionsRetrieved(data) {
+            deferred.resolve(generateHomePage(data[0],data[1]));
         });
 
         return deferred.promise;
@@ -342,7 +353,7 @@ angular.module('navigation').factory('userPageService', ['$injector',
         var pages = [];
 
         // Get home page and settings pages
-        var homePage = generateHomePage(rootGroups);
+        var homePage = generateHomePage(rootGroups, permissions);
         var settingsPages = generateSettingsPages(permissions);
 
         // Only include the home page in the list of main pages if the user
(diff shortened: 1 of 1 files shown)


guacamole/src/main/webapp/app/navigation/services/userPageService.js
```diff
@@ -60,9 +60,13 @@ angular.module('navigation').factory('userPageService', ['$injector',
      * @returns {PageDefinition}
      *     The user's home page.
      */
-    var generateHomePage = function generateHomePage(rootGroups) {
+    var generateHomePage = function generateHomePage(rootGroups,permissions) {
```
mike-jumper: Please include a space following the comma before the `permissions` parameter.

mike-jumper: The newly-added `permissions` parameter needs to be documented.

guacamole/src/main/webapp/app/navigation/services/userPageService.js
```diff
@@ -60,9 +60,13 @@ angular.module('navigation').factory('userPageService', ['$injector',
...
 
         var homePage = null;
+        var settingsPages = generateSettingsPages(permissions);
+
+        if (settingsPages.length > 1)
```
mike-jumper: I'm generally OK with this check (and honestly can't think of a better way), but please include a comment to document what's going on here. The need for this check is not inherently clear.

guacamole/src/main/webapp/app/navigation/services/userPageService.js
```diff
@@ -140,13 +144,20 @@ angular.module('navigation').factory('userPageService', ['$injector',
...
+        );
+        var getPermissionSets = dataSourceService.apply(
+            permissionService.getPermissions,
+            authenticationService.getAvailableDataSources(),
+            authenticationService.getCurrentUsername()
+        );
+
+        $q.all([getRootGroups,getPermissionSets])
+        .then(function rootConnectionGroupsPermissionsRetrieved(data) {
+            deferred.resolve(generateHomePage(data[0],data[1]));
```
mike-jumper: According to [the AngularJS documentation for `$q.all()`](https://code.angularjs.org/1.3.16/docs/api/ng/service/$q#all), the function can accept either an array (in which case `data` will be a parallel array, as you use here) or an object (in which case `data` will be an object with identical keys pointing to the results). For the sake of readability, I suggest switching to the latter.

guacamole/src/main/webapp/app/navigation/services/userPageService.js
necouchman: Commented.

guacamole/src/main/webapp/app/navigation/services/userPageService.js
necouchman: Styled and documented.

guacamole/src/main/webapp/app/navigation/services/userPageService.js
necouchman: Objectified.

necouchman:
Okay, fixed up things as requested.

guacamole/src/main/webapp/app/navigation/services/userPageService.js
```diff
@@ -57,12 +57,24 @@ angular.module('navigation').factory('userPageService', ['$injector',
      *     A map of all root connection groups visible to the current user,
      *     where each key is the identifier of the corresponding data source.
      *
+     * @param {Object.<String, PermissionSet>} permissions
+     *     A map of all permissions granted to the current user, where each
+     *     key is the identifier of the corresponding data source.
+     *
+     * @param {Object.<String,
```
mike-jumper: Looks like some copypasta snuck in here.

guacamole/src/main/webapp/app/navigation/services/userPageService.js
```diff
@@ -57,12 +57,24 @@ angular.module('navigation').factory('userPageService', ['$injector',
...
      * @returns {PageDefinition}
      *     The user's home page.
      */
-    var generateHomePage = function generateHomePage(rootGroups) {
+    var generateHomePage = function generateHomePage(rootGroups, permissions) {
 
         var homePage = null;
+        var settingsPages = generateSettingsPages(permissions);
+
+        // If we have more than one setting page, return the main home page
```
mike-jumper: Better, but the critical fact we want to highlight here is that _more than one settings page indicates that the user has administrative privileges_, and thus redirecting to a single connection would hurt the user experience. The fact that this comparison checks for more than one page is clear from the `> 1`, but the meaning behind that check is non-obvious.

guacamole/src/main/webapp/app/navigation/services/userPageService.js
necouchman: Ha.  I started to type it myself, then copied it from the function that calls this one and forgot to remove the copy I started to type.  Like any good pasta, I've cleaned it up.

guacamole/src/main/webapp/app/navigation/services/userPageService.js
necouchman: Okay, added some more verbiage there.

PR #257, opened 2018-02-23
GUACAMOLE-445: Add client-side changes for configuring RDP Printer Name

necouchman:
This adds the client-side changes that correspond to the changes in guacamole-server to configure the name of the printer passed through in RDP connections.


necouchman:
Merge based change to staging/1.0.0

PR #258, opened 2018-02-23
GUACAMOLE-446: Allow redirected drive name to be configured

necouchman:
Client-side changes to allow the drive name to be configured when redirecting a filesystem using RDP device redirection.


necouchman:
Merge base changed to staging/1.0.0

PR #259, opened 2018-02-26
GUACAMOLE-269: Add fields for controlling Backspace behavior.

necouchman:
Client-side changes to allow the backspace key to be configured for SSH and Telnet sessions.

### DIFF
--- a/guacamole/src/main/webapp/translations/en.json
+++ b/guacamole/src/main/webapp/translations/en.json
@@ -423,8 +423,9 @@
 
     "PROTOCOL_SSH" : {
 
+        "FIELD_HEADER_BACKSPACE"    : "Backspace key sends:",
         "FIELD_HEADER_COLOR_SCHEME" : "Color scheme:",
-        "FIELD_HEADER_COMMAND"     : "Execute command:",
+        "FIELD_HEADER_COMMAND"      : "Execute command:",
         "FIELD_HEADER_CREATE_RECORDING_PATH" : "Automatically create recording path:",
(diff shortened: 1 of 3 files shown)


necouchman:
@mike-jumper Not trying to bug, but just a reminder that this PR goes with the recently-merged changes on the server side for backspace key config.

guacamole/src/main/webapp/translations/en.json
```diff
@@ -447,6 +448,9 @@
         "FIELD_HEADER_TYPESCRIPT_NAME" : "Typescript name:",
         "FIELD_HEADER_TYPESCRIPT_PATH" : "Typescript path:",
 
+        "FIELD_OPTION_BACKSPACE_8"    : "Backspace (Ctrl-H)",
+        "FIELD_OPTION_BACKSPACE_127"  : "Delete (Ctrl-?)",
```
mike-jumper: These options are swapped compared to the telnet options. Which is correct?

necouchman: Fixed.  SSH is correct - ASCII 8 is "Backspace", ASCII 127 is Del.

http://www.physics.udel.edu/~watson/scen103/ascii.html

mike-jumper: You'll also need an entry for the `""` value (`FIELD_OPTION_BACKSPACE_EMPTY`).

necouchman: Added.

PR #261, opened 2018-03-04
GUACAMOLE-516: On screen keyboard - added method to reset all pressed keys

jamhall:
On screen keyboard - added method to reset all pressed keys
https://issues.apache.org/jira/browse/GUACAMOLE-516

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/OnScreenKeyboard.js
+++ b/guacamole-common-js/src/main/webapp/modules/OnScreenKeyboard.js
@@ -456,6 +456,17 @@ Guacamole.OnScreenKeyboard = function(layout) {
     };
 
     /**
+     * Resets the state of this keyboard, releasing all keys, and firing keyup
+     * events for each released key.
+     */
+    this.reset = function() {
+        for (var keysym in pressed) {
+            var key = getActiveKey(keysym);
+            osk.onkeyup(key.keysym);
+        }
+    }
+
+    /**
      * Given the name of a key and its corresponding definition, which may be
      * an array of keys objects, a number (keysym), a string (key title), or a
      * single key object, returns an array of key objects, deriving any missing


guacamole-common-js/src/main/webapp/modules/OnScreenKeyboard.js
```diff
@@ -455,6 +455,17 @@ Guacamole.OnScreenKeyboard = function(layout) {
 
     };
 
+    /**
+     * Resets the state of this keyboard, releasing all keys, and firing keyup
+     * events for each released key.
+     */
+    this.reset = function() {
+        for (var keysym in pressed) {
+            var key = getActiveKey(keysym);
```
mike-jumper: Rather than manually pull the active key, and manually invoke `onkeyup`, it would be better to invoke `release()`:

https://github.com/apache/guacamole-client/blob/5db2e3cae75a658b33515112a39eb2f5f8eef4a8/guacamole-common-js/src/main/webapp/modules/OnScreenKeyboard.js#L346-L377

The `release()` function takes care of invoking `onkeyup` (if it's defined) as well as proper updating of `pressed`.

Directly calling `onkeyup` like this is problematic because:

1. The `pressed` structure tracks all currently-pressed keys, but this function does not update that structure. After manually invoking `onkeyup`, the `pressed` structure will be out of sync with the actual key state.
2. Usages of `Guacamole.OnScreenKeyboard` are not required to provide a function `onkeyup`. If that function is omitted, `reset()` as written here will fail.

guacamole-common-js/src/main/webapp/modules/OnScreenKeyboard.js
```diff
@@ -455,6 +455,17 @@ Guacamole.OnScreenKeyboard = function(layout) {
 
     };
 
+    /**
+     * Resets the state of this keyboard, releasing all keys, and firing keyup
+     * events for each released key.
+     */
+    this.reset = function() {
+        for (var keysym in pressed) {
```
mike-jumper: The `pressed` structure within `Guacamole.OnScreenKeyboard` is indexed by key name, not keysym:

https://github.com/apache/guacamole-client/blob/5db2e3cae75a658b33515112a39eb2f5f8eef4a8/guacamole-common-js/src/main/webapp/modules/OnScreenKeyboard.js#L51-L59

guacamole-common-js/src/main/webapp/modules/OnScreenKeyboard.js
```diff
@@ -455,6 +455,17 @@ Guacamole.OnScreenKeyboard = function(layout) {
 
     };
 
+    /**
+     * Resets the state of this keyboard, releasing all keys, and firing keyup
+     * events for each released key.
+     */
+    this.reset = function() {
```
mike-jumper: Though historical code has not followed this, please define a name for the function:

    this.reset = function reset() {
        ...
    };

This is being done for all new code as doing otherwise results in unreadable stack traces.

guacamole-common-js/src/main/webapp/modules/OnScreenKeyboard.js
```diff
@@ -455,6 +455,17 @@ Guacamole.OnScreenKeyboard = function(layout) {
...
+        for (var keysym in pressed) {
+            var key = getActiveKey(keysym);
+            osk.onkeyup(key.keysym);
+        }
+    }
```
mike-jumper: Missing semicolon.

necouchman:
Ping @jamhall - any chance you can finish this up?

PR #264, opened 2018-03-13
GUACAMOLE-525: Fix issues with SQL Server compatibility

necouchman:
Fixes a couple of issues reported with SQL Server compatibility and the lack of support for the LIMIT keyword.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/resources/org/apache/guacamole/auth/jdbc/user/PasswordRecordMapper.xml
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/resources/org/apache/guacamole/auth/jdbc/user/PasswordRecordMapper.xml
@@ -65,12 +65,11 @@
         );
 
         DELETE FROM [guacamole_user_password_history]
-        WHERE password_history_id IN (
-            SELECT password_history_id
+        WHERE password_history_id NOT IN (
+            SELECT TOP(#{maxHistorySize}) password_history_id
             FROM [guacamole_user_password_history]
             WHERE user_id = #{record.userID,jdbcType=INTEGER}
             ORDER BY password_date DESC
-            OFFSET #{maxHistorySize}
         );
 
     </insert>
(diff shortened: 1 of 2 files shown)


extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/resources/org/apache/guacamole/auth/jdbc/user/PasswordRecordMapper.xml
```diff
@@ -66,12 +65,11 @@
         );
 
         DELETE FROM [guacamole_user_password_history]
-        WHERE password_history_id IN (
-            SELECT password_history_id
+        WHERE password_history_id NOT IN (
```
mike-jumper: Won't this now delete password history for all other users?

mike-jumper: I think I see what you're trying to do here now. Maybe needs an `AND user_id = #{record.userID,jdbcType=INTEGER}` at the level of the outer select?

necouchman: Ummm...yes.  Probably not the behavior we want.

necouchman: > I think I see what you're trying to do here now. Maybe needs an AND user_id = #{record.userID,jdbcType=INTEGER} at the level of the outer select?

Yep!  Commit coming right up...

necouchman: Should be fixed.

necouchman:
@mike-jumper Anything else that needs fixing here?

mike-jumper:
Looks good to me. Will merge shortly.

PR #265, opened 2018-03-14
GUACAMOLE-526: Update webapp to angular 1.6.9.

jmuehlner:


guacamole/pom.xml
```diff
@@ -388,9 +388,14 @@
         <dependency>
             <groupId>org.webjars.bower</groupId>
             <artifactId>angular-translate-loader-static-files</artifactId>
-            <version>2.8.0</version>
+            <version>2.16.0</version>
             <scope>runtime</scope>
         </dependency>
+        <dependency>
+            <groupId>org.webjars.bower</groupId>
+            <artifactId>angular-toArrayFilter</artifactId>
```
jmuehlner: The `orderBy` filter no longer supports objects, which necessitates transforming any object into an array before using it - that's what this new dependency is for.

mike-jumper: The licenses of bundled dependencies will need to be updated (`guacamole/src/licenses`).

mike-jumper: Same for the bundled dependencies which are already documented but whose versions have changed:

https://github.com/apache/guacamole-client/blob/a6579ce4e1c70ea436fa1acc26a9ce2bb1e7fe84/guacamole/src/licenses/LICENSE#L214-L220

jmuehlner: Alright, updated.

mike-jumper:
What are these new "Possibly unhandled rejection" errors from AngularJS?

jmuehlner:
Ah yes, those. After further analysis, those are a new behavior of the way promise rejections are handled as of Angular 1.6, and are surprisingly difficult to work with. 

Additional thought will be required to figure out how to get Angular promises and Guacamole to be friends again.

mike-jumper:
Regarding promise rejections, though it is both nice and annoying that throwing something within a callback results in promise rejection with whatever was thrown (it makes rejection easy, but suddenly you can't be certain that your REST error handling will reject with the JSON error response or a JS error from some bug in a prior callback), maybe we can work around this with some sort of wrapper function.

For example, rather than:

    someService.getThings()
    .then(function gotThings(things) {
        // :)
    })
    .catch(function sorryNoThings(err) {
        // Is it a REST error? Is it a JS error?
    });

we could do something like:

    someService.getThings()
    .then(function gotThings(things) {
        // :)
    })
    .catch(restService.catchError(function sorryNoThings(err) {
        // Always only a REST error. JS errors automatically logged without invoking reject callback.
    }));

We could even provide convenience handling for REST error responses which are not errors in context of the particular request in question, such as some 404's:

    someService.getThings()
    .then(function gotThings(things) {
        // :)
    }).catch(restService.ignoreError(Error.Type.NOT_FOUND)); // Silently ignore given error type, logging all others

guacamole/src/main/webapp/app/index/filters/arrayFilter.js
```diff
@@ -0,0 +1,38 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * A filter for transforming an object into a an array of all non-inherited
```
mike-jumper: "a an array"

guacamole/src/main/webapp/app/index/filters/arrayFilter.js
```diff
@@ -0,0 +1,38 @@
...
+
+    return function toArrayFiter(input) {
+
+        // If no object is available, just return an empty array
+        if (!input) {
+            return [];
+        }
+
+        return Object.keys(input).map(function fetchValueByKey(key) {
+            return input[key]
```
mike-jumper: Missing semicolon.

guacamole/src/main/webapp/app/rest/services/requestService.js
```diff
@@ -0,0 +1,42 @@
...
+ * promises that pass only the data from that response.
+ */
+angular.module('rest').factory('requestService', ['$q', '$http', 'Error',
+        function requestService($q, $http, Error) {
+
+    function wrappedHttpCall(object) {
+        
+        var deferred = $q.defer();
+        
+        $http(object)
```
mike-jumper: Perhaps it would be cleaner to leverage the `return` / `throw` magic provided by promises? Something like:

```js
return $http(object).then(
    function success(request) { return request.data; },
    function failure(request) { throw new Error(request.data); }
);
```

would remove one layer from the promise onion and the need for manual use of `$q`.

guacamole/src/main/webapp/app/rest/services/requestService.js
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+/**
+ * Service for converting $http promises that pass the entire response into
+ * promises that pass only the data from that response.
+ */
+angular.module('rest').factory('requestService', ['$q', '$http', 'Error',
+        function requestService($q, $http, Error) {
+
+    function wrappedHttpCall(object) {
```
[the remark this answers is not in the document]
mike-jumper: Please also use the same

    var someFunc = function someFunc() {
        ...
    };

declaration style used elsewhere within the webapp and JS API.

mike-jumper:
Regarding the "Possibly unhandled rejection" warnings, I believe I have a set of commits which solve things. If we can address the other issues here (such that things are ready for merge and the only problem is log noise), I think we're go.

mike-jumper:
Specifically: https://github.com/jmuehlner/guacamole-client/compare/angular-1.6.9...mike-jumper:fix-angularjs-1.6-warn

guacamole/src/main/webapp/app/rest/services/requestService.js
jmuehlner: Good call. Updated and it looks much cleaner that way.

guacamole/src/main/webapp/app/index/filters/arrayFilter.js
jmuehlner: Fixed.

guacamole/src/main/webapp/app/index/filters/arrayFilter.js
jmuehlner: Fixed.

guacamole/src/main/webapp/app/rest/services/requestService.js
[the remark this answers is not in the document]
jmuehlner: Done.

jmuehlner:
@mike-jumper Handling those rejections with a wrapper sounds like a good idea, if you've got that working.

I think this PR should meet your criteria for merging now.

guacamole/pom.xml
```diff
@@ -324,19 +324,25 @@
         <dependency>
             <groupId>org.webjars.bower</groupId>
             <artifactId>angular</artifactId>
-            <version>1.3.16</version>
+            <version>1.6.9</version>
+            <scope>runtime</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.webjars.bower</groupId>
+            <artifactId>angular-cookies</artifactId>
```
mike-jumper: This dependency was removed via #273.

guacamole/src/main/webapp/index.html
```diff
@@ -61,15 +61,16 @@
         <script type="text/javascript" src="webjars/lodash/2.4.1/dist/lodash.min.js"></script>
 
         <!-- AngularJS -->
-        <script type="text/javascript" src="webjars/angular/1.3.16/angular.min.js"></script>
-        <script type="text/javascript" src="webjars/angular-route/1.3.16/angular-route.min.js"></script>
-        <script type="text/javascript" src="webjars/angular-touch/1.3.16/angular-touch.min.js"></script>
+        <script type="text/javascript" src="webjars/angular/1.6.9/angular.min.js"></script>
+        <script type="text/javascript" src="webjars/angular-cookies/1.6.9/angular-cookies.min.js"></script>
```
mike-jumper: Same here - this dependency was removed via #273.

mike-jumper:
Please also be sure to properly tag the JIRA issue in each commit message, and to not include merge commits within the PR (you'll probably need to rebase). Looking at the overall changes, the first commit looks great, but the commits that follow lack the JIRA issue and include one merge:

pull-265-commits

jmuehlner:
Ok, rebased and removed the `angular-cookies` dependency.

mike-jumper:
Looks great, but please tag the JIRA issue in the commit messages. The first commit (b3eeb36) tags the JIRA issue correctly, but the second (33f537a) does not.

jmuehlner:
Oops! Fixed.

PR #270, opened 2018-04-06
GUACAMOLE-527: Client-side changes for SSH host key checking

necouchman:
The client-side changes for passing through a known host key to guacd.


necouchman:
Merge base updated to staging/1.0.0

necouchman:
@mike-jumper - ping on this one since it is the matching change to the recently-merged apache/guacamole-server#164

guacamole-ext/src/main/resources/org/apache/guacamole/protocols/rdp.json
```diff
@@ -329,6 +329,10 @@
                     "name"  : "sftp-hostname",
                     "type"  : "TEXT"
                 },
+                {
+                    "name"  : "sftp-host-key",
```
mike-jumper: Is there a reason behind the difference in parameter order across protocols? I see `hostname`/`port`/**`host-key`** for SSH but `sftp-hostname`/**`sftp-host-key`**/`sftp-port` for VNC/RDP.

guacamole-ext/src/main/resources/org/apache/guacamole/protocols/ssh.json
```diff
@@ -12,6 +12,10 @@
                 {
                     "name"  : "port",
                     "type"  : "NUMERIC"
+                },
+                {
+                    "name"  : "host-key",
+                    "type"  : "MULTILINE"
```
mike-jumper: Does this value contain multiple lines? I thought each host was defined by a single line.

guacamole/src/main/webapp/translations/en.json
```diff
@@ -364,6 +364,7 @@
         "FIELD_HEADER_SECURITY"        : "Security mode:",
         "FIELD_HEADER_SERVER_LAYOUT"   : "Keyboard layout:",
         "FIELD_HEADER_SFTP_DIRECTORY"             : "Default upload directory:",
+        "FIELD_HEADER_SFTP_HOST_KEY"              : "SSH host key (Base64):",
```
mike-jumper: As the fact this is the key of the SSH host is given from the header for the section (SFTP), but the public vs. private nature of that key is not, I suggest using the same text for VNC/RDP that you use for the SSH parameter ("public host key").

guacamole-ext/src/main/resources/org/apache/guacamole/protocols/rdp.json
necouchman: No, just a mistake.  Reordered.

guacamole-ext/src/main/resources/org/apache/guacamole/protocols/ssh.json
necouchman: Yeah, I think you're right.

necouchman: Fixed.

guacamole/src/main/webapp/translations/en.json
necouchman: Yeah, good point.  Fixed.

PR #274, opened 2018-04-19
GUACAMOLE-220: Add base extension API support for user groups.

mike-jumper:
This change adds the base classes necessary for exposing user groups via extensions, and for declaring the permissions which apply to a user *but are not necessarily directly granted to that user*.

Overall, this involves:

* A new `UserGroup` interface, as well as the usual `DelegatingUserGroup`, and `Directory<UserGroup>` on `UserContext`.
* A new base interface, `Permissions`, which `User` and `UserGroup` share, defining the interface common to any object which can be granted permissions.
* A new `getEffectivePermissions()` function on `User` which returns a `Permissions` that represents all permissions which apply to the user, even if those permissions are inherited through some arbitrary means (such as user groups). The traditional `getUserPermissions()`, `getConnectionPermissions()`, etc. directly on the `User` return permission sets which describe permissions which are directly granted only.
* A new `getEffectiveUserGroups()` function on `AuthenticatedUser` which allows implementations to expose the identity of a user's group memberships in addition to the user's own identity. This forms the means by which membership in groups can be shared across extensions (for example, LDAP group membership can affect a user in MySQL so long as an identical MySQL group exists, even if that user does not exist in MySQL).
* A new `RelatedObjectSet` interface, similar to `PermissionSet`, which abstracts batch add/remove operations on a group of objects sharing some arbitrary relation, such as the member users of a user group, the parent user groups of a user, etc. This allows relations between specific objects to be established or removed without affecting the existence of those objects (as `Directory` would require) nor necessarily other relations to those objects.

The user groups themselves are exposed via a `Directory<UserGroup>` at the `UserContext` level, like all other objects. A new `ObjectPermissionSet` for user group permissions is exposed within `Permissions`. User membership within groups, group membership within groups, the set of groups containing a particular user, and the set of groups containing a particular group can all be maintained through the various getters returning `RelatedObjectSet` instances.

Other than (1) updating the REST services and JavaScript to use *effective* permissions rather than directly-granted permissions and (2) any stubs necessary to allow the database auth to continue working, no changes outside the API are made here.

Assuming these changes move forward, I will open further pull requests for the REST API changes, interface changes, and finally database auth changes.

### DIFF
--- a/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/AbstractAuthenticatedUser.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/AbstractAuthenticatedUser.java
@@ -22,7 +22,6 @@ package org.apache.guacamole.net.auth;
 import java.util.Collections;
 import java.util.Set;
 
-
 /**
  * Basic implementation of an AuthenticatedUser which uses the username to
  * determine equality. Username comparison is case-sensitive.
--- a/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/AbstractAuthenticatedUser.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/AbstractAuthenticatedUser.java
@@ -19,6 +19,9 @@
 
 package org.apache.guacamole.net.auth;
 
+import java.util.Collections;
+import java.util.Set;
+
 
 /**
  * Basic implementation of an AuthenticatedUser which uses the username to
@@ -28,6 +31,11 @@ public abstract class AbstractAuthenticatedUser extends AbstractIdentifiable
         implements AuthenticatedUser {
 
     // Prior functionality now resides within AbstractIdentifiable
+
+    @Override
+    public Set<String> getEffectiveUserGroups() {
+        return Collections.<String>emptySet();
+    }
 
     @Override
     public void invalidate() {
--- a/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/AuthenticatedUser.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/AuthenticatedUser.java
@@ -19,6 +19,7 @@
 
 package org.apache.guacamole.net.auth;
 
+import java.util.Set;
 
 /**
  * A user of the Guacamole web application who has been authenticated by an
@@ -48,6 +49,24 @@ public interface AuthenticatedUser extends Identifiable {
      *     The credentials provided by the user when they authenticated.
      */
     Credentials getCredentials();
+
+    /**
+     * Returns a read-only set of the identifiers of all user groups which
+     * apply to this authenticated user. The exact semantics of what user
+     * groups apply are up to the implementation, and the user groups within
+     * this set may be implied, derived dynamically, inherited through multiple
+     * levels of group membership, etc.
+     *
+     * Note that, as with user identifiers, user group identifiers form the
+     * basis of identity which applies across authentication providers. It is
+     * expected that any two user groups having the same identifier represent
+     * the same group, even if defined by different authentication providers
+     *
+     * @return
+     *     A read-only set of the identifiers of all user groups which apply
+     *     to this authenticated user.
+     */
+    Set<String> getEffectiveUserGroups();
 
     /**
      * Invalidates this authenticated user and their associated token such that
--- a/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/AuthenticatedUser.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/AuthenticatedUser.java
@@ -60,7 +60,7 @@ public interface AuthenticatedUser extends Identifiable {
      * Note that, as with user identifiers, user group identifiers form the
      * basis of identity which applies across authentication providers. It is
      * expected that any two user groups having the same identifier represent
-     * the same group, even if defined by different authentication providers
+     * the same group, even if defined by different authentication providers.
      *
      * @return
      *     A read-only set of the identifiers of all user groups which apply
--- a/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/User.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/User.java
@@ -122,7 +122,7 @@ public interface User extends Identifiable, Attributes, Permissions {
     /**
      * Returns a read-only view of all permissions granted to this user. The
      * exact semantics of what permissions are granted are up to the
-     * implementation, and the permissions within this view may implied,
+     * implementation, and the permissions within this view may be implied,
      * derived dynamically, inherited through multiple levels of group
      * membership, etc.
      *
--- a/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/User.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/User.java
@@ -22,14 +22,11 @@ package org.apache.guacamole.net.auth;
 import java.util.Date;
 import java.util.List;
 import org.apache.guacamole.GuacamoleException;
-import org.apache.guacamole.net.auth.permission.ObjectPermissionSet;
-import org.apache.guacamole.net.auth.permission.SystemPermissionSet;
-
 
 /**
  * A user of the Guacamole web application.
  */
-public interface User extends Identifiable, Attributes {
+public interface User extends Identifiable, Attributes, Permissions {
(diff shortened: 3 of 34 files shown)


guacamole-ext/src/main/java/org/apache/guacamole/net/auth/UserGroup.java
```diff
@@ -0,0 +1,76 @@
...
+     * such modifications should be made through changes to the returned set.
+     *
+     * @return
+     *     The set of all readable user groups that are members of this user
+     *     group, which may be mutable.
+     *
+     * @throws GuacamoleException
+     *     If an error occurs while retrieving the user groups.
+     */
+    RelatedObjectSet getMemberUserGroups() throws GuacamoleException;
```
necouchman: Is there a reason that `getUserGroups()` and `getMemberUserGroups()` are both necessary?  Based on the stubs and descriptions, here, they look identical?  I see that they're both implemented in `DelegatingUserGroup`, but they have identical implementations, there, and `getMemberUserGroups()` does not appear to be used anywhere outside of there?

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleRelatedObjectSet.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.Set;
+import org.apache.guacamole.GuacamoleException;
+import org.apache.guacamole.GuacamoleSecurityException;
+import org.apache.guacamole.net.auth.RelatedObjectSet;
+
+/**
+ * A read-only implementation of RelatedObjectSet which uses a backing Set
+ * of identifiers to determine which objects are present.
+ */
+public class SimpleRelatedObjectSet implements RelatedObjectSet {
```
necouchman: Do you think there's any value to implementing a `AbstractRelatedObjectSet` class like you've done for several of the other `Simple` classes, that implements a bunch of these default behaviors?  Perhaps not initially useful for the JDBC extension, but is it something that would be helpful as other extensions implement the `RelatedObjectSet` class without having to rely on the behavior of `SimpleRelatedObjectSet`?

guacamole/src/main/java/org/apache/guacamole/rest/connection/ConnectionResource.java
```diff
@@ -119,11 +120,13 @@ public ConnectionResource(@Assisted UserContext userContext,
     public Map<String, String> getConnectionParameters()
             throws GuacamoleException {
 
+        // Pull effective permissions
         User self = userContext.self();
+        Permissions effective = self.getEffectivePermissions();
```
necouchman: In the `ConnectionGroupTree` class you did:

    Permissions effective = userContext.self().getEffectivePermissions();

eliminating the need for the existence of the `self` object.  Any reason not to do the same, here?

guacamole/src/main/java/org/apache/guacamole/rest/directory/DirectoryResource.java
```diff
@@ -143,13 +144,14 @@ public DirectoryResource(UserContext userContext, Directory<InternalType> direct
 
         // An admin user has access to all objects
         User self = userContext.self();
-        SystemPermissionSet systemPermissions = self.getSystemPermissions();
+        Permissions effective = self.getEffectivePermissions();
```
necouchman: Same note as above - any reason not to do

    Permissions effective = userContext.self().getEffectivePermissions()

?

guacamole/src/main/java/org/apache/guacamole/rest/sharingprofile/SharingProfileResource.java
```diff
@@ -103,11 +104,13 @@ public SharingProfileResource(@Assisted UserContext userContext,
     public Map<String, String> getParameters()
             throws GuacamoleException {
 
+        // Pull effective permissions
         User self = userContext.self();
+        Permissions effective = self.getEffectivePermissions();
```
necouchman: As with previous places, guessing these two lines could be collapsed...

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/UserGroup.java
mike-jumper: `getUserGroups()` returns the user groups of which the current group is a member, while `getMemberUserGroups()` returns the user groups which are members of the current group.

necouchman: Okay, in that case, the comments above the code need to be updated.  The comment above `getUserGroups()` says:

> Returns a set of all readable user groups that are members of this user group.

and the comment above `getMemberUserGroups()` starts with:

> Returns a set of all readable user groups that are members of this user group.

Also, the `@return` comment seems to be identical for both of them?

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/AbstractAuthenticatedUser.java
```diff
@@ -19,6 +19,9 @@
 
 package org.apache.guacamole.net.auth;
 
+import java.util.Collections;
+import java.util.Set;
+
```
necouchman: Extra line added, here - seems like most files have a single space between `import` and class documentation.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/AuthenticatedUser.java
```diff
@@ -49,6 +50,24 @@
...
+     * Returns a read-only set of the identifiers of all user groups which
+     * apply to this authenticated user. The exact semantics of what user
+     * groups apply are up to the implementation, and the user groups within
+     * this set may be implied, derived dynamically, inherited through multiple
+     * levels of group membership, etc.
+     *
+     * Note that, as with user identifiers, user group identifiers form the
+     * basis of identity which applies across authentication providers. It is
+     * expected that any two user groups having the same identifier represent
+     * the same group, even if defined by different authentication providers
```
necouchman: Missing period at the end of this sentence.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/User.java
```diff
@@ -109,85 +106,33 @@
...
      */
-    ObjectPermissionSet getActiveConnectionPermissions()
-            throws GuacamoleException;
+    RelatedObjectSet getUserGroups() throws GuacamoleException;
 
     /**
-     * Returns all user permissions given to this user.
+     * Returns a read-only view of all permissions granted to this user. The
+     * exact semantics of what permissions are granted are up to the
+     * implementation, and the permissions within this view may implied,
```
necouchman: "may be implied"

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/UserGroup.java
mike-jumper: Oops.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleRelatedObjectSet.java
mike-jumper: In principle, yes, but I'm not sure what that would be. There are only two functions that are effectively unimplemented by `SimpleRelatedObjectSet`:

1. `addObjects()`
2. `removeObjects()`

both of which make changes to the `RelatedObjectSet` in a way that would be implementation dependent. It does make sense to provide default behaviors, but I don't think there are any sensible default behaviors we can provide here.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/UserGroup.java
mike-jumper: OK - this should now be corrected.

mike-jumper:
OK, I believe I've addressed the issues above, with the exception of adding a `AbstractRelatedObjectSet` (as I am unable to define a set of default behaviors that is not already provided by `SimpleRelatedObjectSet`).

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleRelatedObjectSet.java
necouchman: Make sense.

necouchman:
Cool - will try to review one last time tonight and get it merged.

PR #278, opened 2018-04-27
GUACAMOLE-526: Handle absolutely all promise rejections.

mike-jumper:
This change addresses the new "possibly unhandled rejection" error by adding handling for absolutely *all* promise rejections. Each rejection is handled through one of the following mechanisms:

* Ignored via `angular.noop` (if the promise is not REST-related and failure of the promise is expected)
* Ignored via `requestService.IGNORE` (if the promise is REST-related and failure is expected)
* Warning via `requestService.WARN` (if the promise is REST-related and failure of the promise is generally *not* expected, but such failures are better quietly logged to the console than displayed in a modal notification)
* Notification via `guacNotification.SHOW_REQUEST_ERROR` (if the promise is REST-related and any failure should be exposed to the user as a modal notification)
* Handled in some custom fashon via `requestService.createErrorCallback()` (if the promise is REST-related)

The `IGNORE`, `WARN`, and `SHOW_REQUEST_ERROR` constants mentioned above are pre-defined callbacks generated through `requestService.createErrorCallback()`. The `requestService.createErrorCallback()` function produces a new callback function which invokes a given callback *only of the rejection is an instance of the REST `Error` object*. This allows JavaScript failures within the handling of a promise to still result in an error logged to the console.

In addition to the above, I noticed the following issues during my regression testing which had to be addressed:

* A `console.log()` debug statement snuck into the new `toArray` filter.
* An old-style `success()` call (used by the old AngularJS `$http` service) within the password update function.

### DIFF
--- a/guacamole/src/main/webapp/app/rest/services/activeConnectionService.js
+++ b/guacamole/src/main/webapp/app/rest/services/activeConnectionService.js
@@ -25,7 +25,6 @@ angular.module('rest').factory('activeConnectionService', ['$injector',
 
     // Required services
     var requestService        = $injector.get('requestService');
-    var $q                    = $injector.get('$q');
     var authenticationService = $injector.get('authenticationService');
 
     var service = {};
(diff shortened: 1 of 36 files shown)


guacamole/src/main/webapp/app/rest/services/activeConnectionService.js
```diff
@@ -66,68 +65,6 @@ angular.module('rest').factory('activeConnectionService', ['$injector',
...
-     *     result.  If null, no filtering will be performed. Valid values are
-     *     listed within PermissionSet.ObjectType.
-     *
-     * @returns {Promise.<Object.<String, Object.<String, ActiveConnection>>>}
-     *     A promise which resolves with all active connections available to
-     *     the current user, as a map of ActiveConnection maps, as would be
-     *     returned by getActiveConnections(), grouped by the identifier of
-     *     their corresponding data source.
-     */
-    service.getAllActiveConnections = function getAllActiveConnections(dataSources, permissionTypes) {
```
necouchman: Is this method not used?

mike-jumper: Not used at all. I'm guessing it must have been added before `dataSourceService` was created, as the only place in the webapp where `getAllActiveConnections()` would have been called is using `dataSourceService` and `getActiveConnections()` instead:

https://github.com/apache/guacamole-client/blob/9a2e0c60879ce0b7dc14870cdb5fe686e2c22eef/guacamole/src/main/webapp/app/settings/directives/guacSettingsSessions.js#L224-L228

PR #280, opened 2018-04-30
GUACAMOLE-152: Allow zoom/scale to be manually entered.

necouchman:
This change allows the zoom level (display scale) to be manually entered in addition to using the -/+ controls in the menu so that additional zoom levels can be reached that aren't currently possible with the 0.10 (10%) scale stepping.

Not sure if this is the right approach or not, but thought I'd give it a try!

### DIFF
--- a/guacamole/src/main/webapp/app/client/templates/client.html
+++ b/guacamole/src/main/webapp/app/client/templates/client.html
@@ -151,9 +151,11 @@
                     <div class="content">
                         <div id="zoom-settings">
                             <div ng-click="zoomOut()" id="zoom-out"><img src="images/settings/zoom-out.png" alt="-"/></div>
-                            <input type="number" class="zoom-ctrl" guac-zoom-ctrl
-                                    ng-model="client.clientProperties.scale"
-                                    ng-model-options="{ updateOn: 'blur submit' }" />%
+                            <div class="zoom-ctrl">
+                                <input type="number" class="zoom-ctrl" guac-zoom-ctrl
+                                        ng-model="client.clientProperties.scale"
+                                        ng-model-options="{ updateOn: 'blur submit' }" />%
+                            </div>
                             <div ng-click="zoomIn()" id="zoom-in"><img src="images/settings/zoom-in.png" alt="+"/></div>
                         </div>
                         <div><label><input ng-model="menu.autoFit" ng-change="changeAutoFit()" ng-disabled="autoFitDisabled()" type="checkbox" id="auto-fit"/> {{'CLIENT.TEXT_ZOOM_AUTO_FIT' | translate}}</label></div>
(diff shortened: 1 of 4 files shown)


guacamole/src/main/webapp/app/client/templates/client.html
```diff
@@ -151,7 +151,7 @@ <h3>{{'CLIENT.SECTION_HEADER_DISPLAY' | translate}}</h3>
                     <div class="content">
                         <div id="zoom-settings">
                             <div ng-click="zoomOut()" id="zoom-out"><img src="images/settings/zoom-out.png" alt="-"/></div>
-                            <div id="zoom-state">{{formattedScale()}}%</div>
```
mike-jumper: If this is no longer necessary, then we should probably also delete:

https://github.com/apache/guacamole-client/blob/91f19b2e03cce32c964f1f06e810d84aa8b924d5/guacamole/src/main/webapp/app/client/controllers/clientController.js#L734-L736

mike-jumper:
Other than the point above, this looks generally fine to me. I'll pull this and deploy to test the UI changes.

guacamole/src/main/webapp/app/client/templates/client.html
necouchman: Yep, removed.

necouchman:
> Other than the point above, this looks generally fine to me. I'll pull this and deploy to test the UI changes.

I should probably point out that I had to add the `debounce` option to `ng-model-options` to get this to work correctly - it seems like if this option was not present, the attempt to translate the percent to a scale was so instantaneous that it would just default back to 1.0 and never allow you to actually type anything.  I have the debounce at 500ms at the moment - hopefully long enough to allow people to type but short enough that it won't frustrate users.

Anyway, not sure if there's a better way than the debounce delay to handle this??

mike-jumper:
It looks like we can control when exactly the model is updated with respect to changes to the field:

https://code.angularjs.org/1.6.9/docs/api/ng/directive/ngModelOptions

Perhaps restricting that to just "blur" and "submit" would do the trick?

necouchman:
> Perhaps restricting that to just "blur" and "submit" would do the trick?

Brilliant.  Should be good, now.

mike-jumper:
Retesting against the latest, I'm still seeing the issue where the "Automatically fit to browser window" checkbox remains checked despite the zoom level being manually set via the input field. This will result in the zoom resetting if the window is resized.

I do see that there were recent changes to the CSS, but I don't see those changes having any effect. From the latest:

zoom-control

Compared to before:

old-zoom-control

necouchman:
@mike-jumper Style has been tweaked to be much closer to what it was before adding the input box.

Working on the autoFit stuff.

necouchman:
@mike-jumper Okay, style issues and autoFit should be resolved, now.

guacamole/src/main/webapp/app/client/templates/client.html
```diff
@@ -151,7 +151,12 @@ <h3>{{'CLIENT.SECTION_HEADER_DISPLAY' | translate}}</h3>
...
+                            <div class="zoom-ctrl">
+                                <input type="number" class="zoom-ctrl" guac-zoom-ctrl
+                                        ng-model="client.clientProperties.scale"
+                                        ng-model-options="{ updateOn: 'blur submit' }"
+                                        ng-change="zoomSet();" />%
```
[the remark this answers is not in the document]
necouchman: Old (bad?) JavaScript habits. Removed.

PR #282, opened 2018-05-02
GUACAMOLE-220: Extract common base for management UI.

mike-jumper:
A necessary basis for the user group management interface (which adds yet another management tab), these changes extract the common base code used by the various existing management tabs, which up until now has been largely duplicated.

New directives have been added to take care of data source tabs, permission management, and the save/cancel/clone/delete buttons. The permission test logic driving those buttons and other parts of each management tab has been moved into common class, as well.

This is a net increase in code due to the amount of documentation added, and a massive net decrease in the amount of code within the management controllers.

### DIFF
--- a/guacamole/src/main/webapp/app/manage/controllers/manageConnectionController.js
+++ b/guacamole/src/main/webapp/app/manage/controllers/manageConnectionController.js
@@ -29,6 +29,7 @@ angular.module('manage').controller('manageConnectionController', ['$scope', '$i
     var HistoryEntryWrapper   = $injector.get('HistoryEntryWrapper');
     var ManagementPermissions = $injector.get('ManagementPermissions');
     var PermissionSet         = $injector.get('PermissionSet');
+    var Protocol              = $injector.get('Protocol');
 
     // Required services
     var $location                = $injector.get('$location');
@@ -41,7 +42,6 @@ angular.module('manage').controller('manageConnectionController', ['$scope', '$i
     var permissionService        = $injector.get('permissionService');
     var requestService           = $injector.get('requestService');
     var schemaService            = $injector.get('schemaService');
-    var translationStringService = $injector.get('translationStringService');
 
     /**
      * The unique identifier of the data source containing the connection being
@@ -295,51 +295,14 @@ angular.module('manage').controller('manageConnectionController', ['$scope', '$i
     }, angular.noop);
 
     /**
-     * Returns the translation string namespace for the protocol having the
-     * given name. The namespace will be of the form:
-     *
-     * <code>PROTOCOL_NAME</code>
-     *
-     * where <code>NAME</code> is the protocol name transformed via
-     * translationStringService.canonicalize().
-     *
-     * @param {String} protocolName
-     *     The name of the protocol.
-     *
-     * @returns {String}
-     *     The translation namespace for the protocol specified, or null if no
-     *     namespace could be generated.
+     * @borrows Protocol.getNamespace
      */
-    $scope.getNamespace = function getNamespace(protocolName) {
-
-        // Do not generate a namespace if no protocol is selected
-        if (!protocolName)
-            return null;
-
-        return 'PROTOCOL_' + translationStringService.canonicalize(protocolName);
-
-    };
+    $scope.getNamespace = Protocol.getNamespace;
 
     /**
-     * Given the internal name of a protocol, produces the translation string
-     * for the localized version of that protocol's name. The translation
-     * string will be of the form:
-     *
-     * <code>NAMESPACE.NAME<code>
-     *
-     * where <code>NAMESPACE</code> is the namespace generated from
-     * $scope.getNamespace().
-     *
-     * @param {String} protocolName
-     *     The name of the protocol.
-     * 
-     * @returns {String}
-     *     The translation string which produces the localized name of the
-     *     protocol specified.
+     * @borrows Protocol.getName
      */
-    $scope.getProtocolName = function getProtocolName(protocolName) {
-        return $scope.getNamespace(protocolName) + '.NAME';
-    };
+    $scope.getProtocolName = Protocol.getName;
 
     /**
      * Cancels all pending edits, returning to the main list of connections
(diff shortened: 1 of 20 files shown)


guacamole/src/main/webapp/app/manage/controllers/manageConnectionController.js
```diff
@@ -157,205 +137,178 @@ angular.module('manage').controller('manageConnectionController', ['$scope', '$i
...
-        connectionService.getConnectionHistory($scope.selectedDataSource, identifier)
-        .then(function historyReceived(historyEntries) {
+            $scope.connection = values.connection;
+            $scope.parameters = values.parameters;
 
             // Wrap all history entries for sake of display
             $scope.historyEntryWrappers = [];
-            historyEntries.forEach(function wrapHistoryEntry(historyEntry) {
-               $scope.historyEntryWrappers.push(new HistoryEntryWrapper(historyEntry)); 
+            angular.forEach(values.historyEntries, function wrapHistoryEntry(historyEntry) {
```
jmuehlner: Just curious - why do you prefer `angular.forEach` here instead of the built in array `forEach`?

guacamole/src/main/webapp/app/manage/templates/manageConnection.html
```diff
@@ -41,7 +41,7 @@ <h2>{{'MANAGE_CONNECTION.SECTION_HEADER_EDIT_CONNECTION' | translate}}</h2>
     <!-- Connection attributes section -->
     <div class="attributes">
         <guac-form namespace="'CONNECTION_ATTRIBUTES'" content="attributes"
-                   model="connection.attributes" model-only="!canChangeAllAttributes()"></guac-form>
+                   model="connection.attributes" model-only="managementPermissions.canChangeAllAttributes"></guac-form>
```
jmuehlner: Are you missing a `!` here?

guacamole/src/main/webapp/app/manage/controllers/manageConnectionController.js
mike-jumper: I suppose it's mainly out of habit here. Since `historyEntries` should always be defined, either would work here. Any browsers which would lack `Array.prototype.forEach()` (IE8 and older) are also unable to run Guacamole for more fundamental reasons, and we do use it elsewhere in the webapp.

Were that not the case, `angular.forEach()` has the benefit that it automatically deals with `null` and `undefined` values.

guacamole/src/main/webapp/app/manage/templates/manageConnection.html
mike-jumper: Yes!

mike-jumper: OK. I've rebased away the mistake.

guacamole/src/main/webapp/app/manage/controllers/manageConnectionController.js
mike-jumper: I can restore the original usage of built-in `forEach()` if you prefer.

guacamole/src/main/webapp/app/manage/controllers/manageConnectionController.js
```diff
@@ -157,205 +137,178 @@ angular.module('manage').controller('manageConnectionController', ['$scope', '$i
...
+        return $q.resolve();
 
     };
 
     /**
-     * Returns the translation string namespace for the protocol having the
-     * given name. The namespace will be of the form:
-     *
-     * <code>PROTOCOL_NAME</code>
+     * Loads the data requred for performing the management task requested
```
jmuehlner: `requred`?

mike-jumper: Ops.

mike-jumper: Fxed.

guacamole/src/main/webapp/app/manage/controllers/manageConnectionController.js
jmuehlner: I have no problem with `angular.forEach`. :shipit:

jmuehlner:
Aside from some minor typos, this looks pretty good to me, but due to its enormity, it's probably a good idea to have somebody else take a glance as well.

Any concerns @necouchman?

necouchman:
I'll try to run through it today and see if anything jumps out to me, test it out, and merge if I don't run across any issues.

jmuehlner:
👍 sounds good!

mike-jumper:
I can confirm that I see the same behavior.

mike-jumper:
OK - it's because the data for the primary connection is only being loaded for the edit/clone case. I'll make the necessary changes.

guacamole/src/main/webapp/app/manage/directives/connectionPermissionEditor.js
```diff
@@ -0,0 +1,406 @@
...
+
+            // Otherwise, explicitly add the permission
+            else
+                PermissionSet.addConnectionGroupPermission($scope.permissionsAdded, PermissionSet.ObjectPermissionType.READ, identifier);
+
+        };
+
+        /**
+         * Updates the permissionsAdded and permissionsRemoved permission sets
+         * to reflect the removal of the given connection permission.
```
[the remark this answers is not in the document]
mike-jumper: Should now be fixed, though GitHub seems to have magically moved this comment up to the part of the code where "connection" is actually correct.

guacamole/src/main/webapp/app/manage/directives/connectionPermissionEditor.js
```diff
@@ -0,0 +1,406 @@
...
+         *
+         * @param {String} identifier
+         *     The identifier of the connection to remove READ permission for.
```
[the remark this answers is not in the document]
mike-jumper: Same here - fixed, but GitHub decided to be magic.

mike-jumper:
I believe everything should now be addressed.

PR #283, opened 2018-05-03
GUACAMOLE-559: Add support for reading/writing clipboard contents using the Asynchronous Clipboard API.

mike-jumper:
Finally: the Holy Grail.

https://developers.google.com/web/updates/2018/03/clipboardapi
https://www.w3.org/TR/clipboard-apis/#async-clipboard-api


mike-jumper:
Cripes it even works on mobile Chrome.

PR #286, opened 2018-05-06
GUACAMOLE-450: Update Tomcat to version 9

schleyk:
GUACAMOLE-450

Guacamole works fine wiht Version:  tomcat:9.0.7-jre8

all guacamole connections work without problems, I test it with :
Duo-Auth, RDP, SSH, telnet and VNC

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@
 # Use args for Tomcat image label to allow image builder to choose alternatives
 # such as `--build-arg TOMCAT_JRE=jre8-alpine`
 #
-ARG TOMCAT_VERSION=8.5
+ARG TOMCAT_VERSION=9
 ARG TOMCAT_JRE=jre8
 
 # Use official maven image for the build


necouchman:
@schleyk:
Your commit messages also need to have the JIRA issue prepended to them.

mike-jumper:
The issue for updating the Tomcat version is [GUACAMOLE-450](https://issues.apache.org/jira/browse/GUACAMOLE-450), which resulted in the Tomcat version in use being bumped to 8.5 (see #268). [GUACAMOLE-563](https://issues.apache.org/jira/browse/GUACAMOLE-563) requests that the Tomcat version of old releases be retroactively bumped, and was closed as invalid.

Can you speak to why specifically you want to bump all the way to 9, following the bump to 8.5 which has already been completed? If you have a compelling reason why 9 should be preferred for the Docker images going forward (and have tested that the image works correctly), then GUACAMOLE-450 should be reopened citing that reason, with this change referencing GUACAMOLE-450 and not GUACAMOLE-563.

schleyk:
I built the guacamole-client docker image with the latest Tomcat Docker image [tomcat:9.0.7-jre9](https://hub.docker.com/r/library/tomcat/tags/) locally.

All my guacamole connections work without problem on this version:
(RDP, SSH, telnet, VNC with Duo Auth)

I do not see the reason for use an older version of Tomcat?
I change the JIRA issue to GUACAMOLE-450.

mike-jumper:
> I do not see the reason for use an older version of Tomcat?

The 8.5.x series is not an older release; it is an actively-supported version of Tomcat (latest release being 8.5.31 on 2018-04-27) that is well tested with Guacamole. The question is not "why use an older version", but rather "why switch to the latest 9.x rather than the latest 8.5.x".

For this change to be merged, the answer to that question can't simply be "why not." The current state is stable, up-to-date, and known-good. There needs to be a reason behind the change other than the fact the number is bigger.

necouchman:
> There needs to be a reason behind the change other than the fact the number is bigger.

...and, some further testing to make sure everything does work in Tomcat 9.  While it seems like you've done a decent amount of testing, Kevin, there are at least several other authentication extensions that need to be verified against Tomcat 9 and JRE 9.  Having that validated would be great - obviously we'll have to get there eventually, but, at this point I agree with Mike, that there isn't sufficient justification or testing here, to move to Tomcat 9 & JRE 9.

PR #288, opened 2018-05-16
GUACAMOLE-565: Add terminal-type parameter for SSH and Telnet.

darchons:
Add the new terminal-type parameter (apache/guacamole-server#167) to the protocol JSON files.

### DIFF
--- a/guacamole-ext/src/main/resources/org/apache/guacamole/protocols/ssh.json
+++ b/guacamole-ext/src/main/resources/org/apache/guacamole/protocols/ssh.json
@@ -83,6 +83,11 @@
                     "name"    : "backspace",
                     "type"    : "ENUM",
                     "options" : [ "", "127", "8" ]
+                },
+                {
+                    "name"  : "terminal-type",
+                    "type"    : "ENUM",
+                    "options" : [ "xterm", "xterm-256color", "vt220", "vt100", "ansi", "linux" ]
                 }
             ]
         },
--- a/guacamole-ext/src/main/resources/org/apache/guacamole/protocols/telnet.json
+++ b/guacamole-ext/src/main/resources/org/apache/guacamole/protocols/telnet.json
@@ -66,6 +66,11 @@
                     "name"    : "backspace",
                     "type"    : "ENUM",
                     "options" : [ "", "127", "8" ]
+                },
+                {
+                    "name"  : "terminal-type",
+                    "type"    : "ENUM",
+                    "options" : [ "xterm", "xterm-256color", "vt220", "vt100", "ansi", "linux" ]
                 }
             ]
         },


mike-jumper:
I agree with @necouchman here. Adding the ability to explicitly set the terminal type is great, but the UI should probably be restricted to known-good and common values, especially since there is danger that setting this may result in applications sending control characters not yet supported by Guacamole's terminal.

darchons:
Updated to enum. I picked a set of common values that come from xterm's list of terminal emulator types, but beyond that, I don't really know what to look for to make sure each type works with the guacamole terminal.

guacamole-ext/src/main/resources/org/apache/guacamole/protocols/ssh.json
```diff
@@ -83,6 +83,11 @@
                     "name"    : "backspace",
                     "type"    : "ENUM",
                     "options" : [ "", "127", "8" ]
+                },
+                {
+                    "name"  : "terminal-type",
+                    "type"    : "ENUM",
+                    "options" : [ "xterm", "xterm-256color", "vt220", "vt100", "ansi", "linux" ]
```
necouchman: @mike-jumper can comment more on this, but I'm not sure that `xterm` and `xterm-256color` should be included in this list, yet.  There's a JIRA issue, 277 [1] , that deals with providing xterm compatibility, and it isn't completed, yet, AFAIK.  The other values should be fine, and I'm not sure if there are more we can add to the list.

[1] - https://issues.apache.org/jira/browse/GUACAMOLE-277

guacamole-ext/src/main/resources/org/apache/guacamole/protocols/telnet.json
```diff
@@ -66,6 +66,11 @@
                     "name"    : "backspace",
                     "type"    : "ENUM",
                     "options" : [ "", "127", "8" ]
+                },
+                {
+                    "name"  : "terminal-type",
+                    "type"    : "ENUM",
+                    "options" : [ "xterm", "xterm-256color", "vt220", "vt100", "ansi", "linux" ]
```
necouchman: As above - not sure about whether xterm should be supported, yet, but others look good.

guacamole-ext/src/main/resources/org/apache/guacamole/protocols/ssh.json
mike-jumper: As long as they're not the default (the server-side terminal emulator still uses "linux" if this parameter is omitted), it's probably OK. In fact, recent changes adding xterm-specific terminal codes will not have any effect unless "xterm" or "xterm-256color" are used.

I'm personally OK with this as long as it's documented in the manual that xterm and xterm-256color are not fully supported yet.

necouchman:
@darchons - Looks like I missed something relatively important when I reviewed/merged this request - all of the translations for the new fields are missing and need to be added to (at least) the en.json file, location is: guacamole/src/main/webapps/translations/en.json

darchons:
Thanks for the note! I'll submit a new PR.

PR #289, opened 2018-05-16
GUACAMOLE-470: Update "color-scheme" parameter for SSH and Telnet.

darchons:
apache/guacamole-server#132 changes the "color-scheme" parameter to be a configuration string instead of a simple enum, so update the protocol JSON accordingly.


mike-jumper:
@necouchman - while GUACAMOLE-470 is a 1.0.0 change, I don't see these changes on `staging/1.0.0`. Looks like the merge base was set incorrectly?

necouchman:
Oh, crud.  Yes, looks that way.  What needs to be done to fix that?

mike-jumper:
Erg... never mind. My local copy of `staging/1.0.0` was just out of date. 😓

PR #290, opened 2018-05-27
GUACAMOLE-567: Warn if network connection appears unstable.

mike-jumper:
This change adds an additional `UNSTABLE` tunnel status to the JavaScript API, allowing the tunnels to report unstable network conditions in a potentially implementation-dependent way. The web application has been modified to display this as a small warning notification in the lower-left corner of the client screen:

connection-unstable-warning

Both the HTTP and WebSocket tunnels currently implement this by tracking the last time any data was received from the server at all (not the time since last frame, but the time since *any* data was received), flagging themselves as unstable if no data has been received for at least 2 seconds.

Additional handling of HTTP and WebSocket error codes has also been added, such that network errors reported by the browser are recognized as such.


guacamole/src/main/webapp/app/client/types/ManagedClient.js
```diff
@@ -346,6 +346,18 @@ angular.module('client').factory('ManagedClient', ['$rootScope', '$injector',
                             ManagedClientState.ConnectionState.CONNECTING);
                         break;
 
+                    // Connection is being established
```
necouchman: Is being established?  Or is established?

mike-jumper: Oops. Definitely should be "is established".

mike-jumper:
OK - changed "is being established" to "is established" for the `OPEN` state.

PR #291, opened 2018-05-29
GUACAMOLE-566: Replace RESTExceptionWrapper with RESTExceptionMapper

necouchman:
Per the discussion on the mailing list ([1]) this change implements the `GuacamoleExceptionMapper` class that will map all instances of `GuacamoleException` thrown in the web client to a JSON error message.  This works not only for the main REST interfaces but also for any custom-built extension interfaces.  Additionally, I believe I was able to successfully map the rest of the `RESTExceptionWrapper` class into the new class, so I've removed that class entirely.  The interface appears to work fine with that removed, but I very well may have missed corner cases or other items that need to be validated, so that's worth a careful look before this gets merged.

This change does allow the error messages from my custom REST API in GUACAMOLE-38 (#181) to pass through successfully, so, overall, the implementation of the `ExceptionMapper<GuacamoleException>` appears to work well.

[1] - http://mail-archives.apache.org/mod_mbox/guacamole-dev/201805.mbox/%3CCALKeL-PmkxqcmcyC15rBjJ_FZpTHAQqOUHrpyoQsAQA=m9G2xw@mail.gmail.com%3E


guacamole/src/main/java/org/apache/guacamole/rest/GuacamoleExceptionMapper.java
```diff
@@ -0,0 +1,110 @@
...
+     *     The authentication token for the current session, or null if no
+     *     token is present.
+     */
+    private String getAuthenticationToken() {
+
+        @SuppressWarnings("unchecked")
+        Map<String, String[]> parameters = request.getParameterMap();
+
+        for (String paramName : parameters.keySet()) {
+            if (paramName.equals("token")) {
```
mike-jumper: There's no need to iterate over the key set of a `Map` to find the entry having a particular key. Part of the point of a `Map` is the ability to retrieve such an entry by its key.

    parameters.get("token")

In this case, there's no need to go through the map, as you can retrieve the parameter from the request directly:

    request.getParameter("token")

guacamole/src/main/java/org/apache/guacamole/rest/GuacamoleExceptionMapper.java
```diff
@@ -0,0 +1,110 @@
...
+    
+    /**
+     * The logger for this class.
+     */
+    private final Logger logger = LoggerFactory.getLogger(GuacamoleExceptionMapper.class);
+    
+    /**
+     * The request associated with this instance of this mapper.
+     */
+    @Context private HttpServletRequest request;
```
mike-jumper: `@Context` on own line, please.

guacamole/src/main/java/org/apache/guacamole/rest/RESTServiceModule.java
```diff
@@ -83,11 +83,9 @@ protected void configureServlets() {
         bind(AuthenticationService.class);
         bind(AuthTokenGenerator.class).to(SecureRandomAuthTokenGenerator.class);
         bind(DecorationService.class);
-
-        // Automatically translate GuacamoleExceptions for REST methods
-        MethodInterceptor interceptor = new RESTExceptionWrapper();
-        requestInjection(interceptor);
-        bindInterceptor(Matchers.any(), new RESTMethodMatcher(), interceptor);
+        
+        // Get the ExceptionMapper that will rewrite exceptions into JSON.
```
mike-jumper: `bind()` is not "getting" the `ExceptionMapper`. In this case, the original comment is probably still more appropriate.

guacamole/src/main/java/org/apache/guacamole/rest/RESTExceptionWrapper.java
```diff
@@ -1,203 +0,0 @@
...
-        // Translate GuacamoleException subclasses to HTTP error codes
-        catch (GuacamoleException e) {
-            throw new APIException(
-                Response.Status.fromStatusCode(e.getHttpStatusCode()),
-                e
-            );
-        }
-
-        // Rethrow unchecked exceptions such that they are properly wrapped
-        catch (Throwable t) {
```
mike-jumper: There's a good deal of logging and error handling which happens blanketly here (`Throwable`), but will not happen with the new `ExceptionMapper`. This functionality will somehow need to be included.

guacamole/src/main/java/org/apache/guacamole/rest/GuacamoleExceptionMapper.java
necouchman: Ah, of course.

guacamole/src/main/java/org/apache/guacamole/rest/RESTExceptionWrapper.java
necouchman: In that case, seems like this new class should just be renamed to `RESTExceptionMapper` since it isn't just interested in `GuacamoleException` classes.  I'll rework it a bit and add this back in.

guacamole/src/main/java/org/apache/guacamole/rest/GuacamoleExceptionMapper.java
necouchman: That should be better.  Is it worth keeping it in its own method, or better to just throw it in the toResponse() method?

guacamole/src/main/java/org/apache/guacamole/rest/GuacamoleExceptionMapper.java
necouchman: Done.

guacamole/src/main/java/org/apache/guacamole/rest/RESTServiceModule.java
necouchman: Copied/pasted the original comment.

guacamole/src/main/java/org/apache/guacamole/rest/RESTExceptionWrapper.java
necouchman: I reworked it so that it's much more generic overall, and should deal with `Throwable` types.  Let me know if anything needs fixing there.

guacamole/src/main/java/org/apache/guacamole/rest/RESTExceptionMapper.java
```diff
@@ -0,0 +1,118 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A class that maps GuacamoleExceptions in a way that returns a
+ * custom response to the user via JSON rather than allowing the default
+ * web application error handling to take place.
+ */
+@Provider
+@Singleton
```
mike-jumper: Does this have an effect? If so, and `RESTExceptionMapper` is a singleton, that would mean that the `HttpServletRequest` field is shared across all requests (which will cause problems for concurrent requests).

guacamole/src/main/java/org/apache/guacamole/rest/RESTExceptionMapper.java
```diff
@@ -0,0 +1,118 @@
...
+        else
+            logger.error("An internal error occurred, but did not contain "
+                    + "an error message. Enable debug-level logging for "
+                    + "details.");
+            
+        logger.debug("Unexpected error in REST endpoint.", t);
+            
+        return Response
+                .status(Response.Status.INTERNAL_SERVER_ERROR)
+                .entity("Unexpected Internal Error.")
```
mike-jumper: The entity should probably still be an `APIError`, yes? But rather one constructed specifically to represent an internal error?

guacamole/src/main/java/org/apache/guacamole/rest/RESTExceptionMapper.java
```diff
@@ -0,0 +1,118 @@
...
+            
+            if (authenticationService.destroyGuacamoleSession(token))
+                logger.debug("Implicitly invalidated session for token \"{}\"", token);
+        }
+        
+        // Translate GuacamoleException subclasses to HTTP error codes 
+        if (t instanceof GuacamoleException)
+            return Response
+                    .status(((GuacamoleException) t).getHttpStatusCode())
+                    .entity(new APIError((GuacamoleException)t))
```
mike-jumper: Ouch - please add a space between `(GuacamoleException)` and `t`. That `t` is almost invisible. ;)

guacamole/src/main/java/org/apache/guacamole/rest/RESTExceptionMapper.java
necouchman: It has some effect - if you remove it, you get an error during the webapp deployment in Tomcat that it must be a `Singleton`.  What's the best way to test the token expiration?  Log in with two separate sessions, let them expire (60 minutes) and then try to load something and see what happens?  Or is there some other way to trigger the combination of `GuacamoleUnauthorizedException` and destroying the token?

necouchman: According to this StackTrace article, it appears they may use "proxies" (?) to inject the correct request context into the single instance each time an exception is thrown??

https://stackoverflow.com/questions/17766072/jax-rs-jersey-exceptionmapper-context-injection-into-static-singleton-class

So maybe it actually turns out okay??

guacamole/src/main/java/org/apache/guacamole/rest/RESTExceptionMapper.java
necouchman: Yeah, probably so.  Should be good, now.

guacamole/src/main/java/org/apache/guacamole/rest/RESTExceptionMapper.java
necouchman: Spaced.

guacamole/src/main/java/org/apache/guacamole/rest/RESTExceptionMapper.java
mike-jumper: > What's the best way to test the token expiration?

I'd just temporarily modify one of the existing extensions to throw `GuacamoleUnauthorizedException` in some case which you can easily trigger, and verify that doing so invalidates the associated token.

> According to this StackOverfow article, it appears they may use "proxies" (?) to inject the correct request context into the single instance each time an exception is thrown??
>
> https://stackoverflow.com/questions/17766072/jax-rs-jersey-exceptionmapper-context-injection-into-static-singleton-class
>
> So maybe it actually turns out okay??

Yeah, it does seem so. Let's correct the comment in that case, which currently states:

    /**
     * The request associated with this instance of this mapper.
     */
    @Context
    private HttpServletRequest request;

Might be helpful to note that the request object is always request scoped (and will thus always point to the relevant request within `toResponse()`), even for concurrent requests, regardless of the fact it is a field on a singleton.

mike-jumper:
Errr ... side note: might want to reword 85140e6 - it points to GUACAMOLE-513, not GUACAMOLE-566.

BUT! Other than that ... we're good.

necouchman:
> Errr ... side note: might want to reword 85140e6 - it points to GUACAMOLE-513, not GUACAMOLE-566.

Haha, oops.  Should be fixed, now.

PR #292, opened 2018-06-05
GUACAMOLE-524: Add attributes to credentials and token filter

necouchman:
This is the first in a set of at least a couple of pull requests that allows user attributes retrieved in various authentication extensions to be used in the `TokenFilter` class so that they can be used by connections provided by other extensions.  This change adds the attributes `Map` to the `Credentials` object, along with methods for interacting with it, and also adds the changes into the `TokenFilter` and `StandardToken` classes to map those attributes to token values.

With these changes, if an attribute is added to a `Credentials` object with a key of "email" and a value of "nick@guacamole.local", the token `${GUAC_ATTR:EMAIL} would then substitute in the value "nick@guacamole.local".

Assuming this is acceptable, the next pull request or two would add support to various authentication modules for configuring and retrieving attributes and adding them to the `Credentials` object so that they can be used.

### DIFF
--- a/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Credentials.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Credentials.java
@@ -20,6 +20,7 @@
 package org.apache.guacamole.net.auth;
 
 import java.io.Serializable;
+import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


necouchman:
Going to close this for the time being - turns out Travis (original reporter) is still working on some changes and I don't want to take over or squash that effort.

PR #294, opened 2018-06-06
GUACAMOLE-569: add Simplified Chinese translation

FlyingOE:
1. Add translation for Simplified Chinese
2. Fix minor typo error in English translation


FlyingOE:
Yes, that's right. US English layout is indeed the most commonly used keyboard in China. To actually input Chinese characters, one need to use an IME software, as no hardware keyboard can support the vast number of characters in Chinese.

necouchman:
Okay, cool.

@mike-jumper Just to verify, there's no concerns with merging this into master, right, with the 1.0.0 release in progress?  I'm pretty sure that's the point of the separate branches, but just want to make sure - again, been too long and I can't remember what all the caveats are.

mike-jumper:
> Just to verify, there's no concerns with merging this into master, right, with the 1.0.0 release in progress?

@necouchman - nope, no need to worry. The point of having the staging branch is indeed to insulate the contents of the release from any other in-progress changes, and to allow those changes to continue without being blocked until the release is out.

Just be sure not to merge non-1.0.0 changes to `staging/1.0.0`. This one just goes to `master`. ;)

> Yes, that's right. US English layout is indeed the most commonly used keyboard in China. To actually input Chinese characters, one need to use an IME software, as no hardware keyboard can support the vast number of characters in Chinese.

If this is the case, and `zh-cn-qwerty.json` is truly identical to the US one, then we should represent that fact by referencing `en-us-qwerty.json` rather than duplicate things.

necouchman:
> Just be sure not to merge non-1.0.0 changes to staging/1.0.0. This one just goes to master. ;)

Yep, got it.  Thanks for confirming all of that - what I thought, but wanted to make sure.

> If this is the case, and zh-cn-qwerty.json is truly identical to the US one, then we should represent that fact by referencing en-us-qwerty.json rather than duplicate things.

Makes sense.  How does @FlyingOE go about referencing en-us-qwerty.json?

mike-jumper:
> Makes sense. How does @FlyingOE go about referencing en-us-qwerty.json?

Just by pointing the `CLIENT.URL_OSK_LAYOUT` string to the `layouts/en-us-qwerty.json` file. For example:

https://github.com/apache/guacamole-client/blob/0e856d7a212135d71ecfeef19270e4ff534dbc04/guacamole/src/main/webapp/translations/en.json#L145

The relevant line in the Simplified Chinese translation is:

https://github.com/FlyingOE/guacamole-client/blob/fa6a256f0067c8b106bc1f93a34d256470097427/guacamole/src/main/webapp/translations/zh.json#L145

necouchman:
Ah, nice.  @FlyingOE can you make this change, and remove the zh-cn-qwerty.json file?

necouchman:
Looks good.  @mike-jumper any other concerns?

mike-jumper:
Nope. Looks good to me.

PR #295, opened 2018-06-06
GUACAMOLE-565: Add en translations for terminal-type parameter.

darchons:
Follow-up to #288. Add en translations for the terminal-type parameter and its enum values. Also add an empty value to the terminal-type enum lists.

### DIFF
--- a/guacamole/src/main/webapp/translations/en.json
+++ b/guacamole/src/main/webapp/translations/en.json
@@ -446,6 +446,7 @@
         "FIELD_HEADER_RECORDING_PATH" : "Recording path:",
         "FIELD_HEADER_SERVER_ALIVE_INTERVAL" : "Server keepalive interval:",
         "FIELD_HEADER_SFTP_ROOT_DIRECTORY"   : "File browser root directory:",
+        "FIELD_HEADER_TERMINAL_TYPE"   : "Terminal type:",
         "FIELD_HEADER_TYPESCRIPT_NAME" : "Typescript name:",
         "FIELD_HEADER_TYPESCRIPT_PATH" : "Typescript path:",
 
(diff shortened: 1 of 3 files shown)


guacamole/src/main/webapp/translations/en.json
```diff
@@ -475,6 +476,14 @@
         "FIELD_OPTION_FONT_SIZE_96"    : "96",
         "FIELD_OPTION_FONT_SIZE_EMPTY" : "",
 
+        "FIELD_OPTION_TERMINAL_TYPE_EMPTY" : "",
+        "FIELD_OPTION_TERMINAL_TYPE_XTERM" : "xterm",
+        "FIELD_OPTION_TERMINAL_TYPE_XTERM_256COLOR" : "xterm-256color",
+        "FIELD_OPTION_TERMINAL_TYPE_VT220" : "vt220",
+        "FIELD_OPTION_TERMINAL_TYPE_VT100" : "vt100",
+        "FIELD_OPTION_TERMINAL_TYPE_ANSI"  : "ansi",
+        "FIELD_OPTION_TERMINAL_TYPE_LINUX" : "linux",
```
necouchman: This is a style nitpick, but in keeping with the style used throughout the rest of these JSON files, if there are a group of objects like this, the ":" should be lined up in the same column.

guacamole/src/main/webapp/translations/en.json
```diff
@@ -536,6 +546,14 @@
...
```
necouchman: As mentioned above - for style consistency, please line up the ":" in this group of options.

darchons:
Fixed indent and alphabetized the lists.

necouchman:
One more thing - please change the merge base for the pull request from master to staging/1.0.0, since this is going into the 1.0.0 release.

darchons:
Ok done.

PR #296, opened 2018-06-06
GUACAMOLE-470: Remove translations for "color-scheme" enum values.

darchons:
Follow-up to #289. Remove translations for the "color-scheme" enum values, because the field type is now "TEXT" and not "ENUM" anymore.

Changing the field type to "TEXT" does have the disadvantage of losing localized enum choices. Maybe it would be better to keep "color-scheme" an ENUM, and add a "custom-color-scheme" TEXT field for customized color schemes?

### DIFF
--- a/guacamole/src/main/webapp/translations/de.json
+++ b/guacamole/src/main/webapp/translations/de.json
@@ -378,12 +378,6 @@
         "FIELD_HEADER_PRIVATE_KEY" : "Privater Schlüssel:",
         "FIELD_HEADER_READ_ONLY"   : "Nur-Lesen:",
 
-        "FIELD_OPTION_COLOR_SCHEME_BLACK_WHITE" : "Schwarz auf Weiß",
-        "FIELD_OPTION_COLOR_SCHEME_EMPTY"       : "",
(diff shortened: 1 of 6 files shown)


necouchman:
@darchons Please change merge base to staging/1.0.0.

darchons:
Done.

necouchman:
No one else seems to have an opinion on the color-scheme vs. custom-color-scheme thing, so I'll go ahead and merge this.  We can talk about it going forward if we decide it's something we want to change.

mike-jumper:
> Maybe it would be better to keep "color-scheme" an ENUM, and add a "custom-color-scheme" TEXT field for customized color schemes?

Nope. The connection parameters are independent of the user interface, and user interface concerns should not drive the creation of entirely new connection parameters.

I think I have a solution which allows us to hang onto the pre-defined `color-scheme` values while still allowing custom values. If `TEXT` fields are allowed to have defined options just as `ENUM`, that would give us a field with the semantics we're looking for.

I have some WIP changes which provide the above, leveraging the `<datalist>` element:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist

PR #297, opened 2018-06-06
GUACAMOLE-570: Bump version numbers to 1.0.0.

mike-jumper:


mike-jumper:
Nope. No need to delay. In fact, this needs to be merged before the release can happen. ;)

The main important thing to remember for changes that go to `staging/1.0.0` is to also merge `staging/1.0.0` back to master:

http://guacamole.apache.org/pull-requests/#merging-release-specific-changes

necouchman:
> Nope. No need to delay. In fact, this needs to be merged before the release can happen. ;)

I knew it had to happen prior to release, just couldn't remember if there were any other nuances to when it needed to be done, etc.  Anyway, version is now changed to 1.0.0.

> The main important thing to remember for changes that go to staging/1.0.0 is to also merge staging/1.0.0 back to master:

That part I remembered, just wasn't sure about timing of the merge.  This is done, too.

PR #299, opened 2018-06-08
GUACAMOLE-524: Allow custom LDAP attributes to be used as tokens

jaredfrees:
1. Add property for reading list of LDAP attributes in guacamole.properties.
2. Add LDAP attribute name and attribute value into credentials object.
3. Add attribute name and attribute value into Tokens from credentials object.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledAuthenticatedUser.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledAuthenticatedUser.java
@@ -26,6 +26,11 @@ import org.apache.guacamole.net.auth.AuthenticatedUser;
 import org.apache.guacamole.net.auth.AuthenticationProvider;
 import org.apache.guacamole.net.auth.Credentials;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import java.util.Map;
+
+
 /**
  * Associates a user with the credentials they used to authenticate, their
  * corresponding ModeledUser, and the AuthenticationProvider which produced
@@ -79,6 +84,7 @@ public class ModeledAuthenticatedUser extends RemoteAuthenticatedUser {
         super(authenticatedUser.getAuthenticationProvider(), authenticatedUser.getCredentials());
         this.modelAuthenticationProvider = modelAuthenticationProvider;
         this.user = user;
+        this.setAttributes(authenticatedUser.getAttributes());
     }
 
     /**
@@ -93,7 +99,7 @@ public class ModeledAuthenticatedUser extends RemoteAuthenticatedUser {
      *     A ModeledUser object which is backed by the data associated with
      *     this user in the database.
      *
-     * @param credentials 
+     * @param credentials
      *     The credentials given by the user when they authenticated.
      */
     public ModeledAuthenticatedUser(AuthenticationProvider authenticationProvider,
@@ -107,7 +113,7 @@ public class ModeledAuthenticatedUser extends RemoteAuthenticatedUser {
      * Returns a ModeledUser object which is backed by the data associated with
      * this user within the database.
      *
-     * @return 
+     * @return
      *     A ModeledUser object which is backed by the data associated with
      *     this user in the database.
      */
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledAuthenticatedUser.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledAuthenticatedUser.java
@@ -26,11 +26,6 @@ import org.apache.guacamole.net.auth.AuthenticatedUser;
 import org.apache.guacamole.net.auth.AuthenticationProvider;
 import org.apache.guacamole.net.auth.Credentials;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import java.util.Map;
-
-
 /**
  * Associates a user with the credentials they used to authenticate, their
  * corresponding ModeledUser, and the AuthenticationProvider which produced
@@ -84,7 +79,7 @@ public class ModeledAuthenticatedUser extends RemoteAuthenticatedUser {
         super(authenticatedUser.getAuthenticationProvider(), authenticatedUser.getCredentials());
         this.modelAuthenticationProvider = modelAuthenticationProvider;
         this.user = user;
-        this.setAttributes(authenticatedUser.getAttributes());
+        super.setAttributes(authenticatedUser.getAttributes());
     }
 
     /**
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/RemoteAuthenticatedUser.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/RemoteAuthenticatedUser.java
@@ -19,6 +19,8 @@
 
 package org.apache.guacamole.auth.jdbc.user;
 
+import java.util.HashMap;
+import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import javax.servlet.http.HttpServletRequest;
@@ -68,6 +70,21 @@ public abstract class RemoteAuthenticatedUser implements AuthenticatedUser {
     private static final Pattern X_FORWARDED_FOR = Pattern.compile("^" + IP_ADDRESS_REGEX + "(, " + IP_ADDRESS_REGEX + ")*$");
 
     /**
+     * Arbitrary attributes associated with this RemoteAuthenticatedUser object.
+     */
+    private Map<String, String> attributes = new HashMap<String, String>();
+
+    @Override
+    public Map<String, String> getAttributes() {
+        return attributes;
+    }
+
+    @Override
+    public void setAttributes(Map<String, String> attributes) {
+        this.attributes = attributes;
+    }
+
+    /**
      * Derives the remote host of the authenticating user from the given
      * credentials object. The remote host is derived from X-Forwarded-For
      * in addition to the actual source IP of the request, and thus is not
@@ -98,7 +115,7 @@ public abstract class RemoteAuthenticatedUser implements AuthenticatedUser {
         return request.getRemoteAddr();
 
     }
-    
+
     /**
      * Creates a new RemoteAuthenticatedUser, deriving the associated remote
      * host from the given credentials.
@@ -106,7 +123,7 @@ public abstract class RemoteAuthenticatedUser implements AuthenticatedUser {
      * @param authenticationProvider
      *     The AuthenticationProvider that has authenticated the given user.
      *
-     * @param credentials 
+     * @param credentials
      *     The credentials given by the user when they authenticated.
      */
     public RemoteAuthenticatedUser(AuthenticationProvider authenticationProvider,
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
@@ -22,7 +22,14 @@ package org.apache.guacamole.auth.ldap;
 import com.google.inject.Inject;
 import com.google.inject.Provider;
 import com.novell.ldap.LDAPConnection;
+import com.novell.ldap.LDAPAttributeSet;
+import com.novell.ldap.LDAPEntry;
+import com.novell.ldap.LDAPAttribute;
+import com.novell.ldap.LDAPException;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Iterator;
+import java.util.Map;
 import org.apache.guacamole.auth.ldap.user.AuthenticatedUser;
 import org.apache.guacamole.auth.ldap.user.UserContext;
 import org.apache.guacamole.GuacamoleException;
@@ -33,14 +40,6 @@ import org.apache.guacamole.net.auth.credentials.CredentialsInfo;
 import org.apache.guacamole.net.auth.credentials.GuacamoleInvalidCredentialsException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.List;
-import java.util.Iterator;
-import com.novell.ldap.LDAPAttributeSet;
-import com.novell.ldap.LDAPEntry;
-import com.novell.ldap.LDAPAttribute;
-import com.novell.ldap.LDAPException;
 
 /**
  * Service providing convenience functions for the LDAP AuthenticationProvider
@@ -230,22 +229,21 @@ public class AuthenticationProviderService {
             throw new GuacamoleInvalidCredentialsException("Permission denied.", CredentialsInfo.USERNAME_PASSWORD);
 
         try {
-            try {
-                String username = credentials.getUsername();
-                Map<String, String> ldapAttrs = getLDAPAttributes(ldapConnection, username);
-                credentials.setLDAPAttributes(ldapAttrs);
-            }
-            catch (LDAPException e) {
-                throw new GuacamoleServerException("Error while querying for LDAP User Attributes.", e);
-            }
-
             // Return AuthenticatedUser if bind succeeds
             AuthenticatedUser authenticatedUser = authenticatedUserProvider.get();
             authenticatedUser.init(credentials);
+
+            //set attributes
+            String username = credentials.getUsername();
+            Map<String, String> attrs = getLDAPAttributes(ldapConnection, username);
+            authenticatedUser.setAttributes(attrs);
+
             return authenticatedUser;
 
         }
-
+        catch (LDAPException e) {
+            throw new GuacamoleServerException("Error while querying for User Attributes.", e);
+        }
         // Always disconnect
         finally {
             ldapService.disconnect(ldapConnection);
@@ -275,13 +273,13 @@ public class AuthenticationProviderService {
      *     If an error occurs retrieving the user DN.
      */
     private Map<String, String> getLDAPAttributes(LDAPConnection ldapConnection,
-            String username) throws LDAPException, GuacamoleException {
+            String username) throws LDAPException {
 
         // Get attributes from configuration information
         List<String> attrList = confService.getAttributes();
 
         // If there are no attributes there is no reason to search LDAP
-        if (attrList.size() == 0)
+        if (attrList == null || attrList.isEmpty())
             return null;
 
         // Build LDAP query parameters
@@ -294,9 +292,8 @@ public class AuthenticationProviderService {
 
         // Add each attribute into Map
         Map<String, String> attrMap = new HashMap<String, String>();
-        Iterator attrIterator = attrSet.iterator();
-        while (attrIterator.hasNext()) {
-            LDAPAttribute attr = (LDAPAttribute)attrIterator.next();
+        for (Object attrObj : attrSet) {
+            LDAPAttribute attr = (LDAPAttribute)attrObj;
             String attrName = attr.getName();
             String attrValue = attr.getStringValue();
             attrMap.put(attrName, attrValue);
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
@@ -26,9 +26,7 @@ import com.novell.ldap.LDAPAttributeSet;
 import com.novell.ldap.LDAPConnection;
 import com.novell.ldap.LDAPEntry;
 import com.novell.ldap.LDAPException;
-import com.novell.ldap.LDAPReferralException;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import org.apache.guacamole.auth.ldap.user.AuthenticatedUser;
@@ -236,9 +234,7 @@ public class AuthenticationProviderService {
             authenticatedUser.init(credentials);
 
             // Set attributes
-            String username = credentials.getUsername();
-            Map<String, String> attrs = getLDAPAttributes(ldapConnection, username);
-            authenticatedUser.setAttributes(attrs);
+            authenticatedUser.setAttributes(getLDAPAttributes(ldapConnection, credentials.getUsername()));
 
             return authenticatedUser;
 
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
@@ -27,8 +27,8 @@ import com.novell.ldap.LDAPEntry;
 import com.novell.ldap.LDAPAttribute;
 import com.novell.ldap.LDAPException;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import org.apache.guacamole.auth.ldap.user.AuthenticatedUser;
 import org.apache.guacamole.auth.ldap.user.UserContext;
@@ -241,9 +241,6 @@ public class AuthenticationProviderService {
             return authenticatedUser;
 
         }
-        catch (LDAPException e) {
-            throw new GuacamoleServerException("Error while querying for User Attributes.", e);
-        }
         // Always disconnect
         finally {
             ldapService.disconnect(ldapConnection);
@@ -258,6 +255,7 @@ public class AuthenticationProviderService {
      *
      * @param ldapConnection
      *     LDAP connection to find the custom LDAP attributes.
+     *
      * @param username
      *     The username of the user whose attributes are queried.
      *
@@ -273,7 +271,7 @@ public class AuthenticationProviderService {
      *     If an error occurs retrieving the user DN.
      */
     private Map<String, String> getLDAPAttributes(LDAPConnection ldapConnection,
-            String username) throws LDAPException {
+            String username) throws GuacamoleException {
 
         // Get attributes from configuration information
         List<String> attrList = confService.getAttributes();
@@ -286,17 +284,22 @@ public class AuthenticationProviderService {
         String[] attrArray = attrList.toArray(new String[attrList.size()]);
         String userDN = getUserBindDN(username);
 
-        // Get LDAP attributes by querying LDAP
-        LDAPEntry userEntry = ldapConnection.read(userDN, attrArray);
-        LDAPAttributeSet attrSet = userEntry.getAttributeSet();
-
-        // Add each attribute into Map
         Map<String, String> attrMap = new HashMap<String, String>();
-        for (Object attrObj : attrSet) {
-            LDAPAttribute attr = (LDAPAttribute)attrObj;
-            String attrName = attr.getName();
-            String attrValue = attr.getStringValue();
-            attrMap.put(attrName, attrValue);
+        try {
+            // Get LDAP attributes by querying LDAP
+            LDAPEntry userEntry = ldapConnection.read(userDN, attrArray);
+            LDAPAttributeSet attrSet = userEntry.getAttributeSet();
+
+            // Add each attribute into Map
+            for (Object attrObj : attrSet) {
+                LDAPAttribute attr = (LDAPAttribute)attrObj;
+                String attrName = attr.getName();
+                String attrValue = attr.getStringValue();
+                attrMap.put(attrName, attrValue);
+            }
+        }
+        catch (LDAPException e) {
+            throw new GuacamoleServerException("Error while querying for User Attributes.", e);
         }
 
         return attrMap;
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
@@ -269,7 +269,7 @@ public class AuthenticationProviderService {
      *     If an error occurs while searching for the user attributes.
      *
      * @throws GuacamoleException
-     *     If an error occurs retrieving the user DN.
+     *     If an error occurs retrieving the user DN or the attributes.
      */
     private Map<String, String> getLDAPAttributes(LDAPConnection ldapConnection,
             String username) throws GuacamoleException {
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
@@ -197,7 +197,8 @@ public class AuthenticationProviderService {
 
     /**
      * Returns an AuthenticatedUser representing the user authenticated by the
-     * given credentials. Also adds custom LDAP attributes to credentials object.
+     * given credentials. Also adds custom LDAP attributes to the
+     * AuthenticatedUser.
      *
      * @param credentials
      *     The credentials to use for authentication.
@@ -233,7 +234,7 @@ public class AuthenticationProviderService {
             AuthenticatedUser authenticatedUser = authenticatedUserProvider.get();
             authenticatedUser.init(credentials);
 
-            //set attributes
+            // Set attributes
             String username = credentials.getUsername();
             Map<String, String> attrs = getLDAPAttributes(ldapConnection, username);
             authenticatedUser.setAttributes(attrs);
@@ -293,9 +294,7 @@ public class AuthenticationProviderService {
             // Add each attribute into Map
             for (Object attrObj : attrSet) {
                 LDAPAttribute attr = (LDAPAttribute)attrObj;
-                String attrName = attr.getName();
-                String attrValue = attr.getStringValue();
-                attrMap.put(attrName, attrValue);
+                attrMap.put(attr.getName(), attr.getStringValue());
             }
         }
         catch (LDAPException e) {
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
@@ -21,11 +21,12 @@ package org.apache.guacamole.auth.ldap;
 
 import com.google.inject.Inject;
 import com.google.inject.Provider;
-import com.novell.ldap.LDAPConnection;
-import com.novell.ldap.LDAPAttributeSet;
-import com.novell.ldap.LDAPEntry;
 import com.novell.ldap.LDAPAttribute;
+import com.novell.ldap.LDAPAttributeSet;
+import com.novell.ldap.LDAPConnection;
+import com.novell.ldap.LDAPEntry;
 import com.novell.ldap.LDAPException;
+import com.novell.ldap.LDAPReferralException;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
@@ -265,9 +265,6 @@ public class AuthenticationProviderService {
      *     given LDAP connection, as a map of attribute name to
      *     corresponding attribute value.
      *
-     * @throws LDAPException
-     *     If an error occurs while searching for the user attributes.
-     *
      * @throws GuacamoleException
      *     If an error occurs retrieving the user DN or the attributes.
      */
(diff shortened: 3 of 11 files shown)


guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Credentials.java
```diff
@@ -72,6 +72,29 @@
...
+     public Map<String, String> getLDAPAttributes() {
+         return ldapAttrs;
+     }
+
+     /**
+      * Sets the LDAP attributes associated with this set of credentials.
+      * @param attributes The LDAP attributes to associate with this set of
+      *                   credentials.
+      */
+     public void setLDAPAttributes(Map<String, String> attributes) {
```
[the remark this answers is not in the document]
necouchman: > Better perhaps than that would be having AuthenticatedUser implement Attributes (as, semantically, it is the user that has these attributes, not their credentials).

Ah, yes, that makes more sense.  Sorry, @jaredfrees, I may have steered you wrong, here.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -236,6 +253,58 @@ public AuthenticatedUser authenticateUser(Credentials credentials)
...
+        String userDN = getUserBindDN(username);
+
+        // Get LDAP attributes by querying LDAP
+        LDAPEntry userEntry = ldapConnection.read(userDN, attrArray);
+        LDAPAttributeSet attrSet = userEntry.getAttributeSet();
+
+        // Add each attribute into Map
+        Map<String, String> attrMap = new HashMap<String, String>();
+        Iterator attrIterator = attrSet.iterator();
+        while (attrIterator.hasNext()) {
```
necouchman: Might use the Java for-each loop, here - you used it below when dealing with the LDAP attributes in the token - no reason not to use it, here, too! :smile:

jaredfrees:
I just saw your pull request from 3 days ago @necouchman , I should have checked that earlier.
You are definitely correct @mike-jumper. I am going to change this to use AuthenticatedUser inside of the ldap extension. I think having AuthenticatedUser implement Attributes is the best idea. 

Do you think I should still add the attributes in StandardTokens.java?

guacamole-ext/src/main/java/org/apache/guacamole/token/StandardTokens.java
```diff
@@ -23,6 +23,10 @@
 import java.util.Date;
 import org.apache.guacamole.net.auth.AuthenticatedUser;
 import org.apache.guacamole.net.auth.Credentials;
+import java.util.Map;
+import java.util.Set;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
necouchman: Guessing you had these in here for debugging purposes, but doesn't look like there's any logging code left, so probably best to get rid of them.  Also, these should probably be alphabetized.

guacamole-ext/src/main/java/org/apache/guacamole/token/StandardTokens.java
```diff
@@ -74,6 +78,12 @@
      */
     private static final String TIME_FORMAT = "HHmmss";
 
+    /**
+     * Standard prefix to append to beginning of the name of each custom
+    *  LDAP attribute before adding attributes as tokens.
+     */
+    private static final String LDAP_ATTR_PREFIX = "USER_ATTR:";
```
necouchman: As with @mike-jumper's suggestion above, I'd say make this prefix more generic.

necouchman:
> I just saw your pull request from 3 days ago @necouchman , I should have checked that earlier.

Uh, no, I should have waited to submit that until *after* I had given you a chance to respond to see if you were still working on it.  I kind of jumped the gun on that one - sorry about that.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
jaredfrees: the reason I did this is because I couldn't get the casting to work in the for-each loop, but I think I got the for-each loop working now.

jaredfrees:
Thanks for all your comments @necouchman. I am going to try to get this to work today.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
necouchman: Oh, I see what you mean - yeah, that could be a problem, as those custom classes (particularly ancient ones like the JLDAP ones :smile:) don't always plan nice with the for-each loop.

jaredfrees: Here is what I am doing now:
       for (Object attrObj : attrMap) {
            LDAPAttribute attr = (LDAPAttribute)attrObj;
            .....
       }

necouchman: Yep, that's fine.  And, depending on what you're doing with it later, you might not even need that assignment, there, you could do:

    String attrName = ((LDAPAttribute) attrObj).getName();
    String attrValue = ((LDAPAttribute) attrObj).getStringValue();

I'm not certain what penalty there is, if any, to doing the cast twice like that, so it may actually make sense to go ahead and assign it to a new `LDAPAttribute` type.  Either way is probably perfectly fine.

jaredfrees:
Just added the last of my updates. I'm hoping I did that right with the AuthenticatedUser.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -36,6 +36,7 @@
 import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.GuacamoleServerException;
 import org.apache.guacamole.net.auth.AuthenticatedUser;
+//import org.apache.guacamole.auth.ldap.user.AuthenticatedUser;
```
necouchman: This should probably come out :smiley:

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/AuthenticatedUser.java
```diff
@@ -20,15 +20,19 @@
...
+import org.apache.guacamole.net.auth.Attributes;
 import org.apache.guacamole.net.auth.AuthenticationProvider;
 import org.apache.guacamole.net.auth.Credentials;
 
 /**
  * An LDAP-specific implementation of AuthenticatedUser, associating a
  * particular set of credentials with the LDAP authentication provider.
  */
-public class AuthenticatedUser extends AbstractAuthenticatedUser {
+public class AuthenticatedUser extends AbstractAuthenticatedUser
```
necouchman: The way this is implemented, here, these attributes (tokens) would not be available to connections stored in the JDBC module, since the module you've extended, here, is the LDAP-specific one.  In order for this to cover multiple authentication modules, you'll either have to implement this at one of the parent classes (`AbstractAuthenticatedUser` perhaps) or across multiple modules.  Of course, that doesn't all have to be done with this one pull request - this can just tackle LDAP for now - but the base changes that go into it probably need to at least make it available for other extensions to implement.

The other advantage of at least creating the abstract methods within one of the parent classes is avoiding having to do the type-casting when trying to run the `getAttributes()` method.

@mike-jumper Any pointers on the best place to put the various attribute-related stuff such that it can be easily used across different authentication extensions?

guacamole-ext/src/main/java/org/apache/guacamole/token/StandardTokens.java
```diff
@@ -23,6 +23,8 @@
 import java.util.Date;
 import org.apache.guacamole.net.auth.AuthenticatedUser;
 import org.apache.guacamole.net.auth.Credentials;
+import java.util.Map;
+import java.util.Set;
```
necouchman: This should go above the `org.apache` entries, and be alphabetized in the correct order.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/AuthenticatedUser.java
```diff
@@ -20,15 +20,19 @@
 package org.apache.guacamole.auth.ldap.user;
 
 import com.google.inject.Inject;
+import java.util.Map;
+import java.util.HashMap;
```
necouchman: HashMap > Map, at least in the alphabet.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -126,6 +127,10 @@
             TokenFilter tokenFilter = new TokenFilter();
             StandardTokens.addStandardTokens(tokenFilter, user);
 
+            // Add custom attribute tokens
+            Map<String, String> attrs = ( (org.apache.guacamole.auth.ldap.user.AuthenticatedUser) user).getAttributes();
```
necouchman: I'm not sure we want to have to do this type-casting, here, but see my comment below on the implementation within the LDAP-specific `AuthenticatedUser` class.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -221,21 +229,79 @@ public AuthenticatedUser authenticateUser(Credentials credentials)
...
+     *     If an error occurs while searching for the user attributes.
+     *
+     * @throws GuacamoleException
+     *     If an error occurs retrieving the user DN.
+     */
+    private Map<String, String> getLDAPAttributes(LDAPConnection ldapConnection,
+            String username) throws LDAPException {
+
+        // Get attributes from configuration information
+        List<String> attrList = confService.getAttributes();
```
necouchman: Since `getAttributes()` throws a `GuacamoleException`, it will need to be caught, here.  That's why I suggest having the `getLDAPAttributes()` method throw only `GuacamoleException`, and then using `try {} catch {}` within this method to catch and re-throw the `LDAPException` as a `GuacamoleException`.  One way or the other, though, this needs to be handled.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -22,10 +22,18 @@
...
 import com.google.inject.Provider;
 import com.novell.ldap.LDAPConnection;
+import com.novell.ldap.LDAPAttributeSet;
+import com.novell.ldap.LDAPEntry;
+import com.novell.ldap.LDAPAttribute;
+import com.novell.ldap.LDAPException;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Iterator;
+import java.util.Map;
```
necouchman: This is totally a nitpick, but this is *almost* in alphabetical order :-).

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -221,21 +229,79 @@ public AuthenticatedUser authenticateUser(Credentials credentials)
...
     }
 
+    /**
+     * Returns all custom LDAP attributes on the user currently bound under
+     * the given LDAP connection. The custom attributes are specified in
+     * guacamole.properties.
+     *
+     * @param ldapConnection
+     *     LDAP connection to find the custom LDAP attributes.
+     * @param username
```
necouchman: Space between `@param` tags.

jaredfrees:
Sorry for my code sloppiness, I'll fix that up. Yeah I was not sure about how I did the AuthenticatedUser. Initially, I had the AuthenticatedUser interface in guacamole-ext extend Attributes, but there are other classes in extensions implementing AuthenticatedUser, so of course it was giving me a bunch of errors about those classes not defining the methods getAttributes() and setAttributes() defined in the AuthenticatedUser interface. So that's why I decided to have the AuthenticatedUser class implement Attributes instead. 

I guess a better way to do this would be to have AbstractAuthenticatedUser implement Attributes and then implement the getAttributes() and setAttributes classes in AbstractAuthenticatedUser so that there are no errors about not defining the methods in Attributes.
What do you think about that?

jaredfrees:
Do you think I should put 
`private Map<String, String> attributes = new HashMap<String, String>() `
inside of AbstractAuthenticatedUser?

jaredfrees:
Also, I would still need this line of code: 
` Map<String, String> attrs = ( (org.apache.guacamole.auth.ldap.user.AuthenticatedUser) user).getAttributes()`
because the AuthenticatedUser interface in extensions has no way of defining getAttributes() without throwing errors for no implementation of getAttributes() in other modules inside of extensions that implement AuthenticatedUser.

mike-jumper:
> Do you think I should put
>
>     private Map<String, String> attributes = new HashMap<String, String>()
>
> inside of AbstractAuthenticatedUser?

No. Providing `HashMap` storage for any subclass of `AbstractAuthenticatedUser` makes assumptions that we have been unwilling to make for other similar classes (see `AbstractUser` and `AbstractConnection`).

If you wish to provide a default implementation of these, I would return `Collections.<String, String>emptyMap()` and simply do nothing for `setAttributes()` (as the function definition states that all unsupported attributes must be ignored).

If an implementation of `AuthenticatedUser` or `AbstractAuthenticatedUser` wishes to allow other extensions to add attributes, they should be required to collaboratively support that. Similarly, an implementation that wishes to add attributes to another extension's `AuthenticatedUser` should test whether `setAttributes()` succeeds before assuming that the extension supports arbitrary attributes.

> the AuthenticatedUser interface in extensions has no way of defining getAttributes() without throwing errors for no implementation of getAttributes() in other modules inside of extensions that implement AuthenticatedUser.

I believe what we were suggesting here is modifying the core definition of `AuthenticatedUser` (within guacamole-ext) to implement the `Attributes` interface. You shouldn't need to typecast nor make assumptions regarding the internal implementation to invoke `getAttributes()`. The API itself should specify that this function is available.

jaredfrees:
The only problem is if I modify the core definition of `AuthenticatedUser ` (within guacamole-ext) to implement the `Attributes `interface, there are other modules in extensions that already implement `AuthenticatedUser ` but do not have implementations for `getAttributes()`and `setAttributes()` as required by implementing `Attributes`. So this would throw an error. Am I incorrect about this?

mike-jumper:
If other extensions within guacamole-client do not use `AbstractAuthenticatedUser` (or no default implementation is provided within `AbstractAuthenticatedUser`), you will need to modify those extensions to be compatible with your changes, yes.

jaredfrees:
Alright, will do. I just wanted to make sure it would be alright to modify other extensions. I think the only class actually is: `guacamole-client\extensions\guacamole-auth-jdbc\modules\guacamole-auth-jdbc-base\src\main\java\org\apache\guacamole\auth\jdbc\user\RemoteAuthenticatedUser.java`

necouchman:
> Sorry for my code sloppiness, I'll fix that up. 

I wouldn't worry about that - it gets better with time.  Just ask @mike-jumper how many times he had to tell me to change `if(condition)` to `if (condition)`, because "if is not a function!!" :joy:

necouchman:
@jaredfrees I see a new set of commits.  At a glance looks pretty good - will review it in detail, again, in this afternoon.  Think it's getting really close.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -221,21 +229,82 @@ public AuthenticatedUser authenticateUser(Credentials credentials)
...
+            // Get LDAP attributes by querying LDAP
+            LDAPEntry userEntry = ldapConnection.read(userDN, attrArray);
+            LDAPAttributeSet attrSet = userEntry.getAttributeSet();
+
+            // Add each attribute into Map
+            for (Object attrObj : attrSet) {
+                LDAPAttribute attr = (LDAPAttribute)attrObj;
+                String attrName = attr.getName();
+                String attrValue = attr.getStringValue();
+                attrMap.put(attrName, attrValue);
```
necouchman: Can simplify this to:

    attrMap.put(attr.getName(), attr.getStringValue());

and avoid the extra `String` variables that don't get used.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
```diff
@@ -344,4 +344,20 @@ public int getOperationTimeout() throws GuacamoleException {
         );
     }
 
+    /**
+     * Returns names for custom LDAP user attributes.
+     *
+     * @return
+     *     LDAP user attributes as defined in the guacamole.properties file
+     *     as ldap-user-attributes: ''
```
necouchman: This comment just strikes me as a little odd - I would drop the example of its guacamole.properties definition and just allow that to be documented in the manual and make this more like the other `@return` tags for other properties.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/AuthenticatedUser.java
```diff
@@ -53,6 +61,72 @@ public void init(Credentials credentials) {
         setIdentifier(credentials.getUsername());
     }
 
+    /**
+     * Get a map of attributes associated with this AuthenticatedUser.
+     *
+     * @return
+     *     The Map of arbitrary attributes associated with this
+     *     AuthenticatedUser object.
+     */
```
necouchman: Since this is overriding another method, I think it doesn't need the Javadocs, no?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/AuthenticatedUser.java
```diff
@@ -53,6 +61,72 @@ public void init(Credentials credentials) {
...
+    public Map<String, String> getAttributes() {
+        return attributes;
+    }
+
+    /**
+     * Sets a map of attributes associated with this AuthenticatedUser.
+     *
+     * @param attributes
+     *      A map of attribute key/value pairs to add to this AuthenticatedUser.
+     */
```
necouchman: Same thing as above - since we're overriding another already-documented method, here, you should be able to leave out the Javadoc comments.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/AbstractAuthenticatedUser.java
```diff
@@ -41,4 +42,12 @@ public void invalidate() {
         // Nothing to invalidate
     }
 
+    public Map<String, String> getAttributes() {
```
necouchman: So, I think there are two options here:

1. Don't implement these, here, since `AbstractAuthenticatedUser` implements `AuthenticatedUser`, which implements `Attributes`, these should already be defined at that level.  If you don't implement them, here, any classes that extend `AbstractAuthenticatedUser` will need to override them.
2. Implement them, here, so that there's a base implementation, but use the `@Override` tag so that it's understood those are overriding methods from another class (`AuthenticatedUser`).  This means you won't have to implement them in every single class that extends `AbstractAuthenticatedUser` except where you want to change this behavior.

Option 2 is definitely simpler, or at least requires fewer changes to other extensions, so that's probably the easiest to go with for the time being - that just means adding the `@Override` tag for this method and `setAttributes()`.

guacamole-ext/src/main/java/org/apache/guacamole/token/StandardTokens.java
```diff
@@ -21,9 +21,10 @@
 
 import java.text.SimpleDateFormat;
 import java.util.Date;
+import java.util.Map;
+import java.util.Set;
 import org.apache.guacamole.net.auth.AuthenticatedUser;
 import org.apache.guacamole.net.auth.Credentials;
-
```
necouchman: Add this line back in - style-wise, there should be a line between the imports and the class documentation.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -189,7 +197,7 @@ private LDAPConnection bindAs(Credentials credentials)
 
     /**
      * Returns an AuthenticatedUser representing the user authenticated by the
-     * given credentials.
+     * given credentials. Also adds custom LDAP attributes to credentials object.
```
necouchman: Might want to update this comment to reflect the change from storing in `Credentials` to `AuthenticatedUser`

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -221,21 +229,82 @@ public AuthenticatedUser authenticateUser(Credentials credentials)
             throw new GuacamoleInvalidCredentialsException("Permission denied.", CredentialsInfo.USERNAME_PASSWORD);
 
         try {
-
             // Return AuthenticatedUser if bind succeeds
             AuthenticatedUser authenticatedUser = authenticatedUserProvider.get();
             authenticatedUser.init(credentials);
+
+            //set attributes
```
necouchman: `// Set attributes`

jaredfrees:
Thanks for your help @necouchman. I fixed up all of the things you mentioned in your comments.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/AbstractAuthenticatedUser.java
jaredfrees: I think Option 2 is the way to go. Personally, I don't think we should mess so much with the other extensions.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -221,21 +230,80 @@ public AuthenticatedUser authenticateUser(Credentials credentials)
...
+     * @return
+     *     All attributes on the user currently bound under the
+     *     given LDAP connection, as a map of attribute name to
+     *     corresponding attribute value.
+     *
+     * @throws LDAPException
+     *     If an error occurs while searching for the user attributes.
+     *
+     * @throws GuacamoleException
+     *     If an error occurs retrieving the user DN.
```
necouchman: ...or an error occurs retrieving the attributes?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/AuthenticatedUser.java
```diff
@@ -53,6 +61,59 @@ public void init(Credentials credentials) {
...
+    /**
+     * Add the Map of attributes to the current set, without completely
+     * replacing the existing set.  However, if duplicate keys exist the new
+     * values will replace any existing ones.
+     *
+     * @param attributes
+     *     A Map of attributes to add to the existing attributes, without
+     *     completely overwriting them.
+     */
+    public void addAttributes(Map<String, String> attributes) {
```
necouchman: I think this method, the `getAttribute(String key)` method, and the `setAttribute(String key, String value)` method can all be removed, right?  They don't actually ever get called, it doesn't look like.

jaredfrees: I just added them because I thought they might be useful to someone else in the future, but yeah they don't ever get called.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
jaredfrees: yes you're right that is a possibility also, I'll add that.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/AuthenticatedUser.java
```diff
@@ -20,7 +20,10 @@
 package org.apache.guacamole.auth.ldap.user;
 
 import com.google.inject.Inject;
+import java.util.HashMap;
+import java.util.Map;
 import org.apache.guacamole.net.auth.AbstractAuthenticatedUser;
+import org.apache.guacamole.net.auth.Attributes;
```
necouchman: Unused import.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -221,21 +230,80 @@ public AuthenticatedUser authenticateUser(Credentials credentials)
...
+     *
+     * @param username
+     *     The username of the user whose attributes are queried.
+     *
+     * @return
+     *     All attributes on the user currently bound under the
+     *     given LDAP connection, as a map of attribute name to
+     *     corresponding attribute value.
+     *
+     * @throws LDAPException
```
necouchman: Only throws `GuacamoleException` so this should be removed.

jaredfrees:
I was just about to say I cannot get the LDAP connection to work. I've been looking at it all morning and I cannot figure out the problem. This is one error in the logs:
`ERROR o.a.g.extension.ProviderFactory - authentication provider extension failed to start: com/novell/ldap/LDAPReferralException`

I can give you the rest of the logs if you want to look at it.

necouchman:
> ERROR o.a.g.extension.ProviderFactory - authentication provider extension failed to start: com/novell/ldap/LDAPReferralException

Yeah, I think I know what's going on - have a look at the LDAP `UserService` class, specifically, here:

https://github.com/apache/guacamole-client/blob/984ab48ce8dbbb5b9949ce1f5e5f774168b4830b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/user/UserService.java#L108-L142

I'm guessing that, in addition to the `LDAPException` you're catching in the `getLDAPAttributes()` method you'll also want to specifically catch the `LDAPReferralException` types and handle those in a similar fashion to the `UserService` class.  You should just be able to add the `catch (LDAPReferralException e) {}` block above the `catch (LDAPException e) {}` block and deal with it similarly.

jaredfrees:
I tried adding the catch (LDAPReferralException e) block  above the LDAPException block like you suggested but it is still giving me the same errors. Here are some more parts of the log:
`
Jun 13 13:17:16 server: 13:17:16.294 [localhost-startStop-1] DEBUG o.a.g.extension.ExtensionModule - Loading extension: "guacamole-auth-ldap-1.0.0.jar"
Jun 13 13:17:16 server: 13:17:16.364 [localhost-startStop-1] DEBUG o.a.g.extension.ExtensionModule - [0] Binding AuthenticationProvider "org.apache.guacamole.auth.ldap.LDAPAuthenticationProvider".
Jun 13 13:17:16 server: 13:17:16.457 [localhost-startStop-1] ERROR o.a.g.extension.ProviderFactory - authentication provider extension failed to start: com/novell/ldap/LDAPReferralException
Jun 13 13:17:16 server: 13:17:16.466 [localhost-startStop-1] DEBUG o.a.g.extension.ProviderFactory - org.apache.guacamole.auth.ldap.LDAPAuthenticationProvider instantiation failed.
Jun 13 13:17:16 server: Caused by: java.lang.NoClassDefFoundError: com/novell/ldap/LDAPReferralException
Jun 13 13:17:16 server: at org.apache.guacamole.auth.ldap.LDAPAuthenticationProvider.<init>(LDAPAuthenticationProvider.java:60) ~[na:na]
Jun 13 13:17:16 server: Caused by: java.lang.ClassNotFoundException: com.novell.ldap.LDAPReferralException

`

necouchman:
You might want to try a "mvn clean package" and then completely re-deploy everything (extensions and WAR) from the ground up.  The issue you're seeing usually occurs when there is some mis-match between the files.  I've merged this current PR onto a test branch in my setup and deployed it, and it seems to be working fine for me, at least, it's loading the extensions perfectly fine.

necouchman:
Well, it's not crashing for me, but it isn't working quite right, either.  The part that pulls the attributes from LDAP works fine, but the ability to use those across the authentication providers is not working - specifically, if I try to pull them in with a JDBC connection, there's nothing there.

jaredfrees:
Well the part that adds the tokens is in `ConnectionService` in the LDAP module, where `getConnections() ` calls: `StandardTokens.addAttributeTokens(tokenFilter, attrs);.` So I'm not sure how we could make sure you can access them from other modules.

necouchman:
> Well the part that adds the tokens is in ConnectionService in the LDAP module, where getConnections() calls: StandardTokens.addAttributeTokens(tokenFilter, attrs);. So I'm not sure how we could make sure you can access them from other modules.

That's the idea of adding it in to the `AuthenticatedUser` class, so that it is accessible across all of the modules.  I'll take a look and see if I can figure out where it's missing...

jaredfrees:
After doing the "mvn clean package" and fresh installing my LDAP connection works now. 
What are you trying to do to pull in the attributes with a JDBC connection?

jaredfrees:
If a user authenticates with LDAP, the attributes will be added in `AuthenticatedUser`. This object seems to be passed around in JDBC, but it doesn't do anything with the attributes or make them visible unless we change something in the JDBC code. 
The attributes only get added as Tokens if `configurationBaseDN != null` in `onnectionService` in the LDAP module. My understanding is Tokens are the only thing that can be used on the front end by the client, correct me if I am wrong. So I'm not really sure what you are looking for.

jaredfrees:
I see what you are saying now. I think the problem was `addAttributeTokens` in `StandardTokens` was not be called in `addStandardTokens` when it should be. So the JDBC module calls `addStandardTokens` so the custom attribute map should be added now. You can test it now with the new updates.

necouchman:
> This object seems to be passed around in JDBC, but it doesn't do anything with the attributes or make them visible unless we change something in the JDBC code.

So, the change you just made *should* make it work across all of the modules; however, I had made changes which *should* have had the same effect by calling the `addAttributeTokens` method from within the JDBC module, and, from what I could tell, the attributes weren't present on the `RemoteAuthenticatedUser` object (which extends `AuthenticatedUser`), so something about the connection is missing.

What you've just changed is great - it'll allow those to get added transparently across the modules without individual changes.  I'll give it a shot and see if it works and what I can figure out.

jaredfrees:
Okay that makes sense. I'm not sure why the attributes wouldn't be present in the `RemoteAuthenticatedUser` object. I was looking around in the JDBC module and it seems like after LDAP authorizes the user in the LDAP module, it should pass the AuthenticatedUser Object to the JDBC module. But I don't fully understand how the JDBC module works, so I am not totally sure.

necouchman:
> I was looking around in the JDBC module and it seems like after LDAP authorizes the user in the LDAP module, it should pass the AuthenticatedUser Object to the JDBC module.

Yes, I agree.  However, there's some modeling/wrapping of objects done in the JDBC module, and I'm wondering if, in the process of that happening, it's clobbering the attributes variable somewhere.  Anyway, I'll try it out here, momentarily, and see what happens.

guacamole-ext/src/main/java/org/apache/guacamole/token/StandardTokens.java
```diff
@@ -74,6 +76,11 @@
      */
     private static final String TIME_FORMAT = "HHmmss";
 
+    /**
+     * The prefix of the arbitrary attribute tokens.
+     */
+    public static final String ATTR_TOKEN_PREFIX = "GUAC_ATTR:";
```
necouchman: Okay, one of the things I figured out is that the RegEx for tokens won't currently match on ":", so, either one of two things will need to be done:
- Modify the RegEx in `TokenFilter` and add the match for ":" to that.
- Change this so that it fits into the current RegEx, which would probably mean using "GUAC_ATTR_" and then the tokens become GUAC_ATTR_MAIL, GUAC_ATTR_WORKSTATIONNAME, etc.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/RemoteAuthenticatedUser.java
```diff
@@ -67,6 +69,21 @@
      */
     private static final Pattern X_FORWARDED_FOR = Pattern.compile("^" + IP_ADDRESS_REGEX + "(, " + IP_ADDRESS_REGEX + ")*$");
 
+    /**
+     * Arbitrary attributes associated with this RemoteAuthenticatedUser object.
+     */
+    private Map<String, String> attributes = new HashMap<String, String>();
```
necouchman: I'm not sure if it's better to put this here, in this abstract class, or if this (along with the `@Override`s below) should only be done on the implementing classes, like `ModeledAuthenticatedUser`?  @mike-jumper any guidance?

jaredfrees: The problem with the `RemoteAuthenticatedUser` is it implements AuthenticatedUser.

guacamole-ext/src/main/java/org/apache/guacamole/token/StandardTokens.java
jaredfrees: I think I'll change it so it fits into the current RegEx.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/RemoteAuthenticatedUser.java
jaredfrees: and currently `AuthenticatedUser` extends `Attributes`

necouchman: Yes, but `RemoteAuthenticatedUser` is abstract, so the `getAttributes()` and `setAttributes()` methods do not have to be implemented in it - they can be, as you've done, here, but they can also be implemented by all of the implementing classes.  Obviously implementing them in `RemoteAuthenticatedUser` requires fewer lines of code, as then you don't have to implement in both `SharedAuthenticatedUser` and `ModeledAuthenticatedUser`, I'm just not sure how "clean" this is, or if the `RemoteAuthenticatedUser` class is really the right place for implementing these attribute handlers.

jaredfrees: Oh yes you are right. I didn't notice this. I am going to try to set the attributes in `ModeledAuthenticatedUser`

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledAuthenticatedUser.java
```diff
@@ -26,6 +26,11 @@
 import org.apache.guacamole.net.auth.AuthenticationProvider;
 import org.apache.guacamole.net.auth.Credentials;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import java.util.Map;
```
necouchman: Looks like some left-overs from debugging?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledAuthenticatedUser.java
```diff
@@ -79,6 +84,7 @@ public ModeledAuthenticatedUser(AuthenticatedUser authenticatedUser,
         super(authenticatedUser.getAuthenticationProvider(), authenticatedUser.getCredentials());
         this.modelAuthenticationProvider = modelAuthenticationProvider;
         this.user = user;
+        this.setAttributes(authenticatedUser.getAttributes());
```
necouchman: I think this properly should be:

    super.setAttributes(authenticatedUser.getAttributes());

since `setAttributes()` is defined in `RemoteAuthenticatedUser`.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledAuthenticatedUser.java
jaredfrees: yep, my bad.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPGuacamoleProperties.java
```diff
@@ -205,4 +205,14 @@ private LDAPGuacamoleProperties() {}
 
     };
 
+    /**
+     * Custom attribute or attributes in Guacamole user's record in the
```
necouchman: Maybe something more like:
> Custom attribute or attributes to query from Guacamole user's record...

guacamole-ext/src/main/java/org/apache/guacamole/token/StandardTokens.java
```diff
@@ -164,6 +172,33 @@ public static void addStandardTokens(TokenFilter filter, AuthenticatedUser user)
...
+        addAttributeTokens(filter, user.getAttributes());
+    }
+
+    /**
+     * Add attribute tokens to StandardTokens.  These are arbitrary
+     * key/value pairs that may be configured by the various authentication
+     * extensions.
+     *
+     * @param filter
+     *     The TokenFilter to add attributes tokens to.
```
necouchman: "attribute tokens"

jaredfrees:
Thanks for all the help @necouchman !

jaredfrees:
Can I update the Guacamole Manual here: https://guacamole.apache.org/doc/gug/ldap-auth.html#idm140500641772816
or is this someone elses job?

necouchman:
Yes, please do!  The manual is in its own repository (guacamole-manual) in Docbook (=xml) format, and you're welcome to update it and submit a PR for that, as well.

PR #301, opened 2018-06-12
GUACAMOLE-360: Allow users to kill their own sessions

necouchman:
This is the first of a few changes related to GUACAMOLE-360 - this one makes a few minor changes that allow users to kill their own sessions.


extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/activeconnection/ActiveConnectionService.java
```diff
@@ -110,21 +110,21 @@ public TrackedActiveConnection retrieveObject(ModeledAuthenticatedUser user,
...
-        if (!user.getUser().isAdministrator())
-            throw new GuacamoleSecurityException("Permission denied.");
-
+        
         // Close connection, if it exists (and we have permission)
         ActiveConnection activeConnection = retrieveObject(user, identifier);
-        if (activeConnection != null) {
+        if (activeConnection != null && 
+                (user.getUser().isAdministrator() 
+                || user.getIdentifier().equals(activeConnection.getUsername()))) {
```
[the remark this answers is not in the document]
necouchman: Okay, I've taken a stab at implementing this, though it was slightly different from the `ConnectionService` and `UserService` object types.  Those seem to extend the `ModeledChildDirectoryObjectService`, which has abstracts in it for the `hasObjectPermissions()` and `getPermissionSet()` methods.  `ActiveConnectionService`, on the other hand, extends `DirectoryObjectService`, which does not have such methods.  So, I wasn't sure if just implementing them in `ActiveConnectionService` as `private` methods was okay, if `ActiveConnectionService` should somehow be reworked to extend the `ModeledChildDirectoryObjectService` class (this doesn't seem right since `ActiveConnection` objects aren't modeling/modeled by anything), or if `DirectoryObjectService` should add some method declarations for the permission checks?

Also, not sure if there is anywhere in the JS code that I should modify to make use of these permisisons...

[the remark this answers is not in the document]
necouchman: > Also, not sure if there is anywhere in the JS code that I should modify to make use of these permisisons...

Updated the JS code to only show session management if the permission exists.  Answered my own question.

necouchman:
Okay, updated things to use PermissionSet stuff to handle this.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/activeconnection/ActiveConnectionService.java
```diff
@@ -111,20 +113,19 @@ public TrackedActiveConnection retrieveObject(ModeledAuthenticatedUser user,
...
+                && hasObjectPermissions(user, identifier, ObjectPermission.Type.DELETE)) {
 
             // Close connection if not already closed
             GuacamoleTunnel tunnel = activeConnection.getTunnel();
             if (tunnel != null && tunnel.isOpen())
                 tunnel.close();
 
         }
+        else
+            throw new GuacamoleSecurityException("Permission denied.");
```
mike-jumper: This technically goes against [the definition of `deleteObject()` on the interface](https://github.com/apache/guacamole-client/blob/984ab48ce8dbbb5b9949ce1f5e5f774168b4830b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/DirectoryObjectService.java#L103-L104) which states (emphasis added):

> Deletes the object having the given identifier. **If no such object exists, this function has no effect.**

The `GuacamoleSecurityException` should only be thrown if permission is denied to delete the object. If no such object exists at all, the function should silently do nothing.

necouchman: Adjusted to behave as defined.

guacamole/src/main/webapp/app/navigation/services/userPageService.js
```diff
@@ -257,15 +257,23 @@ angular.module('navigation').factory('userPageService', ['$injector',
...
             ) {
-                canManageSessions.push(dataSource);
                 canViewConnectionRecords.push(dataSource);
             }
 
+            // Determine whether the current user needs access to view session management
+            if (
+                    // Permission to manage active sessions.
+                       PermissionSet.hasSystemPermission(permissions,           PermissionSet.SystemPermissionType.ADMINISTER)
+                    || PermissionSet.hasActiveConnectionPermission(permissions, PermissionSet.ObjectPermissionType.DELETE)
```
mike-jumper: This will result in the session management tab only appearing if at least one killable connection is active. Once that connection is killed, the tab will cease to appear (after any cached permissions have been refreshed).

Is this the intended behavior?

necouchman: That's what I was going for, yes, but I don't really have a strong attraction to that behavior if you think it should act some other way?

mike-jumper: I'm not sure. Only displaying the tab if manageable connections exist definitely makes sense in theory, but the behavior in practice feels inconsistent to me:

https://youtu.be/yiVlfRtZRK4

What do you think about always displaying that tab?

necouchman: Always displaying it is fine with me.  I actually originally implemented it that way, but swapped over to checking for active connections and permissions.

The only side-effect that I see to this is dealing with the issue of behavior when only a single connection is present and going to the home screen vs. automatically connecting.  In general, this wouldn't be an issue - we can adjust the number of pages it looks at up one since we're adding one more that is there for every user - however, there's a corner-case, here, that should at least be considered, even if we ignore it.

Say the user only has access to a single connection, and that connection is open on another system.  The user wants to log in and kill that session, but when they log in they get put into the connection automatically, disconnecting the one on the other client, which starts the loop of connect/disconnect between clients.

Again, it's a corner-case, but worth at least figuring out what we want to do overall with automatic connections, whether this tab is always displayed, etc.

mike-jumper: For that case, I think we're relatively stuck until something like the proposed automatic-joining functionality is implemented.

I don't think there should be a disconnect loop, though. IIRC, the recent improvements to RDP error handling will cause Guacamole to not automatically reconnect when the user is kicked off specifically because another user has started a session.

mike-jumper: Also regarding that corner case, such a connection could be configured to allow only one concurrent user. The user attempting to connect would be disallowed with a warning that the connection is in use. If they wish to kick off the existing session (and have permission to do so), they could then open the menu with Ctrl+Alt+Shift, go to Settings, and manually kill it.

necouchman: Okay, changed so it will always show the session management page under Settings.

guacamole/src/main/webapp/app/navigation/services/userPageService.js
```diff
@@ -257,25 +259,16 @@ angular.module('navigation').factory('userPageService', ['$injector',
...
                     PermissionSet.hasSystemPermission(permissions, PermissionSet.SystemPermissionType.ADMINISTER)
             ) {
-                canManageSessions.push(dataSource);
                 canViewConnectionRecords.push(dataSource);
             }
 
         });
 
-        // If user can manage sessions, add link to sessions management page
-        if (canManageSessions.length) {
```
mike-jumper: With these deletions, is `canManageSessions` used anywhere else anymore?

necouchman: Yep, you're right.  Removed.

guacamole/src/main/webapp/app/navigation/services/userPageService.js
```diff
@@ -312,6 +305,12 @@ angular.module('navigation').factory('userPageService', ['$injector',
             url  : '/settings/preferences'
         }));
 
+        // Add link to Session management (always accessible)
+        pages.push(new PageDefinition({
```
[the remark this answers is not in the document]
necouchman: Ah, yeah, that makes sense.  Should be ordered correctly, now.

mike-jumper:
Looks pretty good. Before moving ahead with merge: Do you think any change is necessary to the help text on the session management tab? Currently it states:

> All currently-active Guacamole sessions are listed here. If you wish to kill one or more sessions, check the box next to those sessions and click "Kill Sessions". Killing a session will immediately disconnect the user from the associated connection.

Most of the above seems to still apply, but I don't think "all currently-active Guacamole sessions" is accurate any longer.

necouchman:
Took a stab at updating the language, and gave it the ol' Google Translate try to get it into the 8 or so different languages we have there.

guacamole/src/main/webapp/translations/fr.json
```diff
@@ -624,7 +624,9 @@
         
         "FORMAT_STARTDATE" : "@:APP.FORMAT_DATE_TIME_PRECISE",
 
-        "HELP_SESSIONS" : "Toutes les connexions actives Guacamole sont listées ici. Si vous souhaitez en fermer une ou plusieurs, sélectionner les et cliquer sur \"Fermer Sessions\". La fermeture d'une session déconnectera immédiatement l'utilisateur.", 
+        "HELP_SESSIONS" : "
+137/5000
```
mike-jumper: What happened here?

guacamole/src/main/webapp/translations/en.json
```diff
@@ -750,7 +750,7 @@
         
         "FORMAT_STARTDATE" : "@:APP.FORMAT_DATE_TIME_PRECISE",
 
-        "HELP_SESSIONS" : "All currently-active Guacamole sessions are listed here. If you wish to kill one or more sessions, check the box next to those sessions and click \"Kill Sessions\". Killing a session will immediately disconnect the user from the associated connection.",
+        "HELP_SESSIONS" : "Your currently-active Guacamole sessions are listed here, and Administrators will be able to see currently-active sessions for all users.  If you wish to kill one or more sessions, check the box next to those sessions and click \"Kill Sessions\". Killing a session will immediately disconnect the user from the associated connection.",
```
mike-jumper: This is certainly true for the database authentication extensions, but not necessarily true in the general case. Extensions are free to define their own systems for determining who has permission to view/kill a particular session, so long as they properly expose that permission through the relevant permission sets.

For other parts of the administrative interface, the wording of the help text both describes what may be done and that those actions depend on access level. For example, from the connection interface:

> Click or tap on a connection below to manage that connection. Depending on your access level, connections can be added and deleted, and their properties (protocol, hostname, port, etc.) can be changed.

The new wording is a step in the right direction, but I don't think we should assert that the permissions work a specific way, even if it's the most common implementation of those permissions. Perhaps similar "depending on access level", etc. wording would be useful here?

necouchman: How about:

> This page will be populated with currently-active connections.  The connections listed and the ability to kill those connections is dependent upon your access level. ...

That sound okay?

guacamole/src/main/webapp/translations/fr.json
necouchman: Copypasta.  From Google Translate.  Will fix when I fix the wording.

mike-jumper:
Looks good! I have no way to verify the results of using Google Translate, but hopefully the translation is OK.

necouchman:
> I have no way to verify the results of using Google Translate, but hopefully the translation is OK.

Yeah, me, either, but hopefully others in the community who have contributed to the various translations can jump in and fix things if they're off.

PR #306, opened 2018-07-21
GUACAMOLE-220: Display readable connections only by default within permission editor.

mike-jumper:
In the interest of reducing clutter within the group editor (which has even more sections than the already-somewhat-cluttered user editor), this change splits the "Connections" section into tabs such that the view can be limited to only the connections currently accessible by that user. This reduced view is selected by default for users which have access to at least one connection.

This has the effect that the default view within the editor (1) functions as a report of what the user may currently access and (2) allows direct management of the most relevant permissions:

guac-user-current-connections

To grant access to connections, connection groups, etc. which are not already accessible nor nested within already accessible groups, the administrator can switch to the "All Connections" view:

guac-user-all-connections

For any user that does not have access to any connections, including new users, the "All Connections" view is selected by default:

guac-new-user-all-connections

It is intended that these views represent the actual, current state of things, not the state of unsaved changes made within the editor. Each view is thus populated at the time the editor loads. Connections will not appear within nor disappear from the "Current Connections" view as a result of unsaved changes made within the editor.

Part of all this required modifying the `guacGroupListFilter` directive to preserve the type of its input, which varies between `ConnectionGroup` and `GroupListItem` depending on usage. It previously always converted to `ConnectionGroup` by unwrapping any given `GroupListItem`, thus undoing any filtering performed only on the `GroupListItem` hierarchy.

### DIFF
--- a/guacamole/src/main/webapp/app/groupList/directives/guacGroupListFilter.js
+++ b/guacamole/src/main/webapp/app/groupList/directives/guacGroupListFilter.js
@@ -30,9 +30,10 @@ angular.module('groupList').directive('guacGroupListFilter', [function guacGroup
 
             /**
              * The property to which a subset of the provided map of connection
-             * groups will be assigned.
+             * groups will be assigned. The type of each item within the
+             * original map is preserved within the filtered map.
              *
-             * @type Array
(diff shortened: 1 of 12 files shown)


necouchman:
Concept looks great to me.  Will try to review the code today.

PR #307, opened 2018-07-27
GUACAMOLE-598: Display fatal error if a page is unusable / nonfunctional.

mike-jumper:
These changes add a new `requestService.DIE` constant which can be used in favor of `requestService.WARN` if failure of the request in question will result in the page being useless. The various usages of `requestService.WARN` have been updated to `requestService.DIE` where appropriate.

Unlike `requestService.WARN`, which only logs a warning to the browser console, `requestService.DIE` replaces the current page with a generic error:

guac-generic-error


necouchman:
Just FYI - there does seem to be one minor side-effect to this change.  One of my Guacamole installs is located somewhat remote from me - across a VPN tunnel and WAN connection - and when I initially load the page I get the "triangle-exclamation-point" image for a couple of seconds until it has loaded enough of the content that the login box loads.  Not sure if this is intended or a big deal, but just thought I'd share it.

mike-jumper:
That's a bug alright. I'll copy this over to the JIRA.

PR #309, opened 2018-08-08
GUACAMOLE-220: Add user interface for managing user groups and membership.

mike-jumper:
This change:

* Adds a new tab for managing user groups, similar to the existing user management tab.
* Adds a new directive which provides an interface for manipulating whether an object shares some relation with another object (in this case group membership relations).
* Leverages the above directive to add support for manipulating a user's group memberships, as well as the group memberships of other groups.
* Upgrades both jQuery and Lodash to the latest versions. (The version of Lodash previously included with Guacamole lacks the `sortedIndexOf()` function, which is needed by the new directive mentioned above. If upgrading Lodash, we may as well upgrade jQuery while we're at it, as the version previously included within Guacamole is no longer maintained.)

With these changes in place, administrators will see a new "Groups" tab next to the existing "Users" tab:

guac-groups-tab

When edited, each group appears largely like a user, with sets of permissions that can be granted to the group. In addition to the usual system and connection permission sections, there are three group membership sections which allow parent groups and member users/groups to be manipulated:

guac-group-membership-edit

Clicking the "X" next to the name of a user or group marks that object for removal from that membership relation, to take effect once the admin clicks "Save". A similar new section is now present in the user editor for manipulating group membership at the user level:

guac-user-membership-edit

These membership editor sections are collapsed by default. The intent there is to reduce clutter within the user and group edit screens while also providing admins with a quick view into the current state of a user or group's membership. When expanded, a full list of available users or groups is displayed, with checkboxes allowing those users/groups to be added/removed:

guac-membership-editor-expanded

Each of these sections is displayed only if there are applicable users or groups.

The filters in the section headers have the effect you would expect, reducing the users/groups visible in both the abbreviated and expanded views of the editor.


guacamole/src/main/webapp/app/manage/controllers/manageUserGroupController.js
```diff
@@ -0,0 +1,535 @@
...
+     * Loads skeleton user group data, preparing the interface for creating a
+     * new user group.
+     *
+     * @returns {Promise}
+     *     A promise which is resolved when the interface has been prepared for
+     *     creating a new user group.
+     */
+    var loadSkeletonUserGroup = function loadSkeletonUserGroup() {
+
+        // No user groups exist regardless of data source if there is no identifier
```
jmuehlner: It's a little bit unclear what "if there is no identifier" means in this context.

mike-jumper: Fair enough. I'll reword and amend.

guacamole/src/main/webapp/app/manage/directives/identifierSetEditor.js
```diff
@@ -0,0 +1,299 @@
...
+         * identifier is currently present (true) or absent (false). If an
+         * identifier is absent, it may also be absent from this map.
+         *
+         * @type Object.<String, Boolean>
+         */
+        $scope.identifierFlags = {};
+
+        /**
+         * Map of identifiers to boolean flags indicating whether that
+         * identifier is editabl. If an identifier is not editable, it will be
```
jmuehlner: Typo: "editabl".

mike-jumper: Fixed via rebase.

guacamole/src/main/webapp/app/manage/directives/identifierSetEditor.js
```diff
@@ -0,0 +1,299 @@
...
+            // Convert array of identifiers into set of boolean
+            // presence/absence flags
+            $scope.identifierFlags = {};
+            angular.forEach(identifiers, function storeIdentifierFlag(identifier) {
+                $scope.identifierFlags[identifier] = true;
+            });
+
+        });
+
+        // Keep identifierFlags up to date when identifiers array is replaced
```
jmuehlner: Looks like this  comment is just copypasta'd from the one above? I don't think it applies here.

mike-jumper: 🍝 😢

mike-jumper: OK, I've rebased away the pasta.

jmuehlner:
Other than a few minor issues, this generally looks pretty good to me. Any concerns, @necouchman?

guacamole/src/main/webapp/app/settings/directives/guacSettingsUserGroups.js
```diff
@@ -0,0 +1,270 @@
...
+         *     creation is not allowed.
+         */
+        $scope.getDefaultDataSource = function getDefaultDataSource() {
+
+            // Abort if permissions have not yet loaded
+            if (!permissions)
+                return null;
+
+            // For each data source
+            for (var dataSource in permissions) {
```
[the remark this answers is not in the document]
mike-jumper: Ick. Yeah, this is definitely non-detereministic. The user settings page will have the same issue. I will try to make this reasonable.

guacamole/src/main/webapp/app/settings/directives/guacSettingsUserGroups.js
```diff
@@ -0,0 +1,272 @@
...
+         *     key is the data source identifier from which the user groups
+         *     were retrieved and each value is a map of user group identifiers
+         *     to their corresponding @link{UserGroup} objects.
+         */
+        var setDisplayedUserGroups = function setDisplayedUserGroups(permissions, userGroups) {
+
+            var addedUserGroups = {};
+            $scope.manageableUserGroups = [];
+
+            // For each user in each data source
```
necouchman: For each user or each user group?  Looks like this is looping through user groups in data sources....

guacamole/src/main/webapp/app/manage/templates/manageUserGroup.html
```diff
@@ -0,0 +1,102 @@
+
```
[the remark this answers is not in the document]
mike-jumper: Nope. I'll get rid of it.

guacamole/src/main/webapp/app/rest/services/membershipService.js
```diff
@@ -0,0 +1,385 @@
...
+        // Clear the cache
+        .then(function parentUserGroupsChanged(){
+            cacheService.users.removeAll();
+        });
+
+    };
+
+    /**
+     * Makes a request to the REST API to retrieve the identifiers of all
+     * users which are members of the given user group, returning a promise
```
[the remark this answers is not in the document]
mike-jumper: That aspect is intentionally left to the implementation. While there is a similar call for retrieving permissions which explicitly requires permission inheritance to be taken into account, no assumption is made in general on how that group inheritance is implemented, nor whether it's taken into account here.

In this case, if an extension wants to expose inherited groups at this level, they can. The only requirement is that they make the set of groups mutable if the current user has permission to make changes to those memberships.

guacamole/src/main/webapp/app/rest/types/UserGroup.js
```diff
@@ -0,0 +1,59 @@
...
+        this.identifier = template.identifier;
+
+        /**
+         * Arbitrary name/value pairs which further describe this user group.
+         * The semantics and validity of these attributes are dictated by the
+         * extension which defines them.
+         *
+         * @type Object.<String, String>
+         */
+        this.attributes = {};
```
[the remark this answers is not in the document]
mike-jumper: Not in practice, but it would make sense to do this. I'm not sure what this wasn't done in the first place.

guacamole/src/main/webapp/app/manage/templates/manageUserGroup.html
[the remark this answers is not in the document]
mike-jumper: Extra blank removed. GitHub has unfortunately now reassigned this comment to the intentional blank line following the first `<div>`, but the original issue should be addressed.

mike-jumper:
OK - I believe I've corrected (eaten?) all the pasta.

mike-jumper:
They won't work with the JDBC modules quite yet, but they should fail cleanly (like with the "Permission Denied" message you're seeing). They will work once the schema changes, etc. for the JDBC modules are made available through PR and merge, though, so hopefully these UI changes will ease that upcoming review.

I've kept those changes separate since they're enormous and (technically) independent.

necouchman:
Cool, sounds good.  @jmuehlner any other concerns, or good proceeding with the merge?

jmuehlner:
@necouchman Everything looks good to me! 👍

PR #310, opened 2018-08-24
GUACAMOLE-605: Morphia extension and Morphia-MongoDB extension

daniquir:


daniquir:
The initial idea was to follow the philosophy of the JDBC module, but I do not know that other databases can use the morphia module.

In principle, the generic morphia module has been created to be added to a specific database module, as in the JDBC module.

I think that making it more abstract would not be appropriate if it depends on Morphia and the databases it supports.
I am open to opinions and if they consider that it is better to change it, I will change it without problem.

To finish the intention is to give the possibility of using Guacamole with a database of MongoDB, if someone wants to use it to include other databases that are supported by Morphia, it is welcome.

necouchman:
It looks like Morphia is specific to MongoDB - it appears that perhaps they tried to write it with the concept of abstracting away from MongoDB a bit, but, in practice, there aren't any other supported databases.  The only other one I found that might work is Spring - it looks like Spring has the capability to do CouchDB and MongoDB, and maybe some others.

At this point I think we can push forward with reviewing this and getting it into code - if someone wants to come back later and write a module that supports multiple ones, that'll be fine, but since you have this working, no reason not to get it reviewed and into the code base.  I'll try to review it over the weekend.

mike-jumper:
Before this gets too far through process, I think it's worth looking into whether some of the common functionality could be brought out into a library shared by both, perhaps even something which is available as a dependency via Maven (similar to guacamole-ext).

The base permission enforcement functionality, connection tracking/sharing, etc. theoretically could be extracted such that it could be reused. That would reduce the duplication here, as well as allow downstream extensions to avoid having to reimplement all that.

We could eventually leverage this to add similar functionality to the various other extensions that support connection storage, namely LDAP and some hypothetical future version of the file-based auth (something actually meant at production use, not the current XML and MD5/plaintext incarnation aimed).

necouchman:
>  I think it's worth looking into whether some of the common functionality could be brought out into a library shared by both

I think this is a great idea, and a good time to do it.

daniquir:
It seems to me the best way to integrate both parties and prevent similar cases in the future. Any improvement will be good for the project. I will try to perform some test by extracting the common code.

necouchman:
>  I will try to perform some test by extracting the common code.

We will also want a new JIRA issue for this effort.  I'll create one and add the folks on this PR to it.

necouchman:
GUACAMOLE-617 has been created to track this effort: https://issues.apache.org/jira/browse/GUACAMOLE-617

PR #312, opened 2018-09-07
GUACAMOLE-567: Add specific connection stability test to tunnel implementations.

mike-jumper:
This change corrects the behavior of the connection instability detection by adding tunnel-specific tests which are independent of the underlying Guacamole connection. Regardless of whether the remote desktop connection itself is returning data, these tests verify that the server side of the tunnel is reachable and responding.

For the WebSocket tunnel, these tests leverage the reserved opcode for tunnel-specific instructions to send a "ping" message which the tunnel is required to respond to. By sending these messages roughly every 500ms, the client side guarantees that there should always be traffic over the tunnel unless something is interrupting that traffic.

For the HTTP tunnel, these tests leverage the fact that HTTP already has a request/response pattern that is verified via the HTTP protocol. A simple "nop" instruction is sent to the Guacamole server to force the send half of the tunnel to effectively self test. If the server side of the HTTP tunnel fails to properly handle the HTTP request which results from sending an instruction, the tunnel is flagged as unstable.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Tunnel.js
+++ b/guacamole-common-js/src/main/webapp/modules/Tunnel.js
@@ -153,7 +153,8 @@ Guacamole.Tunnel = function() {
  * use by tunnel implementations. The value of this opcode is guaranteed to be
  * the empty string (""). Tunnel implementations may use this opcode for any
  * purpose. It is currently used by the HTTP tunnel to mark the end of the HTTP
- * response, and by the WebSocket tunnel to transmit the tunnel UUID.
+ * response, and by the WebSocket tunnel to transmit the tunnel UUID and send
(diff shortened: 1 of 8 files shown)


necouchman:
One question - is this code still Java 1.6 compatible?  I was looking up the documentation on the RemoteEndpoint.Basic class, and I don't see it for 1.6 - only 1.7.  Do we need to bump the target up?

mike-jumper:
It is, or at least referencing that class doesn't itself require Java 1.7. That class isn't part of Java 1.7 so much as it is part of recent versions of servlet API, which the servlet container may or may not provide. Dynamic detection of WebSocket support takes place during Guacamole's startup.

PR #336, opened 2018-10-28
GUACAMOLE-641: Add support for populating arbitrary parameter tokens from key vaults.

mike-jumper:
These changes add a new family of extensions, similar to "guacamole-auth-jdbc", which provide support for retrieval of secrets from key vaults: "guacamole-vault". Initial support for Azure Key Vault is present through the "guacamole-vault-azure" module, with the necessary structure in place to allow other implementations to be provided in the future.

The general support works as follows:

1. A YAML file is included within `GUACAMOLE_HOME` which defines a token/secret mapping. Besides defining tokens, the names of each secret may also contain tokens which allow the secret name to vary by connection ID, hostname, username, etc. There is a specific set of tokens available for use within secret names.
2. When a user attempts to connect to a connection or connection group, the key vault implementation is queried to retrieve each defined secret. Tokens are then defined for only those secrets whose names are fully defined within the current context and which have values stored within the vault.
3. The extension defining the connection is given these tokens. If the extension supports substitution of parameter tokens, as all current extensions do, those tokens will be substituted within the configuration of the connection. Guacamole extensions are not required to do this, however.
4. To avoid excessive retrieval requests, the result of retrieving a particular secret is cached (by default for 10 seconds).

This thus allows secret values like passwords and private keys to be stored off-site within the vault and retrieved dynamically based on context.


extensions/guacamole-auth-vault/modules/guacamole-auth-vault-base/src/main/java/org/apache/guacamole/auth/vault/user/VaultUserContext.java
```diff
@@ -0,0 +1,325 @@
...
+    /**
+     * Waits for all pending secret retrieval operations to complete,
+     * transforming each Future within the given Map into its contained String
+     * value.
+     *
+     * @param pendingTokens
+     *     A Map of token name to Future, where each Future represents the
+     *     pending retrieval operation which will ultimately be completed with
+     *     the value of all secret mapped to that token.
+     *
```
necouchman: Missing `@return` tag.

extensions/guacamole-auth-vault/.gitignore
```diff
@@ -0,0 +1,2 @@
+target/
```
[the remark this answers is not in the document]
mike-jumper: I'll recheck - it may well not be needed.

extensions/guacamole-auth-vault/modules/guacamole-auth-vault-azure/src/main/java/org/apache/guacamole/auth/vault/azure/AzureKeyVaultAuthenticationProviderModule.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Guice module which configures injections specific to Azure Key Vault
+ * support.
+ */
+public class AzureKeyVaultAuthenticationProviderModule
+        extends VaultAuthenticationProviderModule {
+
+    /**
+     * Creates a new AzureKeyVaultAuthenticationiProviderModule which
```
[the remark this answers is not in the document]
mike-jumper: OK - will fix.

extensions/guacamole-auth-vault/modules/guacamole-auth-vault-azure/src/main/java/org/apache/guacamole/auth/vault/azure/conf/AzureKeyVaultAuthenticationException.java
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Unchecked exception thrown by AzureKeyVaultCredentials if an error occurs
+ * during the authentication process. Note that the base KeyVaultCredentials
+ * base class does not provide for checked exceptions within the authentication
+ * process.
+ *
+ * @see AzureKeyVaultCredentials#doAuthenticate(java.lang.String, java.lang.String, java.lang.String)
+ */
+public class AzureKeyVaultAuthenticationException extends RuntimeException {
```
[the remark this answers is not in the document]
mike-jumper: Yep. From the top of the class:

https://github.com/apache/guacamole-client/blob/4d90b34732d81efd1fbdeab8df9d9edb939f6266/extensions/guacamole-auth-vault/modules/guacamole-auth-vault-azure/src/main/java/org/apache/guacamole/auth/vault/azure/conf/AzureKeyVaultAuthenticationException.java#L22-L30

The API provided for Azure does not allow for checked exceptions like `GuacamoleException` to be thrown. Only unchecked exceptions can be used for things which must be thrown within callbacks, etc. in their API. We catch this specific exception and translate into a `GuacamoleException` before things make their way back out to Guacamole.

extensions/guacamole-auth-vault/modules/guacamole-auth-vault-azure/src/main/java/org/apache/guacamole/auth/vault/azure/conf/AzureKeyVaultCredentials.java
```diff
@@ -0,0 +1,115 @@
...
+    @Override
+    public String doAuthenticate(String authorization, String resource,
+            String scope) throws AzureKeyVaultAuthenticationException {
+
+        // Read Azure credentials from guacamole.properties
+        ClientCredential credentials;
+        try {
+            credentials = confService.getClientCredentials();
+        }
+        catch (GuacamoleException e) {
```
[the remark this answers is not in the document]
mike-jumper: No rationale, but a hard requirement. `doAuthenticate()` is defined by us but of the Azure API we're using in this extension. We can't add new checked exceptions like `GuacamoleException` to the prototype of a function we're implementing that's defined by someone else's interface. Ideally, that third-party interface would provide some checked exception that we *could* throw, but no dice - only unchecked exceptions are available for use within `doAuthenticate()`.

The code elsewhere in this extension that deals with the part of Azure's API which invokes `doAuthenticate()` has to be conscious of this, catch our special unchecked exception, and translate things back into an appropriate `GuacamoleException.

extensions/guacamole-auth-vault/modules/guacamole-auth-vault-azure/src/main/java/org/apache/guacamole/auth/vault/azure/secret/AzureKeyVaultSecretService.java
```diff
@@ -0,0 +1,121 @@
...
+    /**
+     * Provider for Azure Key Vault credentials.
+     */
+    @Inject
+    private Provider<KeyVaultCredentials> credentialProvider;
+
+    /**
+     * {@inheritDoc}
+     *
+     * <p>Azure Key Vault allows strictly a-z, A-Z, 0-9, and "-". This
```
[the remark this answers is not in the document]
mike-jumper: Yep, exactly right.

extensions/guacamole-auth-vault/modules/guacamole-auth-vault-base/src/main/java/org/apache/guacamole/auth/vault/user/VaultUserContext.java
mike-jumper: Whoops. Will fix.

extensions/guacamole-auth-vault/modules/guacamole-auth-vault-base/src/main/java/org/apache/guacamole/auth/vault/user/VaultUserContext.java
```diff
@@ -0,0 +1,325 @@
...
+/**
+ * UserContext implementation which automatically injects tokens containing the
+ * values of secrets retrieved from a vault.
+ */
+public class VaultUserContext extends TokenInjectingUserContext {
+
+    /**
+     * Logger for this class.
+     */
+    private final Logger logger = LoggerFactory.getLogger(VaultUserContext.class);
```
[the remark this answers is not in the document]
mike-jumper: I'll check SLF4J docs for best practices.

extensions/guacamole-auth-vault/modules/guacamole-auth-vault-azure/pom.xml
```diff
@@ -0,0 +1,174 @@
...
+
+        <!-- Guacamole Extension API -->
+        <dependency>
+            <groupId>org.apache.guacamole</groupId>
+            <artifactId>guacamole-ext</artifactId>
+            <scope>provided</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>slf4j-api</artifactId>
```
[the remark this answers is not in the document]
mike-jumper: You're right - it shouldn't be excluded. It's not that we want it pulled in, though, but that it's provided already by the Guacamole webapp just like guacamole-ext and guacamole-common.

We don't want it fully excluded, but we also don't want it bundled with the extension. The `<scope>provided</scope>` for the guacamole-ext dependency takes care of that.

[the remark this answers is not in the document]
mike-jumper: I'll remove the unnecessary exclusion.

extensions/guacamole-auth-vault/modules/guacamole-auth-vault-base/src/main/java/org/apache/guacamole/auth/vault/VaultAuthenticationProviderModule.java
```diff
@@ -0,0 +1,99 @@
...
+ */
+public abstract class VaultAuthenticationProviderModule extends AbstractModule {
+
+    /**
+     * Guacamole server environment.
+     */
+    private final Environment environment;
+
+    /**
+     * Creates a new AzureKeyVaultAuthenticationProviderModule which
```
necouchman: Copypasta?  I think this should create a new `VaultAuthenticationProviderModule`, not the Azure version?

extensions/guacamole-auth-vault/modules/guacamole-auth-vault-base/src/main/java/org/apache/guacamole/auth/vault/VaultAuthenticationProviderModule.java
```diff
@@ -0,0 +1,99 @@
...
+import org.apache.guacamole.environment.Environment;
+import org.apache.guacamole.environment.LocalEnvironment;
+import org.apache.guacamole.net.auth.UserContext;
+
+/**
+ * Guice module which configures injections specific to the base support for
+ * key vaults. When adding support for a key vault provider, a subclass
+ * specific to that vault implementation will need to be created.
+ *
+ * @see AzureKeyVaultAuthenticationProviderModule
```
necouchman: Is it normal to reference child implementation classes?  Might be, just making sure.

extensions/guacamole-auth-vault/modules/guacamole-auth-vault-base/src/main/java/org/apache/guacamole/auth/vault/conf/VaultConfigurationService.java
```diff
@@ -0,0 +1,106 @@
...
+     * @param tokenMappingFilename
+     *     The name of the JSON file containing the token/secret mapping.
+     */
+    protected VaultConfigurationService(String tokenMappingFilename) {
+        this.tokenMappingFilename = tokenMappingFilename;
+    }
+
+    /**
+     * Returns a mapping dictating the name of the secret which maps to each
+     * parameter token. In the returned mapping, the the value of each entry is
```
necouchman: An extra extra "the", here, I think.

extensions/guacamole-auth-vault/modules/guacamole-auth-vault-base/src/main/java/org/apache/guacamole/auth/vault/user/VaultUserContext.java
```diff
@@ -0,0 +1,329 @@
...
+     * Logger for this class.
+     */
+    private static final Logger logger = LoggerFactory.getLogger(VaultUserContext.class);
+
+    /**
+     * The name of the token which will be replaced with the username of the
+     * current user if specified within the name of a secret. This token
+     * applies to both connections and connection groups.
+     */
+    private static final String USERNAME_TOKEN = "GUAC_USERNAME";
```
necouchman: I think this is probably defined in an upstream token class?  Is there a rationale for redefining it, here, or at least for not referencing the upstream implementation?

extensions/guacamole-auth-vault/modules/guacamole-auth-vault-base/src/main/java/org/apache/guacamole/auth/vault/user/VaultUserContext.java
```diff
@@ -0,0 +1,329 @@
...
+     * Service for retrieving configuration information.
+     */
+    @Inject
+    private VaultConfigurationService confService;
+
+    /**
+     * Service for retrieving the values of secrets stored in a vault.
+     */
+    @Inject
+    private VaultSecretService secretService;
```
necouchman: :dark_sunglasses:

extensions/guacamole-auth-vault/modules/guacamole-auth-vault-base/src/main/resources/translations/en.json
```diff
@@ -0,0 +1,7 @@
+{
+
+    "DATA_SOURCE_AZURE_KEYVAULT" : {
```
necouchman: Maybe copypasta?  This is the translation file from the -base module, but is referencing the Azure keyvault.  This should be in the azure-keyvault module, no?

extensions/guacamole-auth-vault/modules/guacamole-auth-vault-dist/src/licenses/LICENSE
```diff
@@ -0,0 +1,428 @@
...
+
+Apache Commons Lang (http://commons.apache.org/proper/commons-lang/)
+--------------------------------------------------------------------
+
+    Version: 3.4
+    From: 'Apache Software Foundation' (http://apache.org/)
+    License(s):
+        Apache v2.0 (bundled/commons-lang-3.4/LICENSE.txt)
+
+
```
necouchman: The Azure SDK licenses are in the bundled directory for the vault-dist module, but not in this file?  Guessing they should either be here, or removed from dist module and only included in the azure module?

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/TokenInjectingUserContext.java
```diff
@@ -83,8 +83,13 @@ public TokenInjectingUserContext(UserContext userContext) {
      * @return
      *     The tokens which should be added to the in-progress call to
      *     connect().
+     *
+     * @throws GuacamoleException
+     *     If the tokens applicable to the given connection cannot be
+     *     generated.
```
necouchman: Are there implications any of the other existing users of this class for adding this exception, here?  Presumably this represents a change to the API, which might bump these changes into 2.0.0 version?

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/TokenInjectingUserContext.java
```diff
@@ -100,8 +105,13 @@ public TokenInjectingUserContext(UserContext userContext) {
      * @return
      *     The tokens which should be added to the in-progress call to
      *     connect().
+     *
+     * @throws GuacamoleException
+     *     If the tokens applicable to the given connection group cannot be
+     *     generated.
```
necouchman: Same as above comment for adding this to an existing method...

guacamole-ext/src/main/java/org/apache/guacamole/token/GuacamoleTokenUndefinedException.java
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * An exception thrown when a token cannot be substituted because it has no
+ * corresponding value. Additional information describing the undefined token
+ * is provided.
+ */
+public class GuacamoleTokenUndefinedException extends GuacamoleServerException {
+
+    /**
+     * The name off the token that is undefined.
```
necouchman: off -> of

guacamole-ext/src/main/java/org/apache/guacamole/token/TokenFilter.java
```diff
@@ -156,19 +156,31 @@ public void setTokens(Map<String, String> tokens) {
...
      * @return
      *     A copy of the input string, with any tokens replaced with their
      *     corresponding values.
+     *
+     * @throws GuacamoleTokenUndefinedException
+     *     If the strict flag is set to true and at least one token in the
+     *     given string has no corresponding value.
      */
-    public String filter(String input) {
+    private String filter(String input, boolean strict)
```
necouchman: Why `private`?  Is there any reason not to make this implementation available for calling outside this class?  Would it make sense to make this method public and deprecate the other version of the method, leaving it in (for now) for compatibility.

guacamole-ext/src/main/java/org/apache/guacamole/token/TokenFilter.java
```diff
@@ -216,19 +238,71 @@ public String filter(String input) {
...
+     *
+     * @return
+     *     A copy of the input string, with any tokens replaced with their
+     *     corresponding values.
+     *
+     * @throws GuacamoleTokenUndefinedException
+     *     If at least one token in the given string has no corresponding
+     *     value.
+     */
+    public String filterStrict(String input)
```
necouchman: Is there rationale for creating this `filterStrict()` method and leaving the actual `filter()` method `private` as noted above?  Seems like it would be okay to just make the `filter()` method public??

extensions/guacamole-auth-vault/modules/guacamole-auth-vault-base/src/main/java/org/apache/guacamole/auth/vault/VaultAuthenticationProvider.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * AuthenticationProvider implementation which automatically injects tokens
+ * containing the values of secrets retrieved from a vault.
+ */
+public abstract class VaultAuthenticationProvider
+        extends AbstractAuthenticationProvider {
+
+    /**
+     * Injector which will manage the object graph of this authentication
```
necouchman: Is this actually the Injector?

extensions/guacamole-auth-vault/modules/guacamole-auth-vault-base/src/main/java/org/apache/guacamole/auth/vault/VaultAuthenticationProviderModule.java
mike-jumper: I think it's appropriate in our case, where a developer encountering this class will most likely want to see some sort of reference implementation.

extensions/guacamole-auth-vault/modules/guacamole-auth-vault-base/src/main/java/org/apache/guacamole/auth/vault/user/VaultUserContext.java
mike-jumper: Originally, my rationale for this was that the token here is technically defined by the set of vault extensions, though the name does happen to be identical to that defined by the webapp.

That said, there isn't anything in the public API to reference for this token any longer - it's now defined only internally by the webapp, not by guacamole-ext. The old guacamole-ext class that used to define the identically-named parameter token has been deprecated.

mike-jumper: Another important point is that the actual implementation differs here. Within the webapp, `${GUAC_USERNAME}` is replaced with the username provided during authentication. Within the vault extension, `${GUAC_USERNAME}` is replaced with the identifier of the user's corresponding user object, which may not be the same.

mike-jumper: Actually ... with that in mind, perhaps the vault version of this token should be renamed?

extensions/guacamole-auth-vault/modules/guacamole-auth-vault-base/src/main/resources/translations/en.json
mike-jumper: In this case, no. I agree it feels odd when there is only a single implementation, but this approach is what allows the translations of multiple similar extensions to not be duplicated across the codebase. For example, here's the JDBC version of this:

https://github.com/apache/guacamole-client/blob/da495eb7c3a0d0a9bba201089d1df79869d3eedc/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/resources/translations/en.json#L50-L72

And here's SSO:

https://github.com/apache/guacamole-client/blob/da495eb7c3a0d0a9bba201089d1df79869d3eedc/extensions/guacamole-auth-sso/modules/guacamole-auth-sso-base/src/main/resources/translations/en.json#L3-L13

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/TokenInjectingUserContext.java
mike-jumper: There are API compatibility implications, but not ABI compatibility implications. An extension written against the older API will still work as expected against the newer API. From https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html#jls-13.4.21:

>
> ### 13.4.21. Method and Constructor Throws
>
> Changes to the throws clause of methods or constructors do not break compatibility with pre-existing binaries; these clauses are checked only at compile time.
>

An extension that is updated to build against the new API may need code changes if it invokes its own implementation of this class, but that's it.

This is in line with past Guacamole releases where the version number does not need a full major bump so long as extensions written against the older API will continue to function, even if updating an extension to use the newer API will require code changes. For example, from [the 1.3.0 release notes](https://guacamole.apache.org/releases/1.3.0/#deprecation--compatibility-notes):

>
> # Deprecation / Compatibility notes
>
> Each 1.x release of Apache Guacamole should be compatible with components of older 1.x releases. This compatibility is intended at the Guacamole protocol level and at the extension level, but not necessarily at the API level. This means:
>
> * Extensions from older 1.x releases should still work in binary form, but may need code changes before their source will build against a newer version of guacamole-ext.
> ...

guacamole-ext/src/main/java/org/apache/guacamole/token/TokenFilter.java
mike-jumper: (See https://github.com/apache/guacamole-client/pull/336#discussion_r771785428)

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/TokenInjectingUserContext.java
mike-jumper: (See https://github.com/apache/guacamole-client/pull/336#discussion_r771785428)

extensions/guacamole-auth-vault/modules/guacamole-auth-vault-base/src/main/java/org/apache/guacamole/auth/vault/user/VaultUserContext.java
mike-jumper: Renamed to `${USERNAME}` for the vault-specific token.

extensions/guacamole-auth-vault/modules/guacamole-auth-vault-base/src/main/java/org/apache/guacamole/auth/vault/VaultAuthenticationProviderModule.java
mike-jumper: Right you are.

guacamole-ext/src/main/java/org/apache/guacamole/token/TokenFilter.java
mike-jumper: Yes, except that would then require all callers to `filter()` to catch the possible exception. The behavior of `filter()` historically has been the following:

* If a token has an associated value, substitute that value.
* If a token _does not have_ an associated value, perform no substitution and leave the token as-is.

The new strict implementation is all-or-nothing, substituting all tokens and bailing out if any one token has no corresponding value.

The rationale here is that there are two distinct behaviors and two distinct use-cases for `filter()` vs. `filterStrict()`, and so I think it makes more sense to leave the original `filter()` untouched.

guacamole-ext/src/main/java/org/apache/guacamole/token/TokenFilter.java
mike-jumper: My rationale here for keeping `filter(String, boolean)` private and instead relying on a dedicated `filterStrict()` is ease of verifiability. In cases where the behavior of `filterStrict()` is desirable, it's likely that calling `filter()` without the strict behavior would have security implications, and allowing that behavior to vary easily at runtime via a simple boolean parameter could be problematic.

Keeping `filter()` and `filterStrict()` extremely separate avoids that.

extensions/guacamole-auth-vault/modules/guacamole-auth-vault-base/src/main/java/org/apache/guacamole/auth/vault/VaultAuthenticationProvider.java
mike-jumper: Weeeell ... erm ... no. I'll fix that.

mike-jumper:
OK - I believe I've resolved all feedback thus far, as well as brought things up to date. I also:

* Updated to the latest versions of the libraries.
* Switched from JSON to YAML for the token/secret mapping file.
* Renamed from `guacamole-auth-vault-*` to `guacamole-vault-*` (since it's not providing any kind of auth).
* Renamed the only other non-auth extension (`guacamole-auth-quickconnect`) following the same logic.
* Leveraged `getPrivileged()` for parameter retrieval, so that it's more likely that we'll be able to vary things based on connection hostname and connection username.

BUT, I've also converted this to draft because of difficulties in obtaining the license text from the "azure-annotations" dependency. To be clear, the `pom.xml` for that dependency clearly states that the library is MIT-licensed, as does the source for that library, however the full text of the license file can't currently be obtained because the corresponding git repository is private (presumably by mistake). See:

https://github.com/apache/guacamole-client/blob/54150829d3a60137582a912adb982799c39764b5/doc/licenses/azure-annotations-1.10.0/License.txt#L1-L28

As noted in the placeholder above, I've already emailed Microsoft about this, and they are tracking things down, but are unsure whether this will be resolved soon due to the holidays. I've asked whether they can send us the `License.txt` file manually for the time being so that we can move forward. We shall see...

necouchman (changes requested):
Okay, a few more comments while you wait for the feedback from Microsoft....or you don't get any feedback and we just have to pick a route and go for it.

extensions/guacamole-vault/modules/guacamole-vault-base/src/main/java/org/apache/guacamole/auth/vault/user/VaultUserContext.java
```diff
@@ -0,0 +1,366 @@
...
+     */
+    @AssistedInject
+    public VaultUserContext(@Assisted UserContext userContext) {
+        super(userContext);
+    }
+
+    /**
+     * Creates a new TokenFilter instance with token values set for all tokens
+     * which are not specific to connections or connection groups. Currently,
+     * this is only the username token ("GUAC_USERNAME").
```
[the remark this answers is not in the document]
mike-jumper: Nope - you're right. This should be `USERNAME`.

extensions/guacamole-vault/modules/guacamole-vault-base/src/main/java/org/apache/guacamole/auth/vault/user/VaultUserContext.java
```diff
@@ -0,0 +1,366 @@
...
+     * name of a secret. This token only applies only to connections.
+     */
+    private static final String CONNECTION_HOSTNAME_TOKEN = "CONNECTION_HOSTNAME";
+
+    /**
+     * The name of the token which will be replaced with the \"username\"
+     * connection parameter of the current connection if specified within the
+     * name of a secret. This token only applies only to connections.
+     */
+    private static final String CONNECTION_USERNAME_TOKEN = "CONNECTION_USERNAME";
```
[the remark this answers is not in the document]
mike-jumper: Sure, I'll add clarification here.

extensions/guacamole-quickconnect/pom.xml
```diff
@@ -24,10 +24,10 @@
 
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.apache.guacamole</groupId>
-    <artifactId>guacamole-auth-quickconnect</artifactId>
+    <artifactId>guacamole-quickconnect</artifactId>
     <packaging>jar</packaging>
     <version>1.4.0</version>
-    <name>guacamole-auth-quickconnect</name>
+    <name>guacamole-quickconnect</name>
```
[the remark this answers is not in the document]
mike-jumper: Yep, unless you'd like me to split them out. I included them here given the rename from `guacamole-auth-vault` to `guacamole-vault` as `guacamole-auth-quickconnect` was the only other wait-a-moment-thats-not-acually-auth culprit.

extensions/guacamole-vault/modules/guacamole-vault-base/src/main/java/org/apache/guacamole/auth/vault/user/VaultUserContext.java
```diff
@@ -0,0 +1,366 @@
...
+     *     from the given filter.
+     *
+     * @param filter
+     *     The filter to use to substitute values for tokens in the names of
+     *     secrets to be retrieved from the vault.
+     *
+     * @return
+     *     A Map of token name to Future, where each Future represents the
+     *     pending retrieval operation which will ultimately be completed with
+     *     the value of all secret mapped to that token.
```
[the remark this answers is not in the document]
mike-jumper: Yep, I'll fix that.

extensions/guacamole-vault/modules/guacamole-vault-base/src/main/java/org/apache/guacamole/auth/vault/user/VaultUserContext.java
```diff
@@ -0,0 +1,366 @@
...
+
+    /**
+     * Waits for all pending secret retrieval operations to complete,
+     * transforming each Future within the given Map into its contained String
+     * value.
+     *
+     * @param pendingTokens
+     *     A Map of token name to Future, where each Future represents the
+     *     pending retrieval operation which will ultimately be completed with
+     *     the value of all secret mapped to that token.
```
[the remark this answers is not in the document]
mike-jumper: Oy.

extensions/pom.xml
```diff
@@ -45,10 +45,13 @@
...
         <module>guacamole-auth-json</module>
         <module>guacamole-auth-ldap</module>
-        <module>guacamole-auth-quickconnect</module>
         <module>guacamole-auth-sso</module>
         <module>guacamole-auth-totp</module>
 
+        <!-- Additional features -->
+        <module>guacamole-quickconnect</module>
+        <module>guacamole-vault</module>
+
```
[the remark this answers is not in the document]
mike-jumper: >     _________________________________________
>    / It looks like you're making an attempt  \
>    | to distinguish between auth extensions, |
>    | that actually provide authentication,   |
>    \ and those that do not?                  /
>     -----------------------------------------
>     \
>      \
>         __
>        /  \
>        |  |
>        @  @
>        |  |
>        || |/
>        || ||
>        |\_/|
>        \___/

Yep - exactly.

> I think this makes sense, but am unsure if doing all of it in this change is the right thing?

Perhaps so. If you'd like me to remove the `guacamole-auth-connect` change from this, I'm definitely happy to do that and open a separate issue to be dealt with later. Honestly, that's probably best for sake of scope...

> However, the Vault modules still exist in the `auth` namespace

Good point - I'll refactor the package naming.

> The assembly dist.xml still contains a line `<include>org.apache.guacamole:guacamole-auth-vault-azure</include>`

Yikes - I'll fix that, too. I should have reverified the dist archive after refactoring the name.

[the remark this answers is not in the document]
necouchman: > Perhaps so. If you'd like me to remove the guacamole-auth-connect change from this, I'm definitely happy to do that and open a separate issue to be dealt with later. Honestly, that's probably best for sake of scope...

I think this probably would be better left to a separate issue, for two reasons:
* It seems a bit outside the scope of this issue.
* There may be an opportunity to have a more in-depth discussion/examination of the extension code and see if there are other places that this distinction should be also be mode - for example, even non-authentication-provider extensions, like the QuickConnect extension, still have to extend/implement an `AuthenticationProvider` method and class. It seems like if we're going to make the surface change of renaming extensions (guacamole-auth-quickconnect -> guacamole-connect) we should also take a look at some of those other things.

That said, I'm totally fine with this particular extension being named guacamole-vault-azure, without the "auth" part, as I think that's a good start in the right direction.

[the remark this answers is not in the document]
mike-jumper: I agree. The relevant commit has now been rebased away.

extensions/guacamole-vault/modules/guacamole-vault-base/src/main/java/org/apache/guacamole/auth/vault/user/VaultUserContext.java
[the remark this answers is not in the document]
mike-jumper: Fixed via rebase.

extensions/guacamole-vault/modules/guacamole-vault-base/src/main/java/org/apache/guacamole/auth/vault/user/VaultUserContext.java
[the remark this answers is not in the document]
mike-jumper: Fixed via rebase.

extensions/guacamole-vault/modules/guacamole-vault-base/src/main/java/org/apache/guacamole/auth/vault/user/VaultUserContext.java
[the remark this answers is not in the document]
mike-jumper: Fixed via rebase.

extensions/guacamole-vault/modules/guacamole-vault-base/src/main/java/org/apache/guacamole/auth/vault/user/VaultUserContext.java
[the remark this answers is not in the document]
mike-jumper: OK - I've clarified the behavior here, and added additional logic to make the behavior more consistent in the event the parameters can be retrieved but are empty.

PR #337, opened 2018-11-03
GUACAMOLE-220: Add user group support to LDAP.

mike-jumper:
These changes add support for querying user groups from an LDAP directory, exposing those groups using Guacamole's new user group API.

The groups which apply to an authenticated user are exposed as effective groups, allowing other authentication providers to dictate permissions based on their own identically-named groups. As with LDAP users, LDAP user groups are additionally exposed in a read-only manner such that they can be easily selected within the admin interface.

To avoid further duplicating LDAP query code, the common aspects of LDAP queries within Guacamole have been abstracted into `ObjectQueryService`, with existing services refactored accordingly.

Additional convenience objects missing from the user group API were also added (`SimpleUserGroup` and `AbstractUserGroup`). The refactor adding these classes involved:

* Adding `EMPTY_SET` convenience constants to remove the need to create new empty permission sets everywhere.
* Deprecating the built-in permission support of `SimpleUser` to keep things in line with the simpler (and arguably more sensible) design of `SimpleUserGroup`.
* Adding convenience constructors to the various sets to make non-deprecated usage of `SimpleUser` and `SimpleUserGroup` easy.

### DIFF
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ObjectQueryService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ObjectQueryService.java
@@ -1,0 +1,326 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.guacamole.auth.ldap;
+
+import com.google.inject.Inject;
+import com.novell.ldap.LDAPAttribute;
+import com.novell.ldap.LDAPConnection;
+import com.novell.ldap.LDAPEntry;
+import com.novell.ldap.LDAPException;
+import com.novell.ldap.LDAPReferralException;
+import com.novell.ldap.LDAPSearchResults;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import org.apache.guacamole.GuacamoleException;
+import org.apache.guacamole.GuacamoleServerException;
+import org.apache.guacamole.net.auth.Identifiable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Service for executing queries against an LDAP directory intended to retrieve
+ * Guacamole-related objects. Referrals are automatically handled. Convenience
+ * functions are provided for generating the LDAP queries typically required
+ * for retrieving Guacamole objects, as well as for converting the results of a
+ * query into a Map of Guacamole objects.
+ */
+public class ObjectQueryService {
+
+    /**
+     * Logger for this class.
+     */
+    private final Logger logger = LoggerFactory.getLogger(ObjectQueryService.class);
+
+    /**
+     * Service for escaping parts of LDAP queries.
+     */
+    @Inject
+    private EscapingService escapingService;
+
+    /**
+     * Service for retrieving LDAP server configuration information.
+     */
+    @Inject
+    private ConfigurationService confService;
+
+    /**
+     * Returns the identifier of the object represented by the given LDAP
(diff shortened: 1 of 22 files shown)


extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ObjectQueryService.java
```diff
@@ -0,0 +1,326 @@
...
+import org.apache.guacamole.net.auth.Identifiable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Service for executing queries against an LDAP directory intended to retrieve
+ * Guacamole-related objects. Referrals are automatically handled. Convenience
+ * functions are provided for generating the LDAP queries typically required
+ * for retrieving Guacamole objects, as well as for converting the results of a
+ * query into a Map of Guacamole objects.
```
jmuehlner: Incredibly minor quibble: you're capitalizing "Map" here (to refer to the Java class I assume), but in the javadoc for `asMap`, you use the lowercase form "map".

mike-jumper: I'll try to pick a favorite.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
```diff
@@ -0,0 +1,224 @@
...
+     * Service for executing LDAP queries.
+     */
+    @Inject
+    private ObjectQueryService queryService;
+
+    /**
+     * Returns the base search filter which should be used to retrieve user
+     * groups which do not represent Guacamole connections. As excluding the
+     * guacConfigGroup object class may not work as expected (may always return
+     * zero results) if guacConfigGroup object class is not defined, it should
```
jmuehlner: This is a very long sentence that's a bit hard to parse/understand. Perhaps it should be reworded.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/AbstractUser.java
```diff
@@ -44,4 +51,164 @@ public void setPassword(String password) {
...
 
+    /**
+     * {@inheritDoc}
+     *
+     * <p>This implementation simply an immutable, empty map. Implementations
+     * that wish to expose custom attributes should override this function.
+     */
+    @Override
+    public Map<String, String> getAttributes() {
+        return Collections.<String, String>emptyMap();
```
jmuehlner: Are these type qualifiers actually needed?

mike-jumper: I think it can be inferred.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
mike-jumper: True. I'll [rephrase it appropriately](https://www.youtube.com/watch?v=_K-L9uhsBLM).

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleObjectPermissionSet.java
```diff
@@ -37,14 +38,77 @@
...
+    private Set<ObjectPermission> permissions = Collections.emptySet();
 
     /**
-     * Creates a new empty SimpleObjectPermissionSet.
+     * Creates a new empty SimpleObjectPermissionSet. If you are not extending
+     * SimpleObjectPermissionSet and only need an immutable, empty
+     * ObjectPermissionSet, consider using {@link ObjectPermissionSet#EMPTY_SET}
+     * instead.
      */
     public SimpleObjectPermissionSet() {
```
jmuehlner: So I noticed that the other "Simple" classes seem to define (mostly) empty read-only classes that require methods to be overridden if you want them to actually contain data, but this class provides a convenient way of constructing permission sets, much in the way that `SimpleUser` used to before this PR.

I think a more consistent naming scheme might be beneficial here, as at the moment, seeing the prefix "SImple" on a class name might give a developer a false impression about what it does based on previous experience.

mike-jumper: I agree, though I'm not sure how that can be addressed here, at least within this PR. The various "Simple" classes have had this issue for some time, where it's not entirely clear what defining principle guides what each implementation within `org.apache.guacamole.net.auth.simple` should provide. Even the `...net.auth.` part of the package is a bit of a dinosaur.

I think actually fixing what you describe would need to be part of a larger effort to restructure guacamole-ext. Some of what has been done here is a step in a better direction (like the new `EMPTY_SET` convenience constants) while yet more "Simple" classes are more status quo.

jmuehlner: Yeah, I think you're right that this is not the right place to address these concerns. I guess as long as we keep moving in the direction of things being more usable and less confusing this is OK for now.

PR #339, opened 2018-11-09
GUACAMOLE-656: Explicitly depend on maven-surefire-plugin 3.0.0-M1 due to SUREFIRE-1588.

mike-jumper:


jmuehlner:
Yikes! Glad it's an easy fix.

PR #341, opened 2018-12-02
GUACAMOLE-670: Add slf4j logging to CAS and RADIUS modules

necouchman:
Looks like there are some differing versions of slf4j getting pulled into some of the extension modules by other dependencies, and causing errors to occur when loading the modules.  This adds the explicit dependency to the pom.xml files for the modules exhibiting these issues.

### DIFF
--- a/extensions/guacamole-auth-cas/pom.xml
+++ b/extensions/guacamole-auth-cas/pom.xml
@@ -227,6 +227,12 @@
             <groupId>org.jasig.cas.client</groupId>
             <artifactId>cas-client-core</artifactId>
             <version>3.4.1</version>
+            <exclusions>
+                <exclusion>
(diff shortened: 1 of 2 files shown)


mike-jumper:
Do you know which dependencies specifically are pulling in the conflicting slf4j? This might be better achieved by adding an explicit exclusion for the problematic dependency. The OpenID support had a similar issue:

https://github.com/apache/guacamole-client/blob/823bbeace11063b249e3f05c2a1e5c5027107b96/extensions/guacamole-auth-openid/pom.xml#L222-L227

necouchman:
Got it - was able to use the Maven Dependency Tree to find where it was being pulled in and add the exclusions.  PR updated.

PR #342, opened 2018-12-04
GUACAMOLE-598: Ignore if current user has no associated data.

mike-jumper:
Authentication providers are not required to provide data for the users they authenticate. In this case, the data is needed only to display optional human-readable profile attributes instead of the username. If no such attributes are available, the username is used instead.

### DIFF
--- a/guacamole/src/main/webapp/app/navigation/directives/guacUserMenu.js
+++ b/guacamole/src/main/webapp/app/navigation/directives/guacUserMenu.js
@@ -95,12 +95,9 @@ angular.module('navigation').directive('guacUserMenu', [function guacUserMenu() 
              */
             $scope.role = null;
 
-            // Pull user data
+            // Display user profile attributes if available
             userService.getUser(authenticationService.getDataSource(), $scope.username)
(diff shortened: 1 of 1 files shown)


guacamole/src/main/webapp/app/navigation/directives/guacUserMenu.js
```diff
@@ -95,13 +95,10 @@ angular.module('navigation').directive('guacUserMenu', [function guacUserMenu()
              */
             $scope.role = null;
 
-            // Pull user data
+            // Display user profile attributes if available
             userService.getUser(authenticationService.getDataSource(), $scope.username)
                     .then(function userRetrieved(user) {
 
-                // Store retrieved user object
-                $scope.user = user;
```
necouchman: Just want to make sure this doesn't have any impact?  Was this $scope.user not used at all?

necouchman: Ping @mike-jumper - just wanted to verify this and then I think I can push forward with merging this PR.

PR #343, opened 2018-12-05
GUACAMOLE-526: Correct redirect issue with CAS module

necouchman:
These changes clean up code and correct a minor typo that caused the CAS redirect to fail.

### DIFF
--- a/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/AuthenticationProviderService.java
+++ b/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/AuthenticationProviderService.java
@@ -23,12 +23,11 @@ import com.google.inject.Inject;
 import com.google.inject.Provider;
 import java.util.Arrays;
 import javax.servlet.http.HttpServletRequest;
-import org.apache.guacamole.environment.Environment;
 import org.apache.guacamole.form.Field;
 import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.net.auth.Credentials;
 import org.apache.guacamole.net.auth.credentials.CredentialsInfo;
-import org.apache.guacamole.net.auth.credentials.GuacamoleInsufficientCredentialsException;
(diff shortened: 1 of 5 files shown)


mike-jumper:
These changes look good in principle. Glad to hear that all seems to work now, but I think we should understand why lacking these changes fix things before calling the problem solved.

mike-jumper:
I don't have a CAS handy, but I'll try making the opposite of these changes to OpenID and see if the same breakage occurs. Hopefully it does and the underlying issue will suddenly make sense.

mike-jumper:
So, I can reproduce this even with OpenID just by adding `ngRoute` to the dependencies of the module associated with the field. Not sure why yet, but doing so results in the following error:

https://code.angularjs.org/1.6.9/docs/error/$injector/unpr?p0=$rootElementProvider%20%3C-%20$rootElement%20%3C-%20$location%20%3C-%20$route%20%3C-%20$location

This error occurs within `formService.insertFieldElement()` as an exception, resulting in the field insertion promise failing, causing the field controller to apply and resulting in the "Failed to retrieve field with type (blah)" warning.

mike-jumper:
It seems that the manner in which the injector is created for field types means that `$rootElementProvider` is not available for injection in the context of that field. This is generally OK, but `ngRoute` transitively depends on `$rootElementProvider`. Since `$rootElementProvider` cannot be injected within the context of a field type, no field type can depend on `ngRoute` without causing this error.

mike-jumper:
@necouchman - as GUACAMOLE-526 is an issue in 1.0.0, please update the merge base to `staging/1.0.0`.

necouchman:
Merge base updated.

necouchman:
> Since $rootElementProvider cannot be injected within the context of a field type, no field type can depend on ngRoute without causing this error.

Hmmm...that seems like it could be a problematic limitation.  We don't have anything, yet, that depends on both registering a field type and modifying routes, but it seems like it isn't too much of a stretch to imagine that the scenario could arise?

necouchman:
> You can't include ngRoute as a dependency on a module defining a field type, but you can still rely on index already depending on ngRoute.

Aha.  Good to know.

PR #345, opened 2018-12-09
GUACAMOLE-234: Migrate to Apache Directory API for LDAP Extension

necouchman:
This change request takes a new stab at migrating from the legacy Novell JLDAP API to the Apache Directory API for LDAP access.  I've tried to make sure we're using as much of the API as possible for parsing and verification and eliminated overlapping functionality.  I'm sure there are some more optimizations that could be done, but this pretty well takes care of things for now, I think.

I have tested it against ActiveDirectory and it seems to work fine, but it'd be nice if others could test it against other directories.


michaelbarkdoll:
Hi, how do I pull this pull request to test it?  I've built the docker image from source before, I'm just not certain the commands to use to pull this particular pull request and build docker image for testing.  Also, will this resolve issues with LDAP > 1000 user accounts?  Thanks!

necouchman:
@michaelbarkdoll The code from this pull request is in the following branch of my fork of the code:
https://github.com/necouchman/guacamole-client/tree/jira/234

You're welcome to check it out and give it a try -I would be very interested to know whether or not this resolves the 1000 item limit.

michaelbarkdoll:
Hi @necouchman I tested your PR this morning and it is working much better for me than the old LDAP implementation.  On the old ldap implementation, I was only able to view and login ~50 user accounts on the Users tab from one of the handful of our 25000+ accounts.  

Now users can sign-in, but the user's list is only populated with accounts that have had the "New user" procedure run by an administrator user.  I'm using mysql db for connection info and ldap for authentication (no connection data inside ldap).  I'm curious if this is the intended outcome to have to manually add users (with no password set) and manually add a connection for the user prior to them being able see connections upon login.  Also, if only one connection was added upon login they automatically connect to the connection.  Two connections added to new users results in the normal login screen where they can select their desired connection.

I've turned on debugging and can provide logs if it will help.  Let me know, thanks.

necouchman:
> I tested your PR this morning and it is working much better for me than the old LDAP implementation. On the old ldap implementation, I was only able to view and login ~50 user accounts on the Users tab from one of the handful of our 25000+ accounts.

Glad to hear it!

> Now users can sign-in, but the user's list is only populated with accounts that have had the "New user" procedure run by an administrator user. I'm using mysql db for connection info and ldap for authentication (no connection data inside ldap). I'm curious if this is the intended outcome to have to manually add users (with no password set) and manually add a connection for the user prior to them being able see connections upon login. 

Yes, this is expected behavior if you are combining DB and LDAP authentication - Guacamole matches up users in LDAP and DB based on username, and you have to add them to the database and assign permissions in order for them to be able to use connections stored in the DB extension.  Your alternative is to store the connections in LDAP, which allows you to use the security measures built into LDAP (group membership, access to LDAP objects, etc.) to control access to the connections.  If you don't want to do this I suggest using some scripting to load users from LDAP into the database - this can be done either by inserting rows into the DB directly or by leveraging the API.  If you have further questions about this, please take this to the Guacamole mailing list:

> Also, if only one connection was added upon login they automatically connect to the connection. Two connections added to new users results in the normal login screen where they can select their desired connection.

This is also intended functionality for all users except those with administrative privileges.  If users have admin privileges, then the home screen is the one that is presented regardless of number of connections available.

michaelbarkdoll:
> Hi, how do I pull this pull request to test it? I've built the docker image from source before, I'm just not certain the commands to use to pull this particular pull request and build docker image for testing. Also, will this resolve issues with LDAP > 1000 user accounts? Thanks!

# Checkout a particular branch
````
mkdir ldap
cd ldap
git clone https://github.com/necouchman/guacamole-client.git
cd guacamole-client
git checkout master
git pull origin jira/234
````
# Compile
````
sudo docker build -t mbarkdoll-test/guacamole .
````

**Comment by michaelbarkdoll** [comment]
Thanks this PR is beautiful for our AD implementation.  Thank you for your hard work on this.  I hope it is committed soon! :1st_place_medal:

File: extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java [hunk 1]
```diff
@@ -240,17 +244,24 @@ public LDAPAuthenticatedUser authenticateUser(Credentials credentials)
 
         try {
 
+            LdapConnectionConfig ldapConnectionConfig =
+                    ((LdapNetworkConnection) ldapConnection).getConfig();
+            Dn authDn = new Dn(ldapConnectionConfig.getName());
```
mike-jumper: Is there a better way of retrieving the `Dn` than reparsing it from a string? Perhaps it could be stored somewhere by the relevant `bindAs()` implementation? Or perhaps we should be using our own object that provides access to both the `LdapConnection` and the bind `Dn`?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPConnectionService.java
```diff
@@ -156,38 +146,84 @@ public LDAPConnection bindAs(String userDN, String password)
...
+     * @return
+     *     A LdapConnection object that points at the location
+     *     specified in the referralUrl.
+     *     
+     * @throws GuacamoleException
+     *     If an error occurs parsing out the LdapUrl object or the
+     *     maximum number of referral hops is reached.
+     */
+    public LdapConnection referralConnection(LdapUrl referralUrl,
+            LdapConnectionConfig ldapConfig, Integer hop)
```
mike-jumper: Why `Integer` and not `int`?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ObjectQueryService.java
```diff
@@ -188,46 +183,50 @@ public String generateQuery(String filter,
...
             // following referrals if configured to do so
-            List<LDAPEntry> entries = new ArrayList<>(results.getCount());
-            while (results.hasMore()) {
+            List<Entry> entries = new ArrayList<>();
+            while (results.next()) {
 
-                try {
-                    entries.add(results.next());
+                Response response = results.get();
+                if (response instanceof SearchResultEntry) {
```
mike-jumper: Any way to avoid all the `instanceof` and typecasts? I see functions like `isReferral()` and `getReferral()` on `SearchCursor`.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ObjectQueryService.java
```diff
@@ -188,46 +183,50 @@ public String generateQuery(String filter,
...
-                        logger.debug("Got a referral, but configured to not follow them.", e);
+                else if (response instanceof SearchResultReference &&
+                        request.isFollowReferrals()) {
+                    
+                    Referral referral = ((SearchResultReference) response).getReferral();
+                    int referralHop = 0;
+                    for (String url : referral.getLdapUrls()) {
+                        LdapConnection referralConnection = ldapService.referralConnection(
+                            new LdapUrl(url), ldapConnectionConfig, referralHop++);
+                        entries.addAll(search(referralConnection, baseDN, query));
```
mike-jumper: I'm not sure this is correct. Each call to `search()` will effectively reset that `referralHop` counter. Doesn't this need to be tracked recursively such that it limits the depth of the referral tree?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/LdapFilterGuacamoleProperty.java
```diff
@@ -0,0 +1,53 @@
...
+
+        // No value provided, so return null.
+        if (value == null)
+            return null;
+
+        try {
+            return FilterParser.parse(value);
+        }
+        catch (ParseException e) {
+            throw new GuacamoleServerException("Error parsing filter", e);
```
mike-jumper: I think this error should be clearer. "Error parsing filter" is a bit vague, as the important fact for the user/admin is that filter X is invalid. In this case, that fact (filter is invalid) is hidden here behind what is an internal detail (an error occurred parsing it). If it's possible to tell the user why something is failing without requiring them to deduce the reason for the failure, we should do so.

There's a big difference in pain level when troubleshooting `Error parsing filter` vs. something like `"(dn=invalid))" is not a valid LDAP filter`.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/LdapDnGuacamoleProperty.java
```diff
@@ -0,0 +1,50 @@
...
+    public Dn parseValue(String value) throws GuacamoleException {
+
+        if (value == null)
+            return null;
+
+        try {
+            return new Dn(value);
+        }
+        catch (LdapInvalidDnException e) {
+            throw new GuacamoleServerException("Invalid DN specified in configuration.", e);
```
mike-jumper: To avoid confusion, probably better to include the DN in question. Maybe `Specified DN "thednhere" is not valid` or `"thednhere" is not a valid DN.`, etc.?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -240,17 +244,24 @@ public LDAPAuthenticatedUser authenticateUser(Credentials credentials)
 
         try {
 
+            LdapConnectionConfig ldapConnectionConfig =
+                    ((LdapNetworkConnection) ldapConnection).getConfig();
```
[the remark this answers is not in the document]
necouchman: I'll see what makes the most sense - maybe just converting everything to LdapNetworkConnection  is the most straight-forward, since that's really the only implementation that ever gets used.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
necouchman: I'll take a look and try to figure something out.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPConnectionService.java
necouchman: Bad habits?! :smile:

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ObjectQueryService.java
necouchman: I'll take a closer look.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ObjectQueryService.java
necouchman: Yeah, you're probably right.  The entire recursion of this function was kind of a wild stab, so I'm not surprised I made a few mistakes along the way.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/LdapFilterGuacamoleProperty.java
necouchman: Will reword.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/LdapDnGuacamoleProperty.java
necouchman: Sounds good, I'll rework this error a bit.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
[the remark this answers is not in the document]
necouchman: Nevermind...I was able to clean this stuff up.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPConnectionService.java
necouchman: Fixed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ObjectQueryService.java
necouchman: Reworked to avoid this using the calls you mentioned.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/LdapFilterGuacamoleProperty.java
necouchman: Reworded.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/LdapDnGuacamoleProperty.java
necouchman: Reworded.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
necouchman: So I took one stab at a way to do this, storing a `Dn` object called `bindDn` in the `LDAPAuthenticatedUser` class.  I'm not sure it's the right way to go (and haven't tested it to make sure it works), but let me know if that seems reasonable or if I should try another route.

I also thought about creating a new `LDAPCredentials` class that extends `Credentials` and adding the bindDn into that, but that would require importing other things (JAX-RS) into the guacamole-auth-ldap module, and that seemed like a lot of extra stuff just to find a place to store the bind DN.

I'm open to other suggestions.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ObjectQueryService.java
necouchman: Took a stab at this, as well - not entirely sure this is the best/cleanest way, but let me know what you think.  As usual, I'm open to suggestions for how to approach it.

echu2013:
Hi! Been testing this fork for two days.
Authentication works: LDAP bind and user simple bind
I am using mysql backend and NO LDAP config storage: I was never able to asociate an LDAP user with the same user in MySQL.  created a user, exactly the same as an LDAP one, two connections associated with it and when login in the user has any of them.
Also, TOTP 2FA is never shown when loging via LDAP. Using guacadmin which has a password in MySQL is OK.
Groups is misbheaving and couldn´t find out why by now... 
Full debuging is enabled using logback.xml but for example, I couldn´t find any line stating that is querying for groups. For the user login in the full information is retrieved.
Used whireshark for going further in analysis and it is actually querying using (&(obectClass=*)(member=*)) filter but returns zero results.
Thought there was something wrong with either LDAP (which is an MS Windows AD) or configuration so tried it to do the same query using a LDAP tool and results are returned back ok.
Quite lost here... is I can contribute with something let me now

necouchman:
> Used whireshark for going further in analysis and it is actually querying using (&(obectClass=)(member=)) filter but returns zero results.

Is this something you have configured manually or the default filters?  There's at least one typo here - `(obectClass=)` has a misspelling and is missing the "*", but I don't know if that's somewhere in the code (I can't locate it), in your configuration, or just a typo when you typed it in, here?

Also, I pushed several commits last night to fix some things, and this is still very much a work-in-progress, so I appreciate the feedback and we'll try to get things cleared up.

necouchman:
Also worth noting that it is completely broken right now :-)

echu2013:
Sorry, totally right just forgot things. this is the filter found using wireshark:
(&(obectClass=*)(member=*)) 
Perhaps my keyboard was joking me hehe.
Tried the latest comits and I wasn´t able to login at all, some errores where shown that before ever existed.
Let me do some further tests, also let me pull the newer commits you´ve done to provide deeper debuging logs!

necouchman:
So, the latest commit has it working, again, but it has something to do with some of the LDAP filtering that I changed to use the `ExprNode` filters built into the Apache Directory API.  For some reason it doesn't like the `(objectClass=*)` filter - something about the asterisk, I think, is throwing it off.  In the debug messages in the logs it shows up as `(&(objectClass=\2A)(sAMAccountName=nick))`, and it seems like maybe the escaping is throwing it off.  I'll have to play with it some more and see what I can figure out - for now I have it defaulting to `(&(objectClass=user)(uid=username))` which appears to work.

necouchman:
Okay, I believe I have it working, again, with the default `(objectClass=*)` filter -the trick is actually to use `PresenceNode("objectClass")` rather than `EqualityNode("objectClass","*")`.  Need to go back and make sure everything is cleaned up, but I'm able to successfully authenticate to LDAP, both with the default user search filter and defining my own custom one.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
necouchman: This does seem to work (now that I cleared up the other things that were wrong), but let me know if it's a reasonable approach or if I should try something else.

echu2013:
Well now it's working again (auth), plus the list of AD groups are retrieved becuase I can now see them in debugs.
But for me (perhaps there is something I am missing) there is no association with MySQL data and also TOTP 2FA ain´t showing up..
Regarding AD groups, I have a very large LDAP database.. Isn't possible to make a much more specific filter regarding groups? Instead of bringing them all, add something like member=<userDN> ? Wouldn't that be much more performant ?

necouchman:
> TOTP 2FA ain´t showing up..

I'm not certain that TOTP will work as a second factor with LDAP.  Maybe @mike-jumper can weigh in on this, but I think it needs to be combined with JDBC auth.

> Regarding AD groups, I have a very large LDAP database.. Isn't possible to make a much more specific filter regarding groups? Instead of bringing them all, add something like member= ? Wouldn't that be much more performant ?

You can already filter users based on group by using the `ldap-user-search-filter` parameter.

> But for me (perhaps there is something I am missing) there is no association with MySQL data

Not entirely sure what you mean, here?  Users and groups are associated automatically based on matching names, so if you create a group in MySQL with permissions to a certain set of objects, and you have a LDAP group with the same name, the members of that group should be able to see the connections that are granted permissions.

echu2013:
Despite ldap-user-search-filter is defined or not, when trying to authenticate there is an AND operator in user defined filter that queries using **ldap-username-attribute**, this is what I am saying to be specific. (&(**ldap-user-search-filter**)(**ldap-username-attribute**=<login field entered by human>)). 
When going to find out groups, the filter is generic like (&(objectClass=group)(member=*)). Perhaps in member it´s much more convinient to replace wildcard by authenticated user DN.

echu2013:
this is how the current ldap implementation seeks for group, what I was trying to say:
```
11:30:53.944 [http-nio-127.0.0.1-8085-exec-4] DEBUG o.a.g.auth.ldap.ObjectQueryService - Searching "OU=example,DC=local" for objects matching "(&(objectClass=*)(|(sAMAccountName=tesuser)))".
11:30:54.232 [http-nio-127.0.0.1-8085-exec-4] DEBUG o.a.g.auth.ldap.ObjectQueryService - Searching "OU=example,DC=local" for objects matching "(&(objectClass=*)(|(member=CN=Test User,OU=Users,OU=example,DC=local )))".
```

necouchman:
> this is how the current ldap implementation seeks for group, what I was trying to say:

The new code should do that, as well - that part hasn't changed in the code.  I'll have to do some looking and see if there's something that I'm missing, there, but that shouldn't have changed.

mike-jumper:
>     [http-nio-127.0.0.1-8085-exec-4] DEBUG o.a.g.auth.ldap.ObjectQueryService - Searching "OU=example,DC=local" for objects matching "(&(objectClass=*)(|(member=CN=Test User,OU=Users,OU=example,DC=local )))".

I see a mysterious space at the end of the DN in the query.

necouchman:
> this is how the current ldap implementation seeks for group, what I was trying to say:

I guess I should clarify, when you say "current ldap implementation," is that as implemented in this pull request, or in the code prior to this pull request?

echu2013:
> > ```
> > [http-nio-127.0.0.1-8085-exec-4] DEBUG o.a.g.auth.ldap.ObjectQueryService - Searching "OU=example,DC=local" for objects matching "(&(objectClass=*)(|(member=CN=Test User,OU=Users,OU=example,DC=local )))".
> > ```
> 
> I see a mysterious space at the end of the DN in the query.

Sorry that was my fault! I manually replaced real domain by a fake one on purpose. There is no such space.

necouchman:
> Yeap. by current I refer to the one prior to this pull request (Novell JDLAP API).

Okay, that's what I thought, but wanted to make sure.  Even in the current code (prior to this PR), there are two ways that groups are looked up.  The first is looking up groups associated with a particular user:

https://github.com/apache/guacamole-client/blob/fd2198d62bd5181c6a193dbcedc401823e0fbabe/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java#L160-L178

The second is done if LDAP is not being used for storing connections:

https://github.com/apache/guacamole-client/blob/fd2198d62bd5181c6a193dbcedc401823e0fbabe/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java#L75-L87

This second bit of code defines the LDAP filter used to search for groups, and, if there is no DN defined for connections, it just searches for `(objectClass=*)`.  This doesn't change in this PR - it is (should be) consistent between the current code and the code in this PR, so the behavior should be identical assuming the configuration is identical (e.g. not configured to look for connections in LDAP).

echu2013:
Ok right, userDN is present queryservice.search but I am not able to actually see that in debugs.
Ran another wireshark for LDAP. Upon login this is the queries done to server:
(&(objectClass=*)(sAMAccountName=testuser))
(&(objectClass=*)(member=CN=Test User,OU=example,OU=local)) **This is what I was talking about and couldn´t find at debugs**
I don´t now why these..
(&(objectClass=*)(sAMAccountName=[NULL])) 
(&(objectClass=*)(sAMAccountName=[NULL]))

Would love to actively help but I am very noob in Java.

echu2013:
Sorry, I had an error in MySQL db.
User and group asociation between LDAP and MySQL actually works.
TOTP 2FA also does. 

I am very sorry for wasting your time, I´ll keep using your branch which seems to work so far after my mistake.

echu2013:
Hi! Found something weird.
Still using LDAP and MySQL backend. This is what happens:
When using any username present in LDAP, it doesn´t mather whenever I have this user present on Mysql or not, authentication succeeds with any password disregarding if it is wrong or OK. What really makes authentication process fail is entering a non existent username.
10:11:03.392 [http-nio-127.0.0.1-8085-exec-8] INFO  o.a.g.r.auth.AuthenticationService - User "ldapuser" successfully authenticated from x.x.x.x.

Can you reproduce this ?

necouchman:
I can reproduce this when I log in with a user that exists in LDAP and use basically any password.  Soooooo...need to figure that one out :-D.

necouchman:
Okay, that issue should be resolved, although I may try to tweak the handling a little bit.  But it should properly fail out, now, if either an invalid user or an invalid password are specified.

necouchman:
One more commit to tweak that handling a bit, should be good, now.  Tested with incorrect username and incorrect password and things fail as expected, now.

echu2013:
Bravo! Now it´s OK! Do you know why could be this happening?
Apart from the attribute (Active Directory environment) what I don't understand is why for a single log-in 'successfully authenticated...' apears twice. Seems that authentication challenge is repeated while entering 2nd factor authentication code.


11:26:19.232 [NioProcessor-1] WARN  o.a.d.a.l.m.entry.DefaultAttribute - ERR_13207_VALUE_ALREADY_EXISTS The value 'CN=DC1,OU=Sunday,OU=Domain Controllers,DC=example,DC=local' already exists in the attribute (msDS-RevealedDSAs)
11:26:19.233 [NioProcessor-1] WARN  o.a.d.a.l.m.entry.DefaultAttribute - ERR_13207_VALUE_ALREADY_EXISTS The value 'CN=DC1,OU=Sunday,OU=Domain Controllers,DC=example,DC=local' already exists in the attribute (msDS-RevealedDSAs)
11:26:19.233 [NioProcessor-1] WARN  o.a.d.a.l.m.entry.DefaultAttribute - ERR_13207_VALUE_ALREADY_EXISTS The value 'CN=DC1,OU=Sunday,OU=Domain Controllers,DC=example,DC=local' already exists in the attribute (msDS-RevealedDSAs)
11:26:19.233 [NioProcessor-1] WARN  o.a.d.a.l.m.entry.DefaultAttribute - ERR_13207_VALUE_ALREADY_EXISTS The value 'CN=DC1,OU=Sunday,OU=Domain Controllers,DC=example,DC=local' already exists in the attribute (msDS-RevealedDSAs)
11:26:19.467 [http-nio-127.0.0.1-8085-exec-9] INFO  o.a.g.r.auth.AuthenticationService - User "ldapuser" successfully authenticated from 8.8.8.8.
11:26:25.546 [NioProcessor-3] WARN  o.a.d.a.l.m.entry.DefaultAttribute - ERR_13207_VALUE_ALREADY_EXISTS The value 'CN=DC1,OU=Sunday,OU=Domain Controllers,DC=example,DC=local' already exists in the attribute (msDS-RevealedDSAs)
11:26:25.547 [NioProcessor-3] WARN  o.a.d.a.l.m.entry.DefaultAttribute - ERR_13207_VALUE_ALREADY_EXISTS The value 'CN=DC1,OU=Sunday,OU=Domain Controllers,DC=example,DC=local' already exists in the attribute (msDS-RevealedDSAs)
11:26:25.547 [NioProcessor-3] WARN  o.a.d.a.l.m.entry.DefaultAttribute - ERR_13207_VALUE_ALREADY_EXISTS The value 'CN=DC1,OU=Sunday,OU=Domain Controllers,DC=example,DC=local' already exists in the attribute (msDS-RevealedDSAs)
11:26:25.547 [NioProcessor-3] WARN  o.a.d.a.l.m.entry.DefaultAttribute - ERR_13207_VALUE_ALREADY_EXISTS The value 'CN=DC1,OU=Sunday,OU=Domain Controllers,DC=example,DC=local' already exists in the attribute (msDS-RevealedDSAs)
11:26:25.689 [http-nio-127.0.0.1-8085-exec-6] INFO  o.a.g.r.auth.AuthenticationService - User "ldapuser" successfully authenticated from 8.8.8.8.

necouchman:
The LDAP binds happen a couple of different times depending on the operation being done.  I would guess that is what is happening, here - the directory is queried for the user, and that connection disconnects, and then the directory is queried for user groups associated with that user.  Additionally if LDAP is being used to store connections, the directory will be queried for those objects, as well, which may result in another connect/bind/disconnect.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
necouchman: ping @mike-jumper

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ObjectQueryService.java
necouchman: ping @mike-jumper

necouchman:
@mike-jumper Pinging back on this one, in general - thinking it would be nice to have this finished up for the next release...?

necouchman:
@echu2013 I pulled out the commit from GUACAMOLE-717 that added that catch block - I'm not certain whether or not it's necessary in the case of the Apache DS API, so it's worth some additional testing to make sure, but I rebased and removed it, for now.

necouchman:
@mike-jumper Anything else on this one?

@jmuehlner @JoelB @ceharris Anyone else care to poke at this code?

michaelbarkdoll:
This is working for me, any idea when this will get merged?

necouchman:
@michaelbarkdoll: The issue is slated to be closed in the 1.1.0 release, so hopefully in the coming weeks.  Depends on when we're able to get a chance to finish review and reworking any issues.  Thanks for testing it, though - do let me know if you run into any issues with it.

michaelbarkdoll:
@necouchman Thanks for the update.

So, I'm trying to scale this PR with docker swarm inside a docker compose stack.  The infrastrucutre has an nginx reverse proxy that load balances the websockets to two separate guacamole tomcat frontend servers.  I'm not entirely certain at the moment why this started to happen, but I sometimes see the following error occur during an ldap authentication, other login attempts it just works with no issue.
[error.txt](https://github.com/apache/guacamole-client/files/3113767/error.txt)

I'll try and poke around for better info here, but if you have advise as to why I'm seeing a ERROR o.a.d.l.c.api.LdapNetworkConnection - ERR_04122_SSL_CONTEXT_INIT_FAILURE Failed to initialize the SSL context, then please let me know.

michaelbarkdoll:
> @necouchman Thanks for the update.
> 
> So, I'm trying to scale this PR with docker swarm inside a docker compose stack. The infrastrucutre has an nginx reverse proxy that load balances the websockets to two separate guacamole tomcat frontend servers. I'm not entirely certain at the moment why this started to happen, but I sometimes see the following error occur during an ldap authentication, other login attempts it just works with no issue.
> [error.txt](https://github.com/apache/guacamole-client/files/3113767/error.txt)
> 
> I'll try and poke around for better info here, but if you have advise as to why I'm seeing a ERROR o.a.d.l.c.api.LdapNetworkConnection - ERR_04122_SSL_CONTEXT_INIT_FAILURE Failed to initialize the SSL context, then please let me know.

I think my facing a timeout issue possible related to:
https://issues.apache.org/jira/browse/DIRSTUDIO-1197?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel

Is that accurate?  How can I patch this?  Thanks,

michaelbarkdoll:
> [error.txt](https://github.com/apache/guacamole-client/files/3113767/error.txt)
> 
> I'll try and poke around for better info here, but if you have advise as to why I'm seeing a ERROR o.a.d.l.c.api.LdapNetworkConnection - ERR_04122_SSL_CONTEXT_INIT_FAILURE Failed to initialize the SSL context, then please let me know.

I pulled the latest directory-ldap-api into my guacamole tonight and modified the ldap ext pom.xml to use 2.0.0.AM3-SNAPSHOT.  The issue has disappeared and 20 or so login attempts went through without any issue.  I'll check again tomorrow when the campus is very busy, but from what I can tell maybe something in the newer source code fixed the issue.

michaelbarkdoll:
> I'll try and poke around for better info here, but if you have advise as to why I'm seeing a ERROR o.a.d.l.c.api.LdapNetworkConnection - ERR_04122_SSL_CONTEXT_INIT_FAILURE Failed to initialize the SSL context, then please let me know.

[error2.txt](https://github.com/apache/guacamole-client/files/3115534/error2.txt)

I updated my guacamole container with directory-ldap-api 2.0.0.AM3-SNAPSHOT from their github repo when another developer suggested the newer version has more detailed error reporting and he was right.  Looks like I was having:

org.apache.mina.core.RuntimeIoException: Failed to get the session.
Caused by: java.net.NoRouteToHostException: No route to host

I'll try hard coding the active directory ip address tomorrow and test this further.  I'm guessing the DNS request failed for some reason or I have a networking issue.

michaelbarkdoll:
> I'll try hard coding the active directory ip address tomorrow and test this further. I'm guessing the DNS request failed for some reason or I have a networking issue.

[error3.txt](https://github.com/apache/guacamole-client/files/3115581/error3.txt)
[error4.txt](https://github.com/apache/guacamole-client/files/3115614/error4.txt)

Well I hardcoded the ip address and the error is now:
[http-nio-8080-exec-9] ERROR o.a.d.l.c.api.LdapNetworkConnection - ERR_04122_SSL_CONTEXT_INIT_FAILURE Failed to initialize the SSL context
java.lang.NullPointerException: null

I'm not certain why there would be a null pointer?

michaelbarkdoll:
Appears the issue is related to https://github.com/apache/directory-ldap-api

I've emailed their developers (api@directory.apache.org and dev@directory.apache.org) to attempt to understand why I'm receiving the Null Pointer Exception (NPE).  I'll update this thread when I find out more information from them.

michaelbarkdoll:
Appears to mostly be something related to a bad DNS entry for the AD server, since hard coding /etc/hosts I've only managed to get the NPE to come up twice.  I opened a jira for the NPE with apache directory ldap api.  https://issues.apache.org/jira/browse/DIR-333

michaelbarkdoll:
Ok, the issue was that there is a bad dns entry that occasionally causes a connection timeout.  This has been corrected on my end by just using a static ip address.  Sorry for the noise, the PR works great.

mike-jumper:
@necouchman - I just looked over things quickly and things look fine at first glance. I'll do a more thorough review in a bit, but I think this is probably ready to move forward.

There is a merge conflict noted by GitHub for `ConnectionService.java` which will need to be resolved.

Earlier comments noted warnings being repeatedly logged by parts of the Apache Directory API:

> ```
> ...
> 11:26:19.232 [NioProcessor-1] WARN o.a.d.a.l.m.entry.DefaultAttribute - ERR_13207_VALUE_ALREADY_EXISTS The value 'CN=DC1,OU=Sunday,OU=Domain Controllers,DC=example,DC=local' already exists in the attribute (msDS-RevealedDSAs)
> 11:26:19.233 [NioProcessor-1] WARN o.a.d.a.l.m.entry.DefaultAttribute - ERR_13207_VALUE_ALREADY_EXISTS The value 'CN=DC1,OU=Sunday,OU=Domain Controllers,DC=example,DC=local' already exists in the attribute (msDS-RevealedDSAs)
> 11:26:19.233 [NioProcessor-1] WARN o.a.d.a.l.m.entry.DefaultAttribute - ERR_13207_VALUE_ALREADY_EXISTS The value 'CN=DC1,OU=Sunday,OU=Domain Controllers,DC=example,DC=local' already exists in the attribute (msDS-RevealedDSAs)
> ...
> ```

Anyone know whether this happens as frequently as it looks here (or what needs to be fixed to remove these warnings)? Will Guacamole logs be filled with LDAP warnings following merge?

necouchman:
Cool - I'll rebase against current `staging/1.1.0` and try to brush things up a bit.

Regarding the warnings, I don't think I was seeing those, but I also have logging turned up quite verbosely in my test setup, and I see a *lot* of messages - the Apache Directory API definitely does not suffer from the same issue as the Novell LDAP API regarding the lack of log messages.  I'll have to do some testing with it tuned down to more normal levels and see what it spits out.

necouchman:
@mike-jumper Okay, rebase is completed to current `staging/1.1.0`, and I've also pushed a couple of additional commits to resolve issues.  It compiles, now, though I need to review and also test and make sure everything works as expected.

necouchman:
Simple tests appear fine - can log in with my AD username and password on our corporate network.  Here is the log output that I get when I set `logback.xml` to `INFO`:

https://pastebin.com/NpC0VTZY

A little chatty, but nothing too bad.

necouchman:
Resolved recent merge conflicts, rebased on current `staging/1.1.0`.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPConnectionService.java
```diff
@@ -106,86 +125,113 @@ private LDAPConnection createLDAPConnection() throws GuacamoleException {
...
+     * 
+     * @return
+     *     A LdapNetworkConnection object that points at the location
+     *     specified in the referralUrl.
+     *     
+     * @throws GuacamoleException
+     *     If an error occurs parsing out the LdapUrl object or the
+     *     maximum number of referral hops is reached.
+     */
+    public LdapNetworkConnection referralConnection(LdapUrl referralUrl,
```
mike-jumper: To follow convention, this should be named in a manner that reflects what the function does. This feels like naming a function which scrambles eggs `egg()`. ;)

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ObjectQueryService.java
```diff
@@ -178,6 +169,9 @@ public String generateQuery(String filter,
      *
      * @param query
      *     The LDAP query to execute.
+     * 
+     * @param searchHop
+     *     The level of depth for this search, used for tracking referrals.
```
mike-jumper: Reading this description, I'm somewhat confused about what `searchHop` is. Is this the current depth of the search? Or the maximum depth? I think this should be made a bit clearer.

The code within the function looks like the former.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -100,65 +99,90 @@
...
             // Pull the current user DN from the LDAP connection
-            String userDN = ldapConnection.getAuthenticationDN();
+            LdapConnectionConfig ldapConnectionConfig = ldapConnection.getConfig();
+            Dn userDN = new Dn(ldapConnectionConfig.getName());
 
             // getConnections() will only be called after a connection has been
             // authenticated (via non-anonymous bind), thus userDN cannot
             // possibly be null
-            assert(userDN != null);
+            assert (userDN != null);
```
mike-jumper: A wild space appears!

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -100,65 +99,90 @@
...
-                    while (parameters.hasMoreElements()) {
-
-                        String parameter = (String) parameters.nextElement();
+                    while (parameterAttribute.size() > 0) {
+                        String parameter;
+                        try {
+                            parameter = parameterAttribute.getString();
+                        }
+                        catch (LdapInvalidAttributeValueException e) {
+                            logger.warn("Parameter value not valid for {}", cnName, e);
```
mike-jumper: Exceptions should not be logged at any level higher than debug. The general pattern we've used is to use two calls to the logger: one to log the message at the higher level, and another log the debugging details, including the exception.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -100,65 +99,90 @@
...
                 }
 
                 // Set protocol
                 GuacamoleConfiguration config = new GuacamoleConfiguration();
-                config.setProtocol(protocol.getStringValue());
+                try {
+                    config.setProtocol(protocol.getString());
+                }
+                catch (LdapInvalidAttributeValueException e) {
+                    logger.error("Invalid value of the protocol entry.", e.getMessage());
```
mike-jumper: Is `e.getMessage()` included within the output even though `{}` is not present in the string?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -100,65 +99,90 @@
...
                     return null;
                 }
+                
+                String cnName;
+                
+                try {
+                    cnName = cn.getString();
+                }
+                catch (LdapInvalidAttributeValueException e) {
+                    logger.error("Invalid value for CN attribute.", e.getMessage());
```
mike-jumper: Same here - is `e.getMessage()` present in the output without a `{}`?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
```diff
@@ -196,24 +207,29 @@ private String getGroupSearchFilter() throws GuacamoleException {
...
+                if (name != null)
+                    identifiers.add(name);
+
+                // Ignore user groups which lack a name attribute
+                else
+                    logger.debug("User group \"{}\" is missing a name attribute "
+                            + "and will be ignored.", entry.getDn().toString());
+            }
+            catch (LdapInvalidAttributeValueException e) {
+                logger.debug("User group missing identifier.", e.getMessage());
```
mike-jumper: For reference - here, too: `e.getMessage()` might not be logged due to lack of `{}` (unless `{}` is not actually needed).

mike-jumper:
OK, I've re-reviewed. I think everything looks fine with the exception of the comments above. Specifically:

* Naming a function after its relevant noun feels funky (`referralConnection()`)
* The documentation of the `searchHop` parameter is unclear to me.
* One call to the logger included the exception at a level higher than debug. Exceptions should only be logged at the debug level or lower.
* There are a few cases where `e.getMessage()` is included in the call to the logger, yet there is no `{}` to tell the logger where to put that message. I'm not sure if that will work. Even if it does, it's probably better to be explicit and use the formatting we've used elsewhere when intentionally including an exception message.
* A single space seems to have magically materialized.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPConnectionService.java
necouchman: Renamed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ObjectQueryService.java
necouchman: Clarified this.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
necouchman: I'm sure wild horses will follow...

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
necouchman: Okay, I've tried to redo a bunch of these instances to conform to this.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
necouchman: Braced.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
necouchman: Braced, again.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
necouchman: More bracing.

necouchman:
@mike-jumper: Thanks for the re-review!  Addressed those issues - let me know what else needs to be fixed!

extensions/guacamole-auth-ldap/pom.xml
```diff
@@ -141,11 +141,17 @@
...
         </dependency>
 
-        <!-- JLDAP -->
+        <!-- Apache Directory LDAP API -->
         <dependency>
-            <groupId>com.novell.ldap</groupId>
-            <artifactId>jldap</artifactId>
-            <version>4.3</version>
+            <groupId>org.apache.directory.api</groupId>
+            <artifactId>api-all</artifactId>
```
mike-jumper: With things switching from JLDAP to Apache Directory API, the `LICENSE` file and bundled licenses will need to be updated, too:



https://github.com/apache/guacamole-client/blob/8b172a98fc77241664be16e7f40ab4add510c73a/extensions/guacamole-auth-ldap/src/licenses/LICENSE#L232-L243

necouchman: Okay, updated the licenses.

necouchman:
@mike-jumper: Licenses updated.

PR #348, opened 2018-12-24
GUACAMOLE-422: Forward Timezone to RDP and SSH Connections

necouchman:
These are the client-side changes to allow both detection of (via JSTZ) and overriding of the client timezone for remote connections (SSH and RDP).

### DIFF
--- a/guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
@@ -146,17 +146,17 @@ public class ConfiguredGuacamoleSocket implements GuacamoleSocket {
         // Build args list off provided names and config
         List<String> arg_names = args.getArgs();
         
-        // Check for protocol version as first argument
-        if (arg_names.get(0).startsWith("VERSION_")) {
-            String protocolArg = arg_names.get(0);
-            protocol = GuacamoleProtocolVersion.valueOf(protocolArg);
-        }
-        
         String[] arg_values = new String[arg_names.size()];
         for (int i=0; i<arg_names.size(); i++) {
 
             // Retrieve argument name
             String arg_name = arg_names.get(i);
+            
+            // Check for protocol version as first argument
+            if (i == 0 && arg_name.startsWith("VERSION_")) {
+                protocol = GuacamoleProtocolVersion.valueOf(arg_name);
+                arg_values[i] = protocol.toString();
+            }
 
             // Get defined value for name
             String value = config.getParameter(arg_name);
--- a/guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
@@ -19,7 +19,6 @@
 
 package org.apache.guacamole.protocol;
 
-
 import java.util.List;
 import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.GuacamoleServerException;
@@ -55,6 +54,15 @@ public class ConfiguredGuacamoleSocket implements GuacamoleSocket {
      * by the "ready" instruction received from the Guacamole proxy.
      */
     private String id;
+    
+    /**
+     * The protocol version that will be used to communicate with guacd.  The
+     * default is 1.0.0, and, if the server does not provide a specific version
+     * it will be assumed that it operates at this version and certain features
+     * may be unavailable.
+     */
+    private GuacamoleProtocolVersion protocol =
+            GuacamoleProtocolVersion.VERSION_1_0_0;
     
     /**
      * Waits for the instruction having the given opcode, returning that
@@ -137,6 +145,13 @@ public class ConfiguredGuacamoleSocket implements GuacamoleSocket {
 
         // Build args list off provided names and config
         List<String> arg_names = args.getArgs();
+        
+        // Check for protocol version as first argument
+        if (arg_names.get(0).startsWith("VERSION_")) {
+            String protocolArg = arg_names.get(0);
+            protocol = GuacamoleProtocolVersion.valueOf(protocolArg);
+        }
+        
         String[] arg_values = new String[arg_names.size()];
         for (int i=0; i<arg_names.size(); i++) {
 
@@ -185,14 +200,18 @@ public class ConfiguredGuacamoleSocket implements GuacamoleSocket {
                     info.getImageMimetypes().toArray(new String[0])
                 ));
         
-        // Send client timezone, if available
-        String timezone = info.getTimezone();
-        if (timezone != null && !timezone.isEmpty()) {
-            writer.writeInstruction(
-                    new GuacamoleInstruction(
-                        "timezone",
-                        info.getTimezone()
-                    ));
+        // Protocol version 1.1.0 and higher options
+        
+        if (protocol.atLeast(GuacamoleProtocolVersion.VERSION_1_1_0)) {
+            // Send client timezone, if available
+            String timezone = info.getTimezone();
+            if (timezone != null && !timezone.isEmpty()) {
+                writer.writeInstruction(
+                        new GuacamoleInstruction(
+                            "timezone",
+                            info.getTimezone()
+                        ));
+            }
         }
 
         // Send args
--- a/guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
@@ -204,7 +204,7 @@ public class ConfiguredGuacamoleSocket implements GuacamoleSocket {
         if (protocol.isSupported(GuacamoleProtocolCapability.TIMEZONE_HANDSHAKE)) {
             // Send client timezone, if available
             String timezone = info.getTimezone();
-            if (timezone != null && !timezone.isEmpty()) {
+            if (timezone != null) {
                 writer.writeInstruction(
                         new GuacamoleInstruction(
                             "timezone",
--- a/guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
@@ -145,7 +145,6 @@ public class ConfiguredGuacamoleSocket implements GuacamoleSocket {
 
         // Build args list off provided names and config
         List<String> arg_names = args.getArgs();
-        
         String[] arg_values = new String[arg_names.size()];
         for (int i=0; i<arg_names.size(); i++) {
 
--- a/guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
@@ -184,6 +184,13 @@ public class ConfiguredGuacamoleSocket implements GuacamoleSocket {
                     "image",
                     info.getImageMimetypes().toArray(new String[0])
                 ));
+        
+        // Send client timezone
+        writer.writeInstruction(
+                new GuacamoleInstruction(
+                    "timezone",
+                    info.getTimezone()
+                ));
 
         // Send args
         writer.writeInstruction(new GuacamoleInstruction("connect", arg_values));
--- a/guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
@@ -185,12 +185,15 @@ public class ConfiguredGuacamoleSocket implements GuacamoleSocket {
                     info.getImageMimetypes().toArray(new String[0])
                 ));
         
-        // Send client timezone
-        writer.writeInstruction(
-                new GuacamoleInstruction(
-                    "timezone",
-                    info.getTimezone()
-                ));
+        // Send client timezone, if available
+        String timezone = info.getTimezone();
+        if (timezone != null && !timezone.isEmpty()) {
+            writer.writeInstruction(
+                    new GuacamoleInstruction(
+                        "timezone",
+                        info.getTimezone()
+                    ));
+        }
 
         // Send args
         writer.writeInstruction(new GuacamoleInstruction("connect", arg_values));
--- a/guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
@@ -155,6 +155,7 @@ public class ConfiguredGuacamoleSocket implements GuacamoleSocket {
             if (i == 0 && arg_name.startsWith("VERSION_")) {
                 protocol = GuacamoleProtocolVersion.valueOf(arg_name);
                 arg_values[i] = protocol.toString();
+                continue;
             }
 
             // Get defined value for name
--- a/guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
@@ -200,9 +200,8 @@ public class ConfiguredGuacamoleSocket implements GuacamoleSocket {
                     info.getImageMimetypes().toArray(new String[0])
                 ));
         
-        // Protocol version 1.1.0 and higher options
-        
-        if (protocol.atLeast(GuacamoleProtocolVersion.VERSION_1_1_0)) {
+        // Check for support for timezone handshake
+        if (protocol.isSupported(GuacamoleProtocolCapability.TIMEZONE_HANDSHAKE)) {
             // Send client timezone, if available
             String timezone = info.getTimezone();
             if (timezone != null && !timezone.isEmpty()) {
--- a/guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
@@ -153,7 +153,7 @@ public class ConfiguredGuacamoleSocket implements GuacamoleSocket {
             
             // Check for protocol version as first argument
             if (i == 0 && arg_name.startsWith("VERSION_")) {
-                protocol = GuacamoleProtocolVersion.valueOf(arg_name);
+                protocol = GuacamoleProtocolVersion.getVersion(arg_name);
                 arg_values[i] = protocol.toString();
                 continue;
             }
--- a/guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleClientInformation.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleClientInformation.java
@@ -62,7 +62,7 @@ public class GuacamoleClientInformation {
     /**
      * The timezone reported by the client.
      */
-    private String timezone = "";
+    private String timezone;
 
     /**
      * Returns the optimal screen width requested by the client, in pixels.
@@ -151,8 +151,8 @@ public class GuacamoleClientInformation {
     }
     
     /**
-     * Return the timezone as reported by the client, or an empty String if
-     * one is not set.
+     * Return the timezone as reported by the client, or null if the timezone
+     * is not set.  Valid timezones are specified in tz database format.
      * 
      * @return
      *     A string value of the timezone reported by the client.
@@ -162,7 +162,9 @@ public class GuacamoleClientInformation {
     }
     
     /**
-     * Set the string value of the timezone.
+     * Set the string value of the timezone, or null if the timezone will not
+     * be provided by the client.  Valid timezones are specified in tz
+     * database format.
      * 
      * @param timezone
      *     The string value of the timezone reported by the client.
--- a/guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleClientInformation.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleClientInformation.java
@@ -167,7 +167,9 @@ public class GuacamoleClientInformation {
      * database format.
      * 
      * @param timezone
-     *     The string value of the timezone reported by the client.
+     *     The string value of the timezone reported by the client, in tz
+     *     database format, or null if the timezone is not provided by the
+     *     client.
      */
     public void setTimezone(String timezone) {
         this.timezone = timezone;
--- a/guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleClientInformation.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleClientInformation.java
@@ -151,7 +151,8 @@ public class GuacamoleClientInformation {
     }
     
     /**
-     * Return the timezone as reported by the client.
+     * Return the timezone as reported by the client, or an empty String if
+     * one is not set.
      * 
      * @return
      *     A string value of the timezone reported by the client.
--- a/guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleClientInformation.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleClientInformation.java
@@ -60,7 +60,7 @@ public class GuacamoleClientInformation {
     private final List<String> imageMimetypes = new ArrayList<String>();
     
     /**
-     * The timezone report by the client.
+     * The timezone reported by the client.
      */
     private String timezone = "";
 
--- a/guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleClientInformation.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleClientInformation.java
@@ -153,7 +153,7 @@ public class GuacamoleClientInformation {
     /**
      * Return the timezone as reported by the client.
      * 
-     * @returns
+     * @return
      *     A string value of the timezone reported by the client.
      */
     public String getTimezone() {
--- a/guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleClientInformation.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleClientInformation.java
@@ -152,7 +152,8 @@ public class GuacamoleClientInformation {
     
     /**
      * Return the timezone as reported by the client, or null if the timezone
-     * is not set.  Valid timezones are specified in tz database format.
+     * is not set.  Valid timezones are specified in IANA zone key format,
+     * also known as Olson time zone database or TZ Database.
      * 
      * @return
      *     A string value of the timezone reported by the client.
@@ -163,8 +164,8 @@ public class GuacamoleClientInformation {
     
     /**
      * Set the string value of the timezone, or null if the timezone will not
-     * be provided by the client.  Valid timezones are specified in tz
-     * database format.
+     * be provided by the client.  Valid timezones are specified in IANA zone
+     * key format (aka Olson time zone database or tz database).
      * 
      * @param timezone
      *     The string value of the timezone reported by the client, in tz
--- a/guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleClientInformation.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleClientInformation.java
@@ -58,6 +58,11 @@ public class GuacamoleClientInformation {
      * The list of image mimetypes reported by the client to be supported.
      */
     private final List<String> imageMimetypes = new ArrayList<String>();
+    
+    /**
+     * The timezone report by the client.
+     */
+    private String timezone = "";
 
     /**
      * Returns the optimal screen width requested by the client, in pixels.
@@ -143,6 +148,26 @@ public class GuacamoleClientInformation {
      */
     public List<String> getImageMimetypes() {
         return imageMimetypes;
+    }
+    
+    /**
+     * Return the timezone as reported by the client.
+     * 
+     * @returns
+     *     A string value of the timezone reported by the client.
+     */
+    public String getTimezone() {
+        return timezone;
+    }
+    
+    /**
+     * Set the string value of the timezone.
+     * 
+     * @param timezone
+     *     The string value of the timezone reported by the client.
+     */
+    public void setTimezone(String timezone) {
+        this.timezone = timezone;
     }
 
 }
--- a/guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java
@@ -19,6 +19,9 @@
 
 package org.apache.guacamole.protocol;
 
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 /**
  * An enum that defines the available Guacamole protocol versions that can be
  * used between guacd and clients, and provides convenience methods for parsing
@@ -41,6 +44,16 @@ public enum GuacamoleProtocolVersion {
      * for passing the client timezone to the server during the handshake.
      */
     VERSION_1_1_0(1, 1, 0);
+    
+    /**
+     * A regular expression that matches the VERSION_X_Y_Z pattern, where
+     * X is the major version component, Y is the minor version component,
+     * and Z is the patch version component.  This expression puts each of
+     * the version components in their own group so that they can be easily
+     * used later.
+     */
+    private static final Pattern VERSION_PATTERN =
+            Pattern.compile("^VERSION_([0-9]+)_([0-9]+)_([0-9]+)$");
     
     /**
      * The major version component of the protocol version.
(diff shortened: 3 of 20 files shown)


necouchman:
Just a caution on this one - makes changes to guacamole-common and guacamole-ext, so backward compatibility may be an issue.

necouchman:
@mike-jumper: Thanks for the feedback.  I'll take another stab at putting it into the handshake and see what I can get done.

Do we want to go ahead and pull this change out of the 1.1.0 release and push it out to a future version?  Sounds like it's invasive enough that maybe we could add it to whatever release ends up with parameter prompting or something like that - 2.0.0 or whatever we decide?

necouchman:
@mike-jumper: Okay, rebased on staging/1.1.0 for now, though I'm still not sure if this belongs in this release or not?  Anyway, I think I managed to make the changes you were suggesting to make this part of the client handshake, and removed some of the kludgy stuff that glued it together previously.  Also opened a pull on the server-side for the libguac bits.  Need to test it still and make sure it works, but let me know if there is anything obviously wrong with the latest commits.

mike-jumper:
Regarding whether the changes belong in 1.1.0, the base timezone changes are already present on staging/1.1.0, at least for providing "timezone" parameters for RDP and SSH. That much should definitely stay, as it's compatible with 1.0.0 and I believe was included in scope as part of the initial release discussion.

As for the handshake changes ... those are incompatible protocol changes and would technically bump things to 2.0.0. _However_, I think we can alter things there such that the changes are compatible.

If the handling of the handshake phase is altered such that:

1. Order of the various handshake-specific instructions does not matter.
2. Presence/absence of the various handshake-specific instructions does not matter (with the exception of `connect` which terminates the handshake).

then things will inherently be backward compatible, and the protocol becomes a bit more forgiving of implementations going forward. This would require replacing current usage of `guac_parser_expect()` within the handshake with a read loop and array of instruction handler callbacks, similar to:

https://github.com/apache/guacamole-server/blob/ffe0b57faa97272ee1287bcea74c5d787e221953/src/libguac/user-handlers.c#L37-L54

and:

https://github.com/apache/guacamole-server/blob/ffe0b57faa97272ee1287bcea74c5d787e221953/src/libguac/user.c#L170-L186

mike-jumper:
Assuming the above, it would also be important that the `ConfiguredGuacamoleSocket` implementation continue to adhere to the 1.0.0 and older handshake except where the timezone is set, or else it would become incompatible with 1.0.0 and older guacd.

Another option could be to modify the semantics of `args` to include a protocol version number, allowing `ConfiguredGuacamoleSocket` or other implementations to change behavior depending on which version of guacd is being used.

necouchman:
@mike-jumper: So, I think I have backward compatibility from a guacd perspective handled - that is, the newer guacd will accept either older clients that do not provide the "timezone" handshake parameter, or newer clients that do, without any problems.

I'm struggling with backward compatibility from the other angle - that is, if I use the newer client that sends timezone with the older (1.0.0) guacd, I get protocol errors.  I am only sending the `timezone` parameter through the tunnel when set; however, the issue is that, as I have it implemented, today, I'm auto-detecting the timezone via JSTZ and sending that information automatically (=always).  So, not sure what other ways might be available to make this backward-compatible?  I looked at the `args` sent back from the server, but am not really sure the best way to implement sending back some sort of protocol version - right now it looks like the only thing that gets sent back is a list of acceptable parameters for the given protocol?

mike-jumper:
Yes, that's correct. The semantics of `args` would need to be modified in a backward-compatible way. For example, if a simple protocol implementation would currently produce:

    4.args,8.hostname,4.port;

consider a change to `args` which would include the version number as one of the arguments:

    4.args,5.1.1.0,8.hostname,4.port;

If the definition of `args` from that point forward specified that the first argument is the version number, then code that's aware of this would be able to make decisions based on that number. Code that's unaware of this would continue with the old `args` semantics, considering "1.1.0" to be the name of a parameter. With no value defined for such a parameter, a conforming implementation would simply include the empty string in the `connect` response:

    7.connect,0.,9.localhost,4.1234;

whereas newer implementations would know to handle that value specifically. One possibility here would be to similarly modify `connect` to specify that the first argument should be the protocol version of the client (with the exception of older clients that may include an empty string):

    7.connect,5.1.1.0,8.hostname,4.port;

That version information could be logged, included in `guac_user`, compared against the guacd version, etc.

Whether the version number should be the literal release version or some protocol-specific version number (or both?) is also worth considering.

necouchman:
Okay, I've taken a stab at this - I may have gone complete overkill on the client side - implementing a new `enum` with some supporting functions.  Let me know what needs work, here.

necouchman:
@mike-jumper: Okay, client-side of this I've gone back and redone a couple of things, but have confirmed client works both current and backward-compatible.  Let me know what needs attention, here.

guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
```diff
@@ -142,6 +150,12 @@ public ConfiguredGuacamoleSocket(GuacamoleSocket socket,
 
             // Retrieve argument name
             String arg_name = arg_names.get(i);
+            
+            // Check for protocol version as first argument
+            if (i == 0 && arg_name.startsWith("VERSION_")) {
+                protocol = GuacamoleProtocolVersion.valueOf(arg_name);
+                arg_values[i] = protocol.toString();
```
mike-jumper: This will be overwritten by:

https://github.com/apache/guacamole-client/blob/0f4ff0fe468e064a672296af47c0744cf919ee4e/guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java#L160-L167

necouchman: Yeah, that could be a problem...should be fixed.

necouchman:
@mike-jumper Anything else on this one before it can go in following the matching apache/guacamole-server changes?

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java
```diff
@@ -0,0 +1,142 @@
...
+     * version is the same as or greater than the other version, otherwise
+     * false.
+     * 
+     * @param otherVersion
+     *     The version to which this object should be compared.
+     * 
+     * @return 
+     *     True if this object is greater than or equal to the other version.
+     */
+    public boolean atLeast(GuacamoleProtocolVersion otherVersion) {
```
mike-jumper: If what we need is a means of comparing one version against another, it might be better to implement [`Comparable`](https://docs.oracle.com/javase/8/docs/api/java/lang/Comparable.html), an interface meant for achieving exactly that in a generic way.

Now that we're all Java 8, chaining comparisons across several values is easier with [`Comparator`](https://docs.oracle.com/javase/8/docs/api/java/util/Comparator.html) and chains of the various `comparing()` and `thenComparing()` functions. Here, a chain of [`Comparator.comparingInt()`](https://docs.oracle.com/javase/8/docs/api/java/util/Comparator.html#comparingInt-java.util.function.ToIntFunction-) and [`Comparator.thenComparingInt()`](https://docs.oracle.com/javase/8/docs/api/java/util/Comparator.html#thenComparingInt-java.util.function.ToIntFunction-) might condense this down into essentially one line.

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java
```diff
@@ -0,0 +1,142 @@
...
+     *     The integer representation of the patch version component.
+     */
+    GuacamoleProtocolVersion(int major, int minor, int patch) {
+        this.major = major;
+        this.minor = minor;
+        this.patch = patch;
+    }
+    
+    /**
+     * Return the major version number.
```
mike-jumper: You mean the major component of the protocol version number?

guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
```diff
@@ -184,6 +199,20 @@ public ConfiguredGuacamoleSocket(GuacamoleSocket socket,
                     "image",
                     info.getImageMimetypes().toArray(new String[0])
                 ));
+        
+        // Protocol version 1.1.0 and higher options
+        
+        if (protocol.atLeast(GuacamoleProtocolVersion.VERSION_1_1_0)) {
+            // Send client timezone, if available
+            String timezone = info.getTimezone();
+            if (timezone != null && !timezone.isEmpty()) {
```
mike-jumper: Any reason why both `null` and `""` are used rather than just `null`?

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleClientInformation.java
```diff
@@ -144,5 +149,25 @@ public void setOptimalResolution(int optimalResolution) {
...
+     */
+    public String getTimezone() {
+        return timezone;
+    }
+    
+    /**
+     * Set the string value of the timezone.
+     * 
+     * @param timezone
+     *     The string value of the timezone reported by the client.
```
mike-jumper: Same here: if `null` and/or `""` have special meaning, that meaning should be documented.

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleClientInformation.java
```diff
@@ -144,5 +149,25 @@ public void setOptimalResolution(int optimalResolution) {
     public List<String> getImageMimetypes() {
         return imageMimetypes;
     }
+    
+    /**
+     * Return the timezone as reported by the client.
+     * 
+     * @return
+     *     A string value of the timezone reported by the client.
```
mike-jumper: If `null` and/or `""` have special meaning here, that meaning should be documented for the sake of downstream users of this class. I would also suggest settling on a single, specific representation for "no timezone" (presumably `null`?).

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleClientInformation.java
```diff
@@ -144,5 +149,25 @@ public void setOptimalResolution(int optimalResolution) {
     public List<String> getImageMimetypes() {
         return imageMimetypes;
     }
+    
+    /**
+     * Return the timezone as reported by the client.
```
mike-jumper: As it will not otherwise be clear to a user of this class, the standard that these timezone strings should conform to should be documented here.

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleClientInformation.java
```diff
@@ -144,5 +149,25 @@ public void setOptimalResolution(int optimalResolution) {
...
+     * 
+     * @return
+     *     A string value of the timezone reported by the client.
+     */
+    public String getTimezone() {
+        return timezone;
+    }
+    
+    /**
+     * Set the string value of the timezone.
```
mike-jumper: Same here: as it will not otherwise be clear to a user of this class, the standard that these timezone strings should conform to should be documented.

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java
```diff
@@ -0,0 +1,142 @@
...
+import org.apache.guacamole.GuacamoleUnsupportedException;
+
+/**
+ * An enum that defines the available Guacamole protocol versions that can be
+ * used between guacd and clients, and provides convenience methods for parsing
+ * and comparing versions.
+ */
+public enum GuacamoleProtocolVersion {
+    
+    // Version 1.0.0 and older.
```
mike-jumper: Please use JavaDoc comments to document enum values, so that the text will be included in the API documentation generated with the build and ultimately deployed to the website with the release.

With that in mind, I suggest documenting in more detail what "version 1.0.0 and older" means here.

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java
```diff
@@ -0,0 +1,142 @@
...
+    VERSION_1_0_0(1, 0, 0),
+
+    // Version 1.1.0
```
mike-jumper: Same here:

* Needs JavaDoc.
* Description of the value should provide some clarity beyond that "VERSION_1_1_0" means "Version 1.1.0". ;)

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java
```diff
@@ -0,0 +1,142 @@
...
+    VERSION_1_1_0(1, 1, 0);
+    
+    // The major version number.
```
mike-jumper: JavaDoc

guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
```diff
@@ -142,6 +150,13 @@ public ConfiguredGuacamoleSocket(GuacamoleSocket socket,
 
             // Retrieve argument name
             String arg_name = arg_names.get(i);
+            
+            // Check for protocol version as first argument
+            if (i == 0 && arg_name.startsWith("VERSION_")) {
+                protocol = GuacamoleProtocolVersion.valueOf(arg_name);
```
mike-jumper: As `valueOf()` for an enum is specifically defined based on the names of the enum values, I'm uncomfortable directly relying on it for parsing of the Guacamole protocol (something which technically is independent of the Java objects we may use to implement it).

I suggest providing a function within `GuacamoleProtocolVersion` which is specifically defined as parsing the Guacamole protocol constant for that protocol version and returning the corresponding enum value. Even though that function may internally only wrap `valueOf()`, it will decouple the internals of the API from the internals of the protocol.

guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
```diff
@@ -184,6 +199,20 @@ public ConfiguredGuacamoleSocket(GuacamoleSocket socket,
                     "image",
                     info.getImageMimetypes().toArray(new String[0])
                 ));
+        
+        // Protocol version 1.1.0 and higher options
+        
+        if (protocol.atLeast(GuacamoleProtocolVersion.VERSION_1_1_0)) {
```
mike-jumper: Thoughts on providing some sort of capability-style query rather than requiring comparisons on version number? What you've got here will clearly work fine for now, but will get cumbersome as new features are introduced.

Something like `protocol.isSupported(GuacamoleProtocolCapability.HANDSHAKE_TIMEZONE)` would remove the need for repeated version comparison logic here or elsewhere, and would allow us to define what features are supported by what versions in a more readable way.

guacamole/src/main/java/org/apache/guacamole/tunnel/TunnelRequest.java
```diff
@@ -365,5 +370,16 @@ public Integer getDPI() throws GuacamoleException {
     public List<String> getImageMimetypes() {
         return getParameterValues(IMAGE_PARAMETER);
     }
-
+    
+    /**
+     * Returns the value of the timezone parameter declared within the
```
mike-jumper: I suggest wording this from the perspective of the client and the semantics behind the value returned, rather than the fact that it internally pulls the "timezone" parameter.

For example, `getImageMimetypes()` is documented as returning the image mimetypes that the client declares it supports:

https://github.com/apache/guacamole-client/blob/377e93c518fbee77fbfe04970d30d9b721bcaedf/guacamole/src/main/java/org/apache/guacamole/tunnel/TunnelRequest.java#L363-L364

and `getWidth()` is documented as returning the client's desired width:

https://github.com/apache/guacamole-client/blob/377e93c518fbee77fbfe04970d30d9b721bcaedf/guacamole/src/main/java/org/apache/guacamole/tunnel/TunnelRequest.java#L294-L295

If the above were documented instead as "the value of the width parameter" or "the value of the image parameters", we would be missing information that is important for the caller.

guacamole/src/main/webapp/app/settings/services/preferenceService.js
```diff
@@ -98,6 +98,17 @@ angular.module('settings').provider('preferenceService', ['$injector',
         return language.replace(/-/g, '_');
 
     };
+    
+    /**
+     * Return the timezone detected for the current browser session
+     * by the JSTZ timezone library.
+     * 
+     * @returns String
+     *     The name of the currently-detected timezone.
```
mike-jumper: Is the format of the timezone returned by JSTZ known to conform to a particular standard? If so, I suggest documenting that here.

guacamole/pom.xml
```diff
@@ -493,6 +493,13 @@
             </exclusions>
 
         </dependency>
+        
+        <!-- JSTZ for TimeZone Detection -->
```
mike-jumper: This will need to be added to the list of bundled dependencies within `LICENSE`.

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java
necouchman: Ah, but `guacamole-common` is not at Java 8, it's still at Java 6.  So, we'll either have to bring it up, or do this another way...

guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
necouchman: Well, currently in the `GuacamoleClientInformation` class it is pre-initialized to an empty string.  I don't know if there was a particular reason I did this - I think I was trying to follow the pattern of the other config information, like the mimetypes, which are initialized to an empty list.  So, if anything, shouldn't it be only empty and skip the `null` check?

guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
necouchman: Yeah, I like this idea.  I'm guessing another enum (or class?) would be required to implement this?  Or is there a way that I should try to do it within the current enum?

guacamole/src/main/webapp/app/settings/services/preferenceService.js
necouchman: Sounds good.  Is there any precedent to adding a link/URL within the JavaDoc for this to the Wikipedia article on tz database?

mike-jumper:
Ah, good point. Never mind.

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java
necouchman: It looks like Java `enum`s implement the `compareTo()` method, though, using the position of declaration of the enum value as the relative comparison between them.  So, I could leverage that, instead of a function like this, but is that something we really want to rely on?

mike-jumper:
No, I think what you already have is better defined.

guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
necouchman: Okay, I implemented a `getVersion()` method that checks for null/empty and then wraps `valueOf()` for the time being.  I'm happy to do something different in `getVersion()` if that's desirable, but at least it decouples from reliance upon `valueOf()` at this point.

guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
necouchman: Took a stab at an implementation of this with a new `GuacamoleProtocolCapability` enum.  Let me know if it looks sane or if I should go another direction.

guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
necouchman: Reduced it to just the `null` check/value.

guacamole/src/licenses/LICENSE
```diff
@@ -605,6 +605,36 @@ licenses; we recommend you read them, as their terms may differ from the
 terms above.
 
 
+JSTZ (http://webjars.org/)
+--------------------------
+
+    Version: 1.0.10
+    From: 'WebJars' (http://webjars.org/)
```
mike-jumper: "From" here needs to be the copyright holder. Based on the license below, it looks like the correct value would be `'Jon Nylander'`.

If the individual/organization holding the copyright has a website, a GitHub account, etc., that's what has been used for the URL following the name.

guacamole/src/licenses/LICENSE
```diff
@@ -605,6 +605,36 @@ licenses; we recommend you read them, as their terms may differ from the
 terms above.
 
 
+JSTZ (http://webjars.org/)
```
mike-jumper: The URL here should be the URL of the project.

Perhaps https://pellepim.bitbucket.io/jstz/?

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleClientInformation.java
```diff
@@ -144,5 +149,30 @@ public void setOptimalResolution(int optimalResolution) {
     public List<String> getImageMimetypes() {
         return imageMimetypes;
     }
+    
+    /**
+     * Return the timezone as reported by the client, or null if the timezone
+     * is not set.  Valid timezones are specified in tz database format.
```
mike-jumper: The JSTZ documentation states:

> The return value is an IANA zone info key (aka the Olson time zone database).
>
> The IANA timezone database is pretty much standard for most platforms (UNIX and Mac support it natively, and every programming language in the world either has native support or well maintained libraries that support it).

Is the correct term for this "tz database format"? Or "IANA timezone database format"? Or "IANA zone info key"? Or ... ?

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java
```diff
@@ -0,0 +1,193 @@
...
+     *     True if this object is greater than or equal to the other version.
+     */
+    public boolean atLeast(GuacamoleProtocolVersion otherVersion) {
+        
+        // Major version is greater
+        if (major > otherVersion.getMajor())
+            return true;
+        
+        // Major version is less than or equal to.
+        else {
```
mike-jumper: You might be able to clean this up by relying on the fact that code flow stops after that `return`. There's no need for an explicit `else` here.

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java
```diff
@@ -0,0 +1,193 @@
...
+            // Major version is less than
+            if (major < otherVersion.getMajor())
+                return false;
+            
+            // Major version is equal, minor version is greater
+            if (minor > otherVersion.getMinor())
+                return true;
+            
+            // Minor version is less than or equal to.
+            else {
```
mike-jumper: Same here - it may be possible to reduce the complexity of this through relying on the way you've already structured the comparisons.

mike-jumper: Come to think of it, maybe there's yet a more readable, simpler way to achieve this? Just occurred to me (pardon the pseudocode):

    // Compare major with priority
    if (major different)
        return major > otherVersion.major

    // Major is identical, only minor and patch matter
    if (minor different)
        return minor > otherVersion.minor

    // Major and minor are identical, only patch matters
    return patch >= otherVersion.patch

... unless that's wrong. I might be right, or I might need more coffee, or I might be right **and** need more coffee.

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java
```diff
@@ -0,0 +1,193 @@
...
+    /**
+     * Parse the String format of the version provided and return the
+     * the enum value matching that version.  If no value is provided, return
+     * null.
+     * 
+     * @param version
+     *     The String format of the version to parse.
+     * 
+     * @return
+     *     The enum value that matches the specified version.
```
mike-jumper: or `null` if ...

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java
```diff
@@ -0,0 +1,193 @@
...
+     */
+    public static GuacamoleProtocolVersion getVersion(String version) {
+        
+        if (version == null || version.isEmpty())
+            return null;
+        
+        return valueOf(version);
```
mike-jumper: `valueOf()` will throw an `IllegalArgumentException` if the provided version does not match any of the defined versions. Is that intentional behavior for `getVersion()` as well?

Regardless of whether `IllegalArgumentException` or `null` is the route of choice for an unknown version, should the handshake process continue after receiving a version that is newer than what it supports internally?

Currently the behavior is:

guacd version | guacamole-common version
------------- | ------------------------
Blank (Unknown past version) | Handshake succeeds using 1.0.0 features only
1.1.0 | Handshake succeeds and enables 1.1.0 features
Some future unknown version | Handshake fails internally (`IllegalArgumentException`)

If the client should tolerate unknown future versions, relying on guacd to terminate the connection if it cannot support the version of the protocol declared by the client, then we'll need to move away from an `enum` to something that can represent arbitrary future values (presumably a `class` that is otherwise essentially the same).

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleClientInformation.java
mike-jumper: Possibly relevant: https://www.iana.org/time-zones

necouchman: Ah, okay.  I found it, here:
https://en.wikipedia.org/wiki/Tz_database

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java
necouchman: Well, more coffee never hurt anything.

Ran through several version comparisons, and this seems sane, so I'm going with it.  I knew there had to be a simpler way to do this - thanks!!

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java
necouchman: Ah, yesh.

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java
necouchman: Would we really need to move away from enum, or can we just catch the `IllegalArgumentException` and return VERSION_1_0_0 if nothing else matches?

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleClientInformation.java
necouchman: Updated this a bit.

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java
necouchman: Resolved this with your suggestion below.

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java
necouchman: Implemented based on your pseudocode.

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java
necouchman: I went ahead and made 1.0.0 the default return value.  Let me know if this addresses the concerns or if there's a different route we should go.

mike-jumper: I'm not sure. That will definitely stop the error, but if (hypothetically) there is a `VERSION_1_2_0`, that would mean that `VERSION_1_2_0` + `VERSION_1_1_0` = `VERSION_1_0_0`.

necouchman: Yeah, that's true.  I mean, I could go for a more robust selection method that looks for the highest possible version, using something similar to the `atLeast()` method, where it returns the closest known version, equal to or less than the requested version.  I'll see what I can work out - I'm optimistic that I can do something in the `enum` without swapping over to a `class`.

mike-jumper: Hm - alright. I'm intrigued.

guacamole-ext/src/main/resources/org/apache/guacamole/protocols/rdp.json
```diff
@@ -195,6 +195,10 @@
                 {
                     "name"  : "static-channels",
                     "type"  : "TEXT"
+                },
+                {
+                    "name"  : "timezone",
+                    "type"  : "TIMEZONE"
```
mike-jumper: I don't recall if this was mentioned before, but the `TIMEZONE` field type uses the Java IDs for its values:

https://github.com/apache/guacamole-client/blob/e02dcf8c84f0f7f2f9c866b0258bde0551c847b9/guacamole/src/main/webapp/app/form/controllers/timeZoneFieldController.js#L21-L25

This won't match the IANA values required by the `timezone` parameters. The recent changes from #403 use a `TEXT` field for this, however if you'd like to provide convenient drop-downs, etc., a new field type could be defined that uses the IANA values (`IANA_TIMEZONE`?)

Beware also that the #403 changes added this in a different section, so if you feel that this should be in `device-redirection`, the other parameter definition will need to be deleted.

mike-jumper: Or ... do we know that the Java IDs match?

necouchman: According to this post that I found:
https://stackoverflow.com/questions/44582795/java-8-tz-database-time-zones
...the database shipped with Java is at least based on the same one as the IANA/Olson/Tz Database (tzdb - https://en.wikipedia.org/wiki/Tz_database).  My guess is that it's a reasonably safe assumption that these will match up - or, at the very least, that one is going to be just as out-of-date as another.

I'm happy to implement an `IANA_TIMEZONE` text field, as you suggest, if you think that's the right way, but my gut feeling is that the net value of implementing and maintaining it separately from the other timezones already defined within the code is going to be small, and that, by and large, the timezones already defined within the existing timezone field are going to work fine.

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java
necouchman: Ha.  Maybe I was a little _too_ optimistic about that.  Now that I actually start to think critically about how to do it, I'm not sure that's going to work out.  I'm trying to think of other ways to parse this out such that you could get the maximum supported version...

mike-jumper: What about just a good ol' `Pattern` and `Integer.parseInt()`? I think that would still require shifting from `enum` to `class`, though you could still keep around `VERSION_1_0_0`, etc. static convenience constants (assuming the class is immutable).

guacamole-ext/src/main/resources/org/apache/guacamole/protocols/rdp.json
mike-jumper: Ah, OK then. If we're reasonably sure that they line up, I'm happy as long as the `TIMEZONE` field's documentation is updated to match.

mike-jumper: I did a quick search of all the time zones in my zoneinfo database on my Linux system, and absolutely all time zones currently defined within the field match **except one**: `Canada/East-Saskatchewan`.

This is apparently a rare case of a time zone identifier being removed from the database:

https://stackoverflow.com/questions/47618861/error-invalid-value-for-parameter-timezone-canada-east-saskatchewan/47625517#47625517

So ... I think we are indeed safe, but should probably remove `Canada/East-Saskatchewan`.

mike-jumper: OK, I've opened #404 (PR not found!) to hopefully help with this and keep scope narrow.

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java
necouchman: Yeah, that's probably the best option.  I'll give it a go when the $dayjob$ lets up long enough that I have some time to focus on it.

necouchman: So, I've got this implemented to the point where I can parse the provided string for a major, minor, and patch version component, but I'm having trouble coming up with an algorithm that finds the nearest (equal to or less than provided) version...

mike-jumper:
Is there a need to do that kind of search?

I was thinking it would be sufficient to parse the value as a new instance
of the class, compare to the highest supported version, and choose
whichever is less.

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java
necouchman: Okay - I've actually taken a stab at keeping this an `enum` but grabbing the "most compatible" version for unknown future guacd versions.  Let me know what you think...

mike-jumper:
Alrighty. At this point, I think these changes look absolutely fine. I'll try tinkering with class vs. enum for `GuacamoleProtocolVersion` to see if things feel better one way or the other, but I don't see any reason to hold off on these changes further. They look quite good.

PR #349, opened 2018-12-26
GUACAMOLE-678: Implement new UriGuacamoleProperty

necouchman:
This PR implements a new property type, the `UriGuacamoleProperty`, which parses a string into a URI.  Several of the extensions were converted to use the new property, and I also added properties to the JDBC modules that allow for configuring the connection with a single URI property rather than the individual components.  Providing this type of property allows for validation of URI/URLs in the configuration file, catching errors when the configuration is loaded.

The other advantage of using URI is that, from what I have read, it takes care of all of the encoding of the various URL parts internal, which avoids the need to explicitly use `URLEncoder`.

For the JDBC modules, I configured it such that, if the URI is specified, the other module parameters are completely ignored.  This could go the other direction where the URI is the base and the others override it, if desired - it would be a pretty easy change.

### DIFF
--- a/guacamole-ext/src/main/java/org/apache/guacamole/properties/UriGuacamoleProperty.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/properties/UriGuacamoleProperty.java
@@ -1,0 +1,45 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.guacamole.properties;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import org.apache.guacamole.GuacamoleException;
+import org.apache.guacamole.GuacamoleServerException;
+
+/**
+ * A GuacamoleProperty whose value is a URI.
+ */
+public abstract class UriGuacamoleProperty implements GuacamoleProperty<URI> {
+    
+    @Override
+    public URI parseValue(String value) throws GuacamoleException {
+        
+        try {
+            return new URI(value);
+        }
+        catch (URISyntaxException e) {
+            throw new GuacamoleServerException("Value \"" + value
+                + "\" is not a valid URI.");
+        }
+        
+    }
+    
+}
--- a/extensions/guacamole-auth-cas/pom.xml
+++ b/extensions/guacamole-auth-cas/pom.xml
@@ -261,6 +261,14 @@
             <version>2.5</version>
             <scope>provided</scope>
         </dependency>
+        
+        <!-- Jersey - JAX-RS Implementation -->
+        <dependency>
+            <groupId>javax.ws.rs</groupId>
+            <artifactId>jsr311-api</artifactId>
+            <version>1.1.1</version>
+            <scope>provided</scope>
+        </dependency>
 
     </dependencies>
 
--- a/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/CASGuacamoleProperties.java
+++ b/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/CASGuacamoleProperties.java
@@ -19,7 +19,7 @@
 
 package org.apache.guacamole.auth.cas.conf;
 
-import org.apache.guacamole.properties.UriGuacamoleProperty;
+import org.apache.guacamole.properties.URIGuacamoleProperty;
 
 /**
  * Provides properties required for use of the CAS authentication provider.
@@ -36,8 +36,8 @@ public class CASGuacamoleProperties {
     /**
      * The authorization endpoint (URI) of the CAS service.
      */
-    public static final UriGuacamoleProperty CAS_AUTHORIZATION_ENDPOINT =
-            new UriGuacamoleProperty() {
+    public static final URIGuacamoleProperty CAS_AUTHORIZATION_ENDPOINT =
+            new URIGuacamoleProperty() {
 
         @Override
         public String getName() { return "cas-authorization-endpoint"; }
@@ -49,8 +49,8 @@ public class CASGuacamoleProperties {
      * authentication process is complete. This must be the full URL that a
      * user would enter into their browser to access Guacamole.
      */
-    public static final UriGuacamoleProperty CAS_REDIRECT_URI =
-            new UriGuacamoleProperty() {
+    public static final URIGuacamoleProperty CAS_REDIRECT_URI =
+            new URIGuacamoleProperty() {
 
         @Override
         public String getName() { return "cas-redirect-uri"; }
--- a/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/CASGuacamoleProperties.java
+++ b/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/CASGuacamoleProperties.java
@@ -19,7 +19,7 @@
 
 package org.apache.guacamole.auth.cas.conf;
 
-import org.apache.guacamole.properties.StringGuacamoleProperty;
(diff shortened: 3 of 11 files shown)


guacamole-ext/src/main/java/org/apache/guacamole/properties/UriGuacamoleProperty.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import org.apache.guacamole.GuacamoleException;
+import org.apache.guacamole.GuacamoleServerException;
+
+/**
+ * A GuacamoleProperty whose value is a URI.
+ */
+public abstract class UriGuacamoleProperty implements GuacamoleProperty<URI> {
```
mike-jumper: With the competing camps of applying camelcase to variable/class/function names containing abbreviations, the Guacamole codebase has historically gone the `URIGuacamoleProperty` route rather than the `UriGuacamoleProperty` route. I would suggest doing that here for consistency's sake.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/MySQLAuthenticationProviderModule.java
```diff
@@ -58,11 +59,20 @@ public MySQLAuthenticationProviderModule(MySQLEnvironment environment)
...
         // Set the MySQL-specific properties for MyBatis.
         myBatisProperties.setProperty("mybatis.environment.id", "guacamole");
-        myBatisProperties.setProperty("JDBC.host", environment.getMySQLHostname());
-        myBatisProperties.setProperty("JDBC.port", String.valueOf(environment.getMySQLPort()));
-        myBatisProperties.setProperty("JDBC.schema", environment.getMySQLDatabase());
-        myBatisProperties.setProperty("JDBC.username", environment.getMySQLUsername());
-        myBatisProperties.setProperty("JDBC.password", environment.getMySQLPassword());
+        
+        // Check for URI
+        URI mySQLUri = environment.getMySQLUri();
```
mike-jumper: This is likely a good change to have, allowing users to override the entire JDBC URL, but it seems out of scope for the nature of this particular change. The [GUACAMOLE-678](https://issues.apache.org/jira/browse/GUACAMOLE-678) issue is pretty specifically about defining a new property type:

> Implement a new property type, UriGuacamoleProperty, that takes a string and generates a URI object, validating the URI in the process. With the increased number of extensions that take URIs as configuration parameters this will allow for more central validation of the URI before it gets used by the module.

I agree that it makes sense to refactor existing code to migrate to that type wherever appropriate, but this particular change adds a new feature that deserves its own issue.

guacamole-ext/src/main/java/org/apache/guacamole/properties/UriGuacamoleProperty.java
necouchman: Fair enough, I'll rename it.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/MySQLAuthenticationProviderModule.java
necouchman: Yeah, you're probably right - I'll refactor and pull this code out, and open a separate JIRA issue for it.

necouchman: Okay, JDBC URI configuration has been extracted and will be moved to GUACAMOLE-766.

guacamole-ext/src/main/java/org/apache/guacamole/properties/UriGuacamoleProperty.java
necouchman: Renamed to `URIGuacamoleProperty`.

dwoo2dwoo:
Hi @necouchman,

I'm building docker containers from master since I'm interested in https://github.com/apache/guacamole-client/pull/351.  It seems to work just fine as `guacamole.properties` looks exactly the same as my non-Docker v1.0.0 variant.  Unfortunately, the URL redirect being generated is: 

https://openid-connect.onelogin.com/oidc/auth?scope=openid+email+profile&response_type=id_token&client_id=clientID&redirect_uri=https://redacted.com/guacamole/&nonce=bpv87o46gm8dbuetp6knh85krr

Notice the client_id is set to `clientID`.  Browsing through commits since v1.0.0, this PR stands out as a possible cause, and I thought you might know off the top of your head?

Thanks for any ideas ... David

necouchman:
@dwoo2dwoo: Yep, you're correct - this is the cause of the issue.  I'll open a new pull request to fix it, but basically I used the literal `"clientID"` rather than the variable name `clientID` which is the reason this is happening.

necouchman:
@dwoo2dwoo: Opened #393 to fix this.  Thanks for bringing this to our attention!

dwoo2dwoo:
@necouchman: thank **you** for fixing it so quickly!  I applied your fix and everything is working again.

As an aside, I had planned to submit a PR for the Dockerfile to make Guacamole the [default application](https://wiki.apache.org/tomcat/HowTo#How_do_I_make_my_web_application_be_the_Tomcat_default_application.3F).  Do you see any problem with that?
```
--- Dockerfile.orig     2019-04-18 03:24:58.745149145 -0400
+++ Dockerfile  2019-04-18 17:39:20.305615647 -0400
@@ -50,6 +50,11 @@
 # For the runtime image, we start with the official Tomcat distribution
 FROM tomcat:${TOMCAT_VERSION}-${TOMCAT_JRE}

+# Make Guacamole the default application
+RUN cd /usr/local/tomcat/webapps && \
+    mv ROOT ROOT.orig && \
+    ln -s guacamole ROOT
+
 # This is where the build artifacts go in the runtime image
 WORKDIR /opt/guacamole
```
David

necouchman:
@dwoo2dwoo: This actually should already be possible with changes introduced by #300.  This allows the context to be specified using a Docker environment variable (WEBAPP_CONTEXT), which you could specify as ROOT to make it deploy to the root.  As a project we made a conscious decision not to deploy to root by default, but the option is there if you want to.

dwoo2dwoo:
@necouchman: thanks for the pointer!  I tried it and unfortunately it doesn't work.  I still get the default Tomcat page.  Connecting to the Docker container:

```
[root@wsqaguac001a guacamole]# docker exec -it 52f5c4445af4 bash
root@52f5c4445af4:/opt/guacamole# echo $WEBAPP_CONTEXT
ROOT
root@52f5c4445af4:/opt/guacamole# cd /usr/local/tomcat/webapps
root@52f5c4445af4:/usr/local/tomcat/webapps# ls -ald ROOT* /opt/guacamole/guacamole.war
-rw-r--r--. 1 root root 11982934 Apr 19 03:57 /opt/guacamole/guacamole.war
drwxr-xr-x. 3 root root     4096 Apr 13 00:09 ROOT
lrwxrwxrwx. 1 root root       28 Apr 19 03:58 ROOT.war -> /opt/guacamole/guacamole.war
root@52f5c4445af4:/usr/local/tomcat/webapps# grep -i deploy ../conf/server.xml
            unpackWARs="true" autoDeploy="true">
```

The code in https://github.com/apache/guacamole-client/pull/300 works since the symlink is created; however, Tomcat is not re-deploying ROOT even though it is older than the war file.  Some random Googling implies that this is known behaviour; for example, see the reply in [this thread](https://coderanch.com/t/675618/application-servers/Deploying-WAR-Tomcat).

> As a project we made a conscious decision not to deploy to root by default

In this case, my earlier patch would have been rejected, since I'm replacing ROOT.  Unfortunately, https://github.com/apache/guacamole-client/pull/300 doesn't allow me to replace ROOT either.  Would you be amenable to something like:
```
--- start.sh    2019-04-18 03:24:58.834149054 -0400
+++ /tmp/start.sh       2019-04-19 00:40:32.186160710 -0400
@@ -575,8 +575,11 @@
 start_guacamole() {

     # Install webapp
-    ln -sf /opt/guacamole/guacamole.war /usr/local/tomcat/webapps/${WEBAPP_CONTEXT:-guacamole}.war

+    CONTEXT="${WEBAPP_CONTEXT:-guacamole}"
+    ln -sf /opt/guacamole/guacamole.war /usr/local/tomcat/webapps/${CONTEXT}.war
+    rm -rf /usr/local/tomcat/webapps/${CONTEXT}
+
     # Start tomcat
     cd /usr/local/tomcat
     exec catalina.sh run
```
If not, could you suggest an alternative approach?  Thanks again ... David

necouchman:
@dwoo2dwoo: I submitted #394 to correct this, which essentially does what you've suggested.  Hopefully that clears up the issues you're seeing deploying to the root context.

Thanks for pointing this out!

dwoo2dwoo:
@necouchman: you rock :-)!  I tested the change and it's working perfectly.  Much obliged!!  David

PR #351, opened 2018-12-28
GUACAMOLE-683: Add OpenID support in Docker Build Scripts

krisss85:
This change introduces openid authentication support for the docker build.
Openid is sorted as a first module to make it work together with database. 

The patch was tested together with this commit https://github.com/mike-jumper/guacamole-client/commit/a62a3dcb1125f4bc2dc562016e41df82a3d415db
to make it work against keycloak as openid provider.

### DIFF
--- a/guacamole-docker/bin/build-guacamole.sh
+++ b/guacamole-docker/bin/build-guacamole.sh
@@ -126,3 +126,11 @@ if [ -f extensions/guacamole-auth-radius/target/guacamole-auth-radius*.jar ]; th
     mkdir -p "$DESTINATION/radius"
     cp extensions/guacamole-auth-radius/target/guacamole-auth-radius*.jar "$DESTINATION/radius"
 fi
+
+# Copy OPENID auth extension and schema modifications
+#
+
+if [ -f extensions/guacamole-auth-openid/target/guacamole-auth-openid*.jar ]; then
+    mkdir -p "$DESTINATION/openid"
+    cp extensions/guacamole-auth-openid/target/guacamole-auth-openid*.jar "$DESTINATION/openid"
+fi
--- a/guacamole-docker/bin/start.sh
+++ b/guacamole-docker/bin/start.sh
@@ -404,6 +404,62 @@ END
     ln -s /opt/guacamole/radius/guacamole-auth-*.jar "$GUACAMOLE_EXT"
 }
 
+## Adds properties to guacamole.properties which select the OPENID
+## authentication provider, and configure it to connect to the specified OPENID
+## provider.
+##
+associate_openid() {
+
+    # Verify required parameters are present
+    if [ -z "$OPENID_AUTHORIZATION_ENDPOINT" ] || \
+       [ -z "$OPENID_JWKS_ENDPOINT" ]          || \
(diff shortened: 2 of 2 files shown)


guacamole-docker/bin/build-guacamole.sh
```diff
@@ -126,3 +126,15 @@ if [ -f extensions/guacamole-auth-radius/target/guacamole-auth-radius*.jar ]; th
     mkdir -p "$DESTINATION/radius"
     cp extensions/guacamole-auth-radius/target/guacamole-auth-radius*.jar "$DESTINATION/radius"
 fi
+
+# Copy OPENID auth extension and schema modifications
+#
+
+mkdir -p "$DESTINATION/openid"
+tar -xzf extensions/guacamole-auth-openid/target/*.tar.gz \
```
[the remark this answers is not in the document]
krisss85: @necouchman Thx for checking. You are right. It doesn't make sense to complicate it here and just copy the jar file.

guacamole-docker/bin/start.sh
```diff
@@ -404,6 +404,37 @@ END
...
 }
 
+## Adds properties to guacamole.properties which select the OPENID
+## authentication provider, and configure it to connect to the specified OPENID
+## provider.
+##
+associate_openid() {
+
+    # Verify required parameters are present
+    if [ -z "$OPENID_AUTHORIZATION_ENDPOINT" -o -z "$OPENID_CLIENT_ID" ]; then
```
necouchman: According to the OpenID documentation (http://guacamole.apache.org/doc/gug/openid-auth.html), the following parameters are required:
- openid-authorization-endpoint
- openid-jwks-endpoint
- openid-issuer
- openid-client-id
- openid-redirect-uri

Seems like we should probably check for all of these as required, here, and not just the authorization endpoint and client ID?

krisss85: makes sense indeed, pushed the changes

krisss85:
@necouchman do we still want to do this?

mike-jumper:
Looks pretty good at first glance. I will take a deeper peek.

guacamole-docker/bin/start.sh
```diff
@@ -404,6 +404,42 @@ END
...
+    if [ -z "$OPENID_AUTHORIZATION_ENDPOINT" ] || \
+       [ -z "$OPENID_JWKS_ENDPOINT" ]          || \
+       [ -z "$OPENID_ISSUER" ]                 || \
+       [ -z "$OPENID_CLIENT_ID" ]              || \          
+       [ -z "$OPENID_REDIRECT_URI" ]
+    then
+        cat <<END
+FATAL: Missing required environment variables to setup openid authentication.
+-------------------------------------------------------------------------------
+END
```
mike-jumper: The specifics of the error should be logged. For example, take a look at the error displayed if required variables are missing for MySQL auth:

https://github.com/apache/guacamole-client/blob/78f1ae1b4eac25501d532ddee94fd1d8588e56dc/guacamole-docker/bin/start.sh#L131-L143

krisss85: Makes sense of course. Please verify the latest version with amendments in this section.

PR #352, opened 2018-12-31
GUACAMOLE-684: Insufficient credentials should take precedence over other failures

necouchman:
This resolves an issue where, depending on the order of stacked authentication modules, `GuacamoleInsufficientCredentialsException`s are overriden by `GuacamoleInvalidCredentialsException`s that occur prior to the insufficient exceptions.  This particular issue has come up with the RADIUS module when it gets a challenge/response and is stacked on top of the JDBC module, where the JDBC authentication occurs (and fails) prior to the RADIUS authentication attempt, and the additional credentials are never requested.

### DIFF
--- a/guacamole/src/main/java/org/apache/guacamole/rest/auth/AuthenticationService.java
+++ b/guacamole/src/main/java/org/apache/guacamole/rest/auth/AuthenticationService.java
@@ -36,6 +36,7 @@ import org.apache.guacamole.net.auth.Credentials;
 import org.apache.guacamole.net.auth.UserContext;
 import org.apache.guacamole.net.auth.credentials.CredentialsInfo;
 import org.apache.guacamole.net.auth.credentials.GuacamoleCredentialsException;
+import org.apache.guacamole.net.auth.credentials.GuacamoleInsufficientCredentialsException;
 import org.apache.guacamole.net.auth.credentials.GuacamoleInvalidCredentialsException;
 import org.apache.guacamole.net.event.AuthenticationFailureEvent;
 import org.apache.guacamole.net.event.AuthenticationSuccessEvent;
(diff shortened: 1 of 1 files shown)


mike-jumper:
I'm not sure about this. It's a pretty fundamental change and could have unanticipated consequences. Also, while it may make things easier for the case that you describe, it removes the ability to ensure one extension always takes priority over another.

The current mechanism to ensure that an extension is handled with priority over another is ensure they are named such that they sort in that order. If a user needs RADIUS to take priority over JDBC, I'd say the user should rename the extension to explicitly declare that.

`000-guacamole-auth-radius-1.0.0.jar`, `001-guacamole-auth-jdbc-postgresql-1.0.0.jar`, etc.

necouchman:
Okay, I'm fine deferring this to another version and not pushing it into the 1.0.0 release.  Previous conversations that we have had seemed to indicate that ordering the extensions shouldn't be necessary and that this is a bug or at least unanticipated behavior.  But we can take it up in a future version.

necouchman:
@mike-jumper Any more thoughts on this one?  Something I should open for discussion on the dev@ list before we move forward?

necouchman:
@mike-jumper: Ping.

mike-jumper:
I gave this a fair bit of thought over the course of today, and I think the changes you propose are fine. Rather than deviate from existing behavior, I believe these changes actually fill a gap in the original intent. My reasoning:

* An extension throws `GuacamoleInsufficientCredentialsException` specifically to indicate tentative acceptance of the credentials passed thus far.
* Just as such an extension that fully accepts credentials takes priority over an extension that refuses to accept the same, it makes sense to allow an extension that *tentatively* accepts those credentials to also take priority.

With the above perspective, authentication result priorities are layered in the following order, with ties broken by the inherent order of the auth providers (extensions sorted by filename, multiple auth providers within an extension maintained in the order declared within `guac-manifest.json`):

1. Acceptance (returning an `AuthenticedUser` instance).
2. Tentative acceptance (throwing `GuacamoleInvalidCredentialsException`).
3. Complete refusal (any other subclass of `GuacamoleCredentialsException`).
4. Neither refusal nor acceptance (returning `null`).

I don't foresee breakage as a result of this, and expect this will remove the need to rename extensions in all but unusual cases.

necouchman:
> 2. Tentative acceptance (throwing GuacamoleInvalidCredentialsException).

FWIW, I think this should read:
2. Tentative acceptance (throwing GuacamoleInsufficientCredentialsException)

mike-jumper:
Arg. You're right and that typo is baked in the merge commit. I even said "Authenticed" instead of "Authenticated".

PR #353, opened 2019-01-02
GUACAMOLE-577: Support for configuring the Guacamole Proxy in LDAP Connections

necouchman:
This pull request implements the changes necessary to support the `GuacamoleProxyConfiguration` in LDAP connections (and any future extensions that make use of the `SimpleConnection` class.  Not entirely sure this is the right/best way to go, but took a stab at it.

### DIFF
--- a/extensions/guacamole-auth-ldap/schema/guacConfigGroup.ldif
+++ b/extensions/guacamole-auth-ldap/schema/guacConfigGroup.ldif
@@ -20,9 +20,24 @@
 dn: cn=guacConfigGroup,cn=schema,cn=config
 objectClass: olcSchemaConfig
 cn: guacConfigGroup
-olcAttributeTypes: {0}( 1.3.6.1.4.1.38971.1.1.1 NAME 'guacConfigProtocol' SYNTAX 1.3.6.1.4.1.1466
- .115.121.1.15 )
-olcAttributeTypes: {1}( 1.3.6.1.4.1.38971.1.1.2 NAME 'guacConfigParameter' SYNTAX 1.3.6.1.4.1.146
- 6.115.121.1.15 )
-olcObjectClasses: {0}( 1.3.6.1.4.1.38971.1.2.1 NAME 'guacConfigGroup' DESC 'Guacamole config
- uration group' SUP groupOfNames MUST guacConfigProtocol MAY guacConfigParameter )
+
+olcAttributeTypes: ( 1.3.6.1.4.1.38971.1.1.1 NAME 'guacConfigProtocol'
+  SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+olcAttributeTypes: ( 1.3.6.1.4.1.38971.1.1.2 NAME 'guacConfigParameter'
+  SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+olcAttributeTypes: ( 1.3.6.1.4.1.38971.1.1.3 NAME 'guacConfigProxyHostname'
+  SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+olcAttributeTypes: ( 1.3.6.1.4.1.38971.1.1.4 NAME 'guacConfigProxyPort'
+  SINGLE-VALUE
+  SYNTAX 1.3.6.1.4.1.1466.115.121.1.27 )
+olcAttributeTypes: ( 1.3.6.1.4.1.38971.1.1.5 NAME 'guacConfigProxyEncryption'
+  SINGLE-VALUE
+  SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+olcObjectClasses: ( 1.3.6.1.4.1.38971.1.2.1 NAME 'guacConfigGroup'
+  DESC 'Guacamole configuration group'
+  SUP groupOfNames
+  MUST guacConfigProtocol
+  MAY ( guacConfigParameter $
+  guacConfigProxyHostname $
+  guacConfigProxyPort $
+  guacConfigProxyEncryption ) )
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
@@ -42,7 +42,11 @@ import org.apache.guacamole.auth.ldap.ConnectedLDAPConfiguration;
 import org.apache.guacamole.auth.ldap.ObjectQueryService;
 import org.apache.guacamole.auth.ldap.group.UserGroupService;
 import org.apache.guacamole.auth.ldap.user.LDAPAuthenticatedUser;
+import org.apache.guacamole.environment.LocalEnvironment;
+import org.apache.guacamole.net.auth.AuthenticatedUser;
 import org.apache.guacamole.net.auth.Connection;
+import org.apache.guacamole.net.auth.GuacamoleProxyConfiguration;
+import org.apache.guacamole.net.auth.GuacamoleProxyConfiguration.EncryptionMethod;
 import org.apache.guacamole.net.auth.TokenInjectingConnection;
 import org.apache.guacamole.net.auth.simple.SimpleConnection;
 import org.apache.guacamole.protocol.GuacamoleConfiguration;
@@ -59,6 +63,33 @@ public class ConnectionService {
      * Logger for this class.
      */
     private static final Logger logger = LoggerFactory.getLogger(ConnectionService.class);
+    
+    /**
+     * The name of the LDAP attribute that stores connection configuration
+     * parameters for Guacamole.
+     */
+    public static final String LDAP_ATTRIBUTE_PARAMETER = "guacConfigParameter";
+    
+    /**
+     * The name of the LDAP attribute that stores the protocol for a Guacamole
+     * connection.
+     */
+    public static final String LDAP_ATTRIBUTE_PROTOCOL = "guacConfigProtocol";
+    
+    /**
+     * The name of the LDAP attribute that stores guacd proxy hostname.
+     */
+    public static final String LDAP_ATTRIBUTE_PROXY_HOSTNAME = "guacConfigProxyHostname";
+    
+    /**
+     * The name of the LDAP attribute that stores guacd proxy port.
+     */
+    public static final String LDAP_ATTRIBUTE_PROXY_PORT = "guacConfigProxyPort";
+    
+    /**
+     * The name of the LDAP attribute that stores guacd proxy hostname.
+     */
+    public static final String LDAP_ATTRIBUTE_PROXY_ENCRYPTION = "guacConfigProxyEncryption";
 
     /**
      * Service for executing LDAP queries.
@@ -192,9 +223,19 @@ public class ConnectionService {
                     config.setProtocol(protocol.getString());
                 }
                 catch (LdapInvalidAttributeValueException e) {
-                    logger.error("Invalid value of the protocol entry: {}",
-                            e.getMessage());
+                    logger.error("Invalid value of the protocol entry: {}", e.getMessage());
                     logger.debug("LDAP exception when getting protocol value.", e);
+                    return null;
+                }
+                
+                // Get proxy configuration, if any
+                GuacamoleProxyConfiguration proxyConfig;
+                try {
+                    proxyConfig = getProxyConfiguration(entry);
+                }
+                catch (GuacamoleException e) {
+                    logger.error("Failed to retrieve proxy configuration.", e.getMessage());
+                    logger.debug("Guacamole Exception when retrieving proxy configuration.", e);
                     return null;
                 }
 
@@ -209,10 +250,8 @@ public class ConnectionService {
                             parameter = parameterAttribute.getString();
                         }
                         catch (LdapInvalidAttributeValueException e) {
-                            logger.warn("Parameter value not valid for {}: {}",
-                                    cnName, e.getMessage());
-                            logger.debug("LDAP exception when getting parameter value.",
-                                    e);
+                            logger.warn("Parameter value not valid for {}: {}", cnName, e.getMessage());
+                            logger.debug("LDAP exception when getting parameter value.", e);
                             return null;
                         }
                         parameterAttribute.remove(parameter);
@@ -234,7 +273,7 @@ public class ConnectionService {
                 }
 
                 // Store connection using cn for both identifier and name
-                Connection connection = new SimpleConnection(cnName, cnName, config, true);
+                Connection connection = new SimpleConnection(cnName, cnName, proxyConfig, config, true);
                 connection.setParentIdentifier(LDAPAuthenticationProvider.ROOT_CONNECTION_GROUP);
 
                 // Inject LDAP-specific tokens only if LDAP handled user
@@ -300,6 +339,65 @@ public class ConnectionService {
         searchFilter.addNode(groupFilter);
 
         return searchFilter;
+    }
+    
+    /**
+     * Given an LDAP entry that stores a GuacamoleConfiguration, generate a
+     * GuacamoleProxyConfiguration that tells the client how to connect to guacd.
+     * If the proxy configuration values are not found in the LDAP entry the
+     * defaults from the environment are used. If errors occur while trying to
+     * ready or parse values from the LDAP entry a GuacamoleException is thrown.
+     * 
+     * @param connectionEntry
+     *     The LDAP entry that should be checked for proxy configuration values.
+     * 
+     * @return
+     *     The GuacamoleProxyConfiguration that contains information on how
+     *     to contact guacd for the given Guacamole connection configuration.
+     * 
+     * @throws GuacamoleException 
+     *     If errors occur trying to parse LDAP values from the entry.
+     */
+    private GuacamoleProxyConfiguration getProxyConfiguration(Entry connectionEntry)
+            throws GuacamoleException {
+        
+        try {
+                    
+            // Get default proxy configuration values
+            GuacamoleProxyConfiguration proxyConfig = LocalEnvironment.getInstance().getDefaultGuacamoleProxyConfiguration();
+            String proxyHostname = proxyConfig.getHostname();
+            int proxyPort = proxyConfig.getPort();
+            EncryptionMethod proxyEncryption = proxyConfig.getEncryptionMethod();
+
+            // Get the proxy hostname
+            Attribute proxyHostAttr = connectionEntry.get(LDAP_ATTRIBUTE_PROXY_HOSTNAME);
+            if (proxyHostAttr != null && proxyHostAttr.size() > 0)
+                proxyHostname = proxyHostAttr.getString();
+
+            // Get the proxy port
+            Attribute proxyPortAttr = connectionEntry.get(LDAP_ATTRIBUTE_PROXY_PORT);
+            if (proxyPortAttr != null && proxyPortAttr.size() > 0)
+                proxyPort = Integer.parseInt(proxyPortAttr.getString());
+
+            // Get the proxy encryption method
+            Attribute proxyEncryptionAttr = connectionEntry.get(LDAP_ATTRIBUTE_PROXY_ENCRYPTION);
+            if (proxyEncryptionAttr != null && proxyEncryptionAttr.size() > 0) {
+                try {
+                    proxyEncryption = EncryptionMethod.valueOf(proxyEncryptionAttr.getString());
+                }
+                catch (IllegalArgumentException e) {
+                    throw new GuacamoleServerException("Unknown encryption method specified, value must be either \"NONE\" or \"SSL\".", e);
+                }
+            }
+
+            // Return a new proxy configuration
+            return new GuacamoleProxyConfiguration(proxyHostname, proxyPort, proxyEncryption);
+        }
+        catch (LdapInvalidAttributeValueException e) {
+            logger.error("Invalid value in proxy configuration: {}", e.getMessage());
+            logger.debug("LDAP exception fetching proxy attribute value.", e);
+            throw new GuacamoleServerException("Invalid LDAP value in proxy configuration.", e);
+        }
     }
 
 }
--- a/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleConnection.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleConnection.java
@@ -24,7 +24,6 @@ import java.util.Date;
 import java.util.Map;
 import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.GuacamoleServerException;
-import org.apache.guacamole.environment.Environment;
 import org.apache.guacamole.environment.LocalEnvironment;
 import org.apache.guacamole.net.GuacamoleSocket;
 import org.apache.guacamole.net.GuacamoleTunnel;
@@ -53,6 +52,11 @@ public class SimpleConnection extends AbstractConnection {
      * Backing configuration, containing all sensitive information.
      */
     private GuacamoleConfiguration fullConfig;
+    
+    /**
+     * The proxy configuration describing how to connect to guacd.
+     */
+    private GuacamoleProxyConfiguration proxyConfig;
 
     /**
      * Whether parameter tokens in the underlying GuacamoleConfiguration should
@@ -158,6 +162,39 @@ public class SimpleConnection extends AbstractConnection {
         this.interpretTokens = interpretTokens;
 
     }
+    
+    /**
+     * Creates a new SimpleConnection having the given identifier,
+     * GuacamoleConfiguration, and GuacamoleProxyConfiguration. Parameter tokens
+     * will be interpreted if explicitly requested.
+     *
+     * @param name
+     *     The name to associate with this connection.
+     *
+     * @param identifier
+     *     The identifier to associate with this connection.
+     * 
+     * @param proxyConfig
+     *     The Guacamole proxy configuration describing how the connection to
+     *     guacd should be established, or null if the default settings will be
+     *     used.
+     *
+     * @param config
+     *     The configuration describing how to connect to this connection.
+     *
+     * @param interpretTokens
+     *     Whether parameter tokens in the underlying GuacamoleConfiguration
+     *     should be automatically applied upon connecting. If false, parameter
+     *     tokens will not be interpreted at all.
+     */
+    public SimpleConnection(String name, String identifier,
+            GuacamoleProxyConfiguration proxyConfig,
+            GuacamoleConfiguration config, boolean interpretTokens) {
+
+        this(name, identifier, config, interpretTokens);
+        this.proxyConfig = proxyConfig;
+
+    }
 
     /**
      * Returns the GuacamoleConfiguration describing how to connect to this
@@ -201,9 +238,9 @@ public class SimpleConnection extends AbstractConnection {
     public GuacamoleTunnel connect(GuacamoleClientInformation info)
             throws GuacamoleException {
 
-        // Retrieve proxy configuration from environment
-        Environment environment = LocalEnvironment.getInstance();
-        GuacamoleProxyConfiguration proxyConfig = environment.getDefaultGuacamoleProxyConfiguration();
+        // Retrieve proxy configuration from environment if we don't have one
+        if (proxyConfig == null)
+            proxyConfig = LocalEnvironment.getInstance().getDefaultGuacamoleProxyConfiguration();
 
         // Get guacd connection parameters
         String hostname = proxyConfig.getHostname();
--- a/extensions/guacamole-auth-ldap/schema/guacConfigGroup.schema
+++ b/extensions/guacamole-auth-ldap/schema/guacConfigGroup.schema
@@ -18,14 +18,28 @@
 #
 
 attributetype ( 1.3.6.1.4.1.38971.1.1.1 NAME 'guacConfigProtocol'
-	SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+    SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
 
 attributetype ( 1.3.6.1.4.1.38971.1.1.2 NAME 'guacConfigParameter'
-	SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+    SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+
+attributetype ( 1.3.6.1.4.1.38971.1.1.3 NAME 'guacConfigProxyHostname'
+    SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+
+attributetype ( 1.3.6.1.4.1.38971.1.1.4 NAME 'guacConfigProxyPort'
+    SINGLE-VALUE
+    SYNTAX 1.3.6.1.4.1.1466.115.121.1.27 )
+
+attributetype ( 1.3.6.1.4.1.38971.1.1.5 NAME 'guacConfigProxyEncryption'
+    SINGLE-VALUE
+    SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
 
 objectClass ( 1.3.6.1.4.1.38971.1.2.1 NAME 'guacConfigGroup'
     DESC 'Guacamole configuration group'
     SUP groupOfNames
     MUST guacConfigProtocol
-    MAY guacConfigParameter )
+    MAY ( guacConfigParameter $
+          guacConfigProxyHostname $
+          guacConfigProxyPort $
+          guacConfigProxyEncryption ) )
 


extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -167,18 +201,34 @@
                             // Parse name
                             String name = parameter.substring(0, equals);
                             String value = parameter.substring(equals+1);
-
-                            config.setParameter(name, value);
+                            
+                            // Pull out and set proxy parameters, if present
+                            // Otherwise set the parameter.
+                            switch(name) {
+                                case PROXY_HOST_PARAMETER:
```
mike-jumper: These should probably be separate, dedicated LDAP attributes, rather than reusing `guacConfigParameter` (which is meant for connection parameters). This would mean changes to the LDAP schema, which would mean new OID values.

I have the OID number assignment scheme written down on a scrap of paper from years past ... I'll locate my notes and copy them here. There's an OID space which was reserved for Guacamole ages ago (see 38971 within https://www.iana.org/assignments/enterprise-numbers/enterprise-numbers), but there is a system for consistently assigning numbers within that space.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -148,12 +169,25 @@
 
                 // Set protocol
                 GuacamoleConfiguration config = new GuacamoleConfiguration();
+                
+                GuacamoleProxyConfiguration proxyConfig;
+                try {
+                    proxyConfig = new LocalEnvironment().getDefaultGuacamoleProxyConfiguration();
```
mike-jumper: The `LocalEnvironment` should not be repeatedly recreated if it can be avoided.

A singleton instance of it is actually already bound during startup of the LDAP extension, so that would be the instance of choice:

https://github.com/apache/guacamole-client/blob/78f1ae1b4eac25501d532ddee94fd1d8588e56dc/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPAuthenticationProviderModule.java#L74

That singleton instance can be injected within this class if needed, for example:

https://github.com/apache/guacamole-client/blob/78f1ae1b4eac25501d532ddee94fd1d8588e56dc/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java#L41-L45

Since this is configuration, though, I'd think it would make more sense to provide this via a function within `ConfigurationService`.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -148,12 +169,25 @@
...
+                }
+                catch (GuacamoleException e) {
+                    return null;
```
mike-jumper: If retrieval of the proxy config fails, that failure should be logged rather than silently ignored.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Connection.java
```diff
@@ -125,5 +125,17 @@
...
+    /**
+     * Get the GuacamoleProxyConfiguration for this connection.
+     * 
+     * @return
+     *     The GuacamoleProxyConfiguration for this connection.
+     * 
+     * @throws GuacamoleException 
+     *     If configuration information cannot be retrieved or parsed.
+     */
+    public GuacamoleProxyConfiguration getGuacamoleProxyConfiguration()
```
mike-jumper: We have historically avoided making the assumption that extensions will define a means of connecting to guacd at the `Connection` level. The connection to an underlying Guacamole proxy, whether such a proxy is used at all, etc. is all kept abstract behind the `GuacamoleTunnel` returned by a call to `connect()`.

I think it would be better to continue avoiding that assumption.

If we will be providing such a getter for convenience (which is OK), there needs to be well-defined semantics allowing implementations to choose not to define it at that level (returning `null` perhaps).

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Connection.java
```diff
@@ -125,5 +125,17 @@
      *     connection.
      */
     public Set<String> getSharingProfileIdentifiers() throws GuacamoleException;
+    
+    /**
+     * Get the GuacamoleProxyConfiguration for this connection.
```
mike-jumper: There should be more to the documentation of `X.getFoo()` than "Gets the Foo for this X". There are semantics which apply here that should be noted.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Connection.java
```diff
@@ -125,5 +125,17 @@
...
+     * 
+     * @return
+     *     The GuacamoleProxyConfiguration for this connection.
+     * 
+     * @throws GuacamoleException 
+     *     If configuration information cannot be retrieved or parsed.
```
mike-jumper: Implementations will not necessarily be retrieving or parsing this information. We can't assume that `GuacamoleException` will be thrown only due to failures from function calls made against an `Environment`.

Any thoughts on not throwing `GuacamoleException`? `getConfiguration()` does not do so.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleConnection.java
```diff
@@ -68,21 +72,34 @@ public SimpleConnection() {
      * @param identifier The identifier to associate with this connection.
      * @param config The configuration describing how to connect to this
      *               connection.
+     * 
+     * @throws GuacamoleException
+     *     If the default proxy configuration cannot be retrieved.
      */
     public SimpleConnection(String name, String identifier,
-            GuacamoleConfiguration config) {
+            GuacamoleConfiguration config) throws GuacamoleException {
```
mike-jumper: If possible, we should avoid adding new exceptions to an existing constructor, as downstream users of that constructor will suddenly have new things to catch.

Perhaps retrieving the configuration from the local environment could still be put off until `connect()`, being the default behavior if no `GuacamoleProxyConfiguration` is provided?

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleConnection.java
```diff
@@ -68,21 +72,34 @@ public SimpleConnection() {
...
 
         // Set config
-        setConfiguration(config);
+        super.setConfiguration(config);
         this.config = config;
+        this.proxyConfig = new LocalEnvironment().getDefaultGuacamoleProxyConfiguration();
 
     }
+    
+    public SimpleConnection(String name, String identifier,
```
mike-jumper: Please document this constructor.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
necouchman: Sounds good.  I hesitated to modify the LDAP schema, but I'm happy to go that route if that's the best way to go.  Do you think separate LDAP attributes specifically for proxying, or just a generic Guacamole Attribute that is handled similar to the parameters?

For the OID, should we use the Guacamole one, or does ASF have one we should go with?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
necouchman: I originally went this route, but thought it might be a little overkill :-).  I'll go back to this method...

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
necouchman: Got it.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleConnection.java
necouchman: Okay, will refactor.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
mike-jumper: We could switch over to the ASF one: https://cwiki.apache.org/confluence/display/DIRxPMGT/OID+Assignment+Scheme

Probably a good idea to do so. I'll still locate the ancient scrap of paper to server as a reference for assigning numbers from that, though - we need to use some consistent method for assigning the numbers within the Guacamole part of the ASF space (once that exists).

necouchman: Cool.  Do we need to open an INFRA ticket to get that OID assigned under ASF?  Or is there some documentation somewhere on the ASF site that documents this?

mike-jumper: According to https://cwiki.apache.org/confluence/display/DIRxPMGT/OID+Assignment+Scheme:

> Each contact person is the authority for assigning unique OID values and ranges to projects or persons. Contact that person for more assignments.

The contact person listed for the main ASF OID is Alex Karasulu. Feels weird to me to not use a mailing list, but I'll shoot him an email. ;)

mike-jumper: Depending on OID assignment, it looks like we should be fine with the above system. There is no standard limit imposed on OID length, and longer OIDs are already actively used.

necouchman: Cool.  I'm thinking the right way to do this is to continue with the current OID in this PR, and then open a separate JIRA issue (and PR) for converting from the old OID to the ASF OID?

necouchman: Okay, created a new LDAP attribute for the proxy configuration, under the existing OID string.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
necouchman: Okay, switched back to this method.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
necouchman: Logged.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Connection.java
necouchman: Okay, I've removed this entirely from the interfaces and abstracts.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Connection.java
necouchman: Removed.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleConnection.java
necouchman: Backed this change out and just kept things in the `connect()` method.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleConnection.java
necouchman: Documented.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Connection.java
necouchman: Pulled this out.

necouchman:
Okay, fixed a few things up and made some of the requested changes.  Let me know how it looks, now.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
mike-jumper: Alrighty - we now have confirmation from the Apache Directory folks: https://lists.apache.org/thread.html/1897a20401348da14c4b20483464ee8ee06ffb510a29fa86d475d200@%3Cdev.directory.apache.org%3E

We can proceed with `1.3.6.1.4.1.18060.18`, and they will be updating the Apache Directory website to list the various reserved ASF OID branches.

necouchman: Cool.  Would you rather go ahead and make those changes in this PR, or go the route of separating that into another issue (I'd already opened a separate JIRA issue for it...)?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
```diff
@@ -395,5 +396,20 @@ public String getMemberAttribute() throws GuacamoleException {
             "member"
         );
     }
+    
+    /**
+     * Return the default configuration for how a the client connects
```
mike-jumper, line 401: "how the"*

mike-jumper (changes requested):

necouchman:
I may end up closing this PR and just redoing it - I tried to resolve the conflicts between current master and this PR, and it was feeling a little like the kitty-in-a-bathtub gif...

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
necouchman: I think changing OIDs will be saved for a different JIRA issue...

necouchman:
Okay, @mike-jumper, I've reworked this on top of the existing master, so I think it's good, now...

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleConnection.java
```diff
@@ -157,6 +162,43 @@ public SimpleConnection(String name, String identifier,
...
+            GuacamoleConfiguration config, boolean interpretTokens,
+            GuacamoleProxyConfiguration proxyConfig) {
+
+        super.setName(name);
+        super.setIdentifier(identifier);
+        super.setConfiguration(config);
+
+        this.fullConfig = config;
+        this.interpretTokens = interpretTokens;
```
mike-jumper, line 199: I think we should be DRY here and just `this(name, identifier, config, interpretTokens);`.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleConnection.java
```diff
@@ -157,6 +162,43 @@ public SimpleConnection(String name, String identifier,
         this.fullConfig = config;
         this.interpretTokens = interpretTokens;
 
+    }
+    
+        /**
+     * Creates a new SimpleConnection having the given identifier,
```
mike-jumper, line 168: Indentation went wonky here.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -171,9 +210,73 @@
...
+                // Get the default Proxy configuration
+                GuacamoleProxyConfiguration proxyConfig;
+                try {
+                    proxyConfig = confService.getDefaultGuacamoleProxyConfiguration();
+                }
+                catch (GuacamoleException e) {
+                    logger.warn("Could not retrieve Guacamole proxy configuration.",
+                            e.getMessage());
+                    logger.debug("Error when trying to get Guacamole proxy configuration.",
+                            e);
```
mike-jumper, line 223: While I definitely appreciate code wrapped to a nice, reassuring 80 columns, I think these are examples of cases where doing so doesn't improve readability. [Our general style guidelines](http://guacamole.apache.org/guac-style/#general-style) err on the side of caution:

> ### General Style
> ...
> Wrap lines to 80 columns when doing so does not decrease readability.

Buuuut... whether wrapping decreases readability in any particular case is pretty subjective. If you prefer this block of code to be wrapped up all warm and cozy in 80 columns, I'm not going to fight to pick this nit, but I feel it's worth mentioning.

There may be
a
few other cases in this body of code where _not_ wrapping could arguably be
a
better choice.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -59,6 +62,42 @@
...
+     * The proxy parameter name that configures the TCP port that will be used
+     * to connect to guacd.
+     */
+    public static final String PROXY_PARAMETER_PORT = "proxy-port";
+    
+    /**
+     * The proxy parameter name that configures the encryption method that will
+     * be used when connected to guacd.
+     */
+    public static final String PROXY_PARAMETER_ENCRYPTION = "proxy-encryption";
```
mike-jumper, line 100: I think these and the related parameter parsing should be gone in favor of the new `guacConfigProxy` attribute.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -171,9 +210,73 @@
                     logger.debug("LDAP exception when getting protocol value.", e);
                     return null;
                 }
+                
+                // Get the default Proxy configuration
+                GuacamoleProxyConfiguration proxyConfig;
+                try {
+                    proxyConfig = confService.getDefaultGuacamoleProxyConfiguration();
```
mike-jumper, line 217: If `SimpleConnection` will use the default configuration from the environment if no configuration is provided, is this still needed?

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleConnection.java
```diff
@@ -157,6 +162,43 @@ public SimpleConnection(String name, String identifier,
...
+     *     The configuration describing how to connect to this connection.
+     *
+     * @param interpretTokens
+     *     Whether parameter tokens in the underlying GuacamoleConfiguration
+     *     should be automatically applied upon connecting. If false, parameter
+     *     tokens will not be interpreted at all.
+     * 
+     * @param proxyConfig
+     *     The Guacamole proxy configuration used to tell the client how to
+     *     connect to guacd.
```
mike-jumper, line 188: As it may be useful to invoke this constructor for cases where the default environment is desired (rather than conditionally invoke the other constructor to ensure defined behavior), I think we should add an "..., or null to ..." bit to explicitly document that behavior.

mike-jumper, line 188: > The Guacamole proxy configuration used **to tell the client** how to connect to guacd.

I think the wording here is a bit confusing, since the code invoking this constructor _is_ the client. I suggest rewording to something like: "A GuacamoleProxyConfiguration object describing how the connection to guacd should be established."

(I'm partly copy/pasting from https://guacamole.apache.org/doc/1.3.0/guacamole-ext/org/apache/guacamole/net/auth/GuacamoleProxyConfiguration.html)

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleConnection.java
```diff
@@ -157,6 +162,43 @@ public SimpleConnection(String name, String identifier,
...
+     */
+    public SimpleConnection(String name, String identifier,
+            GuacamoleConfiguration config, boolean interpretTokens,
+            GuacamoleProxyConfiguration proxyConfig) {
```
mike-jumper, line 192: Thoughts on parameter order here?

To me, something like:

```java
    public SimpleConnection(String name, String identifier,
            GuacamoleProxyConfiguration proxyConfig
            GuacamoleConfiguration config, boolean interpretTokens) {
```

feels more predictable and logical.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleConnection.java
```diff
@@ -53,6 +52,12 @@
      * Backing configuration, containing all sensitive information.
      */
     private GuacamoleConfiguration fullConfig;
+    
+    /**
+     * The proxy configuration used to tell the client how to connect to
```
mike-jumper, line 57: Same here - I'm iffy on the phrasing "to tell the client how to connect" given the context of this object.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -152,11 +191,11 @@
                 }
 
                 // Get associated protocol
-                Attribute protocol = entry.get("guacConfigProtocol");
+                Attribute protocol = entry.get(LDAP_ATTRIBUTE_PROTOCOL);
                 if (protocol == null) {
                     logger.warn("guacConfigGroup \"{}\" is missing the "
-                              + "required \"guacConfigProtocol\" attribute.",
-                            cnName);
```
mike-jumper, line 159: Heh - looks like I'm guilty of brutal over-wrapping myself. ;)

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/ConfigurationService.java
```diff
@@ -391,5 +392,21 @@ public MemberAttributeType getMemberAttributeType()
...
+     *     The default configuration used by Guacamole Client for connecting
+     *     to guacd.
+     * 
+     * @throws GuacamoleException 
+     *     If guacamole.properties cannot be parsed.
+     */
+    public GuacamoleProxyConfiguration getDefaultGuacamoleProxyConfiguration()
+            throws GuacamoleException {
+        return environment.getDefaultGuacamoleProxyConfiguration();
+    }
```
mike-jumper, line 410: This may not be needed depending on whether we can simply rely on `SimpleConnection` to pull this for us.

mike-jumper (changes requested):

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
necouchman: Under
stood,
makes perfect sense to
me.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleConnection.java
necouchman: I'm good with that.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
necouchman: Hehe - I'll unwrap things throughout this file.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/ConfigurationService.java
necouchman: Removed.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleConnection.java
necouchman: Fixed.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
necouchman: Makes sense.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleConnection.java
necouchman: Sure, sounds good.

necouchman: OKay, makes sense.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleConnection.java
necouchman: Ah, yes.

extensions/guacamole-auth-ldap/schema/guacConfigGroup.ldif
```diff
@@ -20,9 +20,14 @@
...
-olcAttributeTypes: {1}( 1.3.6.1.4.1.38971.1.1.2 NAME 'guacConfigParameter' SYNTAX 1.3.6.1.4.1.146
- 6.115.121.1.15 )
-olcObjectClasses: {0}( 1.3.6.1.4.1.38971.1.2.1 NAME 'guacConfigGroup' DESC 'Guacamole config
- uration group' SUP groupOfNames MUST guacConfigProtocol MAY guacConfigParameter )
+olcAttributeTypes: {0}( 1.3.6.1.4.1.38971.1.1.1 NAME 'guacConfigProtocol'
+ SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+olcAttributeTypes: {1}( 1.3.6.1.4.1.38971.1.1.2 NAME 'guacConfigParameter'
+ SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+olcAttributeTypes: {2}( 1.3.6.1.4.1.38971.1.1.3 NAME 'guacConfigProxy'
+￼SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
```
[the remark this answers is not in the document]
necouchman: Weird.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
necouchman: I'm not sure I understand? These were added as part of the `guacConfigProxy` attribute - what should I be replacing them with?

mike-jumper: Oh, I thought these were holdovers from before `guacConfigProxy` when `guacConfigParameter` was being used. I misread the parsing leveraging these constants as that connection parameter parsing still being there, with `guacConfigProxy` being a direct retrieval elsewhere.

necouchman: Okay - if you've another suggestion for parsing out guacConfigProxy, I'm certainly not opposed to a better way, but this is just the way I had gone about implementing it - with constants for each of the possible parameter names for `guacConfigProxy`.

mike-jumper: Rather than parsing the content of `guacConfigProxy`, would it make more sense to do away with that and instead define `guacConfigProxyHostname`, `guacConfigProxyPort`, and `guacConfigProxyEncryption`?

`guacConfigParameter` has to be special since it defines connection parameters, and those must be arbitrary name/value pairs, but the parameters for connecting to guacd are not arbitrary name/value pairs.

necouchman: Yeah, that makes sense. I'll re-factor it for that.

necouchman:
I reworked the proxy attributes from LDAP as you suggested, but came across a couple of things:
* I used the `INTEGER` syntax for the `guacProxyPort` attribute, but it's unclear to me if there are any implications to this - that is, will the Apache Directory API's `Attribute` `getString()` method actually return a `String` for this, or is it going to throw an exception? I _think_ it'll return a `String`, because this isn't a binary value, but I've no easy way to verify that.
* It ended up being easier to go ahead and grab the default proxy configuration in the LDAP `ConnectionService` code and use that as a base for then trying to pull values from LDAP. I know this is a bit redundant, but it seemed like the cleanest way to me to handle 1) initialization of those variables in `ConnectionService` (vs. assigning `null` and then checking to see if anything had been pulled from LDAP), and 2) cleanly handling situations where one values has been specified in LDAP but the others haven't (e.g. user assigns guacProxyHostname in LDAP, but leaves out guacProxyPort and guacProxyEncryption, and then GuacamolepProxyConfiguration is initialized with one real value and two unknown/other values). If there's a better way, don't hesitate to send that along.

extensions/guacamole-auth-ldap/schema/guacConfigGroup.ldif
```diff
@@ -20,9 +20,24 @@
...
+ SINGLE-VALUE
+ SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+olcObjectClasses: {0}( 1.3.6.1.4.1.38971.1.2.1 NAME 'guacConfigGroup'
+ DESC 'Guacamole configuration group'
+ SUP groupOfNames
+ MUST guacConfigProtocol
+ MAY ( guacConfigParameter $ 
+  guacConfigProxyHostname $
+  guacConfigProxyPort $
+  guacConfigProxyEncryption ) )
```
mike-jumper, line 43: When I run this against OpenLDAP with `ldapadd`, I get:

```
adding new entry "cn=guacConfigGroup,cn=schema,cn=config"
ldap_add: Other (e.g., implementation specific) error (80)
	additional info: olcAttributeTypes: Unexpected token before  1.3.6.1.4.1.1466.115.121.1.15 )
```

I think this is because of how the LDIF format handles indentation, with a single space being a continuation of the previous line. Something like:

```
FOO
 BAR
```

is equivalent to "FOOBAR", whereas:

```
FOO
    BAR
```

is equivalent to "FOO BAR".

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -59,6 +62,33 @@
...
+     */
+    public static final String LDAP_ATTRIBUTE_PROTOCOL = "guacConfigProtocol";
+    
+    /**
+     * The name of the LDAP attribute that stores guacd proxy hostname.
+     */
+    public static final String LDAP_ATTRIBUTE_PROXY_HOSTNAME = "guacConfigProxyHostname";
+    
+    /**
+     * The name of the LDAP attribute that stores guacd proxy port..
```
mike-jumper, line 84: You have two "." here.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/connection/ConnectionService.java
```diff
@@ -166,14 +194,52 @@
...
+                    
+                    // Get the proxy port
+                    Attribute proxyPortAttr = entry.get(LDAP_ATTRIBUTE_PROXY_PORT);
+                    if (proxyPortAttr != null && proxyPortAttr.size() > 0)
+                        proxyPort = Integer.parseInt(proxyPortAttr.getString());
+                    
+                    // Get the proxy encryption method
+                    Attribute proxyEncryptionAttr = entry.get(LDAP_ATTRIBUTE_PROXY_ENCRYPTION);
+                    if (proxyEncryptionAttr != null && proxyEncryptionAttr.size() > 0)
+                        proxyEncryption = EncryptionMethod.valueOf(proxyEncryptionAttr.getString());
```
mike-jumper, line 225: This will throw an `IllegalArgumentException` if the provided value does not identically match one of the internal names of the enum values (currently `NONE` and `SSL`).

mike-jumper, line 225: To avoid making this existing function more complex, I recommend separating out this logic into its own function. Something that returns the `GuacamoleProxyConfiguration` for an arbitrary `Entry` (regardless of whether all or part of the ultimate configuration come from the environment) could abstract this nicely.

mike-jumper (changes requested):

extensions/guacamole-auth-ldap/schema/guacConfigGroup.ldif
mike-jumper: The "schema2ldif" tool should be helpful for producing a basis:

https://manpages.ubuntu.com/manpages/xenial/man1/schema2ldif.1.html

I'm not sure whether that's the same tool that was used to produce the basis for the original LDIF, but the original LDIF was definitely mostly auto-generated from the schema file.

necouchman: Okay, installed schema2ldif and used that to generate it - hopefully that works better!

mike-jumper (approved):

mike-jumper (changes requested):
Merge conflict, but otherwise LGTM.

necouchman:
> Merge conflict, but otherwise LGTM.

Merge conflict is resolved.

mike-jumper (approved):

PR #354, opened 2019-01-09
GUACAMOLE-688: added LDAP_USER_SEARCH_FILTER to docker start script

jolentes:


guacamole-docker/bin/start.sh
```diff
@@ -317,6 +317,10 @@ END
         "ldap-search-bind-password" \
         "$LDAP_SEARCH_BIND_PASSWORD"
 
+    set_optional_property       \
+        "ldap-user-search-filter" \
```
necouchman: Ever so minor nit-pick, but, in keeping with the style on the rest of the document, please line up the `\` at the end of the lines, here.

jolentes: np, aligned it first and then changed the second line indent :-(
I'm with you on clean code style.

PR #355, opened 2019-01-11
GUACAMOLE-271: Duo in docker image

scottpas:
Added Duo to the docker build and start scripts.

### DIFF
--- a/guacamole-docker/bin/start.sh
+++ b/guacamole-docker/bin/start.sh
@@ -467,9 +467,9 @@ END
 ##
 associate_duo() {
     # Verify required parameters are present
-    if [ -z "$DUO_INTEGRATION_KEY" ]    || \
-    [ -z "$DUO_SECRET_KEY" ]            || \
-    [ ${#DUO_APPLICATION_KEY} -lt 40 ]
+    if [ -z "$DUO_INTEGRATION_KEY" ]      || \
+       [ -z "$DUO_SECRET_KEY" ]           || \
+       [ ${#DUO_APPLICATION_KEY} -lt 40 ]
     then
         cat <<END
 FATAL: Missing required environment variables
--- a/guacamole-docker/bin/start.sh
+++ b/guacamole-docker/bin/start.sh
@@ -461,6 +461,45 @@ END
 }
 
 ##
+## Adds properties to guacamole.properties which configure the Duo two-factor
+## authentication service. Checks to see if all variables are defined and makes sure
+## DUO_APPLICATION_KEY is >= 40 characters.
+##
+associate_duo() {
+    # Verify required parameters are present
+    if [ -z "$DUO_INTEGRATION_KEY" ]    || \
+    [ -z "$DUO_SECRET_KEY" ]            || \
+    [ ${#DUO_APPLICATION_KEY} -lt 40 ]
+    then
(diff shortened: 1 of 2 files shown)


mike-jumper:
@scottpas, please rebase things such that:

1. Only commits relevant to your changes are present (no commits from other changes, no merge commits)
2. The commit(s) that remain follow our [contribution guidelines](https://github.com/apache/guacamole-client/blob/master/CONTRIBUTING), specifically:

   >
   > Commits should be made in logical units and must reference the JIRA issue number:
   >
   >
   > Avoid commits which cover multiple, distinct goals that could (and should) be handled separately.
   >

   Something like "add Duo to Docker build" is missing the JIRA issue number, while "fix formatting" and "rebase change" both need the issue number and need to be more specific.

I'd also like to point out that it wouldn't make sense for a commit like "rebase change" to exist. By definition, rebasing replaces the existing commits affected by the rebase. If your changes were originally covered by one commit, rebasing your branch would replace that commit with a new version of the same, not create an additional commit while leaving the rest unaffected. Perhaps something went wrong during rebase?

scottpas:
@mike-jumper Sorry, I'm a git noob and couldn't figure out how to update the changes from master onto my branch. Any idea how I can go back and edit the commits? Or should I just close this PR, fix it, and open a new one?

scottpas:
@mike-jumper Ok, I think it's fixed now. Let me know if I still need to make any changes

guacamole-docker/bin/start.sh
```diff
@@ -591,6 +591,18 @@ END
...
 fi
 
+# Use Duo if specified
+if [ -n "$DUO_API_HOSTNAME" ] &&        \
+    [ -n "$DUO_INTEGRATION_KEY" ] &&    \
+    [ -n "$DUO_SECRET_KEY" ] &&         \
+    [ -n "$DUO_APPLICATION_KEY" ] ; then
+    set_optional_property "duo-api-hostname"        "$DUO_API_HOSTNAME"
+    set_optional_property "duo-integration-key"     "$DUO_INTEGRATION_KEY"
+    set_optional_property "duo-secret-key"          "$DUO_SECRET_KEY"
```
mike-jumper: These properties are actually all required, not optional:

http://guacamole.apache.org/doc/gug/duo-auth.html#guac-duo-config

If any are absent, the startup script should abort and warn the user, as with the other extensions supported by the image. For example:

https://github.com/apache/guacamole-client/blob/9dcee8bdac2d40ae98ebabb53bdba2e9237ca93c/guacamole-docker/bin/start.sh#L128-L146

This should also be set up in a similar manner to the others. We currently have extension-specific setup, parsing, etc. is organized within separate functions, keeping the logic which follows later simple and unobstructed:

https://github.com/apache/guacamole-client/blob/9dcee8bdac2d40ae98ebabb53bdba2e9237ca93c/guacamole-docker/bin/start.sh#L551-L555

You're correct to not update the `INSTALLED_AUTH` environment variable, as things will not work unless some other auth mechanism is set up, but things will be easier to read and more maintainable if the existing pattern is followed.

scottpas: Ok, changes made! Let me know if you think any others are necessary.

PR #358, opened 2019-01-17
[WIP] GUACAMOLE-706: added variable to define link name of RADIUS extension

jolentes:
To combine authentication extension for RADIUS with database authentication extensions it is required to change the order of extensions being loaded. Since by default extensions are being loaded alphabetically and extensions are called in the order loaded, the name of the RADIUS authentication extension needs to be changed when the docker container is created.

### DIFF
--- a/guacamole-docker/bin/start.sh
+++ b/guacamole-docker/bin/start.sh
@@ -405,7 +405,14 @@ END
        "$RADIUS_EAP_TTLS_INNER_PROTOCOL"
 
     # Add required .jar files to GUACAMOLE_EXT
-    ln -s /opt/guacamole/radius/guacamole-auth-*.jar "$GUACAMOLE_EXT"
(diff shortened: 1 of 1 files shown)


jolentes:
This is an example for RADIUS. I would implement for all other auth extension in the same way if ok.

necouchman:
Is this the way to go, or maybe it would be better to add an "order" variable for each extension that allows you to set that particular extension's order?  Something like:

RADIUS_EXT_ORDER
LDAP_EXT_ORDER
etc.

and just have that prepended to each of the extension names, so that the resulting name is the _EXT_ORDER plus the actual JAR file?

necouchman:
Closing this, as, with changes made to 1.3.0, the order of the extensions is no longer an issue to allowing RADIUS to function properly. This was fixed with #352, GUACAMOLE-684.

PR #359, opened 2019-01-19
Guacamole 617: Extract Permission Management from JDBC Authentication Module

daniquir:
New external library with the common code of the authentication module.
Adaptation of the JDBC extensions to use the new library.
Morphia extension with MongoDB using the new library.


mike-jumper:
Please also make sure you do not include merge commits in the set of commits for a pull request (if changes need to be rebased, please rebase them, but no merges), and update your commit messages to follow the [contribution guidelines](https://github.com/apache/guacamole-client/blob/master/CONTRIBUTING):

>
> ... Commits should be made in logical units and must reference the JIRA issue number:
>
>

The git history and the contents of other PRs should serve as a good example for the above.

necouchman:
> ...and update your commit messages to follow the contribution guidelines...

And the title of this PR should also match the same format...

daniquir:
I modified the project, removing the Morphia extension to upload it later in another revision.
I have also deleted the eclipse files, which I forgot to remove.
And finally I changed the title of the PR.

I'm not sure if now that the pull request is done I can eliminate the commits of the previous merges.

daniquir:
Well, it seems that the cleaning of commits has gone well, please check it in case I have made an error.
Thanks for the info.

necouchman:
Commits look much better, thanks!

extensions/guacamole-auth-common-base/modules/guacamole-auth-common-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ObjectEntityModelInterface.java
```diff
@@ -0,0 +1,24 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.guacamole.auth.jdbc.base;
+
+public interface ObjectEntityModelInterface extends ObjectModelInterface, EntityModelInterface {
```
necouchman: Does this class have a purpose?

extensions/guacamole-auth-common-base/modules/guacamole-auth-common-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionAttributeModelInterface.java
```diff
@@ -0,0 +1,24 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.guacamole.auth.jdbc.connection;
+
+public interface ConnectionAttributeModelInterface {
```
necouchman: An empty interface?

daniquir:
Like you have been able to verify it's a very laborious and extensive work. I take into account everything you mention about refactoring the nomenclature of the project, it's something I have pending, along with reviewing the documentation of the JavaDoc, the format, etc.
Before finishing the development of those things I wanted them to be reviewed by you in case I still had to make a big change and I did not have to revise the whole project for these things that you mentioned.

In any case, I will be available for what is necessary in order to modify and give support in this adaptation. I keep revising the code and testing the application, since it practically affects everything.

Regarding your doubts, I answer you below.

The reason why there are interfaces and empty classes is being able to share the code in a generic way with the extension of Morphia/MongoDB, since the data structure is not exactly the same because the relationships between entities are done by objects.

I would like you to clarify a bit this phrase:

"In multiple places you've broken the import blocks into multiple sections, which is not consistent with the style in the rest of the code."

From what I understand you mean the import of the classes and interfaces, in the upper part, this may have been done by the auto-import of eclipse, I will take it into account to correct it.

extensions/guacamole-auth-common-base/modules/guacamole-auth-common-base/pom.xml
```diff
@@ -0,0 +1,104 @@
...
+        <groupId>org.apache.guacamole</groupId>
+        <artifactId>guacamole-auth-common</artifactId>
+        <version>1.0.0</version>
+        <relativePath>../../</relativePath>
+    </parent>
+
+    <build>
+        <plugins>
+
+            <!-- Written for 1.6 -->
```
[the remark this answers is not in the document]
daniquir: The version has been updated

necouchman:
> Before finishing the development of those things I wanted them to be reviewed by you in case I still had to make a big change and I did not have to revise the whole project for these things that you mentioned.

I would advise you to go ahead with these changes.  There are going to be several revisions, I'm sure, but it's hard to review the code partially.

> From what I understand you mean the import of the classes and interfaces, in the upper part, this may have been done by the auto-import of eclipse, I will take it into account to correct it.

Yeah, that makes sense if the IDE is trying to split it up.  I'm guessing it may be doing the same thing to your indentations - there might be some settings in Eclipse that control these behaviors.  I use Netbeans, and some of it is configurable, and some of it it tries to guess from the existing code.  Doesn't always do exactly what I want it to do, but mostly it works.

extensions/guacamole-auth-common-base/modules/guacamole-auth-common-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ObjectEntityModelInterface.java
necouchman: Hmmm...okay, I think I understand.  However, if something is specific to the Morphia extension, I believe those pieces would belong in the code for that extension and not in the common parts.

extensions/guacamole-auth-common-base/modules/guacamole-auth-common-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionAttributeModelInterface.java
necouchman: Okay, it just seems strange to have an interface that doesn't define anything at all.  However, as with the other empty interfaces that extend a couple of other interfaces, I would think this should go in the Morphia-specific code, not in the common extension.

daniquir: Classes that are only for morphia will be moved to the morphia project.

mike-jumper:
> The reason why there are interfaces and empty classes is being able to share the code in a generic way with the extension of Morphia/MongoDB, since the data structure is not exactly the same because the relationships between entities are done by objects.

Empty classes/interfaces are definitely an antipattern. If something about the design of the MongoDB extension is requiring this, that suggests something needs to be addressed in the overall design. It shouldn't be necessary, and the needs of the specific Morphia/MongoDB implementation shouldn't be the guiding factor of the generic extraction of the a permission implementation.

The presence of models in the generic part of this is also a sign that things should be rethought. The models underlying these objects (or if there even are models) should be a concern of the implementation leveraging the generic permission / connection tracking library, not the library itself.

daniquir:
> Empty classes/interfaces are definitely an antipattern. If something about the design of the MongoDB extension is requiring this, that suggests something needs to be addressed in the overall design. It shouldn't be necessary, and the needs of the specific Morphia/MongoDB implementation shouldn't be the guiding factor of the generic extraction of the permission implementation.


You are absolutely right, the use of interfaces marker is not the best solution. No problem, I will study the case to get the best solution to this problem.

> The presence of models in the generic part of this is also a sign that things should be rethought. The models underlying these objects (or if there even are models) should be a concern of the implementation leveraging the generic permission / connection tracking library, not the library itself.

There are no models in the generic part, but there are parts of the code in which the models are used, for that reason there are interfaces that refer to the model located in the specific part of JDBC or Morphia. If this is not expected, I will also revise and modify it if necessary.

daniquir:
I have reviewed all the changes you have proposed.

Refactoring of packages and classes in the common project that referred to JDBC.
Remove the marker interfaces and code revision to not use those interfaces. To make this change, which greatly affected the structure of the code created, I used MapBinder, it is the first time I use Guice, I hope this option is correct.

I will continue with the other changes that do not affect the structure of the code.

daniquir:
I have completed the common project JavaDoc and I have removed three marker interfaces that I had not seen.

daniquir:
I finished with the formatting of the code and restored the blocks of imports that were generated by the IDE Eclipse. The best solution for this has been to create a formatter for Eclipse with the style guidelines, I think this may be useful for other people and I would like to be able to share the formatter so they can use it.

For now, with this I have finished reviewing everything you have mentioned before. I await your feedback to continue reviewing the code.

PR #360, opened 2019-01-22
GUACAMOLE-524: Update connect() API changes for backward compatibility with 1.0.0.

mike-jumper:
[As discussed on dev@](https://lists.apache.org/thread.html/c74221a629bd2da53996b0f6e4cfebc4647ac1239aa8817a45fd0c12@%3Cdev.guacamole.apache.org%3E), these changes modify the `Connectable` interface such that implementations of the previous version of the `connect()` function will continue to work, even if those implementations extended `SimpleConnection` and overrode the old, now-deprecated function. This is achieved through:

1. Continuing to use the old `connect()` as the basis for the implementation.
2. Using a `ThreadLocal` to make the additional `Map<String, String>` parameter available to the old `connect()` within a strict context.

The `SimpleConnection` class and related classes are also modified to provide the same behavior as past releases, particularly for downstream subclasses.

With these changes, things are ABI-compatible with 1.0.0. Things are also API-compatible with 1.0.0, though continuing to use/override the old `connect()` will result in compiler warnings. Behavior of the relevant classes should now be as follows:

| Class | Behavior in 1.0.0 and older | New behavior |
| ----- | --------------------------- | ------------ |
| `SimpleAuthenticationProvider` | Tokens are applied and "baked-in" when the user authenticates. | Tokens are applied dynamically when connecting. |
| `SimpleUserContext` | Tokens are applied only if explicitly implemented. | Tokens are applied only if explicitly implemented *or* if requested via the constructor. |
| `SimpleConnection` | Tokens are applied only if explicitly implemented. | Tokens are applied only if explicitly implemented *or* if requested via the constructor. |

Overall, behavior should be identical to 1.0.0 with the following exceptions:

* Subclasses of `SimpleAuthenticationProvider` will now handle additional tokens and apply those tokens in context of the connection, not at time of authentication.
* The result of calling `getConfiguration()` on a connection accessed via a subclass of `SimpleAuthenticationProvider` will now contain any tokens in uninterpreted form.

### DIFF
--- a/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Connectable.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Connectable.java
@@ -19,7 +19,6 @@
 
 package org.apache.guacamole.net.auth;
 
-import java.util.Collections;
 import java.util.Map;
 import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.net.GuacamoleTunnel;
@@ -30,36 +29,14 @@ import org.apache.guacamole.protocol.GuacamoleClientInformation;
  */
 public interface Connectable {
 
-    /**
-     * Establishes a connection to guacd using the information associated with
-     * this object. The connection will be provided the given client
-     * information.
-     *
-     * @deprecated
-     *     This function has been deprecated in favor of
-     *     {@link #connect(org.apache.guacamole.protocol.GuacamoleClientInformation, java.util.Map)},
-     *     which allows for connection parameter tokens to be injected and
-     *     applied by cooperating extensions, replacing the functionality
-     *     previously provided through the {@link org.apache.guacamole.token.StandardTokens}
-     *     class. It continues to be defined on this interface for
-     *     compatibility. <strong>New implementations should instead implement
-     *     {@link #connect(org.apache.guacamole.protocol.GuacamoleClientInformation, java.util.Map)}.</strong>
-     *
-     * @param info
-     *     Information associated with the connecting client.
-     *
-     * @return
-     *     A fully-established GuacamoleTunnel.
-     *
-     * @throws GuacamoleException
-     *     If an error occurs while connecting to guacd, or if permission to
-     *     connect is denied.
+    /*
+     * IMPORTANT:
+     * ----------
+     * The web application (guacamole) defines its own version of this
+     * interface containing defaults which allow backwards compatibility with
+     * 1.0.0. Any changes to this interface MUST be properly reflected in that
+     * copy of the interface such that they are binary compatible.
      */
-    @Deprecated
-    default GuacamoleTunnel connect(GuacamoleClientInformation info)
-            throws GuacamoleException {
-        return this.connect(info, Collections.emptyMap());
-    }
 
     /**
      * Establishes a connection to guacd using the information associated with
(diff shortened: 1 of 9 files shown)


guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Connectable.java
```diff
@@ -54,8 +97,13 @@
      *     If an error occurs while connecting to guacd, or if permission to
      *     connect is denied.
      */
-    public GuacamoleTunnel connect(GuacamoleClientInformation info,
-            Map<String, String> tokens) throws GuacamoleException;
+    default GuacamoleTunnel connect(GuacamoleClientInformation info,
+            Map<String, String> tokens) throws GuacamoleException {
+
+        // Allow old implementations of Connectable to continue to work
+        return this.connect(info);
```
necouchman: Is there a reason supplying a default of this particular instance of this method is necessary or valuable?  Won't the old implementations already be calling the other default method, without the `tokens` parameter?  And doesn't this introduce the possibility that an implementing class could end up recursively calling these parameters (if they don't actually implement either of them)?  Or will that get caught at compile time?

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleUserContext.java
```diff
@@ -99,7 +129,7 @@ public SimpleUserContext(AuthenticationProvider authProvider,
             GuacamoleConfiguration config = configEntry.getValue();
 
             // Add as simple connection
-            Connection connection = new SimpleConnection(identifier, identifier, config);
+            Connection connection = new SimpleConnection(identifier, identifier, config, true);
```
necouchman: Should this be `true`, or should this be passing the value of `interpretTokens`?

mike-jumper: Good catch - we should pass through `interpretTokens`.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Connectable.java
mike-jumper: > Is there a reason supplying a default of this particular instance of this method is necessary or valuable? Won't the old implementations already be calling the other default method, without the `tokens` parameter?

Some sort of default is necessary to bridge the implementations, as it's the webapp that ultimately calls `connect()`, and in this case will be calling the version which has `tokens`.

In the general case, defaults are needed to bridge in both directions to allow extensions of different versions to interoperate. An older extension (compatible with 1.0.0) which decorates the connections of other extensions will need to be able to invoke `connect()`, even though it will be calling the older version. A newer extension and the webapp need the same ability but in the other direction.

> And doesn't this introduce the possibility that an implementing class could end up recursively calling these parameters (if they don't actually implement either of them)?

Yes, and that's definitely a bad thing. I think I may have a route around this.

mike-jumper: OK - I'm rearranging things a bit such that the interface defined within guacamole-ext is still strictly as defined before: one `connect()` definition which accepts `tokens`.

Within the webapp, I'm adding another version of `Connectable` in the same package which defines both `connect()` versions as well as defaults, and thus an ABI-compatible redefinition of both versions of the interface. Because of the way servlet containers handle classes bundled with the webapp vs. within dependencies, the webapp version of `Connectable` takes priority.

mike-jumper:
Reading through things again, I think we might need to also update `SimpleConnectionGroup`, `DelegatingConnection`, `DelegatingConnectionGroup`, etc.

mike-jumper:
OK - I've gone through and made sure that any guacamole-ext class which provides a concrete implementation of `connect()` does so using the old `connect()` as a basis for the sake of compatibility (same approach as `SimpleConnection`), with the exception of the new token injecting classes that did not exist before.

PR #362, opened 2019-01-22
GUACAMOLE-712: Add Danish to RDP

netromnetrom:
Add Danish


necouchman:
@netromnetrom Thanks for the contribution!  If you could create a single pull request with all of the changes, that would be great.

PR #364, opened 2019-01-22
GUACAMOLE-704: LDAP Follow Referrals setting for Docker containers

JoelB:
This will add the ldap-follow-referrals config setting to the start.sh file for Docker. I've found that setting this to 'false' is necessary for Guacamole to behave correctly in my Active Directory environment.


JoelB:
Closing so I can fix the commit message

PR #365, opened 2019-01-22
GUACAMOLE-704: Add ldap-follow-referrals setting for Docker containers

JoelB:
JIRA Issue: https://issues.apache.org/jira/browse/GUACAMOLE-704

This will add the ldap-follow-referrals config setting to the start.sh file for Docker if the LDAP_FOLLOW_REFERRALS environment variable is set. I've found that setting this to 'false' is necessary for Guacamole to behave correctly in my Active Directory environment.

### DIFF
--- a/guacamole-docker/bin/start.sh
+++ b/guacamole-docker/bin/start.sh
@@ -321,7 +321,7 @@ END
         "ldap-user-search-filter" \
         "$LDAP_USER_SEARCH_FILTER"
 
-    set_optional_property         \
+    set_optional_property       \
         "ldap-follow-referrals" \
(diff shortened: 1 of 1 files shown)


guacamole-docker/bin/start.sh
```diff
@@ -321,6 +321,10 @@ END
         "ldap-user-search-filter" \
         "$LDAP_USER_SEARCH_FILTER"
 
+    set_optional_property         \
+        "ldap-follow-referrals" \
```
necouchman: Very minor nitpick, but please line up the trailing back-slash, for consistency with the rest of the code...

PR #368, opened 2019-01-24
GUACAMOLE-697: Add DISTINCT to selectOne permissions queries.

necouchman:
This should resolve the bug identified in [GUACAMOLE-697](https://issues.apache.org/jira/browse/GUACAMOLE-697) where assigning permissions for a connection to both a user and group results in an error in the web application.  Tested against a PostgreSQL database and it appears to resolve the issue.


mike-jumper:
Looks perfectly decent to me.

I'll quickly run the WIP REST unit tests on https://github.com/mike-jumper/guacamole-client/tree/rest-test to verify things against more databases, but if all is well I think we'll be good to merge.

PR #369, opened 2019-01-24
GUACAMOLE-716: Add all LDAP properties to Docker start script

JoelB:
This will add all possible LDAP guacamole.properties settings to the start.sh script for the Docker container. The five that were missing are:

1. ldap-group-name-attribute
2. ldap-dereference-aliases
3. ldap-max-referral-hops
4. ldap-operation-timeout
5. ldap-max-search-results

I also re-formatted some set_optional_property lines that were < 80 characters to be a single line and re-ordered them single-line entries so they match the order in the documentation.

### DIFF
--- a/guacamole-docker/bin/start.sh
+++ b/guacamole-docker/bin/start.sh
@@ -303,27 +303,31 @@ END
     set_property          "ldap-hostname"           "$LDAP_HOSTNAME"
     set_optional_property "ldap-port"               "$LDAP_PORT"
     set_optional_property "ldap-encryption-method"  "$LDAP_ENCRYPTION_METHOD"
-    set_property          "ldap-user-base-dn"       "$LDAP_USER_BASE_DN"
-    set_optional_property "ldap-username-attribute" "$LDAP_USERNAME_ATTRIBUTE"
-    set_optional_property "ldap-member-attribute"   "$LDAP_MEMBER_ATTRIBUTE"
-    set_optional_property "ldap-group-base-dn"      "$LDAP_GROUP_BASE_DN"
-    set_optional_property "ldap-config-base-dn"     "$LDAP_CONFIG_BASE_DN"
-
-    set_optional_property     \
-        "ldap-search-bind-dn" \
-        "$LDAP_SEARCH_BIND_DN"
+    set_optional_property "ldap-max-search-results" "$LDAP_MAX_SEARCH_RESULTS"
+    set_optional_property "ldap-search-bind-dn"     "$LDAP_SEARCH_BIND_DN"
 
     set_optional_property           \
         "ldap-search-bind-password" \
         "$LDAP_SEARCH_BIND_PASSWORD"
 
-    set_optional_property         \
-        "ldap-user-search-filter" \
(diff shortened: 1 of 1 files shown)


guacamole-docker/bin/start.sh
```diff
@@ -303,27 +303,29 @@ END
...
     set_property          "ldap-user-base-dn"       "$LDAP_USER_BASE_DN"
     set_optional_property "ldap-username-attribute" "$LDAP_USERNAME_ATTRIBUTE"
-    set_optional_property "ldap-member-attribute"   "$LDAP_MEMBER_ATTRIBUTE"
-    set_optional_property "ldap-group-base-dn"      "$LDAP_GROUP_BASE_DN"
+    set_optional_property "ldap-user-search-filter" "$LDAP_USER_SEARCH_FILTER"
     set_optional_property "ldap-config-base-dn"     "$LDAP_CONFIG_BASE_DN"
-
-    set_optional_property     \
-        "ldap-search-bind-dn" \
-        "$LDAP_SEARCH_BIND_DN"
```
necouchman: I like the reformatting overall, but I'm concerned about these items that reorganize properties away from each other - for example, the `ldap-search-bind-dn` property ends up several lines away from `ldap-search-bind-password`.  While this isn't something that users will every see or care about, from a debugging and code maintenance point of view I think it could organized so that it makes a little more sense.

mike-jumper:
Please also rebase such that the only commits present are those from your changes. There shouldn't be any merge commits in the content of pull requests. There will be one after the PR is merged, of course, but there shouldn't be any in the PR itself.

guacamole-docker/bin/start.sh
JoelB: Hi Nick, I actually preferred that when making the change but opted to keep it the same as the previous formatting where the shorter lines were grouped and then the longer lines grouped after.

Just to be clear, would you prefer to have them interspersed like this?

    set_optional_property "ldap-search-bind-dn"     "$LDAP_SEARCH_BIND_DN"

    set_optional_property           \
        "ldap-search-bind-password" \
        "$LDAP_SEARCH_BIND_PASSWORD"

    set_property          "ldap-user-base-dn"       "$LDAP_USER_BASE_DN"
    set_optional_property "ldap-username-attribute" "$LDAP_USERNAME_ATTRIBUTE"


Let me know and I'll amend the PR.

necouchman: IMHO, yes, it makes more sense to group the options together that logically go together, rather than grouping them by line length.  If others want to vote me down on that, that's fine, we can discuss it, and I don't feel uber-passionate about it, but it does make more sense to me.

JoelB:
@mike-jumper I'm relatively new to git. Can you recommend the best way to do this without submitted a new PR?

guacamole-docker/bin/start.sh
JoelB: I've made the changes and squashed my commits to remove the merge (I think!). Commit 26d3d51 should have the changes you requested.

PR #374, opened 2019-02-07
GUACAMOLE-513: Implement Wake-on-LAN extension

necouchman:
This PR implements a wake-on-lan extension that decorates connections from other extensions, providing attributes for storing MAC and Broadcast addresses for connections.


extensions/guacamole-auth-wol/src/main/resources/services/wolService.js
```diff
@@ -0,0 +1,158 @@
...
+        
+            // Send the WOL request
+            return requestService({
+                method  : 'GET',
+                url     : 'api/session/ext/wol/check/' + hostData['hostname'] + '/' + hostData['port'],
+                params  : httpParameters
+            })
+            .then(function gotStatus(statusData) {
+                return statusData;
+            }, function(error) {
```
mike-jumper: For the sake of readable stack traces, some sort of reasonable names should be provided to the callback functions.

extensions/guacamole-auth-wol/src/main/resources/services/wolService.js
```diff
@@ -0,0 +1,158 @@
...
+        var status = hostInfo.then(function(hostData) {
+        
+            // Send the WOL request
+            return requestService({
+                method  : 'GET',
+                url     : 'api/session/ext/wol/check/' + hostData['hostname'] + '/' + hostData['port'],
+                params  : httpParameters
+            })
+            .then(function gotStatus(statusData) {
+                return statusData;
```
mike-jumper: Isn't this what will already happen (ie: effectively a no-op)?

extensions/guacamole-auth-wol/src/main/resources/services/wolService.js
```diff
@@ -0,0 +1,158 @@
...
+            }, function(error) {
+                $log.error(error);
```
mike-jumper: If a warning is sufficient, you may be able to leverage `requestService.WARN`:

https://github.com/apache/guacamole-client/blob/3089e71e60541731a49c8d8a5d69a7425780ce6f/guacamole/src/main/webapp/app/rest/services/requestService.js#L133-L144

pom.xml
```diff
@@ -57,11 +57,11 @@
         <module>extensions/guacamole-auth-openid</module>
         <module>extensions/guacamole-auth-quickconnect</module>
         <module>extensions/guacamole-auth-totp</module>
+        <module>extensions/guacamole-auth-wol</module>
 
         <!-- Example web applications using the Guacamole APIs -->
         <module>doc/guacamole-example</module>
         <module>doc/guacamole-playback-example</module>
-
```
mike-jumper: Poor blank line. It was just minding its own business, being blank.

extensions/guacamole-auth-wol/src/main/java/org/apache/guacamole/auth/wol/connection/WOLConnection.java
```diff
@@ -0,0 +1,210 @@
...
+     *     The Connection object to decorate.
+     * 
+     * @param canUpdate
+     *     Whether or not the user can update this connection.
+     */
+    public WOLConnection(Connection connection, Boolean canUpdate) {
+
+        super(connection);
+        this.undecorated = connection;
+        this.attributes = super.getAttributes();
```
mike-jumper: Caching the attributes of the superclass could be problematic. I would recommend instead calling `getAttributes()` where needed, saving the result of that call only locally to the function that needs it.

extensions/guacamole-auth-wol/src/main/java/org/apache/guacamole/auth/wol/rest/WOLUserContextResource.java
```diff
@@ -0,0 +1,127 @@
...
+    @Path("check/{hostname}/{port}")
+    public Boolean checkHost(@PathParam("hostname") String hostname,
+            @PathParam("port") int port) {
+        
+        try (Socket socket = new Socket()) {
+            socket.connect(new InetSocketAddress(hostname, port), DEFAULT_TIMEOUT);
+            return true;
+        }
+        catch (IOException | IllegalArgumentException | SecurityException e) {
+            return false;
```
mike-jumper: Any need to log that exception? Perhaps at least at the debug level?

extensions/guacamole-auth-wol/src/main/java/org/apache/guacamole/auth/wol/user/WOLUserContext.java
```diff
@@ -0,0 +1,114 @@
...
+    @Override
+    public Directory<Connection> getConnectionDirectory() throws GuacamoleException {
+        return new DecoratingDirectory<Connection>(super.getConnectionDirectory()) {
+
+            @Override
+            protected Connection decorate(Connection object) throws GuacamoleException {
+                Permissions effective = self().getEffectivePermissions();
+                SystemPermissionSet systemPermissions = effective.getSystemPermissions();
+                ObjectPermissionSet objectPermissions = effective.getConnectionPermissions();
+                Boolean canUpdate = false;
```
mike-jumper: Why `Boolean` rather than `boolean`?

extensions/guacamole-auth-wol/src/main/resources/templates/wol.html
```diff
@@ -0,0 +1,11 @@
+<meta name="after" content="a.home-connection" />
+<span class="wol"
+      ng-controller="wolController"
```
mike-jumper: It looks like you're essentially implementing a directive. It would be better to implement such a directive and reference that here, rather than use `ng-controller`.

extensions/guacamole-auth-wol/src/main/java/org/apache/guacamole/auth/wol/connection/WOLConnection.java
```diff
@@ -0,0 +1,210 @@
...
+    public Map<String, String> getAttributes() {
+        
+        // Create a mutable copy of the attributes
+        Map<String, String> effectiveAttributes = new HashMap<>(attributes);
+        
+        // Check to see if any need to be added or removed
+        for (String attr : WOL_ATTRIBUTES) {
+            if (canUpdate && !effectiveAttributes.containsKey(attr))
+                effectiveAttributes.put(attr, null);
+            else if (!canUpdate && effectiveAttributes.containsKey(attr))
```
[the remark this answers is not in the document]
necouchman: Safety, first.

extensions/guacamole-auth-wol/src/main/java/org/apache/guacamole/auth/wol/connection/WOLConnection.java
```diff
@@ -0,0 +1,210 @@
...
+
+    }
+
+    @Override
+    public void setAttributes(Map<String, String> setAttributes) {
+
+        // Create a multable copy of the attributes
+        setAttributes =  new HashMap<>(setAttributes);
+
+        if (!canUpdate)
```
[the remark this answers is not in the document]
necouchman: Okay, I've refactored this a bit - hopefully it's okay, now.

extensions/guacamole-auth-wol/src/main/java/org/apache/guacamole/auth/wol/rest/WOLUserContextResource.java
necouchman: Yeah, that's reasonable.

extensions/guacamole-auth-wol/src/main/java/org/apache/guacamole/auth/wol/user/WOLUserContext.java
necouchman: Because CAPS is better than lowercase.

Not really.

necouchman:
Okay, I've fixed up several things, just need to deal with the last couple of AngularJS issues.

One thing I was thinking, though, for this functionality - I'm happy to keep going this route, but is there any sense in implementing the WoL stuff inside of guacd and providing a way to trigger it that way, instead of an extension?  Since guacd is assumed to be "closer" to the actual remote hosts, and, indeed, may be across a firewall or WAN link, I'm wondering if this makes any more sense?  Thoughts?

mike-jumper:
Were you thinking of adding some sort of base WoL functionality to libguac or some common part of guacamole-server, and leveraging that via additional parameters for each protocol that might benefit?

necouchman:
> Were you thinking of adding some sort of base WoL functionality to libguac or some common part of guacamole-server, and leveraging that via additional parameters for each protocol that might benefit?

Yeah, that was what I was wondering.  I mean, I have no problem continuing on this direction if that's the best way to go, but if it would be better to do it on the libguac side, I'm happy to take another look and refactor.  It seems like it would be doable - just might have to take it out of the 1.2.0 scope and take another look...

mike-jumper:
Sure - it feels like a libguac-based implementation may even be simpler, perhaps even just a single function which implements what you've done with `sendMagicPacket()` here. The various protocols would then just need a pair of additional parameters and poof: WoL support.

necouchman:
Sounds good.  I'll take a stab at doing it that way, then.

vampywiz17:
@necouchman 

Hi there, 

I recognise, that this implementation not working, if we use Guacamole inside docker container.

I think the only way to working well if add the --net=host flag to magicpacket. (Not broadcast. direct IP to host)

Possible to add the option to use this method?

mike-jumper:
@vampywiz17, if you have questions, please post to the user@guacamole.apache.org list rather than commenting on a closed pull request:

Thanks.

PR #381, opened 2019-02-21
GUACAMOLE-696: In JDBC module, merge effective groups

necouchman:
This pull request overrides the `getEffectiveUserGroups()` method for the `ModeledAuthenticatedUser` class such that the user groups are merged between ones passed in from the provider that authenticated the user and the ones present for the same user within the JDBC module.

I don't know if this is the best way to do this, if it fully resolves all of the issues associated with this particular JIRA issue, or if it introduces any adverse behavior, but my testing indicates that it at least clarifies/resolves the issue reported in GUACAMOLE-696.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledAuthenticatedUser.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledAuthenticatedUser.java
@@ -20,6 +20,7 @@
 package org.apache.guacamole.auth.jdbc.user;
 
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import org.apache.guacamole.net.auth.AuthenticatedUser;
@@ -167,6 +168,13 @@ public class ModeledAuthenticatedUser extends RemoteAuthenticatedUser {
     @Override
     public void setIdentifier(String identifier) {
         user.setIdentifier(identifier);
+    }
+    
+    @Override
+    public Set<String> getEffectiveUserGroups() {
+        Set<String> allGroups = new HashSet<>(user.getEffectiveUserGroups());
+        allGroups.addAll(super.getEffectiveUserGroups());
+        return Collections.unmodifiableSet(allGroups);
     }
 
 }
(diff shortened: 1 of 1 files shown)


mike-jumper:
From what I can tell, retrieval of effective user groups is already guaranteed to produce a set of distinct user groups (no duplicates) as it returns a `Set<String>`:

https://github.com/apache/guacamole-client/blob/87aa2e6c34d11bb30626c6c27a42796506493ffd/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/EntityService.java#L76-L96

Is this not the case?

necouchman:
> From what I can tell, retrieval of effective user groups is already guaranteed to produce a set of distinct user groups (no duplicates) as it returns a Set<String>...Is this not the case?

Yes, this is the case; however, the behavior that GUACAMOLE-696 deals with is the following:
- User is defined in JDBC module.
- User is member of groups in JDBC module.
- Groups in JDBC module are assigned permissions to connections, system, etc.
- User with matching username authenticates with LDAP
- Permissions assigned to the JDBC group are not applied, because no matching group is found in LDAP, and the user did not authenticate through the JDBC module.

What this change does (I think?) is to combine the user groups returned for a user, whether they authenticate through the JDBC module or not, by pulling both the groups for the  `AuthenticatedUser (may or may not be JDBC user) and the `ModeledUser` object (from the database, matching the name of the logged in user) so that permissions apply to the user from group membership, regardless of whether the user logged in with the JDBC module.

If there's a different/better way I should do this, I'm open to it - this just seems to work and is pretty minimal in terms of changes to accomplish it.

mike-jumper:
Alrighty. That makes sense to me.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledAuthenticatedUser.java
```diff
@@ -168,5 +169,12 @@ public String getIdentifier() {
     public void setIdentifier(String identifier) {
         user.setIdentifier(identifier);
     }
+    
+    @Override
+    public Set<String> getEffectiveUserGroups() {
+        Set<String> allGroups = new HashSet<>(user.getEffectiveUserGroups());
+        allGroups.addAll(super.getEffectiveUserGroups());
+        return Collections.unmodifiableSet(allGroups);
```
mike-jumper: Since we're using Guava, I suggest using `Sets.union()` to produce a read-only merged view of the two sets, rather than creating a new merged set through copying each element:

https://google.github.io/guava/releases/27.0.1-jre/api/docs/com/google/common/collect/Sets.html#union-java.util.Set-java.util.Set-

Other than that, I'm happy with these changes.

necouchman: Done.

PR #382, opened 2019-02-21
GUACAMOLE-736: Remove JAXB data conversion in favor of Guava

necouchman:
This pull request replaces all of the instances of `DatatypeConverter` from the deprecated `java.xml.bind` package with the equivalent Guava packages from the `BaseEncoding` class(es).

I've tested the following items:
- Log in with JDBC
- Log in with RADIUS, including 2FA
- Start SSH connection

These tests have worked fine.  There are probably a couple of other corner cases that need to be tested - I have no way of testing the Duo module, TOTP should probably be validated, and changing passwords in JDBC should be verified, as well.  But, initial tests look good.


keers:
Seems guava library is used only for base64 encoding/decoding. Why not use Apache Commons Codec for that instead?

necouchman:
@keers: Actually, it's used for several Hex (base-16) operations, as well.  I went with Guava because it's used in other places in the project, so it just keeps it simple and consistent.

extensions/guacamole-auth-duo/src/main/java/org/apache/guacamole/auth/duo/api/SignedDuoCookie.java
```diff
@@ -221,7 +221,7 @@ private static String sign(String key, String data) throws GuacamoleException {
             mac.init(new SecretKeySpec(key.getBytes("UTF-8"), SIGNATURE_ALGORITHM));
 
             // Return signature as hex
-            return DatatypeConverter.printHexBinary(mac.doFinal(data.getBytes("UTF-8"))).toLowerCase();
+            return BaseEncoding.base16().encode(mac.doFinal(data.getBytes("UTF-8"))).toLowerCase();
```
mike-jumper: Guava provides a mechanism for controlling the case of the resulting string prior to encoding:

https://google.github.io/guava/releases/27.0-jre/api/docs/com/google/common/io/BaseEncoding.html#lowerCase--

Fairly minor, but probably worth switching over to `BaseEncoding.base16().lowerCase()`.

necouchman: Okay, swapped this over to lower prior to encoding.

PR #383, opened 2019-02-22
GUACAMOLE-715: Change how null JDBC users are handled

necouchman:
This pull request allows the JDBC module to create empty users when authentication has succeeded but the user doesn't exist in the database.  This allows group membership to be processed for users authenticated with other modules, where the user does not actually exist in the database.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/UserService.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/UserService.java
@@ -423,6 +423,43 @@ public class UserService extends ModeledDirectoryObjectService<ModeledUser, User
         return user;
 
     }
+    
+    /**
+     * Generates an empty (skeleton) user corresponding to the given
+     * AuthenticatedUser.  The user will not be stored in the database, and
+     * will only be available in-memory during the time the session is
+     * active.
+     * 
+     * @param authenticationProvider
+     *     The AuthenticationProvider on behalf of which the user is being
+     *     retrieved.
+     *
+     * @param authenticatedUser
+     *     The AuthenticatedUser to generate the skeleton account for.
+     *
+     * @return
+     *     The empty ModeledUser which corresponds to the given
+     *     AuthenticatedUser.
+     *
+     * @throws GuacamoleException
+     *     If a ModeledUser object for the user corresponding to the given
+     *     AuthenticatedUser cannot be created.
+     */
+    public ModeledUser retrieveSkeletonUser(AuthenticationProvider authenticationProvider,
+            AuthenticatedUser authenticatedUser) throws GuacamoleException {
+        
+        // Set up an empty user model
+        ModeledUser user = getObjectInstance(null,
+                new UserModel(authenticatedUser.getIdentifier()));
+        
+        // Create user object, and configure cyclic reference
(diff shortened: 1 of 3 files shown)


extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/UserService.java
```diff
@@ -411,8 +411,13 @@ public ModeledUser retrieveUser(AuthenticationProvider authenticationProvider,
 
         // Retrieve corresponding user model, if such a user exists
         UserModel userModel = userMapper.selectOne(username);
-        if (userModel == null)
-            return null;
+        
+        // If no user exists, create a temporary one to map
+        // the authenticated user.
+        if (userModel == null) {
+            userModel = new UserModel();
```
mike-jumper: Unfortunately, this will break handling of the `mysql-user-required`, `postgresql-user-required`, etc. properties, which rely on this function returning `null` if no user exists:

http://guacamole.apache.org/doc/gug/jdbc-auth.html#jdbc-auth-restrict

See:

https://github.com/apache/guacamole-client/blob/d1e928bea79ca81c827e9b6adedabc98eefdf701/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/JDBCAuthenticationProviderService.java#L84-L140

This function's documented return behavior will also need to be updated if this change (or something similar to it) are implemented. It is currently documented as returning `null` if no such user exists.

necouchman: Okay, this should be handled, now.  And return tag has been updated.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/UserService.java
```diff
@@ -392,8 +399,12 @@ public ModeledAuthenticatedUser retrieveAuthenticatedUser(AuthenticationProvider
      *     The AuthenticatedUser to retrieve the corresponding ModeledUser of.
      *
      * @return
-     *     The ModeledUser which corresponds to the given AuthenticatedUser, or
-     *     null if no such user exists.
+     *     The ModeledUser which corresponds to the given AuthenticatedUser, if
+     *     the user exists in the database.  If the user does not exist, and
+     *     the module has been configured to require JDBC authentication via
+     *     a "[db]-user-required" parameter, null will be returned.  Otherwise
```
mike-jumper: This now looks OK, but this feels like a separation of concerns issue to me. The `UserService` has up until now not dealt with configuration details from the `JDBCEnvironment` at all, and this function is aimed at actual retrieval of data.

What do you think about a separate function that embodies strictly the semantics of creating a skeleton, in-memory `AuthenticatedUser`? That way `retrieveAuthenticatedUser()` can truly be about retrieval and can remain independent of configuration concerns?

necouchman: Yeah, you're right.  I've taken a stab at redoing this way - creating a new `retrieveSkeletonUser()` method in the `UserService` class, and then doing all of the checks (that already happen, anyway) for configuration items in `JDBCAUthenticationProviderSrevice`.

PR #386, opened 2019-03-01
fixed typo in german translation file

kbudde:


necouchman:
@kbudde: We definitely appreciate your contribution to the project, in particular correcting this translation issue.  A couple of things:
- All contributions and pull requests need to be tied to a JIRA issue.  You'll need to open a JIRA issue for this and then tag both this pull request and your commit with that issue.  See the other pull requests for examples.
- The change you're proposing is actually already covered in #375, which involves some pretty substantial updates to the German translation files across the client.

kbudde:
Thank you for your response. It's always niceto see how fast projects are moving.

PR #389, opened 2019-03-12
GUACAMOLE-708: Enable auto-creation of users in JDBC modules

necouchman:
As noted in the JIRA issue, it seems like it cold be useful to allow the JDBC module to auto-create users that are logged in successfully with other modules.  This PR introduces the option to enable it (and leaves it disabled by default), along with the creation of those users.

A little unsure if this is desirable behavior, and if this is the best way to go about it, but figured I'd give it a shot :smile:.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/JDBCAuthenticationProviderService.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/JDBCAuthenticationProviderService.java
@@ -27,6 +27,7 @@ import org.apache.guacamole.auth.jdbc.sharing.user.SharedAuthenticatedUser;
 import org.apache.guacamole.auth.jdbc.user.ModeledAuthenticatedUser;
 import org.apache.guacamole.auth.jdbc.user.ModeledUser;
 import org.apache.guacamole.auth.jdbc.user.ModeledUserContext;
+import org.apache.guacamole.auth.jdbc.user.PrivilegedModeledAuthenticatedUser;
 import org.apache.guacamole.auth.jdbc.user.UserService;
 import org.apache.guacamole.language.TranslatableGuacamoleClientException;
 import org.apache.guacamole.net.auth.AuthenticatedUser;
@@ -98,7 +99,7 @@ public class JDBCAuthenticationProviderService implements AuthenticationProvider
         ModeledUser user = userService.retrieveUser(authenticationProvider, authenticatedUser);
         ModeledUserContext context = userContextProvider.get();
         if (user != null && !user.isDisabled()) {
-
+            
             // Enforce applicable account restrictions
             if (databaseRestrictionsApplicable) {
 
@@ -126,9 +127,15 @@ public class JDBCAuthenticationProviderService implements AuthenticationProvider
         }
         
         // If no user account is found, and database-specific account
-        // restrictions do not apply, get an empty user.
+        // restrictions do not apply, get a skeleton user.
         else if (!databaseRestrictionsApplicable) {
             user = userService.retrieveSkeletonUser(authenticationProvider, authenticatedUser);
+            
+            // If auto account creation is enabled, add user to DB.
+            if(environment.autoCreateAbsentAccounts()) {
+                userService.createObject(new PrivilegedModeledAuthenticatedUser(user.getCurrentUser()), user);
+            }
+            
         }
 
         // Veto authentication result only if database-specific account
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/JDBCAuthenticationProviderService.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/JDBCAuthenticationProviderService.java
@@ -132,7 +132,7 @@ public class JDBCAuthenticationProviderService implements AuthenticationProvider
             user = userService.retrieveSkeletonUser(authenticationProvider, authenticatedUser);
             
             // If auto account creation is enabled, add user to DB.
-            if(environment.autoCreateAbsentAccounts()) {
+            if (environment.autoCreateAbsentAccounts()) {
                 userService.createObject(new PrivilegedModeledAuthenticatedUser(user.getCurrentUser()), user);
             }
             
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
@@ -410,9 +410,9 @@ public abstract class ModeledDirectoryObjectService<InternalType extends Modeled
     }
 
     /**
-     * Returns a collection of permissions that should be granted due to the
-     * creation of the given object. These permissions need not be granted
-     * solely to the user creating the object.
+     * Returns an immutable collection of permissions that should be granted due
+     * to the creation of the given object.  These permissions need not be
+     * granted solely to the user creating the object.
      * 
      * @param user
      *     The user creating the object.
@@ -427,16 +427,22 @@ public abstract class ModeledDirectoryObjectService<InternalType extends Modeled
     protected Collection<ObjectPermissionModel> getImplicitPermissions(ModeledAuthenticatedUser user,
             ModelType model) {
         
+        // Get the user model and check for an entity ID.
+        UserModel userModel = user.getUser().getModel();
+        Integer entityId = userModel.getEntityID();
+        if (entityId == null)
+            return Collections.emptyList();
+        
         // Build list of implicit permissions
         Collection<ObjectPermissionModel> implicitPermissions =
-                new ArrayList<ObjectPermissionModel>(IMPLICIT_OBJECT_PERMISSIONS.length);
+                new ArrayList<>(IMPLICIT_OBJECT_PERMISSIONS.length);
 
-        UserModel userModel = user.getUser().getModel();
+        
         for (ObjectPermission.Type permission : IMPLICIT_OBJECT_PERMISSIONS) {
 
             // Create model which grants this permission to the current user
             ObjectPermissionModel permissionModel = new ObjectPermissionModel();
-            permissionModel.setEntityID(userModel.getEntityID());
+            permissionModel.setEntityID(entityId);
             permissionModel.setType(permission);
             permissionModel.setObjectIdentifier(model.getIdentifier());
 
@@ -445,7 +451,7 @@ public abstract class ModeledDirectoryObjectService<InternalType extends Modeled
 
         }
         
-        return implicitPermissions;
+        return Collections.unmodifiableCollection(implicitPermissions);
 
     }
 
@@ -464,7 +470,9 @@ public abstract class ModeledDirectoryObjectService<InternalType extends Modeled
         object.setIdentifier(model.getIdentifier());
 
         // Add implicit permissions
-        getPermissionMapper().insert(getImplicitPermissions(user, model));
+        Collection<ObjectPermissionModel> implicitPermissions = getImplicitPermissions(user, model);
+        if (implicitPermissions != null && !implicitPermissions.isEmpty())
+            getPermissionMapper().insert(implicitPermissions);
 
         // Add any arbitrary attributes
         if (model.hasArbitraryAttributes())
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
@@ -427,11 +427,13 @@ public abstract class ModeledDirectoryObjectService<InternalType extends Modeled
     protected Collection<ObjectPermissionModel> getImplicitPermissions(ModeledAuthenticatedUser user,
             ModelType model) {
         
+        // Check to see if the user granting permissions is a skeleton user,
+        // thus lacking database backing.
+        if (user.getUser().isSkeleton())
+            return Collections.emptyList();
+
         // Get the user model and check for an entity ID.
         UserModel userModel = user.getUser().getModel();
-        Integer entityId = userModel.getEntityID();
-        if (entityId == null)
-            return Collections.emptyList();
         
         // Build list of implicit permissions
         Collection<ObjectPermissionModel> implicitPermissions =
@@ -442,7 +444,7 @@ public abstract class ModeledDirectoryObjectService<InternalType extends Modeled
 
             // Create model which grants this permission to the current user
             ObjectPermissionModel permissionModel = new ObjectPermissionModel();
-            permissionModel.setEntityID(entityId);
+            permissionModel.setEntityID(userModel.getEntityID());
             permissionModel.setType(permission);
             permissionModel.setObjectIdentifier(model.getIdentifier());
 
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
@@ -471,7 +471,7 @@ public abstract class ModeledDirectoryObjectService<InternalType extends Modeled
 
         // Add implicit permissions
         Collection<ObjectPermissionModel> implicitPermissions = getImplicitPermissions(user, model);
-        if (implicitPermissions != null && !implicitPermissions.isEmpty())
+        if (!implicitPermissions.isEmpty())
             getPermissionMapper().insert(implicitPermissions);
 
         // Add any arbitrary attributes
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/UserService.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/UserService.java
@@ -52,6 +52,7 @@ import org.apache.guacamole.net.auth.ActivityRecord;
 import org.apache.guacamole.net.auth.AuthenticatedUser;
 import org.apache.guacamole.net.auth.AuthenticationProvider;
 import org.apache.guacamole.net.auth.User;
+import org.apache.guacamole.net.auth.UserContext;
 import org.apache.guacamole.net.auth.credentials.CredentialsInfo;
 import org.apache.guacamole.net.auth.permission.ObjectPermission;
 import org.apache.guacamole.net.auth.permission.ObjectPermissionSet;
@@ -407,11 +408,8 @@ public class UserService extends ModeledDirectoryObjectService<ModeledUser, User
         if (authenticatedUser instanceof ModeledAuthenticatedUser)
             return ((ModeledAuthenticatedUser) authenticatedUser).getUser();
 
-        // Get username
-        String username = authenticatedUser.getIdentifier();
-
         // Retrieve corresponding user model, if such a user exists
-        UserModel userModel = userMapper.selectOne(username);
+        UserModel userModel = userMapper.selectOne(authenticatedUser.getIdentifier());
         if (userModel == null)
             return null;
 
@@ -448,6 +446,8 @@ public class UserService extends ModeledDirectoryObjectService<ModeledUser, User
      */
     public ModeledUser retrieveSkeletonUser(AuthenticationProvider authenticationProvider,
             AuthenticatedUser authenticatedUser) throws GuacamoleException {
+        
+        logger.info(">>>JDBC<<< Creating skeleton user {}", authenticatedUser.getIdentifier());
         
         // Set up an empty user model
         ModeledUser user = getObjectInstance(null,
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/UserService.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/UserService.java
@@ -52,7 +52,6 @@ import org.apache.guacamole.net.auth.ActivityRecord;
 import org.apache.guacamole.net.auth.AuthenticatedUser;
 import org.apache.guacamole.net.auth.AuthenticationProvider;
 import org.apache.guacamole.net.auth.User;
-import org.apache.guacamole.net.auth.UserContext;
 import org.apache.guacamole.net.auth.credentials.CredentialsInfo;
 import org.apache.guacamole.net.auth.permission.ObjectPermission;
 import org.apache.guacamole.net.auth.permission.ObjectPermissionSet;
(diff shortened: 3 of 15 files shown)


necouchman:
Okay, this has been updated to behave more like and integrate better with recent changes introduced by GUACAMOLE-715 for handling user accounts that do not exist in the database (#383).  Should be a little more streamlined at this point.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/UserService.java
```diff
@@ -460,6 +460,47 @@ public ModeledUser retrieveSkeletonUser(AuthenticationProvider authenticationPro
...
+            
+        // Auto-generate a password
+        user.setPassword(null);
+            
+        // Set up cyclic reference
+        user.setCurrentUser(new ModeledAuthenticatedUser(authenticatedUser,
+            authenticationProvider, user));
+            
+        // Insert the user object
+        userMapper.insert(user.getModel());
```
mike-jumper: The creation process will need to take into account the default set of permissions intended to be granted to each user upon creation:

https://github.com/apache/guacamole-client/blob/68c542787c092aae22decc40307b1927119f6e4b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/UserService.java#L73-L79

I'm a little worried about duplicating creation logic here. Is there a way to leverage `createObject()`?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerGuacamoleProperties.java
```diff
@@ -181,5 +181,13 @@ private SQLServerGuacamoleProperties() {}
         public String getName() { return "sqlserver-driver"; }
 
     };
+    
+    public static final BooleanGuacamoleProperty SQLSERVER_AUTO_CREATE_ACCOUNTS =
```
mike-jumper: Missing JavaDoc.

mike-jumper:
> ...
> A little unsure if this is desirable behavior, and if this is the best way to go about it, but figured I'd give it a shot 😄.

Definitely desirable, I would say. Without this, enabling TOTP across the board for users backed by LDAP is cumbersome.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/UserService.java
```diff
@@ -401,25 +401,25 @@ public ModeledAuthenticatedUser retrieveAuthenticatedUser(AuthenticationProvider
      */
     public ModeledUser retrieveUser(AuthenticationProvider authenticationProvider,
             AuthenticatedUser authenticatedUser) throws GuacamoleException {
+        
+        if (authenticatedUser == null)
```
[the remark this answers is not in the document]
necouchman: Don't remember...will yank that out.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/SQLServerGuacamoleProperties.java
necouchman: Oops.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/UserService.java
necouchman: @mike-jumper: Ping - suggestions on the best way to go for this?

mike-jumper: Tweaking `createObject()` (and pretty much everything else) such that it has defined behavior for some unrestricted internal system user makes sense. It would also ease the implementation of some future addition to the extension API which allows third-party extensions to perform actions without requiring the logged-in user to have sufficient privileges (enrolling in TOTP despite lacking `UPDATE` permission on your own user object, for example).

Defining behavior for `null` would be simple but dangerous. It would be all to easy for a bug to result in an unexpected `null` and then bam: privilege escalation.

I like the idea of an internal system user definition. That would require very explicit usage which would not be error prone. I can think of two possible approaches:

1. Creating some wrapper class which *might* contain a `ModeledAuthenticatedUser` (but isn't guaranteed to) and which has some well-defined way to declare that an instance represents a privileged system user that lacks a model.
2. Creating some base interface which both `RemoteAuthenticatedUser` and `ModeledAuthenticatedUser` implement and which abstracts away permission retrieval. Existing usages of those classes can then be replaced with the interface, and a concrete, privileged implementation of that interface can be created which simply returns true for all permission checks.

mike-jumper: A third option:

Check whether it's possible to refactor things which accept `ModeledAuthenticatedUser` to instead accept `User`. Callers would then either pass `modeledAuthenticatedUser.getUser()` or the previously-mentioned concrete implementation of the interface (in this case `User`) which simply returns true for all permission checks.

necouchman: I’m getting back to working on this, hopefully have something figured out in a couple of days.  Any of the options are going to take some significant refactoring, so just trying to flesh out which makes the most sense.

necouchman: Okay, so, been taking a look at the options and trying some things out - I'm determined to figure this out...
1. The first option - a wrapper class - seems like the least desirable to me, so I've been mostly ignoring this route.  Still a possibility, but the other two seem a little more elegant.
2. The second option - an interface for permission gathering that can be implemented by `RemoteAuthenticatedUser` and `ModeledAuthenticatedUser` - seems like the best way to do this.  **Question: Is there any reason the interface to be implemented shouldn't be the `Permissions` interface, since it already contains the permissions-gathering methods?**
3. I took a look at the third option - replacing `ModeledAuthenticatedUser` with `User`, and, while it probably would be feasible in the end, it isn't quite as simple as just swapping them out - particularly since `ModeledAuthenticatedUser.getUser()` passes a `ModeledUser`, not a plain `User`, which has some extra methods like `isAdministrator()` that have to be factored in.  So, either the `User` class (interface?) would have to have at least the `isAdministrator()` method defined (stubbed), or we'd have to do `ModeledUser` instead of `User`, but then this System user couldn't be defined in the `guacamole-ext` package, it would have to be in the `guacamole-jdbc-base` package, which may not be the best place for it.  Hopefully that makes sense.
4. I guess this brings up a fourth option that is a combination of two and three - either add a handful of the permissions methods to the `User` interface, and replace `ModeledAuthenticatedUser` with `User` such that swapping out `ModeledAuthenticatedUser` with `User` would actually work.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/sharing/user/SharedAuthenticatedUser.java
```diff
@@ -99,7 +102,57 @@ public String getIdentifier() {
...
+        throw new UnsupportedOperationException(
+                "Users authenticated via share keys are immutable.");
+    }
+
+    @Override
+    public ObjectPermissionSet getActiveConnectionPermissions()
+            throws GuacamoleException {
+        throw new UnsupportedOperationException(
+                "Shared users do not have any assigned permissions.");
+    }
```
mike-jumper: The `ObjectPermissionSet` representation of no permissions is an implementation of `ObjectPermissionSet` which returns no permissions, such as [`ObjectPermissionSet.EMPTY_SET`](http://guacamole.apache.org/doc/guacamole-ext/org/apache/guacamole/net/auth/permission/ObjectPermissionSet.html#EMPTY_SET).

Retrieving user permissions is a common operation and needs to succeed. Throwing `UnsupportedOperationException` doesn't indicate that the user has no permissions, but rather that permission retrieval is not supported, and will cause an internal error whenever any part of the webapp attempts to retrieve the permissions of this user.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/UserService.java
mike-jumper: I'll have to reread and reacquaint myself with these changes.

I don't think it will make sense to expose permission sets from `AuthenticatedUser`, nor that there should need to be any fundamental changes to guacamole-ext. It should be possible for the internals of the JDBC auth to have their own internal representation of a system-level user without requiring that permissions be re-anchored to `AuthenticatedUser`.

There are semantic issues with doing so:

1. The `AuthenticatedUser` interface is a pure means of representing _identity_. Mixing that with identity (authentication) with permissions (authorization) collapses an established separation of concerns.
2. There is no definition of cross-extension identity except for `AuthenticatedUser`. As such, the identifiers used by `ObjectPermissionSet` will not have a well-defined meaning. Given a particular `AuthenticatedUser`, I can retrieve the corresponding `User` for each other extension and discover what extension-specific permissions have been granted. If those permissions are moved to `AuthenticatedUser`, this becomes impossible, and the meaning of the identifiers becomes ambiguous.

mike-jumper: Looking over creation of a user and the functionality of `createObject()`, the main beast appears to be the automatic permission creation. When a user `X` attempts to create a user `Y`, the following permissions are automatically added:

This user... | ...can do this... | ...to this object
------------ | ----------------- | -----------------
`X` | `READ` | User `Y`
`X` | `UPDATE` | User `Y`
`X` | `DELETE` | User `Y`
`X` | `ADMINISTER` | User `Y`
`Y` | `READ` | User `Y`

This will be problematic where `X` is the special system user, as then `X` does not exist and we really only want to grant the following permissions:

This user... | ...can do this... | ...to this object
------------ | ----------------- | -----------------
`Y` | `READ` | User `Y`

Perhaps we really can't avoid having a special-case `createMissingUser()`-type function specific to `UserService`...

necouchman: Okay - I'll refocus it on solely the JDBC extension for now.

I was thinking that it actually might dovetail nicely with GUACAMOLE-617, which deals with extracting the permissions management portions of the JDBC module out into a common place usable by multiple modules, but at this point it probably just makes sense to go ahead and stick in the JDBC module and work about those changes later.  Let me see what I can figure out.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/sharing/user/SharedAuthenticatedUser.java
necouchman: Yeah, this was one of my many attempts to figure out some elegant or sexy way to do this.  I'll refactor it, here, soon.

mike-jumper:
Welp, with the merge of #523, I suppose we should dive back into this one.

necouchman:
> Welp, with the merge of #523, I suppose we should dive back into this one.

Thanks for the assist.  I'll start reworking this from the new changes you made...should be significantly easier this time around...

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/UserService.java
necouchman: Okay, @mike-jumper, jumping back into this.  The changes you made have vastly simplified the code....

....but, I've run aground of another issue.  Your first request, long, long ago, was to try to re-use the existing `userService.createObject()` method, both to avoid duplicating code, and to make sure that implicit permissions are granted.  I'm now hitting an issue with this, because the `PrivilegedModeledAuthenticatedUser` doesn't have an `entity_id`, which means when it goes to run `getImplicitPermissions()`, it sets the `entity_id` to `null`, which then causes a database constraint violation trying to insert into `guacamole_user_permission` with a `null` `entity_id` value.

Suggestions?  Refactor `createObject()` to handle these cases?  Or scrap that and just manually do the `entityMapper.insert()` and `userMapper.insert()` calls, and then update permissions after those?

necouchman: Okay, with a couple of minor tweaks to `getImplicitPermissions()`, I think I've got this working.

necouchman:
Alrighty, then - @mike-jumper, I think this is ready for review, again.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/JDBCAuthenticationProviderService.java
```diff
@@ -126,9 +127,15 @@ public ModeledUserContext getUserContext(AuthenticationProvider authenticationPr
         }
         
         // If no user account is found, and database-specific account
-        // restrictions do not apply, get an empty user.
+        // restrictions do not apply, get a skeleton user.
         else if (!databaseRestrictionsApplicable) {
             user = userService.retrieveSkeletonUser(authenticationProvider, authenticatedUser);
+            
+            // If auto account creation is enabled, add user to DB.
+            if(environment.autoCreateAbsentAccounts()) {
```
mike-jumper, line 135: Your `if` nemesis returns. ;)

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
```diff
@@ -429,14 +429,19 @@ public InternalType retrieveObject(ModeledAuthenticatedUser user,
         
         // Build list of implicit permissions
         Collection<ObjectPermissionModel> implicitPermissions =
-                new ArrayList<ObjectPermissionModel>(IMPLICIT_OBJECT_PERMISSIONS.length);
+                new ArrayList<>();
```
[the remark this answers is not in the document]
necouchman: The way I implemented it, the length gets checked in `createObject()` to see if there's any reason at all to try to create the implicit permissions.  I was unsure what the behavior of that would be if an initial length is actually set?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
```diff
@@ -429,14 +429,19 @@ public InternalType retrieveObject(ModeledAuthenticatedUser user,
...
 
         UserModel userModel = user.getUser().getModel();
         for (ObjectPermission.Type permission : IMPLICIT_OBJECT_PERMISSIONS) {
 
+            Integer entityId = userModel.getEntityID();
+            
+            if (entityId == null)
+                continue;
```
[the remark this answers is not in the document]
necouchman: As far as I know, the only circumstances where `entityId` is null is where the system user is doing the work rather than an actual user.  That said, it should always be null, so I can definitely re-arrange this a bit.  I just wasn't sure where the permission comes from that grants the new entity `READ` access to itself...

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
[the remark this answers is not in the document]
necouchman: Okay, I'v re-arranged the way that method works a bit more to keep this and not check entity ID on each iteration of the loop.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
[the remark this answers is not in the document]
necouchman: Okay, I've re-arranged this a bit - let me know if that looks okay.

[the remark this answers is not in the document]
necouchman: Okay, took another couple of tries, but I've tweaked this a bit more.  Let me know if that looks okay - essentially, in the `super` version of the `getImplicitPermissions()` method I'm either returning an `emptyList()` if the entity id is null, or an `unmodifiableCollection()` of permissions.  In the `UserService` version of `getImplicitPermissions()` I have it making a copy and then adding anything to it (`READ` for self) that it needs, and return another `unmodifiableCollection()` of those.  This makes the behavior consistent across both of those methods.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/JDBCAuthenticationProviderService.java
necouchman: -Sigh-

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
```diff
@@ -427,16 +427,22 @@ public InternalType retrieveObject(ModeledAuthenticatedUser user,
     protected Collection<ObjectPermissionModel> getImplicitPermissions(ModeledAuthenticatedUser user,
             ModelType model) {
         
+        // Get the user model and check for an entity ID.
+        UserModel userModel = user.getUser().getModel();
+        Integer entityId = userModel.getEntityID();
+        if (entityId == null)
+            return Collections.emptyList();
```
mike-jumper, line 434: I'm still unclear on the circumstances that would cause `entityId` to be `null`. The entity serves as the permission-receiving base of either a user or a group, and referential integrity constraints guarantee that neither a user nor a group can be inserted without a corresponding entity.

For users, the entity is currently automatically inserted during creation of the user within `beforeCreate()` (invoked by `createObject()`):

https://github.com/apache/guacamole-client/blob/e1f95b87636705931fff50fd37598bb63a26ae48/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/UserService.java#L253-L254

Is there something going on which leads to this check being required? If so, and `entityId` can indeed be `null` here, how are things working without an entity?

necouchman: Okay, I think I see what you're getting at.  Basically...
- `createObject()` requires two parameters - the user creating the object, and the object being created.
- In `JDBCAuthenticationService` I make the call to `userService.createObject()` to create the user.  The second argument for that call is obviously the user record that is being automatically created, here.  The first argument (as currently implemented) is a new `PrivilegedModeledAuthenticatedUser` based on the user that was authenticated and is currently being added.  So, essentially, we're giving the user that is currently being added the privilege to create their own account in the database.
- However, because that user doesn't already exist (presumably _at the time of_ the `new PrivilegedModeledAuthenticatedUser() call), there is no entity, so the `entityId` field is null, resulting in a constraint violation when trying to insert those implicit privileges.
- Thus, the only implicit privilege that actually ever gets created with these checks is the one for the user to `READ` their own account.  Which I think is actually okay and desirable - otherwise, the user will end up with full privileges on their own account, which might be okay but also might be more than what the sysadmin desires, and I'm all about least privilege.

Let me know if there's something different I should do, here:
- Move the `new PrivilegedModeledAuthenticatedUser()` to a different place.
- Use the `userContext.getPrivileged()` instead of manually creating that user.
- Use a different based account for the `new PrivilegedModeledAuthenticatedUser()` call such that an entity exists.
- Something else?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
```diff
@@ -464,7 +470,9 @@ public InternalType createObject(ModeledAuthenticatedUser user, ExternalType obj
         object.setIdentifier(model.getIdentifier());
 
         // Add implicit permissions
-        getPermissionMapper().insert(getImplicitPermissions(user, model));
+        Collection<ObjectPermissionModel> implicitPermissions = getImplicitPermissions(user, model);
+        if (implicitPermissions != null && !implicitPermissions.isEmpty())
```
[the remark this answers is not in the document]
necouchman: Should be never.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
mike-jumper: Ah, OK ... I think I see what you're saying.

For users, `getImplicitPermissions()` contains essentially two distinct sets of permissions:

1. Those from `IMPLICIT_USER_PERMISSIONS`, which are the permissions that a newly-created user should have _for themselves_. This consists only of `READ`, and should successfully be granted to an automatically-created user.
2. Those from `super.getImplicitPermissions()`, which are the permissions granted to the user creating the object. This consists of several admin-level permissions, and in this case both can't and shouldn't be granted:
   * "Can't" because the user creating the object does not exist.
   * "Shouldn't" because automatically-created users shouldn't be magically granted permission to administer themselves.

Is that correct?

necouchman: Yep, exactly.  So, I think the way it's working right now is desirable, as the `IMPLICIT_USER_PERMISSIONS` are granted (user can `READ` self), and the administrative permissions granted by `super.getImplicitPermissions()` are not granted.  I just don't know if everything is set up in the best possible way or if there's a different/better way to accomplish that.

mike-jumper: I think what you have makes sense, then, but that check on whether `entityId` is `null` should be documented to clarify this need.

I suggest adding something like `isSkeleton()` to `ModeledUser` to abstract away this check. Then the code itself would embody the idea behind what you're checking, and the check within the superclass version of `getImplicitPermissions()` would be that much clearer.

necouchman: Okay, stab has been taken at your suggestion.

PR #391, opened 2019-04-03
GUACAMOLE-723: Add interface for switching between multiple active connections.

mike-jumper:
This change replaces the connection header in the Guacamole menu with a dropdown menu:

guac-multi-conn-001-dropdown

Clicking the menu reveals a list of all available connections. The user may use this to switch connections without navigating all the way back to the home screen:

guac-multi-conn-002-dropdown-expanded

As with the list of connections on the home screen, this list both paginated and filterable:

guac-multi-conn-003-dropdown-filtered

When multiple connections are opened this way within the same tab, connections running in the background are displayed as thumbnails in a collapsible panel at the bottom-right corner of the client interface:

guac-multi-conn-004-panel

Each thumbnail updates in real-time and may be manually closed from the panel by clicking the "X" in the upper-right corner of the thumbnail. If a connection running in the background encounters an error or is disconnected by the server, a warning icon is displayed to call the user's attention to the connection:

guac-multi-conn-005-panel-error

The notification containing the details of the failure will only be visible once the user navigates back to that connection. Alternatively, if the failure is expected, the user can simply click the "X" to dismiss the connection and remove the thumbnail.

The arrow button on the left edge of the panel functions as a show/hide toggle, allowing the panel to be collapsed against the right edge of the screen if it is preventing interaction with the display underneath.

If so many connections are open that the panel expands to the width of the browser window, the panel will cease expanding and become scrollable. Scrolling with the mouse scroll wheel will work even though the scroll direction is horizontal, as vertical scrolling via the mouse has been manually redirected horizontally with JavaScript.


mike-jumper:
Interesting, but I suppose that's not an unwelcome effect. I didn't test against QuickConnect, but I expect this must be due to the `<meta ...>` tag inserting the connection form.

As long as that form works, I have no issue with this.

necouchman:
QuickConnect appears to work fine from there.  Going to review this once more but I suspect it'll get merged here, shortly...

PR #392, opened 2019-04-09
GUACAMOLE-774: Add in MD4 support for MSCHAPv1/2

necouchman:
Takes a stab at manually pulling in the MD4 provider for Java versions that do not include that support, which is required for MSCHAPv1/2 RADIUS conversations.

Full Disclosure: I do not have a RADIUS server running MSCHAP, so I cannot test these changes.  The code compiles, and this seems to be the method that will work based on other changes, but if someone can test against a Windows RADIUS server running MSCHAP, that would be helpful.

### DIFF
--- a/extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/RadiusAuthenticationProviderModule.java
+++ b/extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/RadiusAuthenticationProviderModule.java
@@ -66,12 +66,10 @@ public class RadiusAuthenticationProviderModule extends AbstractModule {
         // Check for MD4 requirement
         RadiusAuthenticationProtocol authProtocol = environment.getProperty(RadiusGuacamoleProperties.RADIUS_AUTH_PROTOCOL);
         RadiusAuthenticationProtocol innerProtocol = environment.getProperty(RadiusGuacamoleProperties.RADIUS_EAP_TTLS_INNER_PROTOCOL);
-        if ((authProtocol != null 
-                    && (authProtocol == RadiusAuthenticationProtocol.MSCHAPv1 
-                    || authProtocol == RadiusAuthenticationProtocol.MSCHAPv2)) 
-                || (innerProtocol != null 
-                    && (innerProtocol == RadiusAuthenticationProtocol.MSCHAPv1 
-                    || innerProtocol == RadiusAuthenticationProtocol.MSCHAPv2))) {
+        if (authProtocol == RadiusAuthenticationProtocol.MSCHAPv1 
+                    || authProtocol == RadiusAuthenticationProtocol.MSCHAPv2
+                    || innerProtocol == RadiusAuthenticationProtocol.MSCHAPv1 
+                    || innerProtocol == RadiusAuthenticationProtocol.MSCHAPv2) {
             
             Security.addProvider(new Provider("MD4", 0.00, "MD4 for MSCHAPv1/2 Support") {
                 {
--- a/extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/RadiusAuthenticationProviderModule.java
+++ b/extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/RadiusAuthenticationProviderModule.java
@@ -20,8 +20,12 @@
 package org.apache.guacamole.auth.radius;
 
 import com.google.inject.AbstractModule;
+import java.security.Provider;
+import java.security.Security;
 import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.auth.radius.conf.ConfigurationService;
+import org.apache.guacamole.auth.radius.conf.RadiusAuthenticationProtocol;
+import org.apache.guacamole.auth.radius.conf.RadiusGuacamoleProperties;
 import org.apache.guacamole.environment.Environment;
 import org.apache.guacamole.environment.LocalEnvironment;
 import org.apache.guacamole.net.auth.AuthenticationProvider;
@@ -58,6 +62,24 @@ public class RadiusAuthenticationProviderModule extends AbstractModule {
 
         // Get local environment
         this.environment = new LocalEnvironment();
+        
+        // Check for MD4 requirement
+        RadiusAuthenticationProtocol authProtocol = environment.getProperty(RadiusGuacamoleProperties.RADIUS_AUTH_PROTOCOL);
+        RadiusAuthenticationProtocol innerProtocol = environment.getProperty(RadiusGuacamoleProperties.RADIUS_EAP_TTLS_INNER_PROTOCOL);
+        if ((authProtocol != null 
+                    && (authProtocol == RadiusAuthenticationProtocol.MSCHAPv1 
+                    || authProtocol == RadiusAuthenticationProtocol.MSCHAPv2)) 
+                || (innerProtocol != null 
+                    && (innerProtocol == RadiusAuthenticationProtocol.MSCHAPv1 
+                    || innerProtocol == RadiusAuthenticationProtocol.MSCHAPv2))) {
+            
+            Security.addProvider(new Provider("MD4", 0.00, "MD4 for MSCHAPv1/2 Support") {
+                {
+                    this.put("MessageDigest.MD4", org.bouncycastle.jce.provider.JDKMessageDigest.MD4.class.getName());
+                }
+            });
+            
+        }
 
         // Store associated auth provider
         this.authProvider = authProvider;
--- a/extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/RadiusAuthenticationProviderModule.java
+++ b/extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/RadiusAuthenticationProviderModule.java
@@ -21,6 +21,7 @@ package org.apache.guacamole.auth.radius;
 
 import com.google.inject.AbstractModule;
 import org.apache.guacamole.GuacamoleException;
+import org.apache.guacamole.auth.radius.conf.ConfigurationService;
 import org.apache.guacamole.environment.Environment;
 import org.apache.guacamole.environment.LocalEnvironment;
 import org.apache.guacamole.net.auth.AuthenticationProvider;
--- a/extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/RadiusConnectionService.java
+++ b/extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/RadiusConnectionService.java
@@ -25,6 +25,8 @@ import java.io.IOException;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.security.NoSuchAlgorithmException;
+import java.security.Provider;
+import java.security.Security;
 import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.GuacamoleServerException;
 import org.slf4j.Logger;
@@ -41,6 +43,8 @@ import net.jradius.packet.AccessRequest;
 import net.jradius.packet.attribute.AttributeList;
 import net.jradius.client.auth.EAPTLSAuthenticator;
 import net.jradius.client.auth.EAPTTLSAuthenticator;
+import net.jradius.client.auth.MSCHAPv1Authenticator;
+import net.jradius.client.auth.MSCHAPv2Authenticator;
 import net.jradius.client.auth.RadiusAuthenticator;
 import net.jradius.client.auth.PEAPAuthenticator;
 import net.jradius.packet.attribute.AttributeFactory;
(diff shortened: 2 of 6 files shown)


extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/RadiusConnectionService.java
```diff
@@ -129,6 +133,18 @@ private RadiusAuthenticator setupRadiusAuthenticator(RadiusClient radiusClient)
         if (radAuth == null)
             throw new GuacamoleException("Could not get a valid RadiusAuthenticator for specified protocol: " + confService.getRadiusAuthProtocol());
 
+        // For MSCHAPv1/2, we need MD4 support
+        if (radAuth instanceof MSCHAPv1Authenticator
+                || radAuth instanceof MSCHAPv2Authenticator) {
+            
+            Security.addProvider(new Provider("MD4", 0.00, "MD4 for MSCHAPv1/2 RADIUS") {
```
mike-jumper: `setupRadiusAuthenticator()` is invoked from within `authenticate()`, so this will result in `Security.addProvider(...)` being invoked each time someone authenticates. It's unclear to me within the documentation for `addProvider()` whether this will result in multiple copies of the MD4 implementation being added.

Perhaps this should be moved to a location where it occurs only once?

Is it known that no other auth protocols use MD4?

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/RadiusConnectionService.java
```diff
@@ -129,6 +133,18 @@ private RadiusAuthenticator setupRadiusAuthenticator(RadiusClient radiusClient)
         if (radAuth == null)
             throw new GuacamoleException("Could not get a valid RadiusAuthenticator for specified protocol: " + confService.getRadiusAuthProtocol());
 
+        // For MSCHAPv1/2, we need MD4 support
+        if (radAuth instanceof MSCHAPv1Authenticator
+                || radAuth instanceof MSCHAPv2Authenticator) {
```
mike-jumper: Rather than checking `instanceof` everywhere, and thus relying on the internals of JRadius, it would be better if `getRadiusAuthProtocol()` leveraged an `enum` that could be checked independently of whatever JRadius decides to return. Such an `enum` could provide a method for obtaining a corresponding instance of `RadiusAuthenticator` given a `RadiusClient`.

Rechecking `getRadiusAuthProtocol()`, it is somewhat scary that the raw string value is passed straight to JRadius. I think there should be some validation on the value on the Guacamole side, and a decoupling from the underlying RADIUS library.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/RadiusConnectionService.java
necouchman: Yeah, probably a good point.  I could move it to the constructor for the `RadiusConnectionService` class, but not sure if that would still happen each time someone logged in, or would be a one-shot thing?

> Is it known that no other auth protocols use MD4?

I do not know one way or the other if others require it.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/RadiusConnectionService.java
necouchman: Okay, I'll work up a new property type and `enum` for this.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/RadiusConnectionService.java
necouchman: Okay, I've put the initialization for this into the Guice module, which I think will suffice to load it only a single time.

I queried the JRadius repository for any other instances of "MD4", and the MSCHAP library is the only place it shows up, so I think that covers the requirement:
https://github.com/coova/jradius/search?q=MD4&unscoped_q=MD4

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/RadiusConnectionService.java
necouchman: This should be implemented as requested.

necouchman:
@mike-jumper: Okay, addressed those two issues - let me know if anything else needs work.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/RadiusConnectionService.java
```diff
@@ -68,6 +70,27 @@
     private ConfigurationService confService;
 
 
+    /**
+     *
```
mike-jumper: Empty comment.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/conf/RadiusAuthenticationProtocol.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+package org.apache.guacamole.auth.radius.conf;
+
+/**
+ * This enum represents supported RADIUS authentication protocols for
+ * the guacamole-auth-radius extension.
+ */
+public enum RadiusAuthenticationProtocol {
+    
+    // Password authentication protocol
```
mike-jumper: The comments documenting `enum` values should be JavaDoc style.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/conf/RadiusAuthenticationProtocol.java
```diff
@@ -0,0 +1,64 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
```
mike-jumper: Oops - Netbeans default template snuck in. I think this may also fail the build when the RAT plugin kicks in.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/conf/RadiusAuthenticationProtocol.java
```diff
@@ -0,0 +1,64 @@
...
+    // Extensible authentication protocol with MD5 hashing.
+    EAP_MD5("eap-md5"),
+
+    // Extensible authentication protocol with TLS
+    EAP_TLS("eap-tls"),
+
+    // Extensible authentication protocol with Tunneled TLS
+    EAP_TTLS("eap-ttls");
+
+    // Store the string value used in the configuration file.
```
mike-jumper: JavaDoc.

Besides being the value used in `guacamole.properties`, it looks like we're intentionally using the value used by JRadius (and relying on this later on). That may a detail be worth documenting.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/conf/RadiusAuthenticationProtocol.java
```diff
@@ -0,0 +1,64 @@
...
+    private final String strValue;
+    
+    /**
+     * Create a new RadiusAuthenticationProtocol object having the
+     * given string value.
+     * 
+     * @param strValue
+     *     The string value of the protocol.
```
mike-jumper: If I'm going to invoke this constructor (defining a new enum value perhaps), "the string value of the protocol" doesn't give me an understanding of what needs to go here. I would need to go through the code using the value to determine what actually is needed.

Is there a better way to describe this?

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/conf/RadiusAuthenticationProtocol.java
```diff
@@ -0,0 +1,64 @@
...
+     */
+    public RadiusAuthenticationProtocol(String strValue) {
+        this.strValue = strValue;
+    }
+    
+    @Override
+    public String toString() {
```
mike-jumper: Though it's not normally needed to document `toString()`, I would suggest doing so here since we are depending on this function having specific behavior (returning the value used by both `guacamole.properties` and specific parts of JRadius).

Something like:

```java
/**
 * {@inheritDoc}
 * <p>
 * Doooocuuumentaaationnn...
 */
```

would probably do well.

The alternative would be to not use `toString()` but our own function with well defined semantics regarding the return value and `guacamole.properties` and JRadius.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/conf/RadiusAuthenticationProtocol.java
```diff
@@ -0,0 +1,64 @@
...
+        return strValue;
+    }
+    
+    @Override
+    public static RadiusAuthenticationProtocol valueOf(String value) {
```
mike-jumper: `valueOf()` for enums is explicitly defined as returning the enum constant whose name exactly matches the given string:

https://docs.oracle.com/javase/8/docs/api/java/lang/Enum.html#valueOf-java.lang.Class-java.lang.String-

I don't think we should override the default implementation of a function such that it no longer matches the contract of the original interface. This should probably be our own function.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/conf/RadiusAuthenticationProtocolProperty.java
```diff
@@ -0,0 +1,38 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
```
mike-jumper: Double oops. Netbeans template.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/RadiusConnectionService.java
necouchman: Not seeing this in the current set of commits...were you looking at an earlier commit??

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/conf/RadiusAuthenticationProtocol.java
necouchman: Again, not sure if you were looking at a specific commit instead of the latest round of code, but the latest has `getEnum()` defined (which I failed to document, so that needs fixing) - turns out that I don't think it's even valid to override `valueOf()` for `enum`s.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/conf/RadiusAuthenticationProtocolProperty.java
necouchman: As with others, I think this is actually okay in the most recent code.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/conf/RadiusAuthenticationProtocol.java
necouchman: I think this is good in most recent code.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/conf/RadiusAuthenticationProtocol.java
necouchman: JavaDoc'd.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/conf/RadiusAuthenticationProtocol.java
necouchman: JavaDoc'd, and added some detail.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/conf/RadiusAuthenticationProtocol.java
necouchman: Took a better slice at this documentation.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/conf/RadiusAuthenticationProtocol.java
necouchman: Doc'd.

necouchman:
@mike-jumper: Okay, fixed up the documentation in that enum where needed.

necouchman:
@mike-jumper: Any more work needed, here?

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/RadiusAuthenticationProviderModule.java
```diff
@@ -57,6 +62,24 @@ public RadiusAuthenticationProviderModule(AuthenticationProvider authProvider)
...
+        
+        // Check for MD4 requirement
+        RadiusAuthenticationProtocol authProtocol = environment.getProperty(RadiusGuacamoleProperties.RADIUS_AUTH_PROTOCOL);
+        RadiusAuthenticationProtocol innerProtocol = environment.getProperty(RadiusGuacamoleProperties.RADIUS_EAP_TTLS_INNER_PROTOCOL);
+        if ((authProtocol != null 
+                    && (authProtocol == RadiusAuthenticationProtocol.MSCHAPv1 
+                    || authProtocol == RadiusAuthenticationProtocol.MSCHAPv2)) 
+                || (innerProtocol != null 
+                    && (innerProtocol == RadiusAuthenticationProtocol.MSCHAPv1 
+                    || innerProtocol == RadiusAuthenticationProtocol.MSCHAPv2))) {
```
mike-jumper: You might be able to clean up this logic if you remove the `null` checks, which are unnecessary in this case. The `==` comparisons will just fail if `authProtocol` or `innerProtocol` are `null`.

necouchman: Cleaned up.

necouchman:
@mike-jumper: Okay, that's been cleaned up.

PR #393, opened 2019-04-18
GUACAMOLE-678: Assign clientID variable instead of string literal

necouchman:
Fix issue introduced with #349.


mike-jumper:
Yikes. Good catch.

PR #398, opened 2019-04-26
GUACAMOLE-784: Tolerate port number within X-Forwarded-For header

sporeno:


necouchman:
@sporeno:
* This PR is still in draft mode.
* You still need to create a JIRA issue and tag both the PR and commits with that issue (see other pull requests for examples)

guacamole/src/main/java/org/apache/guacamole/rest/auth/AuthenticationService.java
```diff
@@ -93,12 +93,12 @@
...
      * Regular expression which matches any IPv4 address.
      */
-    private static final String IPV4_ADDRESS_REGEX = "([0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3})";
+    private static final String IPV4_ADDRESS_REGEX = "([0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3})(?::[0-9]{1,4})?";
 
     /**
      * Regular expression which matches any IPv6 address.
      */
-    private static final String IPV6_ADDRESS_REGEX = "([0-9a-fA-F]*(:[0-9a-fA-F]*){0,7})";
+    private static final String IPV6_ADDRESS_REGEX = "([0-9a-fA-F]*(:[0-9a-fA-F]*){0,7})(?::[0-9]{1,4})?";
```
mike-jumper: Both `IPV4_ADDRESS_REGEX` and `IPV6_ADDRESS_REGEX` are documented here as matching IP addresses. Altering them such that they also accept port numbers will mean that the documentation becomes incorrect. If the change proposed here is correct, then that documentation needs to be updated to match. However:

* Duplicating the same pattern across both `IPV4_ADDRESS_REGEX` and `IPV6_ADDRESS_REGEX` is suboptimal. There are other patterns which be better homes for this change and avoid duplication, but again: modifying something that is essentially named "IP address" and documented as matching IP addresses such that it also matches port numbers isn't complete in itself. That change would need to be followed through such that the documentation and naming are correct.
* The de facto `X-Forwarded-For` header is defined as a list of IP addresses, not a list of IP addresses with optional port numbers:

  https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For
  https://en.wikipedia.org/wiki/X-Forwarded-For

  If there are real world cases where a port number is included, please provide some background information when you open the corresponding issue in JIRA so the reasoning for your proposed change can be understood.

necouchman:
@sporeno Commit messages should be tagged with the same `GUACAMOLE-784:` prefix that the pull request has.

guacamole/src/main/java/org/apache/guacamole/rest/auth/AuthenticationService.java
```diff
@@ -93,12 +93,12 @@
     /**
      * Regular expression which matches any IPv4 address.
      */
-    private static final String IPV4_ADDRESS_REGEX = "([0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3})";
+    private static final String IPV4_ADDRESS_REGEX = "([0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3})(?::[0-9]{1,5})?";
```
mike-jumper: The documentation for this constant currently states:

> Regular expression which matches any IPv4 address.

Modifying this constant such that it also matches port numbers isn't complete in itself. That change would need to be followed through such that the documentation is correct. The same goes for the naming of the constant (`IPV4_ADDRESS_REGEX` is pretty specific to IPv4 addresses).

It may be much simpler to modify `IP_ADDRESS_REGEX` instead. This would remove the duplicated pattern and may be easier to rename and document.

sporeno: Another way may would be to create e new constant PORT_NUMBER_REGEX. And combine it within the X_FORWARDED_FOR Constant. Then you have an impended Port Regex and the renaming can be avoided.

mike-jumper: Sure. That sounds good and satisfies the criteria.

guacamole/src/main/java/org/apache/guacamole/rest/auth/AuthenticationService.java
```diff
@@ -105,11 +105,16 @@
      */
     private static final String IP_ADDRESS_REGEX = "(" + IPV4_ADDRESS_REGEX + "|" + IPV6_ADDRESS_REGEX + ")";
 
+     /**
```
mike-jumper: It looks like the start of this comment was indented one space too far. There is a standard formatting for JavaDoc comments. For example:

https://github.com/apache/guacamole-client/blob/8f12ae4fcb1c3385d2f214f0cfcaa29f5ff8bf64/guacamole/src/main/java/org/apache/guacamole/rest/auth/AuthenticationService.java#L113-L116

mike-jumper:
I'd also like to highlight:

Screenshot 2019-05-02 at 11 07 44 PM

Regardless of whether you choose to rebase and squash things such that there is only one commit or prepend the JIRA issue to your existing commits, the commit messages need to describe the nature of the change being made, ideally at a high level. They should provide some context/justification that is not otherwise available through git itself.

A message like "Update AuthenticationService.java" does not provide any more information than git already has. It may as well say "potato". ;)

Your first and last commits, on the other hand, are quite nice. "Modify regex for X-Forwarded-For to parse IP:Port" is very clear and describes the nature of the change you're making, as does "Tolerate port number within X-Forwarded-For header" (though they really sound like they should be the same commit).

sporeno:
I managed to change the commit messages.

necouchman:
Changes look okay to me, overall, just a couple of issues with the commits:
* First commit has a message of "Initial commit", which isn't a very descriptive message.
* For six lines of code changes there are 5 commits, which seems a little much - maybe consider squashing all of your commits into a single one with a more descriptive message?

sporeno:
I change the squashed the commits and change the message.

PR #401, opened 2019-04-28
GUACAMOLE-742: Provide feedback while login is in progress.

mike-jumper:
These changes:

* Add a new `disabled` attribute to the `guacForm` and `guacFormField` directives which allow fields and forms to be disabled.
* Updates existing field types to support being disabled.
* Disables the login form while the authentication request is being processed and while the initial route following login has not yet been satisfied.

The CSS associated with the login dialog has also been updated such that the dialog doesn't repeatedly flash/fade with each state change.


necouchman:
:shipit:

PR #402, opened 2019-05-10
GUACAMOLE-649: Add definition and translation for SSH "locale" parameter.

mike-jumper:


mike-jumper:
I have the `timezone` parameter definitions for [GUACAMOLE-422](https://issues.apache.org/jira/browse/GUACAMOLE-422) prepared, as well, though that will have to wait until after this since I've placed the `timezone` parameter directly under `locale` in `ssh.json` (see 212a39b0250ca18632c0c3ccc21229daed3981e8).

PR #406, opened 2019-05-20
GUACAMOLE-643: Show connection groups during filter operations

InfocertAutomationTeam:
If a connection group has the 'filter' attribute to 'false', it won't be hidden during a search.

Example of connection group that won't be hidden:

```
ConnectionGroup group = new MyConnectionGroup(groupName, groupName, Collections.<String>emptyList(), Collections.<String>emptyList());
group.setParentIdentifier(ROOT_IDENTIFIER);
group.setAttributes(Collections.singletonMap("filter", "false"));
rootGroup.getConnectionGroupIdentifiers().add(groupName);
connectionGroups.add(group);
```


mike-jumper:
1. Beware that your pull request includes three commits with identical commit messages.
2. I don't believe adding a new attribute that makes connection groups oblivious to filtering is a good approach to the need described by GUACAMOLE-643; it's too hacky. Assuming that JIRA issue is something that should be improved, discussion on the approach is needed. One such alternative approach might be to alter the way the flattened connections are rendered such that they can be distinguished despite being in separate groups.

PR #419, opened 2019-06-24
GUACAMOLE-302: Incorrect password field focus

m-khan-glyptodon:
Note: This is not specific to just password fields, but any field type in general.

### DIFF
--- a/guacamole/src/main/webapp/app/form/templates/timeZoneField.html
+++ b/guacamole/src/main/webapp/app/form/templates/timeZoneField.html
@@ -3,6 +3,7 @@
     <!-- Available time zone regions -->
     <select class="time-zone-region"
             ng-disabled="disabled"
+            guac-focus="focused"
             ng-model="region"
             ng-options="name for name in regions | orderBy: name"></select>
 
--- a/guacamole/src/main/webapp/app/login/directives/login.js
+++ b/guacamole/src/main/webapp/app/login/directives/login.js
@@ -151,7 +151,7 @@ angular.module('login').directive('guacLogin', [function guacLogin() {
                     $scope.enteredValues[field.name] = '';
             });
 
-            console.log(mostRelevantField());
+            $scope.relevantField = getRelevantField();
 
         });
 
@@ -203,8 +203,6 @@ angular.module('login').directive('guacLogin', [function guacLogin() {
                             delete $scope.enteredValues[field.name];
 
                     });
-
-                    console.log(mostRelevantField());
                 }
 
             }));
@@ -212,15 +210,14 @@ angular.module('login').directive('guacLogin', [function guacLogin() {
         };
 
         /**
-         * Returns the field most relevant field to the user given the current
-         * state of the login process. This will normally be the first empty
-         * field.
+         * Returns the field most relevant to the user given the current state
+         * of the login process. This will normally be the first empty field.
          *
          * @return {Field}
          *     The field most relevant, null if there is no single most relevant
          *     field.
          */
-        var mostRelevantField = function findMostRelevantField() {
+        var getRelevantField = function getRelevantField() {
 
             for (var i = 0; i < $scope.remainingFields.length; i++) {
                 var field = $scope.remainingFields[i];
--- a/guacamole/src/main/webapp/app/login/directives/login.js
+++ b/guacamole/src/main/webapp/app/login/directives/login.js
@@ -103,6 +103,13 @@ angular.module('login').directive('guacLogin', [function guacLogin() {
         $scope.submitted = false;
 
         /**
+         * The field that is most relevant to the user.
+         *
+         * @type Field
+         */
+        $scope.relevantField = null;
+
+        /**
          * Returns whether a previous login attempt is continuing.
          *
          * @return {Boolean}
@@ -143,6 +150,8 @@ angular.module('login').directive('guacLogin', [function guacLogin() {
                 if (!$scope.enteredValues[field.name])
                     $scope.enteredValues[field.name] = '';
             });
+
+            console.log(mostRelevantField());
 
         });
 
@@ -194,9 +203,32 @@ angular.module('login').directive('guacLogin', [function guacLogin() {
                             delete $scope.enteredValues[field.name];
 
                     });
+
+                    console.log(mostRelevantField());
                 }
 
             }));
+
+        };
+
+        /**
+         * Returns the field most relevant field to the user given the current
+         * state of the login process. This will normally be the first empty
+         * field.
+         *
+         * @return {Field}
+         *     The field most relevant, null if there is no single most relevant
+         *     field.
+         */
+        var mostRelevantField = function findMostRelevantField() {
+
+            for (var i = 0; i < $scope.remainingFields.length; i++) {
+                var field = $scope.remainingFields[i];
+                if (!$scope.enteredValues[field.name])
+                    return field;
+            }
+
+            return null;
 
         };
 
--- a/guacamole/src/main/webapp/app/form/directives/form.js
+++ b/guacamole/src/main/webapp/app/form/directives/form.js
@@ -187,6 +187,15 @@ angular.module('form').directive('guacForm', [function form() {
 
             });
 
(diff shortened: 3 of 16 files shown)


guacamole/src/main/webapp/app/login/directives/login.js
```diff
@@ -188,17 +190,36 @@ angular.module('login').directive('guacLogin', [function guacLogin() {
...
             }));
 
         };
+        
+        /**
+         * Returns the field most relevant to the user given the current state
+         * of the login process. This will normally be the first empty field.
+         *
+         * @return {Field}
+         */
```
mike-jumper: Missing documentation for the return value.

guacamole/src/main/webapp/app/login/directives/login.js
```diff
@@ -137,13 +137,15 @@ angular.module('login').directive('guacLogin', [function guacLogin() {
...
             });
-
+            
             // Set default values for all unset fields
             angular.forEach($scope.remainingFields, function setDefault(field) {
                 if (!$scope.enteredValues[field.name])
                     $scope.enteredValues[field.name] = '';
             });
 
+            $scope.relevantField = getRelevantField();
```
mike-jumper: Properties which will be set on the scope need to be documented, just as functions, objects, etc. This is done through assigning an initial value to the property earlier in the directive/controller and using that assignment as the location for the property's JSDoc:

For example:

https://github.com/apache/guacamole-client/blob/829aac98c7c6b04140d97c17be7347039a93ef11/guacamole/src/main/webapp/app/login/directives/login.js#L74-L93

guacamole/src/main/webapp/app/login/directives/login.js
```diff
@@ -102,6 +102,13 @@ angular.module('login').directive('guacLogin', [function guacLogin() {
          */
         $scope.submitted = false;
 
+        /**
+         * The field that needs to be focused
```
mike-jumper: JSDoc comments need to end with a period.

> The field that needs to be focused

This feels too tightly coupled to the actual internal implementation. It is ultimately used to focus a field, yes, but I don't think this level of things should be aware of that. The property is named nicely based on its high-level use: `relevantField`. Is there a better way to phrase this which captures that high-level intent?

guacamole/src/main/webapp/app/login/directives/login.js
```diff
@@ -102,6 +102,13 @@ angular.module('login').directive('guacLogin', [function guacLogin() {
...
+         *
+         * @type Field
+         */
+        $scope.relevantField = {};
```
mike-jumper: `{}` is not a `Field` and will also fail simple tests for whether the value has been set (`{}` is truthy). If you're looking for a default value prior to actual assignment later, `null` would be a better choice.

m-khan-glyptodon:
Update: Added the two pieces of documentation requested above.

guacamole/src/main/webapp/app/login/directives/login.js
mike-jumper: > Update: Added the two pieces of documentation requested above.

I see the new documentation for `$scope.relevantField`, but this return value still appears undocumented.

m-khan-glyptodon:
Update: Fixed the default value of $scope.relevantField = {}; as well as the documentation.

guacamole/src/main/webapp/app/login/directives/login.js
```diff
@@ -188,17 +197,38 @@ angular.module('login').directive('guacLogin', [function guacLogin() {
...
             }));
 
         };
+        
+        /**
+         * Returns the field most relevant to the user given the current state
+         * of the login process. This will normally be the first empty field.
+         *
+         * @return {Field}
+         *     A field object if an incorrect ligin is attempted, null if no
```
mike-jumper: "ligin"

Typo aside, what does this function have to do with incorrect login? This function looks only at the current state of the login process - what values have been entered, what values remain, etc. It isn't aware of whether login was incorrect.

I'd also like to point out that "a field object" doesn't provide much more information beyond what is already given by `@return {Field}`. The documentation for the function itself covers this, but the documentation for the return value should also.

m-khan-glyptodon:
Update: Changes ready for review.

guacamole/src/main/webapp/app/form/templates/timeZoneField.html
```diff
@@ -3,12 +3,14 @@
...
     <select class="time-zone-region"
             ng-disabled="disabled"
+            guac-focus="focused"   
             ng-model="region"
             ng-options="name for name in regions | orderBy: name"></select>
 
     <!-- Time zones within selected region -->
     <select class="time-zone"
             ng-disabled="disabled || !region"
+            guac-focus="focused"
```
mike-jumper: Having `guac-focus="focused"` for both `<select>` fields doesn't make sense, as you can only have one field focused at a time. What is the intended effect?

guacamole/src/main/webapp/app/login/directives/login.js
```diff
@@ -188,17 +197,37 @@ angular.module('login').directive('guacLogin', [function guacLogin() {
...
             }));
 
         };
+        
+        /**
+         * Returns the field most relevant to the user given the current state
+         * of the login process. This will normally be the first empty field.
+         *
+         * @return {Field}
+         *     A field most relevant, null if no relevant field exist.
```
mike-jumper: > A field most relevant

Do you mean "the field most relevant?"

> null if no relevant field exist.

Is this the case? My reading of the logic is that it will return null if all fields appear equally relevant (ie: there is no single "most relevant" field), not necessarily that are absolutely no relevant fields whatsoever. In fact, for the login process, it's more likely that all fields are relevant; it's a question of degrees.

guacamole/src/main/webapp/app/login/directives/login.js
```diff
@@ -188,17 +197,37 @@ angular.module('login').directive('guacLogin', [function guacLogin() {
...
                     angular.forEach($scope.remainingFields, function clearEnteredValueIfPassword(field) {
-
+                        
                         // If field is not username field, delete it.
                         if (field.type !== Field.Type.USERNAME && field.name in $scope.enteredValues)
                             delete $scope.enteredValues[field.name];
-
+                        
                     });
+
```
mike-jumper: Not a deal-breaker, but if you can, I suggest rebasing away these whitespace changes. I'm not sure where they came from, but they don't appear related to the changes.

guacamole/src/main/webapp/app/form/templates/timeZoneField.html
m-khan-glyptodon: What you point out is indeed correct and there was no legit reasoning behind this besides the fact that I was applying this `guac-focus` on every field template out there. I will make appropriate changes based on actual reasoning this time around.

m-khan-glyptodon: The new approach that I believe is suitable for this field is to have the `guac-focus` directive applied only to the "time-zone-region" and there are a few reasons why I believe it is the right to do.
1. In most cases, it is impossible to determine which part of the field is relevant so it makes sense to naturally focus the beginning of this field.
2. Also, the "time-zone" field is disabled until some content is selected in the "time-zone-region". Because of this dependency, I think it is best to focus the field that does not depend on the other.
3. And last but not the least, it is simple.

mike-jumper: Sure. Focusing just the first `<select>` makes sense to me.

m-khan-glyptodon:
Update: Unneeded whitespaces removed as well the extra `guac-focus`. Also the documentation is altered as per your suggestion. Ready for review.

guacamole/src/main/webapp/app/login/directives/login.js
mike-jumper: > Unneeded whitespaces removed

It's partly fixed, yes, but the two instances of added whitespace above (lines 204 and 206) are still here.

PR #420, opened 2019-06-25
GUACAMOLE-302: Refocus relevant in-progress login fields after auth failure

m-khan-glyptodon:
Note: There were two parts of this issue this time around. The first(error message) has been fixed and described in the commit itself. As far as the inconsistency in focusing the username field on the initial page load is concerned, that seems to no longer happen as a result of this fix. I do not understand why that is the case, but that is the behavior I have observed. For that reason, I am uncertain whether we can deem that part as resolved.

### DIFF
--- a/guacamole/src/main/webapp/app/element/directives/guacFocus.js
+++ b/guacamole/src/main/webapp/app/element/directives/guacFocus.js
@@ -52,20 +52,6 @@ angular.module('element').directive('guacFocus', ['$parse', function guacFocus($
                 });
             });
 
-            // Set focus flag when focus is received
-            element.addEventListener('focus', function focusReceived() {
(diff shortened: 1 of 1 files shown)


mike-jumper:
> As far as the inconsistency in focusing the username field on the initial page load is concerned, that seems to no longer happen as a result of this fix. I do not understand why that is the case, but that is the behavior I have observed. For that reason, I am uncertain whether we can deem that part as resolved.

I'm still seeing the same behavior following merge. The issue specifically addressed by these changes is fixed, but the inconsistent focus remains. If things were working for you after making this change, that was likely a coincidence.

Testing against things before the merge, the focus behavior doesn't correlate with the error message. The error message appeared when fields were manually blurred, but didn't appear when the username failed to receive focus or when it successfully received focus.

PR #421, opened 2019-06-25
GUACAMOLE-302: Do not rely on $evalAsync() for assigning focus.

mike-jumper:
It appears that we cannot rely on `focus()` working as expected unless its execution is deferred via `window.setTimeout()` or equivalent. If not deferred, the target element may not be ready to receive focus, and the attempt will silently fail.

Note that while there is a _chance_ that deferred execution will happen with the current usage of `$evalAsync()` within the `guacFocus` directive, this is not guaranteed. If it can, AngularJS will execute the function provided to `$evalAsync()` immediately after the current `$digest` loop and without control returning to the browser.

### DIFF
--- a/guacamole/src/main/webapp/app/element/directives/guacFocus.js
+++ b/guacamole/src/main/webapp/app/element/directives/guacFocus.js
@@ -20,7 +20,11 @@
 /**
  * A directive which allows elements to be manually focused / blurred.
  */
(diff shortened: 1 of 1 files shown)


jmuehlner:
Well, I don't fully understand this, but testing confirms that this fixes the issue on several different browsers. LGTM.

PR #422, opened 2019-07-01
GUACAMOLE-300: Support posixGroup in LDAP Authentication and Group-based Session Admission

mlewissmith:
To support groups where group members are specified by usercode, rather than by DN
- new guacamole.property 'ldap-member-attribute-type'.  Can take value "dn" (as now) or "uid"


extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
```diff
@@ -165,14 +165,44 @@ private String getGroupSearchFilter() throws GuacamoleException {
         if (groupBaseDN == null)
             return Collections.emptyList();
 
+        // memberAttribute specified in properties could contain DN or username 
+        String memberAttributeType = confService.getMemberAttributeType();
+        String userID = userDN;
+        switch (memberAttributeType) {
```
necouchman: Might be worth doing a new `enum` type for this and a `GuacamoleProperty` that validates it.

mike-jumper:
> ... I suggest doing an `enum` and `GuacamoleProperty` to cover this new property so that it can be validated correctly.

+1

I also suggest holding on merge until after #345, as this will likely need to be rebased and slightly modified following the migration from JLDAP to the Apache Directory API. These changes will otherwise end up needing to be rewritten during merge conflict resolution, which will be awkward.

mlewissmith:
yes. of course it should be an enum...

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
necouchman: Looks much better with the `enum`, here, but, now I think we can probably get rid of the `switch()` statement, no?  Based on the way the logic is constructed, here, I think you can just do the `String userID = userDN;` assignment, and then do a `if (memberAttributeType == MemberAttributeType.UID)` with and re-assign as necessary?

mlewissmith:
switch is gone as requested.  I'm happy to wait and deal with the API changes when the time comes

mlewissmith:
Sorry -- forgot to tag the commit messages appropriately.  I hope force-push doesn't cause problems

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
```diff
@@ -396,4 +396,21 @@ public String getMemberAttribute() throws GuacamoleException {
         );
     }
 
+    /**
+     * Returns a string indicating the type of data contained in the LDAP
```
necouchman: No longer returns a `String` - is now a `MemberAttributeType`

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
```diff
@@ -396,4 +396,21 @@ public String getMemberAttribute() throws GuacamoleException {
...
+     * attribute used to enumerate members in a group, "dn" by default.
+     *
+     * @return
+     *     The type of data contained in the LDAP attribute used to enumerate
+     *     members in a group, as configured in guacamole.properties
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties connect be parsed.
```
necouchman: connect -> cannot

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/MemberAttributeTypeProperty.java
```diff
@@ -0,0 +1,53 @@
...
+        // dn
+        if (value.equals("dn"))
+            return MemberAttributeType.DN;
+
+        // uid
+        if (value.equals("uid"))
+            return MemberAttributeType.UID;
+
+        // The provided value is not legal
+        throw new GuacamoleServerException("Encryption method must be one of \"dn\" or \"uid\".");
```
necouchman: Copypasta, I think - this is not the encryption method ;-).

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ConfigurationService.java
mlewissmith: ... also copypasta, from getMemberAttribute()

necouchman:
@mlewissmith: Okay, changes for the LDAP API in use (JLDAP -> Apache) have been committed, so should be safe to clean this up, now, and take another look.

mlewissmith:
I may have messed up the rebase.  Do you want me to close this PR and start again?

necouchman:
> I may have messed up the rebase. Do you want me to close this PR and start again?

Up to you - you might be able to rebase, again, on top of the current master, and pull out all of the changes except yours, or you can just close it and open a new one.  It does seem to be a little messed up - 6K lines of code changed :-).

PR #423, opened 2019-07-06
GUACAMOLE-837: Add RDP keymap for Hungarian keyboard layout

P-Zs:
Please add these changes to code base to add Hunagarian keyboard support to Guacamole.

### DIFF
--- a/guacamole-ext/src/main/resources/org/apache/guacamole/protocols/rdp.json
+++ b/guacamole-ext/src/main/resources/org/apache/guacamole/protocols/rdp.json
@@ -99,6 +99,7 @@
                         "failsafe",
                         "fr-fr-azerty",
                         "fr-ch-qwertz",
+                        "hu-hu-qwertz",                        
                         "it-it-qwerty",
                         "ja-jp-qwerty",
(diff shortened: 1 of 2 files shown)


necouchman:
@P-Zs:
A couple of slight nit-picks here off the bat:
- The JIRA issue in the Pull Request title should be in all caps - GUACAMOLE-837.
- The JIRA issue should be tagged in the commit messages, too.
- The commit messages should probably be more descriptive (e.g. "Add Hungarian keyboard option to rdp.json").

mike-jumper:
> The commit messages should probably be more descriptive (e.g. "Add Hungarian keyboard option to rdp.json").

There's a good blog post covering what makes a bad vs. good message here:

A good rule of thumb is to make sure the message provides high-level information not otherwise captured by git already. Messages like "Update rdp.json" and "Update en.json" only serve to duplicate what git already tells us in the diff and will not help someone looking through the logs, git blame, git bisect, etc. in the future.

It should be possible to tell what you generally meant to do with your change from reading the commit messages, even if the changes themselves turn out to be incorrect some time in the future.

PR #426, opened 2019-07-16
GUACAMOLE-846: tunnel.uuid not initialized if tunnel becomes UNSTABLE

avwx:
This patch will set the `tunnel.uuid` if it's undefined.

### DIFF
--- a/guacamole-ext/src/main/resources/org/apache/guacamole/protocols/rdp.json
+++ b/guacamole-ext/src/main/resources/org/apache/guacamole/protocols/rdp.json
@@ -99,6 +99,7 @@
                         "failsafe",
                         "fr-fr-azerty",
                         "fr-ch-qwertz",
+                        "hu-hu-qwertz",                        
                         "it-it-qwerty",
                         "ja-jp-qwerty",
                         "pt-br-qwerty",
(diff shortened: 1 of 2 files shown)


guacamole-common-js/src/main/webapp/modules/Tunnel.js
```diff
@@ -1015,7 +1015,7 @@ Guacamole.WebSocketTunnel = function(tunnelURL) {
                     var opcode = elements.shift();
 
                     // Update state and UUID when first instruction received
-                    if (tunnel.state === Guacamole.Tunnel.State.CONNECTING) {
+                    if (tunnel.state === Guacamole.Tunnel.State.CONNECTING || tunnel.uuid === null) {
```
mike-jumper: If the issue is that the value of `tunnel.uuid` is being inadvertently overly coupled to tunnel state, and that the tunnel UUID should be accepted when sent via `INTERNAL_DATA_OPCODE` so long as it hasn't yet been set, perhaps this should just be `if (!tunnel.uuid)`?

avwx: agreed. Do you prefer `!tunnel.uuid` over `tunnel.uuid === null`?

mike-jumper: Eh ... I have nothing against keeping `=== null` in this case. `tunnel.uuid` is documented specifically as being `null` if not known, and the WebSocket implementation can certainly be aware of its own use of `null`.

PR #427, opened 2019-07-16
GUACAMOLE-846: tunnel.uuid not initialized if tunnel becomes UNSTABLE

avwx:
This patch will set the `tunnel.uuid` if it's undefined.


mike-jumper:
FYI - you don't need to keep closing and opening new pull requests for the same change. You can always just push new commits, force push following a rebase, etc.

PR #428, opened 2019-07-18
GUACAMOLE-827: If the FileReader object is busy, don't continue with …

jcaple:
…the next block of blob data when transferring files to remote server until it is free to continue.

I've demonstrated this problem and solution in two separate projects, which combined, provide a web client to a local guacamole server that allows file transfers to a remote RDP Windows Desktop.  The new BlobWriter.js does not throw errors when transferring files resulting in complete file transfers.

To test, run these projects using the existing BlobWriter.js code and transfer image files of decent size (>800kb or so) to the remote desktop by drag-and-drop of local files to the remote desktop:

https://github.com/jcaple/guacamole-webclient-test

https://github.com/jcaple/guacamole-server-test

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/BlobWriter.js
+++ b/guacamole-common-js/src/main/webapp/modules/BlobWriter.js
@@ -110,6 +110,10 @@ Guacamole.BlobWriter = function BlobWriter(stream) {
         var offset = 0;
         var reader = new FileReader();
 
+        var sleep = function sleep(ms) {
+            return new Promise(resolve => setTimeout(resolve, ms));
+        }
+
         /**
          * Reads the next chunk of the blob provided to
          * [sendBlob()]{@link Guacamole.BlobWriter#sendBlob}. The chunk itself
@@ -119,6 +123,12 @@ Guacamole.BlobWriter = function BlobWriter(stream) {
          * @private
          */
         var readNextChunk = function readNextChunk() {
+
+            while (reader.readyState == 1) {
+                // The reader is busy so give it time to finish before proceeding.
+                // 1 second might be a bit generous, but it seems functional.
+                await sleep(1000);
+            }
 
             // If no further chunks remain, inform of completion and stop
             if (offset >= blob.size) {


guacamole-common-js/src/main/webapp/modules/BlobWriter.js
```diff
@@ -120,6 +124,12 @@ Guacamole.BlobWriter = function BlobWriter(stream) {
          */
         var readNextChunk = function readNextChunk() {
 
+            while (reader.readyState == 1) {
+                // The reader is busy so give it time to finish before proceeding.
+                // 1 second might be a bit generous, but it seems functional.
+                await sleep(1000);
```
mike-jumper: We cannot safely use `await` at this level of the API. Doing so would require dropping support for IE10+ and would require some significant discussion. I believe the same goes for `Promise` and the arrow notation.

That said, this may not matter: I don't believe we should use this approach. It may technically work, but there has to be a better way than polling `readyState` every second.

Are there no callbacks, events, etc. that can be used here such that we can rely on a call to `readNextChunk()` when the `FileReader` is actually ready?

guacamole-common-js/src/main/webapp/modules/BlobWriter.js
```diff
@@ -120,6 +124,12 @@ Guacamole.BlobWriter = function BlobWriter(stream) {
          */
         var readNextChunk = function readNextChunk() {
 
+            while (reader.readyState == 1) {
```
mike-jumper: If such a check is needed, the operator of choice is `===`.

guacamole-common-js/src/main/webapp/modules/BlobWriter.js
```diff
@@ -110,6 +110,10 @@ Guacamole.BlobWriter = function BlobWriter(stream) {
         var offset = 0;
         var reader = new FileReader();
 
+        var sleep = function sleep(ms) {
```
mike-jumper: I expect this function will simply disappear (see below), but for reference: all functions need to be documented. In this case, this would mean a JSDoc block comment.

guacamole-common-js/src/main/webapp/modules/BlobWriter.js
```diff
@@ -110,6 +110,10 @@ Guacamole.BlobWriter = function BlobWriter(stream) {
...
+            return new Promise(resolve => setTimeout(resolve, ms));
+        }
```
mike-jumper: Beware you're missing a semicolon here. Not an issue if this function is ultimately deleted (see below).

PR #430, opened 2019-07-22
GUACAMOLE-221: Client-side changes for parameter prompting

necouchman:
This is a re-work of the client-side changes required to get parameter prompting going.  Very much a work-in-progress at this point, but I'd appreciate some pointers getting things finished up.  This matches with the changes in apache/guacamole-server#228.

As with the changes on the server-side, this does not handle multiple parameters concurrently, but does one at a time.  That can probably stand to be fixed, plus the style needs a little work.

### DIFF
--- a/guacamole/src/main/webapp/app/client/types/ManagedClient.js
+++ b/guacamole/src/main/webapp/app/client/types/ManagedClient.js
@@ -37,6 +37,7 @@ angular.module('client').factory('ManagedClient', ['$rootScope', '$injector',
 
     // Required services
     var $document               = $injector.get('$document');
+    var $log                    = $injector.get('$log');
     var $q                      = $injector.get('$q');
     var $rootScope              = $injector.get('$rootScope');
     var $window                 = $injector.get('$window');
@@ -51,6 +52,7 @@ angular.module('client').factory('ManagedClient', ['$rootScope', '$injector',
     var guacAudio               = $injector.get('guacAudio');
     var guacHistory             = $injector.get('guacHistory');
     var guacImage               = $injector.get('guacImage');
+    var guacPrompt              = $injector.get('guacPrompt');
     var guacVideo               = $injector.get('guacVideo');
 
     /**
@@ -367,6 +369,9 @@ angular.module('client').factory('ManagedClient', ['$rootScope', '$injector',
             client : client,
             tunnel : tunnel
         });
+        
+        // Parse connection details from ID
+        var clientIdentifier = ClientIdentifier.fromString(id);
 
         // Fire events for tunnel errors
         tunnel.onerror = function tunnelError(status) {
@@ -577,12 +582,27 @@ angular.module('client').factory('ManagedClient', ['$rootScope', '$injector',
                 managedClient.filesystems.push(ManagedFilesystem.getInstance(object, name));
             });
         };
+        
+        // Handle any received prompts
+        client.onrequired = function onrequired(parameters) {
+            
+            guacPrompt.getUserInput(parameters.reduce((a,b)=> (a[b]='',a),{}))
+                    .then(function gotUserInput(data) {
+                        for (var parameter in data) {
+                            var stream = client.createArgumentValueStream("text/plain", parameter);
+                            var writer = new Guacamole.StringWriter(stream);
+                            writer.sendText(data[parameter]);
+                            writer.sendEnd();
+                        }
+
+                }, function errorUserInput() {
+                    $log.error('Error gathering user input.');
+                    client.disconnect();
+                });
+        };
 
         // Manage the client display
         managedClient.managedDisplay = ManagedDisplay.getInstance(client.getDisplay());
-
-        // Parse connection details from ID
-        var clientIdentifier = ClientIdentifier.fromString(id);
 
         // Connect the Guacamole client
         getConnectString(clientIdentifier, connectionParameters)
--- a/guacamole-common-js/src/main/webapp/modules/Client.js
+++ b/guacamole-common-js/src/main/webapp/modules/Client.js
@@ -715,16 +715,16 @@ Guacamole.Client = function(tunnel) {
     this.onpipe = null;
     
     /**
-     * This method is fired when a prompt (required) instruction is received
-     * from the server, which indicates that guacd requires further information
+     * This method is fired when a required instruction is received from the
+     * server, which indicates that guacd requires additional parameters
      * from the client in order to continue the connection.
      * 
      * @event
-     * @param {String} parameter The name of a connection parameter that needs
+     * @param {String} parameters The names of connection parameters that need
      *      to be provided by the client to guacd before the connection can
      *      proceed.
      */
-    this.onprompt = null;
+    this.onrequired = null;
 
     /**
      * Fired whenever a sync instruction is received from the server, indicating
@@ -1350,8 +1350,8 @@ Guacamole.Client = function(tunnel) {
 
         },
                 
-        "required": function receivePrompt(parameters) {
-            if (guac_client.onprompt) guac_client.onprompt(parameters[0]);
+        "required": function required(parameters) {
+            if (guac_client.onrequired) guac_client.onrequired(parameters);
         },
         
         "reset": function(parameters) {
--- a/guacamole-common-js/src/main/webapp/modules/Client.js
+++ b/guacamole-common-js/src/main/webapp/modules/Client.js
@@ -713,6 +713,18 @@ Guacamole.Client = function(tunnel) {
      * @param {String} name The name of the pipe.
      */
     this.onpipe = null;
+    
+    /**
+     * This method is fired when a prompt (required) instruction is received
+     * from the server, which indicates that guacd requires further information
+     * from the client in order to continue the connection.
+     * 
+     * @event
+     * @param {String} parameter The name of a connection parameter that needs
+     *      to be provided by the client to guacd before the connection can
+     *      proceed.
+     */
+    this.onprompt = null;
 
     /**
      * Fired whenever a sync instruction is received from the server, indicating
@@ -1336,6 +1348,10 @@ Guacamole.Client = function(tunnel) {
 
             display.rect(layer, x, y, w, h);
 
+        },
+                
+        "required": function receivePrompt(parameters) {
+            if (guac_client.onprompt) guac_client.onprompt(parameters[0]);
         },
         
         "reset": function(parameters) {
--- a/guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolCapability.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolCapability.java
@@ -42,6 +42,15 @@ public enum GuacamoleProtocolCapability {
     PROTOCOL_VERSION_DETECTION(GuacamoleProtocolVersion.VERSION_1_1_0),
     
     /**
+     * Support for the "required" instruction.  The "required" instruction
+     * allows the server to request additional information from the client
+     * during the course of a connection.  Support for this instruction was
+     * introduced in
+     * {@link GuacamoleProtocolVersion#VERSION_1_3_0}.
+     */
+    REQUIRED_INSTRUCTION(GuacamoleProtocolVersion.VERSION_1_3_0),
+    
+    /**
      * Support for the "timezone" handshake instruction. The "timezone"
      * instruction allows the client to request that the server forward their
      * local timezone for use within the remote desktop session. Support for
--- a/guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java
@@ -45,12 +45,19 @@ public class GuacamoleProtocolVersion {
      * for passing the client timezone to the server during the handshake.
      */
     public static final GuacamoleProtocolVersion VERSION_1_1_0 = new GuacamoleProtocolVersion(1, 1, 0);
+    
+    /**
+     * Protocol version 1.3.0, which introduces support for the "required"
+     * instruction, allowing the server to request additional required
+     * information from the client in the course of a connection.
+     */
+    public static final GuacamoleProtocolVersion VERSION_1_3_0 = new GuacamoleProtocolVersion(1, 3, 0);
 
     /**
      * The most recent version of the Guacamole protocol at the time this
      * version of GuacamoleProtocolVersion was built.
      */
-    public static final GuacamoleProtocolVersion LATEST = VERSION_1_1_0;
+    public static final GuacamoleProtocolVersion LATEST = VERSION_1_3_0;
     
     /**
      * A regular expression that matches the VERSION_X_Y_Z pattern, where
--- a/guacamole/src/main/webapp/app/client/controllers/clientController.js
+++ b/guacamole/src/main/webapp/app/client/controllers/clientController.js
@@ -39,6 +39,7 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
     var dataSourceService      = $injector.get('dataSourceService');
     var guacClientManager      = $injector.get('guacClientManager');
     var guacNotification       = $injector.get('guacNotification');
+    var guacPrompt             = $injector.get('guacPrompt');
     var iconService            = $injector.get('iconService');
     var preferenceService      = $injector.get('preferenceService');
     var requestService         = $injector.get('requestService');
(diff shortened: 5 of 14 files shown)


docdawning:
Nice work @necouchman. My roll-out of guac has died on the hill of needing to prompt during RDP connection opening for a username + password (domain too, would be fine).

trengri:
Could you please advise how to use it?
I have applied this and apache/guacamole-server#228 patches in my test environment, but cannot get credentials prompt. Should I put some magic tokens to username and password fields?

necouchman:
@trengri 
A couple of things to consider:
1) This is a work in progress, and things are changing a bit in it at the moment.  Some things may work, others may not.
2) It works slightly different ways based on the protocol, but, essentially, each of the RDP, SSH, and VNC protocols have call-back functions that are invoked if authentication fails, and those functions are called within the code.  So, it'll try to authenticate, fail, and then *should* issue the credential prompt from there for the required credentials.
3) It definitely has some bugs.  I was actually just reviewing it the last couple of days to go back over it and try to give a go at fixing things up, again.

necouchman:
@trengri After reviewing to figure out where I left off, the prompts are getting successfully passed from guacd back to Guacamole Client, but are not being displayed properly at the moment.  I'm trying to sort it out...

necouchman:
@trengri This Pull Request is cleaned up, now, and is now working in my testing.  Feel free to re-pull, re-build, and give it another shot.

@mike-jumper  Ready for some more review when you have a free moment.

trengri:
@necouchman, thank you.
I checked it out and performed some tests.
Our Windows 10 desktops are configured to accept only NLA connections.
The first problem is that the prompt appears only when I explicitly set "Security mode" to NLA in the connection settings. If the "Security mode" is set to "Any" or unset (blank), then the password prompt does not appear - I immediately get "CONNECTION ERROR - The remote desktop server is currently unreachable". I expected that "Any" mode will allow me to connect, but OK. I set "Security mode" to NLA and started testing.

The second problem is more serious.
After I click on the session, the white message box "CONNECTING - Connected to Guacamole. Waiting for response..." appears. This message hands for a while (up to 5-10 seconds). Then credentials prompt appears.
As long as I'm typing my credentials, another message box "CONNECTION ERROR - The connection has been closed because the server is taking too long to respond" appears behind the credentials prompt. Of course, I can see only the part of it, because it is almost hidden by credentials window.
When I complete typing my credentials and press the "Connect" button, the credentials window disappears and I can fully see the window that was hidden behind it - it's still counting down the seconds to reconnect, as if the credentials I have entered had no effect. After the counter reaches zero, all the described sequence of events repeats.

A couple of times I was lucky enough to enter my credentials before that background window with countdown appears - and after that I was able to connect. But most of times, this countdown window appears long way before I finish entering my credentials.

And a small usability enhancement:
I noticed that if the username and domain fields are set in the connection properties, I'm prompted only for the password and not prompted for the username and domain. This might be expected behavior, but to my mind, it's better to prompt the user for all 3 fields (login, password, domain),  but pre-populate them in the dialog box, so the user can just leave them as is, or edit them (for example, if he wants to connect as another user). Right now, if you are not showing the username in the prompt at all, the user does not have an opportunity to login as another user.

mike-jumper:
> I noticed that if the username and domain fields are set in the connection properties, I'm prompted only for the password and not prompted for the username and domain. This might be expected behavior, but to my mind, it's better to prompt the user for all 3 fields (login, password, domain), but pre-populate them in the dialog box, so the user can just leave them as is, or edit them (for example, if he wants to connect as another user). Right now, if you are not showing the username in the prompt at all, the user does not have an opportunity to login as another user.

Re-showing the username, etc. if no username was originally part of the connection parameters could make sense.

The username definitely must _not_ be shown if the value is provided when the connection is started, as at that point the administrator has dictated the account to be used, and the user should not be able to override that.

necouchman:
@trengri @mike-jumper 
> The first problem is that the prompt appears only when I explicitly set "Security mode" to NLA in the connection settings. If the "Security mode" is set to "Any" or unset (blank), then the password prompt does not appear - I immediately get "CONNECTION ERROR - The remote desktop server is currently unreachable". I expected that "Any" mode will allow me to connect, but OK. I set "Security mode" to NLA and started testing.

Yes, this is due to how the NLA connection security is negotiated, along with how the back-end FreeRDP library handles negotiating security modes.  Basically, NLA requires a username and password to connect, so if that is not provided, and the security mode not explicitly given, then that mode will be discarded during negotiation.  So, if you need NLA and prompting for username/password/domain, you'll need to force the mode to NLA.

> After I click on the session, the white message box "CONNECTING - Connected to Guacamole. Waiting for response..." appears. This message hands for a while (up to 5-10 seconds). Then credentials prompt appears...

Yeah, I'm going to need some help (ahem, @mike-jumper ) on this - I'm basically not doing something correctly during the prompting process to "pause" the connection such that it is kept alive in the background while waiting on the credentials.  Again, this is still a work-in-progress.

>> I noticed that if the username and domain fields are set in the connection properties, I'm prompted only for the password and not prompted for the username and domain. This might be expected behavior, but to my mind, it's better to prompt the user for all 3 fields (login, password, domain), but pre-populate them in the dialog box, so the user can just leave them as is, or edit them (for example, if he wants to connect as another user). Right now, if you are not showing the username in the prompt at all, the user does not have an opportunity to login as another user.

> Re-showing the username, etc. if no username was originally part of the connection parameters could make sense.

> The username definitely must not be shown if the value is provided when the connection is started, as at that point the administrator has dictated the account to be used, and the user should not be able to override that.

I'll take a look at the logic and see if I can figure out how to handle this a little more cleanly, but, at present, it just looks for any "missing" (=blank) credentials and prompts for those.  If values are present, it will continue to use the configured values.  I suspect this behavior will stay largely the same (for the reasons Mike mentions), but I will definitely have to document it once it is ready.

trengri:
@necouchman Why do you try to authenticate, fail, then issue credential prompt?
Isn't it better to prompt for missing fields first, then authenticate? If I get it correctly, it will allow to try NLA when "Security mode" is set to "Any".
Also each failed login attempt leaves an unsuccessful authentication event in security logs (additional questions from security guys) and may provoke account lockout.

necouchman:
@trengri It is implemented this way because that's the way the FreeRDP API works - it tries authentication, first, and, if that fails, it issues a callback to an authentication function, if that function is defined.  See:

https://github.com/apache/guacamole-server/pull/228/files#diff-b9d603d00d2586ba7230ea4ba2267ff1L198-L232

and the original code...

https://github.com/apache/guacamole-server/blob/bbb794966bdebdea1c670062d573561c066a7504/src/protocols/rdp/rdp.c#L371
https://github.com/apache/guacamole-server/blob/bbb794966bdebdea1c670062d573561c066a7504/src/protocols/rdp/rdp.c#L198-L232

And, aside from authentication itself, it is part of the negotiation process for the security protocol, as mentioned before.  I'm not dead set against tweaking the behavior, but I believe the FreeRDP API is operating in the most compatible fashion to determine the security mechanisms supported by the server and the information required to make that authentication happen.  I'm just working within the API :-).  If there's a solid reason to modify this behavior, then we can do that, but I don't see that it's all that big a deal to set the Security Type to NLA for connections that are going to be prompting for credentials and where the server actually requires NLA??

trengri:
@necouchman, Sure, not a big deal. I was just curious why it is implemented this way, not more.
But one more idea came to my mind - if there are some difficulties with "pausing" connection, won't prompting credentials _before_ the connection help here?

necouchman:
@trengri:
> But one more idea came to my mind - if there are some difficulties with "pausing" connection, won't prompting credentials before the connection help here?

Maybe, but not necessarily - the issue isn't in pausing the connection between guacd and the RDP server, it's in pausing the connection between guacamole client and guacd, which will still be an issue no matter where in the connection process it gets paused.  Even if it's before the initial connection to the RDP server, I still have to figure out the proper way to have that session between guacamole client and guacd kept alive while waiting for the user.

mike-jumper:
> ... I still have to figure out the proper way to have that session between guacamole client and guacd kept alive while waiting for the user.

Aha! I can help here. For guacd --> webapp --> client, you need `guac_socket_require_keep_alive()`:

http://guacamole.apache.org/doc/libguac/socket_8h.html#a1d8c5111a8234c7785ae23f379f7db03

Invoking that function on the relevant `guac_socket` will result in an occasional `nop` instruction being sent from server to client every few seconds. The instruction is sent frequently enough that the connection is kept alive in the server -> client direction.

For client -> server, this should already be happening. The client implementation automatically sends `nop` every few seconds, as well.

necouchman:
Thanks @mike-jumper !  Change has been made to apache/guacamole-server#228 to include this in the `__guac_protocol_send_required()` function, so hopefully that does the trick!

ralfbecker:
It would be great if that makes it in the next release, as for some installations NLA is a requirement, which is hard to meet without storing clear-text passwords in the database (when authentication is done via OpenID Connect).

Ralf

necouchman:
@ralfbecker: This will not be in the next release, as the scope for 1.2.0 has been set, already, and this is not included.  Hopefully it'll follow pretty quickly after that, but for 1.2.0 we are focusing on fixing some of the issues introduced by 1.1.0 and the shift to FreeRDP 2, along with a handful of other functionality enhancements.

ralfbecker:
Thanks for the quick reply :)

Need to check how difficult / well documented building own containers with the above patch is.

Ralf

necouchman:
@ralfbecker: First, I wouldn't use these patches (you need both client and server ones) just yet, since it hasn't been fully reviewed or finalized.

Once it is reviewed and finalized we'll get it merged, and at that point it would be safe to try to patch an existing version of code (or just check out the version that has it).  These changes are going to be somewhat-invasive in terms of the amount of stuff they change, so I'm not sure how easy it's going to be to patch a previous version.

ralfbecker:
Thanks again.

We are currently integrating Guacamole in our WebApp to allow our users to make desktop computers available to to their employees now force tod work from home:

https://github.com/EGroupware/guacamole

Thanks again :)

Ralf

guacamole/src/main/webapp/app/client/types/ManagedClient.js
```diff
@@ -577,13 +582,28 @@ angular.module('client').factory('ManagedClient', ['$rootScope', '$injector',
...
+                            var stream = client.createArgumentValueStream("text/plain", parameter);
+                            var writer = new Guacamole.StringWriter(stream);
+                            writer.sendText(data[parameter]);
+                            writer.sendEnd();
+                        }
+
+                }, function errorUserInput() {
+                    $log.error('Error gathering user input.');
+                    client.disconnect();
+                });
```
knacktim, line 601: @necouchman In my initial testing I am getting a popup that looks like this:
image

Looks like we may be missing the client.forms for the content field in guacPrompt.html, where was that intended to be populated?

knacktim: Upon further testing this morning, this use case happens when the user clicks on a CONNECTION_GROUP to initiate a connection rather than a CONNECTION.

necouchman: Ah, yes - okay, good to know.  I'll have to figure out how to pull the forms for the underlying connection...

knacktim: I was having difficulty with that as well.  I couldn't find an easy way to identify the childrenConnections protocol.  The only way I saw was to just pull the protocol from the first child connection, which breaks pretty easily (especially since I think this is meant to be able to support connection-groups with sub-connection-groups.

Is there a reason the Connection-group is passed around as the selected object and then later assigned a specific connection?  Would there be a way get the 'assigned' connection first and then pass around that clientIdentifier?

necouchman: > Is there a reason the Connection-group is passed around as the selected object and then later assigned a specific connection? Would there be a way get the 'assigned' connection first and then pass around that clientIdentifier?

I think at least one of the reasons is that the user accessing the connection need not have any knowledge of or even access to the underlying connection when using a load balancing connection group.  That is, you can assign users access to the top-level connection group and not to anything else, and they'll still be able to connect.  The current behavior insures that this works, and I think what you're suggesting would require that the user accessing the connection group also be able to see/read the underlying connection.  But, I'm not sure if there's any good way around this.

@mike-jumper Could use your expertise, here :-).

knacktim: Will there ever be a case where a Connection-group has more than one protocol?

necouchman: Well, it is certainly possible today, yes, and I'm not sure that we should go the route of forcing things into that mold (one protocol per Connection Group) just yet.  While I do not have any scenarios in my current usage of Guacamole that require that, there may well be others that do.

knacktim:
@necouchman I rebased jira/221 locally on top of the 1.2.0 release.  I know that you guys have been actively changing the server portion of this.

I am now getting this error when trying to connect. Is the code base in a testable state or do I have a different rabbit hole to dive into? :)

```
guacd[15]: INFO:        Loading keymap "base"
guacd[15]: INFO:        Loading keymap "en-us-qwerty"
guacd[15]: DEBUG:        Support for CLIPRDR (clipboard redirection) registered. Awaiting channel connection.
guacd[15]: DEBUG:        Support for static channel "rdpdr" loaded.
guacd[15]: DEBUG:        Support for static channel "rdpsnd" loaded.
guacd[15]: DEBUG:        Local framebuffer format  PIXEL_FORMAT_BGRX32
guacd[15]: DEBUG:        Remote framebuffer format PIXEL_FORMAT_RGB16
guacd[15]: DEBUG:        creating directory /home/guacd/.config/freerdp
guacd[15]: DEBUG:        creating directory [/home/guacd/.config/freerdp/certs]
guacd[15]: DEBUG:        created directory [/home/guacd/.config/freerdp/server]
guacd[15]: DEBUG:        InitializeSecurityContextA status SEC_E_OUT_OF_SEQUENCE [0x80090310]
guacd[15]: DEBUG:        SPNEGO failed with NTSTATUS: 0x80090308
guacd[15]: DEBUG:        freerdp_set_last_error ERRCONNECT_AUTHENTICATION_FAILED [0x00020009]
guacd[15]: DEBUG:        rdp_recv_callback: CONNECTION_STATE_NLA - nla_recv_pdu() fail
guacd[15]: DEBUG:        transport_check_fds: transport->ReceiveCallback() - -1
guacd[15]: DEBUG:        SVC "rdpdr" disconnected.
guacd[15]: DEBUG:        SVC "rdpsnd" disconnected.
guacd[15]: ERROR:        Error connecting to RDP server
guacd[15]: INFO:        User "@9d3de152-bc69-40be-ab64-d4eb9b20db55" disconnected (0 users remain)
guacd[15]: INFO:        Last user of connection "$467a78c6-5064-4013-9d10-69f47c98112e" disconnected
guacd[15]: DEBUG:        Requesting termination of client...
guacd[15]: DEBUG:        Client terminated successfully.
guacd[1]: INFO:        Connection "$467a78c6-5064-4013-9d10-69f47c98112e" removed.
```

knacktim:
@necouchman Do we need to add:
```
    /**
     * Protocol version 1.3.0, which supports the "required" instruction,
     * allowing connections in guacd to request information from the client and
     * await a response.
     */
    public static final GuacamoleProtocolVersion VERSION_1_3_0 = new GuacamoleProtocolVersion(1, 3, 0);
```

to GuacamoleProtocolVersion.java for this?  (this is for my personal understanding for how all of this is working :) )

necouchman:
@knacktim Yeah, these changes are sort of mid-stream at this point.  My day job is killing me these days, but I'm trying to find some time to get back to this in the next few days or couple of weeks.

knacktim:
@necouchman Totally understand!  I am willing to try and contribute if you’d like help.  If so, do you have a TODO list tracking what else needs to be done?

necouchman:
@knacktim: Thanks.  I've added the new protocol version to this PR, so that part should be good, now.  Other than continued testing of the code, I don't know what to tell you at this point - I'm trying to get myself back up to speed with my changes and what needs to happen to clean it up.

knacktim:
> @knacktim: Thanks. I've added the new protocol version to this PR, so that part should be good, now. Other than continued testing of the code, I don't know what to tell you at this point - I'm trying to get myself back up to speed with my changes and what needs to happen to clean it up.

I pulled down the latest and everything looks good functionality-wise except for the following:

- Need to address the connection group not being aware what protocol form to show
- The prompt width is massive, could probably polish that up.

necouchman:
Thanks @knacktim.  I've been focusing on squaring away the guacd-side changes - will get back to the client changes for style and protocol, here, soon.

mike-jumper:
@necouchman, do you want me to re-review?

necouchman:
> @necouchman, do you want me to re-review?

@mike-jumper: Sure, that'd be great. I did start another branch in my fork of the tree to try to just roll the prompt changes into the `Notification` dialog and avoid a bunch of extra duplicate code, so I'm working through that right now. If you want to review this and make suggestions I can try to see if that will help me in my mortal combat against this issue :laughing:.

necouchman:
This was closed by #571

PR #431, opened 2019-07-26
GUACAMOLE-853: Add/Update of the german translations

dwydler:
I tested the new release (v 1.0.0) new days ago. I have see that that some strings are not translate or not correct. Additionally the translation for the extensions guacamole-auth-radius and guacamole-auth-jdbc are lacked.

### DIFF
--- a/extensions/guacamole-auth-radius/src/main/resources/guac-manifest.json
+++ b/extensions/guacamole-auth-radius/src/main/resources/guac-manifest.json
@@ -11,7 +11,8 @@
 
     "translations" : [
         "translations/en.json",
-        "translations/ja.json"
+        "translations/ja.json",
+        "translations/de.json"
     ],
 
     "js" : [
--- a/guacamole/src/main/webapp/translations/de.json
+++ b/guacamole/src/main/webapp/translations/de.json
@@ -443,7 +443,7 @@
         "FIELD_OPTION_SECURITY_ANY"   : "Jeder",
         "FIELD_OPTION_SECURITY_EMPTY" : "",
         "FIELD_OPTION_SECURITY_NLA"   : "NLA (Authentifizierung auf Netzwerkebene)",
-        "FIELD_OPTION_SECURITY_RDP"   : "RDP-Verschlüsselungn",
+        "FIELD_OPTION_SECURITY_RDP"   : "RDP-Verschlüsselung",
         "FIELD_OPTION_SECURITY_TLS"   : "TLS-Verschlüsselung",
 
         "FIELD_OPTION_SERVER_LAYOUT_DE_DE_QWERTZ" : "Deutsch (Qwertz)",
--- a/guacamole/src/main/webapp/translations/de.json
+++ b/guacamole/src/main/webapp/translations/de.json
@@ -4,12 +4,16 @@
     
     "APP" : {
 
+        
+        
+
         "ACTION_ACKNOWLEDGE"        : "OK",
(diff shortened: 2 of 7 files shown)


guacamole/src/main/webapp/translations/de.json
```diff
@@ -138,7 +153,7 @@
 
     "FORM" : {
 
-        "FIELD_PLACEHOLDER_DATE" : "YYYY-MM-DD",
```
mike-jumper: `FIELD_PLACEHOLDER_DATE` shows the format of date required by the application, which is always YYYY-MM-DD (ISO 8601). If the letters, etc. used within the date format need to be different for the language at hand, then it should be updated accordingly, but the format does still need to equate to YEAR-MONTH-DAY.

I'm guessing that the German version of the required format would be "JJJJ-MM-TT", but I don't honestly know.

guacamole/src/main/webapp/translations/de.json
```diff
@@ -4,103 +4,117 @@
...
+        "HELP_INPUT_METHOD_OSK"    : "Anzeigen und Akzeptieren von Eingaben über die eingebaute Guacamole-Bildschirmtastatur. Die Bildschirmtastatur ermöglicht die Eingabe von Tastenkombinationen, die ansonsten möglicherweise nicht möglich sind (z. B. Strg-Alt-Entf).",
+        "HELP_INPUT_METHOD_TEXT"   : "Ermöglicht die Eingabe von Text und emuliert Tastaturereignisse basierend auf dem eingegebenen Text. Dies ist für Geräte wie Mobiltelefone erforderlich, denen eine physische Tastatur fehlt.",
+        "HELP_MOUSE_MODE"          : "Legt fest, wie sich die Remote-Maus in Bezug auf Berührungen verhält.",
+        "HELP_MOUSE_MODE_ABSOLUTE" : "Tippe um zu klicken. Das Klicken erfolgt am Ort der Berührung.",
+        "HELP_MOUSE_MODE_RELATIVE" : "Ziehen Sie, um den Mauszeiger zu bewegen und tippen Sie, um zu klicken. Der Klick erfolgt an der Position des Mauszeigers.",
+        "HELP_SHARE_LINK"          : "Die aktuelle Verbindung wird gemeinsam genutzt und kann von jedem mit den folgenden {LINKS, Plural, einem {Link} anderen {Links} aufgerufen werden:",
```
mike-jumper: I don't believe this will work, as it's not correct ICU format for pluralization. Shouldn't this be `... mit {LINKS, plural, one{foldendem Link} other{den folgenden Links}} ...`?

extensions/guacamole-auth-radius/src/main/resources/guac-manifest.json
```diff
@@ -11,7 +11,8 @@
 
     "translations" : [
         "translations/en.json",
-        "translations/ja.json"
+        "translations/ja.json",
+        "translations/de.json"
```
mike-jumper: Please maintain alphabetical order of the filenames listed under `translations` for any particular `guac-manifest.json`.

dwydler:
Oh Shit, i uploaded some of my old translation versions for the files. Please drop the Pull Request.
I'am now on holiday for some weeks. After it i start a new attempt.

necouchman:
@dwydler: There is also an ongoing effort to update the German translation(s) as GUACAMOLE-699, including pull request #375.  I'd suggest taking a look at that PR and suggesting changes on that so that we're not duplicating efforts.

PR #433, opened 2019-08-03
GUACAMOLE-514: Client-side fields and translations for VNC authentication enhancements

necouchman:
This pull request implements the fields and translations for the VNC authentication enhancements introduced in apache/guacamole-server#232.

### DIFF
--- a/guacamole-ext/src/main/resources/org/apache/guacamole/protocols/vnc.json
+++ b/guacamole-ext/src/main/resources/org/apache/guacamole/protocols/vnc.json
@@ -20,8 +20,28 @@
             "name"  : "authentication",
             "fields" : [
                 {
+                    "name"  : "username",
+                    "type"  : "TEXT"
+                },
+                {
                     "name"  : "password",
                     "type"  : "PASSWORD"
+                },
+                {
+                    "name"  : "client-cert",
+                    "type"  : "MULTILINE"
+                },
+                {
+                    "name"  : "client-key",
+                    "type"  : "MULTILINE"
+                },
+                {
+                    "name"  : "ca-cert",
+                    "type"  : "MULTILINE"
+                },
+                {
+                    "name"  : "ca-crl",
+                    "type"  : "TEXT"
                 }
             ]
         },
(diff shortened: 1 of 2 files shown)


BongoEADGC6:
+1

LGTM

guacamole-ext/src/main/resources/org/apache/guacamole/protocols/vnc.json
```diff
@@ -19,6 +19,10 @@
         {
             "name"  : "authentication",
             "fields" : [
+                {
+                    "name"  : "username",
+                    "type"  : "TEXT"
```
mike-jumper: This should actually be `USERNAME`. The two fields are technically implemented identically, but `USERNAME` is defined with username-specific meaning.

See:

https://github.com/apache/guacamole-client/blob/339c920a89ba6d2363bd442d77490b00237ce650/guacamole-ext/src/main/resources/org/apache/guacamole/protocols/rdp.json#L22-L25

guacamole/src/main/webapp/translations/en.json
```diff
@@ -677,6 +677,7 @@
         "FIELD_HEADER_SFTP_ROOT_DIRECTORY"        : "File browser root directory:",
         "FIELD_HEADER_SFTP_USERNAME"              : "Username:",
         "FIELD_HEADER_SWAP_RED_BLUE"    : "Swap red/blue components:",
+        "FIELD_HEADER_USERNAME"     : "Username:",
```
mike-jumper: "Username" is common enough that there are likely translations in the other supported languages which can be safely copied without having to be a native speaker. Thoughts on adding those?

guacamole-ext/src/main/resources/org/apache/guacamole/protocols/vnc.json
necouchman: Ah, yes.  Fixed.

guacamole/src/main/webapp/translations/en.json
necouchman: Definitely - I've shamelessly copied the Telnet protocol version of the translation for each language to the VNC protocol.  Two things I noticed as I was doing this:
* There is some level of inconsistency within the translations as to whether they are organized alphabetically or logically.  I tried to be consistent within the inconsistency - but I can try to correct the inconsistency if you'd rather.
* The Italian translation looks like most of the items probably need to be removed - I don't think they're actually translated.  Again, I just copied it and carried on, but if it's a good time to strip out untranslated strings within that translation, I can go for that...

PR #435, opened 2019-08-04
GUACAMOLE-630: Allow parameters received via "argv" streams to be edited within the Guacamole menu.

mike-jumper:
This change leverages apache/guacamole-server#234 to automatically display a subset of connection parameters within the Guacamole menu:

guacamole-630-002-menu-before

Connection parameters are displayed only if each of the following is true:

1. The current value of the connection parameter was received via an inbound `argv` stream.
2. Creating an outbound `argv` stream for that same parameter resulted in a successful `ack`.

Changes to any parameters are applied once the menu is closed. After any parameter is changed, the above conditions are rechecked.

### DIFF
--- a/guacamole/src/main/webapp/app/client/templates/client.html
+++ b/guacamole/src/main/webapp/app/client/templates/client.html
@@ -96,6 +96,14 @@
                     </div>
                 </div>
 
+                <!-- Connection parameters which may be modified while the connection is open -->
+                <div class="menu-section connection-parameters" id="connection-settings" ng-show="client.protocol">
+                    <guac-form namespace="getProtocolNamespace(client.protocol)"
+                               content="client.forms"
+                               model="menu.connectionParameters"
+                               model-only="true"></guac-form>
+                </div>
+
                 <!-- Input method -->
                 <div class="menu-section" id="keyboard-settings">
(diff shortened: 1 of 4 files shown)


guacamole/src/main/webapp/app/client/templates/client.html
```diff
@@ -96,6 +96,15 @@ <h3>{{'CLIENT.SECTION_HEADER_DEVICES' | translate}}</h3>
                     </div>
                 </div>
 
+                <!-- Connection parameters which may be modified while the connection is open -->
+                <div class="menu-section connection-parameters" id="connection-settings" ng-show="client.protocol">
+                    <guac-form namespace="getProtocolNamespace(client.protocol)"
+                               content="client.forms"
+                               model="menu.connectionParameters"
+                               model-only="true"
+                               data-disabled="true"></guac-form>
```
jmuehlner: Why is this disabled?

mike-jumper: I ... don't know. I will check, but I'm pretty sure it shouldn't be. This might be from testing.

mike-jumper: OK. Should be fixed now. That was from an earlier version of the interface where I was toying with having the form disabled by default and definitely shouldn't have still been there.

PR #436, opened 2019-08-08
GUACAMOLE-630: Add field type for editing terminal color schemes.

mike-jumper:
This change adds a more user-friendly field type specific to the terminal color schemes used by Guacamole's terminal emulator. Initially, the field appears no different from a normal dropdown menu:

guacamole-630-001-color-scheme-dropdown

When "Custom..." is selected instead of a pre-defined color scheme, a palette editor appears below the dropdown:

guacamole-630-002-custom-color-scheme

Clicking any of the colors within the editor will pop up a color picker dialog to edit the selected color:

guacamole-630-003-edit-with-color-picker

If the user wishes to directly edit the color scheme as text, this can be done by clicking "Show" next to the "Details" header. Numbers will appear within the palette for reference:

guacamole-630-004-edit-directly

Changes to the text are reflected immediately within the palette editor, and changes within the palette editor immediately update the text.

For browsers that do not support the JavaScript features required by the color picker (Internet Explorer), the palette editor becomes read-only, and changes to the color scheme can be made only via text. The palette editor functions as a real-time preview in this case.


guacamole/src/main/webapp/app/form/directives/guacInputColor.js
```diff
@@ -0,0 +1,204 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/* global Pickr */
```
jmuehlner: What is this?

guacamole/src/main/webapp/app/form/types/ColorScheme.js
```diff
@@ -0,0 +1,949 @@
...
+        'yellow1'               : '#FFFF00',
+        'yellow2'               : '#EEEE00',
+        'yellow3'               : '#CDCD00',
+        'yellow4'               : '#8B8B00',
+        'yellowgreen'           : '#9ACD32'
+    };
+
+    return ColorScheme;
+
+}]);
```
jmuehlner: No newline at the end of the file? Do we have a standard around this?

guacamole/src/main/webapp/app/form/directives/guacInputColor.js
mike-jumper: It's an annotation to advise the IDE that `Pickr` is a legitimate global and not something to warn about.

guacamole/src/main/webapp/app/form/types/ColorScheme.js
mike-jumper: We don't, but I generally try to always have a newline. Will add.

mike-jumper: OK. Added.

guacamole/src/main/webapp/app/form/directives/guacInputColor.js
jmuehlner: Do you think that's worth documenting?

mike-jumper: The annotation or that `Pickr` is coming from an external library?

The `/* global ... */` syntax is relatively standard, recognized by jslint and eslint:

* https://www.jslint.com/help.html#global
* https://eslint.org/docs/rules/no-undef#rule-details

jmuehlner: Both I guess? Just mostly a comment about why that line is there, since it wasn't super obvious (to me anyway).

mike-jumper: Eh ... I'll just remove the annotation. If we move to using eslint or similar during build or CI at some point, we can then address if/how such annotations are added.

PR #437, opened 2019-08-10
GUACAMOLE-769: Polish the RADIUS challenge box

necouchman:
This cleans up the RADIUS challenge box a bit, such that the challenge message sent by the RADIUS server appears above the box, and the headers on the message are removed.


extensions/guacamole-auth-radius/src/main/resources/templates/radiusResponseField.html
```diff
@@ -1 +1,6 @@
-<input type="password" ng-model="model" ng-trim="false" autocorrect="off" autocapitalize="off" placeholder="{{radiusPlaceholder}}" />
+<input
+    type="password"
```
mike-jumper: If moving to a simple `<input type="password" ...>`, any reason to continue using the custom `RadiusChallengeResponseField` instead of the standard `PasswordField`?

necouchman: Yeah, that makes sense.  Removed.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/AuthenticationProviderService.java
```diff
@@ -53,6 +54,12 @@
      * Logger for this class.
      */
     private final Logger logger = LoggerFactory.getLogger(AuthenticationProviderService.class);
+    
+    /**
+     * The name of the password field for the user to response to the RADIUS
+     * challenge.
+     */
+    private static final String CHALLENGE_RESPONSE_PARAM = "radisuChallenge";
```
mike-jumper: radius*

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/AuthenticationProviderService.java
```diff
@@ -97,13 +106,17 @@ private CredentialsInfo getRadiusChallenge(RadiusPacket challengePacket) {
         }
 
         // We have the required attributes - convert to strings and then generate the additional login box/field
-        String replyMsg = replyAttr.toString();
+        logger.debug("Received challenge: {}", replyAttr.toString());
+        String replyMsg = replyAttr.toString().split(" = ", 2)[1];
```
mike-jumper: Would `replyAttr.getValue()` work here? Formatting the `RadiusAttribute` as a string and then re-parsing it back into an attribute feels like it shouldn't be necessary.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/AuthenticationProviderService.java
```diff
@@ -67,18 +74,20 @@
...
      *     server.
      *
      * @return
-     *     A CredentialsInfo object that represents fields that need to
-     *     be presented to the user in order to complete authentication.
-     *     One of these must be the RADIUS state.
+     *     A GuacamoleRadiusChallenge object that contains the challenge message
+     *     sent by the RADIUS server and the expected credentials that should
+     *     be requested of the user in order to continue authentication.  One
+     *     of the expected credentials *must* be the RADIUS state.
```
mike-jumper: I see further down that `getRadiusChallenge()` is expected to return `null` under some circumstances. If that is indeed the case, then the circumstances for returning `null` should be documented here.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/AuthenticationProviderService.java
```diff
@@ -53,6 +54,12 @@
      * Logger for this class.
      */
     private final Logger logger = LoggerFactory.getLogger(AuthenticationProviderService.class);
+    
+    /**
+     * The name of the password field for the user to response to the RADIUS
```
mike-jumper: > The name of the password field for the user to response to the RADIUS challenge.

I was initially going to say that "response" should be "respond" here, but with that change the sentence still reads a bit off ("the password field for the user to respond").

Did you mean the password field that will contain the user's response?

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/form/GuacamoleRadiusChallenge.java
```diff
@@ -0,0 +1,77 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.guacamole.auth.radius.form;
+
+import org.apache.guacamole.net.auth.credentials.CredentialsInfo;
+
+/**
+ * A class for storing the RADIUS challenge message and expected credentials
```
mike-jumper: I'm OK with keeping this as is, but in case you feel inspired to change it:

If all our documentation for all Java classes started with "A class for ...", things would be pretty redundant. If there's a way to rephrase this such that this is avoided, that would be better.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/form/GuacamoleRadiusChallenge.java
```diff
@@ -0,0 +1,77 @@
...
+    private final String challengeText;
+    
+    /**
+     * The expected credentials that need to be provided to satisfy the
+     * RADIUS authentication challenge.
+     */
+    private final CredentialsInfo expectedCredentials;
+    
+    /**
+     * Create a new GuacamoleRadiusChallenge object with the provided
```
mike-jumper: Convention for the first part of a function's JavaDoc documentation is a present tense sentence fragment - the portion that would follow the implied "This function" beginning of that sentence. To follow that convention, you'll want "Creates a new ..." here, not "Create a new ...". The latter is more the style of git commit messages. ;)

Same with the "Return the ..." phrasing below.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/AuthenticationProviderService.java
necouchman: Well, two things, here:
* In order to get the "Reply-Message = " out of here, I need to split it.  Not sure how feasible this is with an `AttributeValue` type?
* It ends up getting passed through the form on the Web side.  Is this going to play okay with using `AttributeValue` vs. `String`?

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/AuthenticationProviderService.java
necouchman: Fixed.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/AuthenticationProviderService.java
necouchman: Null documented.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/AuthenticationProviderService.java
necouchman: Fixed this up.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/form/GuacamoleRadiusChallenge.java
necouchman: I felt inspired.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/form/GuacamoleRadiusChallenge.java
necouchman: Present tensed.

necouchman:
@mike-jumper: Okay, cleaned up most of those issues, just the one question on the issue above.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/AuthenticationProviderService.java
mike-jumper: Checking the source of `RadiusAttribute`, [the `toString()` function](https://github.com/coova/jradius/blob/master/core/src/main/java/net/jradius/packet/attribute/RadiusAttribute.java#L155-L158) says:

```java
public String toString()
{
    return attributeName + " " + Operator.operatorToString(attributeOp) + " " + attributeValue.toString();
}
```

with `attributeValue` being exposed through [`getValue()`](https://github.com/coova/jradius/blob/4ee91e33fa155843fe428559d81f036051d30515/core/src/main/java/net/jradius/packet/attribute/RadiusAttribute.java#L94-L97):

```java
public AttributeValue getValue()
{
    return attributeValue;
}
```

From the above, it looks like calling `replyAttr.toString()` and splitting that to obtain the value is equivalent to `replyAttr.getValue().toString()` (the value used in the `logger.debug()` call), with the latter avoiding the back-and-forth conversion and assumption of the contents of `attributeOp`.

Is this not the case?

necouchman: Ah, okay, i see.  I'll give it a shot.

PR #439, opened 2019-08-12
GUACAMOLE-234: Fix resource leaks in new LDAP code.

necouchman:
I believe I've got the two resource leak issues resolved from the new LDAP code...

### DIFF
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ObjectQueryService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ObjectQueryService.java
@@ -20,6 +20,7 @@
 package org.apache.guacamole.auth.ldap;
 
 import com.google.inject.Inject;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ObjectQueryService.java
```diff
@@ -223,10 +224,11 @@ else if (results.isReferral() && request.isFollowReferrals()) {
                 
             }
 
+            results.close();
```
mike-jumper: Will this still leak if an exception is thrown above?

mike-jumper: [`try`-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) might be an option.

necouchman: Okay, gave this another shot with a `try...finally` block.  I don't like the nested `try` statements - I could declare the `results` variable outside of the first `try` block, instead, if you think that would be a better route?

mike-jumper: Well, I think you can likely avoid all that with Java 8 fanciness (`try`-with-resources). Something like:

```java
// Search within subtree of given base DN
LdapConnectionConfig ldapConnectionConfig = ldapConnection.getConfig();
SearchRequest request = ldapService.getSearchRequest(baseDN, query);

try (SearchCursor results = ldapConnection.search(request)) {
    ...
}
catch (CursorException | IOException | LdapException e) {
    ...
}
```

mike-jumper: In fact ... I think the same implicit `close()` from `try`-with-resources would work with `LdapNetworkConnection`, as well. It implements `AutoCloseable` just as `SearchCursor` does.

necouchman: Cool - I'll rework it, give me 5 minutes :smile:

necouchman:
@mike-jumper: Okay, reworked both those instances with `try...finally` blocks - let me know if that looks okay.  It'll change the error reporting in the method from `LDAPConnectionService` that was changed, but I think it's still acceptable.

PR #440, opened 2019-08-13
GUACAMOLE-774: Fix deprecation issue with Java 9 and up.

necouchman:
This is probably not the best solution, but I'm having a hard time finding another.  The only `Provider` constructor present in Java 8 and lower is:

    Provider(String name, double version, String info)

In Java 9, that constructor is deprecated, and the recommended one is:

    Provider​(String name, String versionStr, String info)

Since the code introduced with that constructor is designed to handle MD4 requirements in *newer* versions of Java, it needs to work with newer versions.  Unless I'm missing some sort of way in Java you can say..."if running with Java version X and earlier, do this, else do this?"

### DIFF
--- a/extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/RadiusAuthenticationProviderModule.java
+++ b/extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/RadiusAuthenticationProviderModule.java
@@ -20,9 +20,12 @@
 package org.apache.guacamole.auth.radius;
 
 import com.google.inject.AbstractModule;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
 import java.security.Provider;
 import java.security.Security;
 import org.apache.guacamole.GuacamoleException;
+import org.apache.guacamole.GuacamoleServerException;
 import org.apache.guacamole.auth.radius.conf.ConfigurationService;
 import org.apache.guacamole.auth.radius.conf.RadiusAuthenticationProtocol;
 import org.apache.guacamole.auth.radius.conf.RadiusGuacamoleProperties;
@@ -57,7 +60,6 @@ public class RadiusAuthenticationProviderModule extends AbstractModule {
      *     If an error occurs while retrieving the Guacamole server
      *     environment.
      */
-    @SuppressWarnings("deprecation")
     public RadiusAuthenticationProviderModule(AuthenticationProvider authProvider)
             throws GuacamoleException {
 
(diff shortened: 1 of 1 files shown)


mike-jumper:
The only way I can think of would be reflection. For example:

```java
Constructor providerConstructor = Provider.class.getConstructor(String.class, String.class, String.class);
```

and then call it if found, otherwise resorting to calling the older version.

The downside to just suppressing the warning would be the eventual build failure when the deprecated function is removed, which would then require either doing the above or removing support for Java 8.

necouchman:
Yeah, I'd much rather avoid suppressing warnings and go with a solution that handles the cases.  I'll try reflection....

necouchman:
Okay, stab taken at using reflection...

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/RadiusAuthenticationProviderModule.java
```diff
@@ -71,11 +74,28 @@ public RadiusAuthenticationProviderModule(AuthenticationProvider authProvider)
...
-                    this.put("MessageDigest.MD4", org.bouncycastle.jce.provider.JDKMessageDigest.MD4.class.getName());
-                }
-            });
+            try {
+                Provider md4Provider;
+                Constructor providerConstructor = Provider.class
+                        .getConstructor(String.class, String.class, String.class);
+                if (providerConstructor != null)
+                    md4Provider = (Provider) providerConstructor
+                            .newInstance("MD4", "0.00", "MD4 for MSCHAPv1/2 Support");
```
mike-jumper: I haven't checked myself, but is "0.00" the correct format for the version string expected by the new constructor?

necouchman: I'm not sure - details on that seem to be few and far between.  But maybe there's an easier way than this, even...
```
try {
    MessageDigest.getInstance("MD4");
}
catch (NoSuchAlgorithmException e)
{
    Security.addProvider(new BouncyCastleProvider());
}
```

mike-jumper: If the above works, that looks so much better.

necouchman: Yeah...problem is, I don't have much of a way to test it :-).  I can reach out to the folks who reported the issue and see if they're able to test it.

necouchman: Well, I implemented the above code, and verified that the module loads successfully when configured for `mschapv1` or `mschapv2`, which should indicate that it's able to load the MD4 code.  I'm reasonably certain that MD4 support was removed in Java 1.8, which is what I'm running, so I'm pretty sure this is working as intended.  Just hard to say for sure.

drhy-nz:
Tested https://github.com/necouchman/guacamole-client/tree/jira/774 against Windows Server 2019 Network Policy Server (NPS) by compiling all of 774 on my separate maven VM. The NPS had all possible connection types checked, including pap, chap, mschap and mschapv2.
guacamole774 was running on CentOS 7, Tomcat 8.5, and Java 8.

Tried all "radius-auth-protocol" options in the guacamole.properties file, including all options for eap-ttls.
The only options that worked were pap and mschapv2.
Haven't tested with tomcat 9 and java 9.

-David

PR #446, opened 2019-09-30
GUACAMOLE-884: Leverage createImageBitmap() for reading image data where supported.

mike-jumper:
Some browsers suffer from a memory leak when reading image data repeatedly using the `Image` object. Reading from `Blob` objects does not exhibit the same behavior. While reading from `Blob` objects has previously been seen to perform poorly compared to data URIs, this was observed when reading using `createObjectURL()`. The `createImageBitmap()` function appears to perform identically to reading data URIs using `Image`.

This change:

* Updates `drawBlob()` to leverage `createImageBitmap()` where available.
* Adds a new `drawStream()` function to `Guacamole.Display` which automatically selects between reading from the given stream using data URIs or `Blob` objects depending on what the platform best supports.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Display.js
+++ b/guacamole-common-js/src/main/webapp/modules/Display.js
@@ -512,11 +512,17 @@ Guacamole.Display = function() {
      * Draws the specified image at the given coordinates. The image specified
      * must already be loaded.
      * 
-     * @param {Guacamole.Layer} layer The layer to draw upon.
-     * @param {Number} x The destination X coordinate.
-     * @param {Number} y The destination Y coordinate.
-     * @param {Image} image The image to draw. Note that this is an Image
-     *                      object - not a URL.
+     * @param {Guacamole.Layer} layer
+     *     The layer to draw upon.
+     *
+     * @param {Number} x
+     *     The destination X coordinate.
+     *
+     * @param {Number} y 
+     *     The destination Y coordinate.
+     *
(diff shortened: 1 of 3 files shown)


guacamole-common-js/src/main/webapp/modules/Display.js
```diff
@@ -543,26 +549,97 @@ Guacamole.Display = function() {
...
+                guac_display.drawBlob(layer, x, y, reader.getBlob());
+            };
+        }
+
+        // Lacking createImageBitmap(), fall back to data URIs and the Image
+        // object
+        else {
+            var reader = new Guacamole.DataURIReader(stream, mimetype);
+            reader.onend = function drawImageDataURI() {
+                guac_display.draw(layer, x, y, reader.getURI());
```
necouchman: Is `guac_display.draw()` correct here, or should it be `guac_display.drawImage()`?

mike-jumper: Yep, `draw()` is correct here as we are using a data URI.

`drawImage()` takes an `Image` (though the documentation is now updated to acknowledge it can take anything considered a `CanvasImageSource`). The `draw()` function is the one that accepts a URI.

See:

* http://guacamole.apache.org/doc/guacamole-common-js/Guacamole.Display.html#draw
* http://guacamole.apache.org/doc/guacamole-common-js/Guacamole.Display.html#drawImage

PR #448, opened 2019-10-08
GUACAOMLE-893: Fix regression in LDAP filter with null value

necouchman:
This fixes a regression introduced in GUACAMOLE-234 where undefined attribute values are passed through rather than being checked and changed to wildcards, as was the behavior prior to swapping over to the Apache LDAP API.

I've requested this go into the staging/1.1.0 branch since it is a regression introduced by changes in that branch.

### DIFF
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ObjectQueryService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ObjectQueryService.java
@@ -142,16 +142,21 @@ public class ObjectQueryService {
         AndNode searchFilter = new AndNode();
         searchFilter.addNode(filter);
 
-        // Include all attributes within OR clause if there are more than one
+        // If no attributes provided, we're done.
+        if (attributes.size() < 1)
+            return searchFilter;
+
+        // Include all attributes within OR clause
         OrNode attributeFilter = new OrNode();
-       
+
         // Add equality comparison for each possible attribute
         attributes.forEach(attribute ->
-            attributeFilter.addNode(new EqualityNode(attribute, attributeValue))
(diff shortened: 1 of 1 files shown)


extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ObjectQueryService.java
```diff
@@ -141,16 +141,21 @@ public ExprNode generateQuery(ExprNode filter,
         // the given search filter
         AndNode searchFilter = new AndNode();
         searchFilter.addNode(filter);
-
-        // Include all attributes within OR clause if there are more than one
-        OrNode attributeFilter = new OrNode();
+        
+        // Add in attribute list
+        if (attributes.size() > 0) {
```
mike-jumper: If we should be adding nothing in the case where there are no attributes, would it make more sense to simply `return filter` if no attributes are given?

mike-jumper:
Yep, definitely should fix any known regressions prior to release.

Also, beware that your commit message says "guacaomle". ;)

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ObjectQueryService.java
necouchman: Yep, that's fair - implemented as suggested.

necouchman:
> Yep, definitely should fix any known regressions prior to release.
> Also, beware that your commit message says "guacaomle". ;)

Fixed the commit message - thanks!

necouchman:
...and one more commit to fix an issue with testing attribute presence (`attribute=*`) vs. equality (`attribute=value`).

necouchman:
This has been confirmed by the requester on the JIRA issue to resolve the problem.

PR #449, opened 2019-10-12
Modify the start time of the history session to match the GUAC_DATE and GUAC_TIME in the parameter tokens

trgbishi:
Modify the start time of the history session to match the GUAC_DATE and GUAC_TIME in the parameter tokens. This historical session can correspond to screen recording files using GUAC_DATE and GUAC_TIME.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,6 +45,9 @@ COPY guacamole-docker/bin/ /opt/guacamole/bin/
 COPY . "$BUILD_DIR"
 
 # Run the build itself
+RUN ["chmod", "777", "/opt/guacamole/bin/build-guacamole.sh"]
+RUN ["chmod", "777", "/opt/guacamole/bin/start.sh"]
+
 RUN /opt/guacamole/bin/build-guacamole.sh "$BUILD_DIR" /opt/guacamole "$BUILD_PROFILE"
 
 # For the runtime image, we start with the official Tomcat distribution
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
@@ -21,6 +21,9 @@ package org.apache.guacamole.auth.jdbc.tunnel;
 
 import com.google.inject.Inject;
 import com.google.inject.Provider;
+
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -119,7 +122,7 @@ public abstract class AbstractGuacamoleTunnelService implements GuacamoleTunnelS
      */
     private final Map<String, ActiveConnectionRecord> activeTunnels =
             new ConcurrentHashMap<String, ActiveConnectionRecord>();
-    
+
     /**
      * All active connections to a connection having a given identifier.
      */
@@ -245,7 +248,7 @@ public abstract class AbstractGuacamoleTunnelService implements GuacamoleTunnelS
             config.setParameter(parameter.getName(), parameter.getValue());
 
         return config;
-        
+
     }
 
     /**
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -45,6 +45,9 @@ COPY guacamole-docker/bin/ /opt/guacamole/bin/
 COPY . "$BUILD_DIR"
 
 # Run the build itself
+RUN ["chmod", "777", "/opt/guacamole/bin/build-guacamole.sh"]
+RUN ["chmod", "777", "/opt/guacamole/bin/start.sh"]
```
necouchman: These changes seem out-of-scope for this particular pull request.  If the pull request is focused on modifying how the GUAC_DATE and GUAC_TIME tokens behave, then it should be kept to that, without these extra changes.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
```diff
@@ -392,7 +395,29 @@ private GuacamoleSocket getUnconfiguredGuacamoleSocket(
...
-        
+
+
+
+        /***
+         * @Author zero
+         * @Description 将活动会话记录里的 会话开始时间变更为GUAC_DATE,GUAC_TIME
+         * @InitDate 10:53 2019/10/12
+         * @Param [activeConnection, tokens]
+         **/
```
necouchman: The javadoc comments should follow the format of the other comments throughout the code.  Please be consistent with the style used in the code.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
```diff
@@ -392,7 +395,29 @@ private GuacamoleSocket getUnconfiguredGuacamoleSocket(
...
+        public ConnectionCleanupTask(ActiveConnectionRecord activeConnection,Map<String ,String > tokens) {
+            this.activeConnection = activeConnection;
+
+            String yearMonthDay = tokens.get("GUAC_DATE");
+            String hourMinSecMs = tokens.get("GUAC_TIME");
+            SimpleDateFormat simpleDateFormat=new SimpleDateFormat("yyyyMMddHHmmss-SSS");
```
necouchman: I'm not sure about using the token to set the `ActiveConnectionRecord` date - in mind, this should actually work the other way around - the token should (somehow) be set using the date generated for the `ActiveConnectionRecord`, making the connection record the reference point and not the other way around.

@mike-jumper @ceharris @jmuehlner Any opinions on this??

guacamole/src/main/java/org/apache/guacamole/tunnel/StandardTokenMap.java
```diff
@@ -73,7 +73,7 @@
      * The date format that should be used for the time token. This format must
      * be compatible with Java's SimpleDateFormat.
      */
-    private static final String TIME_FORMAT = "HHmmss";
+    private static final String TIME_FORMAT = "HHmmss-SSS";
```
necouchman: Why is the format being changed?

Dockerfile
mike-jumper: These changes would also add far more permissions to the scripts than they should be given. The build and startup scripts definitely do not need to be world-writable.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
mike-jumper: Documentation will also need to be in English.

guacamole/src/main/java/org/apache/guacamole/tunnel/StandardTokenMap.java
mike-jumper: I agree that this change need to be justified.

Assuming we do want such a token, it will probably need to be a different token. Modifying a well-established token to use a different format will result in breakage in downstream systems that rely on the format of the old token.

trgbishi:
I just provided a solution.
I am not familiar with the source code, and assigning tokens to ActiveConnectionRecord is the fastest way to solve the problem.
Sorry for the document format not being checked carefully.

PR #450, opened 2019-10-18
GUACOMOLE-300: Support posixGroup in LDAP Authentication and Group-based Session Admission

mlewissmith:
This supersedes PR #422 
(NB it also supersedes long inactive #245 )


necouchman:
@mlewissmith: Looks much better.  A couple of things off the bat:
* Several of your commits lack the "GUACAMOLE-300" prefix - all commits should have it.
* There are a couple of rebranch and merge commits.  These should be eliminated.
* Probably worth squashing a bunch of the commits into a smaller number of them, as there are quite a few commits for the relatively small amount of code changing.
* Please double-check style - overall it looked okay, but there were a couple of places where you used cuddled `if...else` statements (`} else {`).

mlewissmith:
@necouchman Implemented suggestions

PR #452, opened 2019-10-23
GUACAMOLE-234: Make LDAP user and group identifiers lowercase.

necouchman:
Not sure this is the right place/time to do this, or if it merits its own more comprehensive JIRA issue, but case sensitivity seems to cause some issues/confusion in the link between LDAP and JDBC modules.  Since LDAP is, in general, case insensitive, one possible solution is to push the LDAP username and group identifiers all to lower-case.

### DIFF
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
@@ -132,7 +132,8 @@ public class UserGroupService {
 
             // Translate entry into UserGroup object having proper identifier
             try {
-                String name = queryService.getIdentifier(entry, attributes);
+                String name =
+                        queryService.getIdentifier(entry, attributes).toLowerCase();
                 if (name != null)
                     return new SimpleUserGroup(name);
             }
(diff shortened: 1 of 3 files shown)


mike-jumper:
First, as you note, I'm not sure this is the right issue for this. Is there a reason why this should be part of [GUACAMOLE-234](https://issues.apache.org/jira/browse/GUACAMOLE-234)?

As for the implementation itself, I think this may be problematic:

* LDAP DNs may be case-insensitive by default, but I don't believe this is required nor guaranteed. If the LDAP implementation is configured to be case-sensitive, I think we should honor that.
* It is not guaranteed that the database will be case-sensitive (I don't believe MySQL is case sensitive), nor that all usernames within a case-sensitive database will be lowercase.

I think it boils down to the overall definition of identity as dictated by the administrator/system. Guacamole may be deployed in environments where usernames are case-sensitive or case-insensitive, and Guacamole's definition of identity should be configurable to the extent that it can be compatible with either.

Off the top of my head, one possibility might be to provide an option within the JDBC auth which defines whether usernames are case-sensitive. Another might be to provide a global option at the `Environment` level.

necouchman:
Yes, probably merrits its own issue/discussion.  I do like the idea of making it configurable.  I'll take the discussion to the mailing list and close this PR out.

PR #453, opened 2019-11-29
GUACAMOLE-903: Improved Chinese internationalization support

XYUU:


necouchman:
@XYUU: Thank you for the contribution to improving the support for non-English languages!

Before we go further with this, you need to do a couple of things:
1) Open a JIRA issue for these changes.  All changes to the code must have an associated JIRA issue.
2) Once the JIRA issue is open, tag both the pull request and the git commit messages with the JIRA issue.  Please have a look at other pull requests to see how this is done.
3) Make the commit message(s) a little more descriptive of the changes being done.  The commit message should tell us something meaningful about the changes - more details than just the title of the pull request, but not so detailed as to describe the actual code changes.

XYUU:
Okay, I did.

necouchman:
> Okay, I did.

@XYUU: Almost :-).  The commit messages need to be edited and tagged with this same "GUACAMOLE-903:" JIRA issue.

guacamole/src/main/webapp/app/form/controllers/languageFieldController.js
```diff
@@ -40,7 +40,7 @@ angular.module('form').controller('languageFieldController', ['$scope', '$inject
 
     // Retrieve defined languages
     languageService.getLanguages().then(function languagesRetrieved(languages) {
-        $scope.$apply(function updateLanguageOptions() {
+        $scope.$applyAsync(function updateLanguageOptions() {
```
necouchman: I'm not really sure this change should be, here.  We can talk about changing this; however, I believe it is out of scope for a set of changes targeted at improving Chinese language support.

XYUU: This is a bug that I fixed that would cause an existing script error, I fixed it, and if I don't modify it here, my commit may not work.

I did a simple test, it was fixed, and it worked fine.

Unfortunately I don't have the ability to write unit tests for this modification at this time because it is beyond my ability to do so.

XYUU: I have explained the reason for this change in jira.

I found that there is a script error in the front end when setting the default internationalization. This error is caused by an ajax language available in angular. I checked the information and fixed the problem.

mike-jumper: > I have explained the reason for this change in JIRA.

Not yet - the JIRA issue currently says "fixed ... replace "$apply" to "$applyAsync"" and links to this commit. Please clarify on the issue, when you can, the actual underlying nature of the issue being fixed.

> I found that there is a script error in the front end when setting the default internationalization. This error is caused by the wrong use of angular ajax request.

`$apply()` and `$applyAsync()` do not deal with AJAX, nor is AJAX involved with the internationalization system. The error in question (infinite digest) results from `$apply()` being used within a digest cycle.

I see the same errors you do, but please describe on the issue the nature of the issue being fixed (why current usage of `$apply()` leads to an infinite digest error). The reason that `$apply()` is incorrect here (ie: what circumstances would cause this particular call to be invoked within an active digest) needs to be known.

Splitting this specific fix into its own pull request (rather than piggybacking it on GUACAMOLE-903) should also be done.

> Unfortunately I don't have the ability to write unit tests for this modification at this time because it is beyond my ability to do so.

I'm not sure what unit test would need to be written for this, but don't worry - there's no need in this case. All that is needed is (1) to split this change such that it only covers the scope of the specific issue and (2) to document the nature of the issue in JIRA.

XYUU: okay!I'll do as you say. I will open another PR after this PR merge to submit this bug fix.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/resources/translations/zh.json
```diff
@@ -0,0 +1,113 @@
...
+
+    "PASSWORD_POLICY" : {
+
+        "ERROR_CONTAINS_USERNAME"      : "密码可能不包含用户名。",
+        "ERROR_REQUIRES_DIGIT"         : "密码必须至少包含一位数字。",
+        "ERROR_REQUIRES_MULTIPLE_CASE" : "密码必须同时包含大写和小写字符。",
+        "ERROR_REQUIRES_NON_ALNUM"     : "密码必须包含至少一个符号。",
+        "ERROR_REUSED"                 : "此密码已被使用。请勿重复使用以前的{HISTORY_SIZE} {HISTORY_SIZE, plural, one{password} other{passwords}}。",
+        "ERROR_TOO_SHORT"              : "密码必须至少{LENGTH} {LENGTH, plural, one{character} other{characters}} 位长度。",
+        "ERROR_TOO_YOUNG"              : "此帐户的密码已被重设。请至少再等待 {WAIT}个{WAIT, plural, one{day} other{days}}，然后再次更改密码。"
```
mike-jumper: These pluralization substitutions will result in English plurals being included within Chinese text. Perhaps these should be translated, as well?

guacamole/src/main/webapp/translations/zh.json
```diff
@@ -430,30 +572,38 @@
...
         "FIELD_HEADER_RECORDING_EXCLUDE_MOUSE"  : "排除鼠标：",
         "FIELD_HEADER_RECORDING_EXCLUDE_OUTPUT" : "排除图像/数据流：",
         "FIELD_HEADER_RECORDING_INCLUDE_KEYS"   : "包含按键事件：",
         "FIELD_HEADER_RECORDING_NAME" : "录像名：",
         "FIELD_HEADER_RECORDING_PATH" : "录像路径：",
         "FIELD_HEADER_SERVER_ALIVE_INTERVAL" : "服务器keepalive时间间隔：",
+
         "FIELD_HEADER_SFTP_ROOT_DIRECTORY"   : "文件浏览器根目录：",
-        "FIELD_HEADER_TYPESCRIPT_NAME" : "打字稿名：",
-        "FIELD_HEADER_TYPESCRIPT_PATH" : "打字稿路径：",
```
mike-jumper: Is "打字稿" incorrect vs. "Typescript"?

guacamole/src/main/webapp/translations/zh.json
```diff
@@ -475,7 +625,15 @@
...
         "FIELD_OPTION_FONT_SIZE_EMPTY" : "",
-
+	
+        "FIELD_OPTION_TERMINAL_TYPE_ANSI"           : "ansi",
+        "FIELD_OPTION_TERMINAL_TYPE_EMPTY"          : "",
+        "FIELD_OPTION_TERMINAL_TYPE_LINUX"          : "linux",
+        "FIELD_OPTION_TERMINAL_TYPE_VT100"          : "vt100",
+        "FIELD_OPTION_TERMINAL_TYPE_VT220"          : "vt220",
+        "FIELD_OPTION_TERMINAL_TYPE_XTERM"          : "xterm",
+        "FIELD_OPTION_TERMINAL_TYPE_XTERM_256COLOR" : "xterm-256色",
```
mike-jumper: "xterm-256color" is the value that `$TERM` would be set to. There is no "xterm-256色".

guacamole/src/main/webapp/translations/zh.json
```diff
@@ -490,31 +648,40 @@
...
+        "FIELD_HEADER_SCROLLBACK"     : "最大回滚尺寸:",
+        "FIELD_HEADER_TERMINAL_TYPE"   : "终端类型:",
+        "FIELD_HEADER_TYPESCRIPT_NAME" : "Typescript 名称:",
+        "FIELD_HEADER_TYPESCRIPT_PATH" : "Typescript 路径:",
 
         "FIELD_OPTION_BACKSPACE_EMPTY" : "",
-        "FIELD_OPTION_BACKSPACE_8"     : "退格键（Ctrl-H）",
-        "FIELD_OPTION_BACKSPACE_127"   : "删除键（Ctrl-?）",
+        "FIELD_OPTION_BACKSPACE_8"     : "Backspace (Ctrl-H)",
+        "FIELD_OPTION_BACKSPACE_127"   : "Delete (Ctrl-?)",
```
mike-jumper: Why are these being changed back into English?

guacamole/src/main/webapp/translations/zh.json
```diff
@@ -678,14 +856,15 @@
...
+        "FIELD_HEADER_TIMEZONE"           : "时区:",
         "FIELD_HEADER_USERNAME"           : "用户名：",
         
         "HELP_DEFAULT_INPUT_METHOD" : "缺省输入法决定了Guacamole如何接收键盘事件。当使用移动设备或使用IME输入时，有可能需要更改设置。本设置可在Guacamole菜单内被单个连接的设定覆盖。",
         "HELP_DEFAULT_MOUSE_MODE"   : "缺省鼠标模拟方式决定了新连接内的远程鼠标如何响应屏幕触控。本设置可在Guacamole菜单内被单个连接的设定覆盖。",
         "HELP_INPUT_METHOD_NONE"    : "@:CLIENT.HELP_INPUT_METHOD_NONE",
         "HELP_INPUT_METHOD_OSK"     : "@:CLIENT.HELP_INPUT_METHOD_OSK",
         "HELP_INPUT_METHOD_TEXT"    : "@:CLIENT.HELP_INPUT_METHOD_TEXT",
-        "HELP_LANGUAGE"             : "在下方列表中选择Guacamole界面所使用的语言。可选用的语言决定于系统安装了什么语言。",
+        "HELP_LOCALE"             : "以下选项与用户的语言环境有关，并将影响界面各部分的显示方式。",
```
mike-jumper: Why is the nature of the text here changing?

guacamole/src/main/webapp/translations/zh.json
```diff
@@ -418,7 +560,7 @@
         "SECTION_HEADER_PERFORMANCE"        : "性能",
         "SECTION_HEADER_PRECONNECTION_PDU"  : "预连接PDU / Hyper-V",
         "SECTION_HEADER_RECORDING"          : "屏幕录像",
-        "SECTION_HEADER_REMOTEAPP"          : "RemoteApp",
+        "SECTION_HEADER_REMOTEAPP"          : "远程应用",
```
mike-jumper: Is "远程应用" correct for "RemoteApp"? I'm surprised to see a Microsoft brand-like term like "RemoteApp" translated, while other translations below like "typescript", "backspace", and "delete" are being changed back to English.

guacamole/src/main/webapp/translations/zh.json
```diff
@@ -300,14 +318,129 @@
...
         "SECTION_HEADER_EDIT_USER"   : "编辑用户",
         "SECTION_HEADER_PERMISSIONS" : "使用权限",
-
+        "SECTION_HEADER_USER_GROUPS"         : "用户组",
+	
         "TEXT_CONFIRM_DELETE" : "将不能恢复已被删除的用户。确定要删除这个用户吗？"
 
+    },
+    
+        "MANAGE_USER_GROUP" : {
```
mike-jumper: Beware - it looks like you accidentally indented this an extra level.

guacamole/src/main/webapp/translations/zh.json
XYUU: Okay,I correct it.

guacamole/src/main/webapp/translations/zh.json
XYUU: Okay,I correct it.

guacamole/src/main/webapp/translations/zh.json
XYUU: Okay,I correct it.

PR #454, opened 2019-11-29
GUACAMOLE-793: CAS Provider returns Group - like LDAP Provider

siacali:
Here's a functioning solution to this PR.  I'm not a java programmer (first project), but it looked easy enough to take a crack at, so I did.  It works for my own use case (CAS backed by LDAP), but definitely needs more testing.  I figured I'd submit it for review while I do that testing..


necouchman:
@siacali: Thanks for taking a stab at this!  I'll try to get the code reviewed before too long.

A couple of minor house-keeping issues, though, ahead of that review:
- Please make sure that the PR title matches the format we use consistently throughout the project - that is "GUACAMOLE-793:" (vs. "Guacamole 793").
- Please make sure that all commit messages contain the "GUACAMOLE-793:" prepended to the message.
- Please make sure the commit messages are meaningful and descriptive summaries of the commits.

siacali:
Thanks, I've done the housekeeping fix-ups you've requested.

necouchman:
@siacali: Thanks for updating things - unfortunately you actually need to go back and rebase and/or squash commits together.  Right now you've added two commits, and one of them is a "merge" of a branch - you'll want to go back and rebase everything on top of the current master and then edit the original commit messages.  Afterward you'll need to do a "git push -f" as git will complain about the branches having diverged.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/AuthenticationProviderService.java
```diff
@@ -58,6 +63,7 @@
     @Inject
     private Provider<CASAuthenticatedUser> authenticatedUserProvider;
 
+
```
necouchman: Extra line added - not sure this needs to be here :-).

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/AuthenticationProviderService.java
```diff
@@ -81,11 +87,13 @@ public CASAuthenticatedUser authenticateUser(Credentials credentials)
         if (request != null) {
             String ticket = request.getParameter(CASTicketField.PARAMETER_NAME);
             if (ticket != null) {
-                Map<String, String> tokens = ticketService.validateTicket(ticket, credentials);
+		TokensAndGroups tokensAndGroups = ticketService.validateTicket(ticket, credentials);
+		Map<String, String> tokens = tokensAndGroups.tokens;
+                Set<String> effectiveGroups = tokensAndGroups.effectiveGroups;
```
necouchman: Style needs to conform to the existing standards, which are four space tab stops.  Looks like a couple of these lines have tabs.

I'm also not sure that an entirely separate class is required here for parsing out tokens and groups.  It seems like maybe it would be better to just have the `validateTicket()` method return the raw attributes, and then have a  utility method - maybe a private one within this class - that parses the returned attributes for whatever one is configured for group membership, returning the `effectiveGroups`.  The method could either leave the group membership on for use as a token (if that's of any value) or remove the group membership attributes(s) from the `Map` such that only the non-group attributes are left as tokens.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/CASGuacamoleProperties.java
```diff
@@ -68,5 +69,30 @@ private CASGuacamoleProperties() {}
...
+  
+   /**
+    * The attribute used for group membership
+    * example:  memberOf  (case sensitive)
+    */
+    public static final StringGuacamoleProperty CAS_GROUP_ATTRIBUTE =
+         new StringGuacamoleProperty() {
 
+      @Override
+      public String getName() { return "cas-group-attribute"; }
```
necouchman: Indentation issues, here...

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/CASGuacamoleProperties.java
```diff
@@ -68,5 +69,30 @@ private CASGuacamoleProperties() {}
...
+    * The attribute used for group DN Format
+    * such as CN=%s,OU=myou,DC=example,DC=com  (case sensitive)
+    * used to strip all but %s.  This is only necessary when
+    * CAS backend is LDAP. 
+    */
+    public static final StringGuacamoleProperty CAS_GROUP_DN_FORMAT =
+         new StringGuacamoleProperty() {
+
+      @Override
+      public String getName() { return "cas-group-dn-format"; }
```
necouchman: Indentation issues...

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/ConfigurationService.java
```diff
@@ -85,4 +85,34 @@ public PrivateKey getClearpassKey() throws GuacamoleException {
         return environment.getProperty(CASGuacamoleProperties.CAS_CLEARPASS_KEY);
     }
 
+    /**
+     * Returns the attribute used to determine group memberships
+     * in CAS, or null if not defined.
+     *
+     * @return
+     *     The attribute name
```
necouchman: This should probably be more along the lines of "The attribute name used to determine group memberships in CAS, or null if not defined."  I know it seems redundant based on what's already a couple of lines above it, but when people are looking at the JavaDoc documentation for a Guacamole method there are situations where it can be useful.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/ConfigurationService.java
```diff
@@ -85,4 +85,34 @@ public PrivateKey getClearpassKey() throws GuacamoleException {
...
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed.
+     */
+    public String getCasGroupAttribute() throws GuacamoleException {
+        return environment.getProperty(CASGuacamoleProperties.CAS_GROUP_ATTRIBUTE);
+    }
+
+    /**
+     * Returns the attribute used to format group DN's
+     * in CAS, or null if not defined.
```
necouchman: This seems a little unclear to me.  I know you described the behavior in the body of the pull request, so I think I get what you're trying to say, but try to think of it from the perspective of someone reading JavaDocs without having ever seen what you wrote on the pull request, and maybe without having read any additional stuff that will be added to the manual.  So, this should be a little more descriptive of what the property is used for.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
```diff
@@ -68,6 +72,7 @@
     @Inject
     private ConfigurationService confService;
 
+
```
necouchman: No need for the extra line.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
```diff
@@ -122,15 +128,41 @@
...
                     tokens.put(tokenName, value.toString());
+                    if (attr.getKey().equals(groupAttribute)) {
+                        String valueWithoutBrackets = value.toString().substring(1,value.toString().length()-1);
+                        // probably could move this up further for efficiency
+                        Pattern pattern = Pattern.compile(groupTemplate);
+                        Matcher matcher = pattern.matcher(valueWithoutBrackets);
+                        while (matcher.find()) {
+                            effectiveGroups.add(matcher.group(1));
+                        }
+                    }
```
necouchman: Keep in mind that, with the way you have written this code, here, the group attribute will continue to be available in the tokens, as well.  Which is fine - this may be what you were going for, or it may not, just something to discuss about the merits of doing things this way or not.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/user/CASAuthenticatedUser.java
```diff
@@ -50,6 +51,12 @@
      */
     private Map<String, String> tokens;
 
+    /**
+     * The unique identifiers of all user groups which affect the permissions
+     * available to this user.
```
necouchman: I'm not sure that this should reference permissions, mainly because the CAS module does not implement connections or permissions at all, so it may be a little misleading to start referencing it within the context of this module.  While we do fully expect that these groups will be used to assign permissions in other modules (JDBC), within this module it's simply the groups the user is a member of.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/util/TokensAndGroups.java
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * A class to facilitate the return values of both Tokens and Groups
+ */
+public class TokensAndGroups {
+      public final Map<String, String> tokens;
+      public final Set<String> effectiveGroups;
+      public <TokensAndGroups> TokensAndGroups(Map<String, String> tokens, Set<String> effectiveGroups) {
+             this.tokens=tokens;
+             this.effectiveGroups=effectiveGroups;
+      }
```
necouchman: If we're going to keep this class (see other comments), style here needs to match up, and comments need to be added.  That may be moot, though...

siacali:
Hi Nick,
Thanks for the review!  I'll get these cleaned up and resubmitted shortly.  On the thinking behind both the creation of the TokensAndGroups class, and leaving the group attribute in Tokens:  

1. I have no idea how folks may be using (or planning to use) Tokens, so I left their implementation alone.  I didn't want to change the end-user apparent functionality of the existing code, especially since the Jira story was to "implement" not to clean up.  Did I take that too literally?  :-)   I did briefly consider re-implementing the underlying library so it would return the group attributes in a more sensible/structured fashion...decided to "fix" that if/when there's a Jira story raised on it...

2. I had considered letting ValidateTicket just pass back Tokens and handle the parsing of groups there, but that would make it less obvious to someone reading the code (perhaps with intentions of modifying the "Token" code) that Groups depended upon that functionality.  This way, dependencies are reduced and the dirty work of dealing with both is right out in the open.   Normally I wouldn't make a class to use once, but given the choice of "hiding" dependencies in two separate modules or making a simple class to make things clearer, well...you know what I chose.   Considering that thinking, would you still prefer I restructure away TokensAndGroups?

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/ConfigurationService.java
siacali: I've taken a crack at a correction to this, please review and let me know if it's clearer.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
siacali: I've left a comment below as to my reasoning.  If you agree, I'll leave it unchanged.  If you disagree, I will agree to disagree and change it.  :-)

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/user/CASAuthenticatedUser.java
siacali: Please review my changes to see if they're more on the mark....

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/util/TokensAndGroups.java
siacali: I;ve corrected the formatting and am awaiting your reply on my comments below to resolve.

necouchman:
> I have no idea how folks may be using (or planning to use) Tokens, so I left their implementation alone.

That's fine - I have no issue with these remaining available as a token.

> I had considered letting ValidateTicket just pass back Tokens and handle the parsing of groups there, but that would make it less obvious to someone reading the code (perhaps with intentions of modifying the "Token" code) that Groups depended upon that functionality.

My current view is that this just adds more complexity than it's worth, and that doing a private method within the current class to parse it out is probably the way to go.  I'd appreciate feedback from some of the other project members, though - @mike-jumper @jmuehlner @ceharris, any thoughts on this?  I don't want to be the single voice pushing in one direction or another...

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/util/TokensAndGroups.java
```diff
@@ -0,0 +1,44 @@
...
+     *
+     * @return an instance of itself
+     *
+     */
+    public class TokensAndGroups {
+        public final Map<String, String> tokens;
+        public final Set<String> effectiveGroups;
+        public <TokensAndGroups> TokensAndGroups(Map<String, String> tokens, Set<String> effectiveGroups) {
+            this.tokens=tokens;
+            this.effectiveGroups=effectiveGroups;
```
necouchman: Should be spaces around the `=` here.

siacali:
> > I have no idea how folks may be using (or planning to use) Tokens, so I left their implementation alone.
> 
> That's fine - I have no issue with these remaining available as a token.
> 
> > I had considered letting ValidateTicket just pass back Tokens and handle the parsing of groups there, but that would make it less obvious to someone reading the code (perhaps with intentions of modifying the "Token" code) that Groups depended upon that functionality.
> 
> My current view is that this just adds more complexity than it's worth, and that doing a private method within the current class to parse it out is probably the way to go. I'd appreciate feedback from some of the other project members, though - @mike-jumper @jmuehlner @ceharris, any thoughts on this? I don't want to be the single voice pushing in one direction or another...

@necouchman ,

I revisited this with the intentions of accepting your guidance and implementing it the way you wished to see it done.  In getting into it, I discovered that he returned token names are all capitalized (and group names are case sensitive), so parsing those to create groups won't work quite right.   It looks like the extra class is actually not a bad way to get it done.  Also has the benefit of being much more efficient than the other method would be in terms of fewer loops, no string operations to undo combines that were just done, etc...

Can we agree to keep it or do you have other guidance?

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/util/TokensAndGroups.java
siacali: Fixing...

necouchman:
@siacali - There seems to be an extra commit with a binary file in it (ebebb69).  This should probably be removed.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/ConfigurationService.java
```diff
@@ -85,4 +85,40 @@ public PrivateKey getClearpassKey() throws GuacamoleException {
...
+     *
+     * @return
+     *     The attribute name user to determine group memberships in CAS,
+     *     null if not defined.
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed.
+     */
+    public String getCasGroupAttribute() throws GuacamoleException {
```
necouchman: In sticking with the way the rest of these methods are named, I'd suggest taking the "Cas" out of this and making it `getGroupAttribute()`, and similarly with the method below for the group DN format.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
```diff
@@ -33,10 +33,14 @@
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
+import java.util.HashSet;
+import java.util.regex.*;
```
necouchman: We generally try to avoid importing classes using wildcards like this, and prefer to list out the specific ones you need.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
```diff
@@ -88,7 +92,7 @@
      *     If the ID ticket is not valid or guacamole.properties could
      *     not be parsed.
      */
-    public Map<String, String> validateTicket(String ticket,
+    public TokensAndGroups  validateTicket(String ticket,
```
necouchman: I hate to be a pain about this, but I'm still in favor of having this `validateTicket()` method just return the `Map<String, String>` data, and parsing it out in the body of the `authenticateUser()` method in the `AutenticationProviderService` class.  This leaves the `validateTicket()` method to do ticket validation, keeping it free from the extra logic introduced here.

I'd suggest doing one of two things:
* Keep the `TokenAndGroups` utility class, but, rather than using it to store the information, implement a static method (or two) that takes in the `Map<String, String>` input from `validateTicket()` and parses out the groups, returning the `List` of effective groups.  Call this method within the `authenticateUser()` method using the output received from `validateTicket()` and using the output to set up the effective groups
* Get rid of the `TokensAndGroups` class and implement the parser in a `private` method within the `AuthenticationProviderService` class that does the same as above.

I'm completely on-board with leaving the group data in the tokens, as we discussed earlier.

siacali: Ok, the problem with your suggestion is that the `Map<String, String>` **_presently_** returned by `validateTicket()` doesn't contain valid information for me to derive groups.  In particular, it returns capitalized keys with a string prepended (CAS-FOOBAR), and groups is case sensitive (So, CAS-FOOBAR, would not be enough for me to differentiate groups such as FooBar from fooBar - I had mentioned this before).  So, to do the processing within `authenticateUser()`, I would need to:

- Refactor the previous author's code to return what I need ("raw" SAML attributes released by CAS)
- Move/restructure their code to create tokens from SAML attributes to `authenticateUser()`
- Derive both tokens and groups in `authenticateUser()`

Certainly possible, but refactoring/restructuring someone elses code (and then having to QA it) does seem like a different task and a bit more risk than adding a feature to existing code.   

Also, does that really fit into the philosophy of "one jira ticket, one change?" or is it "ok" to "fix" someone elses non-broken code as part of adding a feature?  I'm not even a Java programmer and I can see a lot of things I'd like to "fix" given a license like that...  :-)

I'll do it (given the time to do so), if that's what you'd like to see.

mike-jumper: > ...
>
> Certainly possible, but refactoring/restructuring someone elses code (and then having to QA it) does seem like a different task and a bit more risk than adding a feature to existing code.
>
> Also, does that really fit into the philosophy of "one jira ticket, one change?" or is it "ok" to "fix" someone elses non-broken code as part of adding a feature? ...

It does occasionally happen that additional code needs to be touched in order to implement a feature. I don't think there's a universal "this way is safe, this way is unsafe" answer to your question, but we do tend to look for ways to improve established code when a new change indicates that it would be a better direction. Doing otherwise would mean we trend toward greater and greater technical debt.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/util/TokensAndGroups.java
```diff
@@ -0,0 +1,44 @@
...
+     *
+     * @param tokens, groups
+     *     A Map<String, String> of tokens, parsed from attributes returned
+     *     by CAS.
+     *     A Set<String> of groups, parsed from attributes returned by CAS.
+     *
+     * @return an instance of itself
+     *
+     */
+    public class TokensAndGroups {
```
mike-jumper: Why is this class declaration indented by a level?

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/util/TokensAndGroups.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * A class to facilitate the return values of both Tokens and Groups
+     *
+     * @param tokens, groups
+     *     A Map<String, String> of tokens, parsed from attributes returned
+     *     by CAS.
+     *     A Set<String> of groups, parsed from attributes returned by CAS.
+     *
+     * @return an instance of itself
```
mike-jumper: The `class` is just a `class`. It doesn't accept parameters nor return a value.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/AuthenticationProviderService.java
```diff
@@ -81,11 +86,13 @@ public CASAuthenticatedUser authenticateUser(Credentials credentials)
         if (request != null) {
             String ticket = request.getParameter(CASTicketField.PARAMETER_NAME);
             if (ticket != null) {
-                Map<String, String> tokens = ticketService.validateTicket(ticket, credentials);
+                TokensAndGroups tokensAndGroups = ticketService.validateTicket(ticket, credentials);
```
mike-jumper: I'm not a fan of defining a new class just for the sake of returning a pair of values when there's no higher level concept behind that class besides being a pair. `TokensAndGroups` is a type-specific pair, but that's about it. There has to be a better way.

Perhaps `validateTicket()` should return the `CASAuthenticatedUser` instance?

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/CASGuacamoleProperties.java
```diff
@@ -68,5 +69,30 @@ private CASGuacamoleProperties() {}
...
+    public String getName() { return "cas-group-attribute"; }
+
+    };
+
+   /**
+    * The attribute used for group DN Format
+    * such as CN=%s,OU=myou,DC=example,DC=com  (case sensitive)
+    * used to strip all but %s.  This is only necessary when
+    * CAS backend is LDAP. 
+    */
```
mike-jumper: Are there other cases where CAS might have group names in a specific format? If so, perhaps this should be more generic and not specific to an LDAP DN?

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
```diff
@@ -122,15 +127,41 @@
...
                 String tokenName = TokenName.canonicalize(attr.getKey(),
                         CAS_ATTRIBUTE_TOKEN_PREFIX);
                 Object value = attr.getValue();
-                if (value != null)
+                if (value != null) {
                     tokens.put(tokenName, value.toString());
+                    if (attr.getKey().equals(groupAttribute)) {
+                        String valueWithoutBrackets = value.toString().substring(1,value.toString().length()-1);
+                        // probably could move this up further for efficiency
+                        Pattern pattern = Pattern.compile(groupTemplate);
```
mike-jumper: What if the value of the property that `groupTemplate` is derived from already contains characters with special meaning to Java regex?

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
```diff
@@ -122,15 +127,41 @@
...
+                // if CAS is backended to LDAP, groups come in as RFC4514 DN
+                // syntax.  If cas-group-dn-format is set, this strips
+                // an entry such as "CN=Foo,OU=Bar,DC=example,DC=com" to
+                // "Foo"
+                groupTemplate = groupDnFormat.replace("%s","([A-Za-z0-9_\\(\\)\\-\\.\\s+]+)"); 
+                // the underlying parser aggregates all instances of the same
+                // attribute, so we need to be able to parse them out
+                groupTemplate=groupTemplate+",*\\s*";
+            } else {
+               groupTemplate = "([A-Za-z0-9_\\(\\)\\-\\.\\s+]+,*\\s*)";
```
mike-jumper: The `getProperty()` function already has a variant for providing a default value. There's no need to test for `null` and manually select that default.

I'd also suggest that even if there were such a need, it would be better to abstract that within the relevant function within `ConfigurationService`.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/util/TokensAndGroups.java
```diff
@@ -0,0 +1,44 @@
...
+     *
+     */
+    public class TokensAndGroups {
+        public final Map<String, String> tokens;
+        public final Set<String> effectiveGroups;
+        public <TokensAndGroups> TokensAndGroups(Map<String, String> tokens, Set<String> effectiveGroups) {
```
mike-jumper: Why is this constructor declared as if it leverages generics (`<TokensAndGroups>`)?

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/util/TokensAndGroups.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+package org.apache.guacamole.auth.cas.util;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.HashSet;
+
+    /**
+     * A class to facilitate the return values of both Tokens and Groups
```
mike-jumper: There should be a better way to achieve this (more on this above). I don't think we should be declaring a class just to be a type-specific pair when there is no other high-level concept behind that class.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/user/CASAuthenticatedUser.java
```diff
@@ -79,9 +85,10 @@ public void init(String username, Credentials credentials) {
      *     as tokens when connections are established with this user.
      */
     public void init(String username, Credentials credentials,
-            Map<String, String> tokens) {
+            Map<String, String> tokens, Set<String> effectiveGroups) {
```
mike-jumper: If adding a new parameter, documentation needs to be added for that parameter (see the documentation for other parameters of this function for examples of this).

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/user/CASAuthenticatedUser.java
```diff
@@ -50,6 +51,11 @@
      */
     private Map<String, String> tokens;
 
+    /**
+     * The unique identifiers of all user groups which a user is a member of
```
mike-jumper: A period is missing here (JavaDoc requires that comments are terminated with punctuation - see the other JavaDoc comments in the vicinity for examples).

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
```diff
@@ -122,15 +127,41 @@
...
             for (Entry <String, Object> attr : ticketAttrs.entrySet()) {
                 String tokenName = TokenName.canonicalize(attr.getKey(),
                         CAS_ATTRIBUTE_TOKEN_PREFIX);
                 Object value = attr.getValue();
-                if (value != null)
+                if (value != null) {
                     tokens.put(tokenName, value.toString());
+                    if (attr.getKey().equals(groupAttribute)) {
+                        String valueWithoutBrackets = value.toString().substring(1,value.toString().length()-1);
+                        // probably could move this up further for efficiency
```
mike-jumper: Yes. Or abstract all that within `ConfigurationService` so this part of the code doesn't need to deal with the formatting of the property.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
```diff
@@ -122,15 +127,41 @@
...
+           }
             for (Entry <String, Object> attr : ticketAttrs.entrySet()) {
                 String tokenName = TokenName.canonicalize(attr.getKey(),
                         CAS_ATTRIBUTE_TOKEN_PREFIX);
                 Object value = attr.getValue();
-                if (value != null)
+                if (value != null) {
                     tokens.put(tokenName, value.toString());
+                    if (attr.getKey().equals(groupAttribute)) {
+                        String valueWithoutBrackets = value.toString().substring(1,value.toString().length()-1);
```
mike-jumper: Why?

siacali:
Gentlemen, Thank you for taking the time for a detailed review.  I will try to make some time to respond to/incorporate the changes requested.  Realistically, It probably will be a few more weeks as I've got a pretty full plate.  Meanwhile, as a side note (and admittedly pretty irrelevant to your requests) we do have these changes running in our production environment and they seem to function well as intended.  Regardless, I do want to make this contribution fit the high standards you've established for this project, and wherever possible make it fit general purpose needs.  As such I will comply (or debate) your review as appropriate.   Thanks for the patience and your time and efforts.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/util/TokensAndGroups.java
siacali: stupid formatting on my part?  :-). Wil fix.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/util/TokensAndGroups.java
siacali: Will fix.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/AuthenticationProviderService.java
siacali: Nick had suggested that I pass the tokens up and deal with groups a level up.  I felt that makes one unnecessarily dependent on the other (more fragile) and less obvious as to function.  It also ensures that we iterate over these items multiple times (less efficient).  When I get the time to do so, I'll look at and propose alternatives to this discussion.  Classes appear to be very efficient both  performance-wise, and to show what's going on.  I'm not sure why both you and Nick are not fans of using them, but I'm willing to try to work around that.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/CASGuacamoleProperties.java
siacali: CAS groups come from LDAP, databases, and Grouper.  AFAIK LDAP is the only place you have the notion of DNs with potentially irrelevant RDNs that you *may* want to strip.  Database entries are usually singular group names, and grouper entries are fully significant.  Do you know of any other cases?

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
siacali: They would get processed by regex and this wouldn't work right.  While It's possible, it's not likely.  (ok, yes, I realize by saying that, I've probably just started a cascade within the universe that will make it happen soon...).  Is it worth escaping such characters (which may lead to unintended consequences), throwing an error, or just writing a caveat in the doc?

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
siacali: Will try to fix.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/util/TokensAndGroups.java
siacali: lack of knowing any better.   This is my first attempt at any java whatsoever, so I'm an admitted rank amateur.   How should I do it?

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/user/CASAuthenticatedUser.java
siacali: ok, will fix.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/user/CASAuthenticatedUser.java
siacali: Will fix, thank you.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
siacali: the object comes in as a string of comma separated values inside brackets.  Need to remove the brackets to address the comma separated values only.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/AuthenticationProviderService.java
necouchman: I think my record will show that I like classes as much as the next guy, as I've implemented plenty of them! ;-)

In this case, though, it just seems like a lot of overhead to essentially accomplish splitting out the contents of an array into groups and other stuff.  To me it is simpler and more maintainable to have a private method within the existing class accomplish the same thing.

Just my two bits...okay, I've put in way more than that on this topic...

necouchman:
Thanks @siacali - we are happy to have your contributions to the project.  It's great to know that it's working well in your environment, too.  I'm always happy to find someone else using CAS and interested in making stuff work with it!

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/util/TokensAndGroups.java
necouchman: Just `public TokensAndGroups(...` without any type specification.  Since the name of the method is the same as the name of the class, Java knows it is a constructor and that it is returning that as its type.

necouchman:
@siacali:Ping - any chance to further refine these changes?

mike-jumper:
Bumping merge base from `staging/1.2.0` to `master` so that we can proceed with a 1.2.0 RC.

@siacali, hope all is well. Let us know if you have any questions on the feedback thus far.

siacali:
Sorry guys, I had gotten caught up in other things and had let this slip.  I'll put it on my list and see if I can get to it within the next week or two.  Feedback is clear.

doctorfree:
Hi, @siacali has asked me to assist with this. After reviewing the changes and discussion here I agree with the approach suggested by @mike-jumper "Perhaps validateTicket() should return the CASAuthenticatedUser instance?". I've created a patch that does this and it is going through some testing. If all works as expected I will resubmit this, rebasing on top of the current master. Does this seem like the right way to go?

necouchman:
Closed by #579.

PR #457, opened 2019-12-25
GUACAMOLE-550: Add autocomplete attributes to applicable form fields.

necouchman:
Took a shot at adding autocomplete attributes to the form fields.  Let me know if this looks okay, or if things need tweaking.

### DIFF
--- a/guacamole/src/main/webapp/app/form/templates/dateField.html
+++ b/guacamole/src/main/webapp/app/form/templates/dateField.html
@@ -8,5 +8,6 @@
            guac-focus="focused"
            placeholder="{{'FORM.FIELD_PLACEHOLDER_DATE' | translate}}"
            autocorrect="off"
-           autocapitalize="off"/>
+           autocapitalize="off"
+           autocomplete="off"/>
 </div>
--- a/guacamole/src/main/webapp/app/form/templates/passwordField.html
+++ b/guacamole/src/main/webapp/app/form/templates/passwordField.html
@@ -6,6 +6,7 @@
            ng-trim="false"
            guac-focus="focused"
            autocorrect="off"
-           autocapitalize="off"/>
+           autocapitalize="off"
+           autocomplete="new-password"/>
     <div class="icon toggle-password" ng-click="togglePassword()" title="{{getTogglePasswordHelpText() | translate}}"></div>
 </div>
--- a/guacamole/src/main/webapp/app/form/templates/emailField.html
+++ b/guacamole/src/main/webapp/app/form/templates/emailField.html
@@ -6,6 +6,7 @@
            ng-hide="readOnly"
            guac-focus="focused"
            autocorrect="off"
-           autocapitalize="off"/>
+           autocapitalize="off"
+           autocomplete="email"/>
     <a href="mailto:{{model}}" ng-show="readOnly">{{model}}</a>
 </div>
--- a/guacamole/src/main/webapp/app/form/templates/numberField.html
+++ b/guacamole/src/main/webapp/app/form/templates/numberField.html
@@ -4,4 +4,5 @@
        ng-model="typedValue"
        guac-focus="focused"
        autocorrect="off"
-       autocapitalize="off"/>
+       autocapitalize="off"
+       autocomplete="off"/>
--- a/guacamole/src/main/webapp/app/form/templates/textField.html
+++ b/guacamole/src/main/webapp/app/form/templates/textField.html
@@ -6,7 +6,8 @@
            ng-disabled="disabled"
            guac-focus="focused"
            autocorrect="off"
-           autocapitalize="off"/>
+           autocapitalize="off"
+           autocomplete="off"/>
     <datalist ng-if="dataListId" ng-attr-id="{{ dataListId }}">
         <option ng-repeat="option in field.options | orderBy: option"
                 value="{{ option }}">{{ getFieldOption(option) | translate }}</option>
--- a/guacamole/src/main/webapp/app/form/templates/timeField.html
+++ b/guacamole/src/main/webapp/app/form/templates/timeField.html
@@ -8,5 +8,6 @@
            guac-lenient-time
            placeholder="{{'FORM.FIELD_PLACEHOLDER_TIME' | translate}}"
            autocorrect="off"
-           autocapitalize="off"/>
+           autocapitalize="off"
+           autocomplete="off"/>
 </div>


guacamole/src/main/webapp/app/form/templates/passwordField.html
```diff
@@ -6,6 +6,7 @@
            ng-trim="false"
            guac-focus="focused"
            autocorrect="off"
-           autocapitalize="off"/>
+           autocapitalize="off"
+           autocomplete="new-password"/>
```
mike-jumper: I'm not immediately sure how to deal with this, but I think this may cause trouble on login pages which use this field. In those cases, we're looking for an existing password, and "new-password" might not be the correct choice.

In your testing, did this disrupt login behavior at all?

guacamole/src/main/webapp/app/form/templates/dateField.html
```diff
@@ -8,5 +8,6 @@
            guac-focus="focused"
            placeholder="{{'FORM.FIELD_PLACEHOLDER_DATE' | translate}}"
            autocorrect="off"
-           autocapitalize="off"/>
+           autocapitalize="off"
+           autocomplete="off"/>
```
mike-jumper: Any idea whether these input fields will also need proper `name` attributes for browser autocomplete to function reliably?

The MDN article on disabling autocomplete across the board has a section at the bottom titled ["Tools for disabling autocompletion"](Tools_for_disabling_autocompletion) which seems to imply that the `name` attribute plays an important role:

> ## Tools for disabling autocompletion
>
> The [`jquery.disableAutoFill`](https://terrylinooo.github.io/jquery.disableAutoFill/) plugin randomizes an input's `name` attribute by default. When the form is submitted, the plugin restores the original name. This prevents auto-completion for all browsers (includes third-party auto-completion extensions) but doesn't necessarily help with login fields.

guacamole/src/main/webapp/app/form/templates/passwordField.html
mike-jumper: If we provide a way to enable/disable autocomplete at the `guac-form` level, we could enable autocomplete only during login and default to `autocomplete="off"` (or `autocomplete="new-password"`) otherwise.

necouchman: Good idea - I've taken a stab at implementing this. Let me know what you think.

guacamole/src/main/webapp/app/form/templates/dateField.html
necouchman: Good question - I'm not sure. However, the fields do get a "name" attribute that I think is pushed down from the `ng-attr-name` property? Maybe this is enough?

mike-jumper: Ah, yep - this changed in the time since the first review via #690.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledUser.java
```diff
@@ -105,10 +105,10 @@ public class ModeledUser extends ModeledPermissions<UserModel> implements User {
      * form.
      */
     public static final Form PROFILE = new Form("profile", Arrays.<Field>asList(
-        new TextField(User.Attribute.FULL_NAME),
+        new TextField(User.Attribute.FULL_NAME, Field.Autocomplete.NAME),
```
mike-jumper, line 108: If we set these autocomplete attributes for all users, will that result in the browser assuming that the details of the _current_ user are appropriate for any user being edited?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
```diff
@@ -98,7 +98,7 @@ public class ModeledConnection extends ModeledChildDirectoryObject<ConnectionMod
      * connection.
      */
     public static final Form GUACD_PARAMETERS = new Form("guacd", Arrays.<Field>asList(
-        new TextField(GUACD_HOSTNAME_NAME),
+        new TextField(GUACD_HOSTNAME_NAME, Field.Autocomplete.ON),
```
mike-jumper, line 101: Should this be `Field.Autocomplete.OFF`?

guacamole/src/main/frontend/src/app/form/templates/dateField.html
```diff
@@ -9,5 +9,6 @@
            guac-focus="focused"
            placeholder="{{'FORM.FIELD_PLACEHOLDER_DATE' | translate}}"
            autocorrect="off"
-           autocapitalize="off">
+           autocapitalize="off"
+           autocomplete="{{ field.autocomplete }}"/>
```
mike-jumper, line 13: Here and elsewhere: should this be `ng-attr-autocomplete`?

necouchman: Maybe so, but:
* I didn't find any references to `ng-attr-autocomplete` in the AngularJS documentation, so wasn't sure if it was valid. Is it just a general rule that `ng-attr-<anything>` maps to the `<anything>` tag property?
* The above code seems to work, at least in my somewhat limited testing, so it doesn't seem to be necessary, though maybe it's preferred?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java
necouchman: I'm open to suggestions - I was tempted to leave this one `ON` so that it would auto-complete for admins filling in this field, but I can be swayed...

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledUser.java
necouchman: Yes, probably so. As with the other ones, if you think it's better to set it to `OFF`, I'm okay with that, too.

PR #460, opened 2019-12-27
GUACAMOLE-760: Add support for configuring server timezone.

necouchman:
This adds support for the various JDBC modules for configuring the database server timezone.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/MySQLAuthenticationProviderModule.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/MySQLAuthenticationProviderModule.java
@@ -24,6 +24,7 @@ import com.google.inject.Module;
 import com.google.inject.name.Names;
 import java.io.File;
 import java.util.Properties;
+import java.util.TimeZone;
 import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.auth.mysql.conf.MySQLDriver;
 import org.apache.guacamole.auth.mysql.conf.MySQLEnvironment;
@@ -114,6 +115,11 @@ public class MySQLAuthenticationProviderModule implements Module {
         
         // Get the MySQL-compatible driver to use.
         mysqlDriver = environment.getMySQLDriver();
+
+        // If timezone is present, set it.
+        TimeZone serverTz = environment.getServerTimeZone();
+        if (serverTz != null)
+            driverProperties.setProperty("serverTimezone", serverTz.getID());
 
     }
 
--- a/guacamole-ext/src/main/java/org/apache/guacamole/properties/TimeZoneGuacamoleProperty.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/properties/TimeZoneGuacamoleProperty.java
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.guacamole.properties;
+
+import java.util.TimeZone;
+import org.apache.guacamole.GuacamoleException;
+
+/**
+ * A GuacamoleProperty whose value is a TimeZone.
+ */
+public abstract class TimeZoneGuacamoleProperty
+        implements GuacamoleProperty<TimeZone> {
+    
+    @Override
+    public TimeZone parseValue(String value) throws GuacamoleException {
+        
+        // Nothing in, nothing out
+        if (value == null || value.isEmpty())
+            return null;
+        
+        // Attempt to return the TimeZone of the provided string value.
+        return TimeZone.getTimeZone(value);
+        
+    }
+    
+}
--- a/guacamole-ext/src/main/java/org/apache/guacamole/properties/TimeZoneGuacamoleProperty.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/properties/TimeZoneGuacamoleProperty.java
@@ -20,13 +20,21 @@
 package org.apache.guacamole.properties;
 
 import java.util.TimeZone;
+import java.util.regex.Pattern;
 import org.apache.guacamole.GuacamoleException;
+import org.apache.guacamole.GuacamoleServerException;
 
 /**
  * A GuacamoleProperty whose value is a TimeZone.
  */
 public abstract class TimeZoneGuacamoleProperty
         implements GuacamoleProperty<TimeZone> {
+    
+    /**
+     * A regex that matches valid variants of GMT timezones.
+     */
+    public static final Pattern GMT_REGEX =
+            Pattern.compile("^GMT([+-](0|00)((:)?00)?)?$");
     
     @Override
     public TimeZone parseValue(String value) throws GuacamoleException {
@@ -36,7 +44,16 @@ public abstract class TimeZoneGuacamoleProperty
             return null;
         
         // Attempt to return the TimeZone of the provided string value.
-        return TimeZone.getTimeZone(value);
+        TimeZone tz = TimeZone.getTimeZone(value);
+        
+        // If the input is not GMT, but the output is GMT, the TimeZone is not
+        // valid and we throw an exception.
+        if (!GMT_REGEX.matcher(value).matches()
+                && GMT_REGEX.matcher(tz.getID()).matches())
+            throw new GuacamoleServerException("Property \"" + getName()
+                + "\" does not specify a valid time zone.");
+
+        return tz;
         
     }
     
--- a/guacamole-ext/src/test/java/org/apache/guacamole/properties/TimeZoneGuacamolePropertyTest.java
+++ b/guacamole-ext/src/test/java/org/apache/guacamole/properties/TimeZoneGuacamolePropertyTest.java
@@ -1,0 +1,234 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.guacamole.properties;
+
+import java.util.Arrays;
+import java.util.List;
+import org.apache.guacamole.GuacamoleException;
+import static org.junit.Assert.*;
+import org.junit.Test;
+
+/**
+ * Tests that validate Time Zone property input.
+ */
+public class TimeZoneGuacamolePropertyTest {
+    
+    /**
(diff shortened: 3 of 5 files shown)


necouchman:
Okay...
* Removed any default value for the server timezone to avoid causing problems to existing installs.
* Removed PostgreSQL and SQL Server implementations of this and left it with MySQL, only.
* Placed the setting of the JDBC driver property in the correct place.

mike-jumper (changes requested):
Depending on the answer to the question regarding timezone string format, this looks fine to me. All that's for sure needed here is a rebase to resolve conflicts against `staging/1.3.0`.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/MySQLAuthenticationProviderModule.java
```diff
@@ -82,6 +83,11 @@ public MySQLAuthenticationProviderModule(MySQLEnvironment environment)
         // Get the MySQL-compatible driver to use.
         mysqlDriver = environment.getMySQLDriver();
 
+        // If timezone is present, set it.
+        TimeZone serverTz = environment.getServerTimeZone();
+        if (serverTz != null)
+            driverProperties.setProperty("serverTimezone", serverTz.toString());
```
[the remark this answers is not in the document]
necouchman: I think `getID()` will be safer.  According to MySQL J/Connector documentation([1]):

> The MySQL server is configured with a canonical time zone that is recognizable by Java (for example, Europe/Paris, Etc/GMT-5, UTC, etc.)

This seems to match with Java's TimeZone documentation on ID([2]):

> You can also get a TimeZone using getTimeZone along with a time zone ID. For instance, the time zone ID for the U.S. Pacific Time zone is "America/Los_Angeles".

[1] - https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-other-changes.html
[2] - https://docs.oracle.com/javase/7/docs/api/java/util/TimeZone.html

necouchman:
Okay, rebase completed, and slight tweak to which method gets used, there...

guacamole-ext/src/main/java/org/apache/guacamole/properties/TimeZoneGuacamoleProperty.java
```diff
@@ -0,0 +1,43 @@
...
+    
+    @Override
+    public TimeZone parseValue(String value) throws GuacamoleException {
+        
+        // Nothing in, nothing out
+        if (value == null || value.isEmpty())
+            return null;
+        
+        // Attempt to return the TimeZone of the provided string value.
+        return TimeZone.getTimeZone(value);
```
mike-jumper, line 39: What if the provided timezone name is invalid?

necouchman: According to the `TimeZone` documentation, it simply returns GMT if the provided value is invalid.  Guess the Java developers were tired that day and decided not to use any Exception handling :laughing:.  Anyway, I can either document it as such, or throw in some logic that tries to detect when that happens (GMT is not provided as the string value, but ends up as the `TimeZone` value) and throw an exception.

mike-jumper: Yikes, that is a bit of a horror show: https://stackoverflow.com/questions/13092865/timezone-validation-in-java

If possible, I think proper validation is worth doing ... but I get that it's going to be real hairy. Would probably need a few tests that verify the validation works correctly for legal ways of specifying "GMT".

I hate to ask, but is there a finite, known set of strings for GMT within Java?

mike-jumper: > I hate to ask, but is there a finite, known set of strings for GMT within Java?

Judging from [the grammar in the documentation for `TimeZone`](https://docs.oracle.com/javase/7/docs/api/java/util/TimeZone.html), it looks like valid GMT strings are:

* `GMT`
* `GMT+0`
* `GMT-0`
* `GMT+0:00`
* `GMT-0:00`
* `GMT+000`
* `GMT-000`
* `GMT+00`
* `GMT-00`
* `GMT+00:00`
* `GMT-00:00`
* `GMT+0000`
* `GMT-0000`

We might be able to safely validate by checking that either of the following is true:

1. The resulting `TimeZone` is not equivalent to `GMT`.
2. The resulting `TimeZone` _is_ equivalent to `GMT`, and is either explicitly listed by `getAvailableIDs()` or matches a valid "custom time zone ID" for GMT (regex?).

necouchman: Is there any reason not to just loop through the `getAvailableIDs()` output and compare the string value we receive to the available IDs, and throw an error if don't match on that? It's probably a slightly expensive operation from a compute perspective, but probably not all that bad, and then we can just document that the value for the timezone property must be a valid Java TimeZone ID. Something like:
```
for (String tzStr : TimeZone.getAvailableIDs()) {
    if (value.equals(tzStr))
        return TimeZone.getTimeZone(value);
}

throw GuacamoleServerException("Invalid timezone specified.");
```

It's not pretty or elegant, but it's fairly simple. Or is there some downfall I'm missing with that?

mike-jumper: I believe so: not _all_ valid GMT representations are within that list. The grammar for "custom time zone IDs" allows for quite a few equivalent ways to say GMT / GMT+0.

I think the best way would to accept any value that does not equate to GMT, and for all others do a quick comparison for known-good values (static `HashSet` of `getAvailableIDs()`?) and known-good representations of GMT (regex?). Anything that equals GMT, isn't a valid custom ID for GMT, and isn't explicitly listed by `getAvailableIDs()` must be invalid.

necouchman: Okay, I think I understand - will give it a go.

necouchman: Wow, the Java `TimeZone` implementation is really awful.  Writing the `JUnit` tests is fun:
* If you use the string "GMT" for the input, the ID you get back is "GMT" - okay, good so far.
* If you use the string "GMT+00:00" for the input, the ID you get back is "GMT+00:00" - not ideal, but not the worst.
* If you use the string "GMT+0000" for the input, the ID you get back is "GMT+00:00" - :exploding_head: 

So, cannot check that valid input strings generate an ID of "GMT", and cannot check that valid input strings generate an ID equal to the input string.

necouchman: Okay, well, I have given it a shot, so let me know how that looks.

necouchman: I feel like there might be two RegEx matches required, here - one for the valid GMT inputs, and one for TimeZone IDs that are returned for either GMT or custom offsets.  It seems that any time an input value is in the format "GMT+hhmm" it gets converted with `getID()` to "GMT+hh:mm" (adds the colon).  So, it seems like the checks for validity should be:
* Matches the list of known TimeZones
* Input value is the same as the `getID()` output
* Valid GMT value (GMT+/- and some number of 0s)
* Some sort of RegEx that matches custom offset of GMT[+-][hhmm] to GMT[+-][hh:mm]

mike-jumper: What about something like `^GMT[+-](0|00)((:)?00)?$` ?

guacamole-ext/src/main/java/org/apache/guacamole/properties/TimeZoneGuacamoleProperty.java
```diff
@@ -0,0 +1,78 @@
...
+        if (GMT_REGEX.matcher(value).matches())
+            return tz;
+        
+        // If the output of the TimeZone matches the input, it's probably valid
+        if (tz.getID().equals(value))
+            return tz;
+        
+        // Nothing valid is found, so throw exception.
+        throw new GuacamoleServerException("Property \"" + getName()
+                + "\" does not specify a valid time zone.");
```
mike-jumper, line 72: I think we should test for invalidity only if the resulting `TimeZone` is not identical to GMT. Any other value returned by `TimeZone.getTimeZone()` would mean that `value` must be valid. That reduces the scope of the problem from "is this timezone valid?" to "did the user really mean GMT?".

guacamole-ext/src/main/java/org/apache/guacamole/properties/TimeZoneGuacamoleProperty.java
necouchman: Yeah, one slight modification - since "GMT" is also valid, had to make the entire section after that optional - but that works pretty well.

guacamole-ext/src/main/java/org/apache/guacamole/properties/TimeZoneGuacamoleProperty.java
necouchman: Makes sense.  Implemented as so.

guacamole-ext/src/main/java/org/apache/guacamole/properties/TimeZoneGuacamoleProperty.java
mike-jumper: Strictly speaking, it isn't valid for a "custom time zone ID". From [the documentation for `TimeZone`](https://docs.oracle.com/javase/7/docs/api/java/util/TimeZone.html):

> You can use the getAvailableIDs method to iterate through all the supported time zone IDs. You can then choose a supported ID to get a TimeZone. If the time zone you want is not represented by one of the supported IDs, then a custom time zone ID can be specified to produce a TimeZone. The syntax of a custom time zone ID is:
>
> ```
>  CustomID:
>          GMT Sign Hours : Minutes
>          GMT Sign Hours Minutes
>          GMT Sign Hours
>  Sign: one of
>          + -
>  Hours:
>          Digit
>          Digit Digit
>  Minutes:
>          Digit Digit
>  Digit: one of
>          0 1 2 3 4 5 6 7 8 9
> ```

"GMT" is a valid time zone ID, but because `TimeZone` explicitly accepts it as one of the various supported IDs defined by `getAvailableIDs()`.

necouchman: So, are you good with the code as-is, or would you rather see a loop through `getAvailableIDs()` and put the RegEx back to requiring `GMT[+-]0`?

guacamole-ext/src/main/java/org/apache/guacamole/properties/TimeZoneGuacamoleProperty.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+public abstract class TimeZoneGuacamoleProperty
+        implements GuacamoleProperty<TimeZone> {
+    
+    /**
+     * A regex that matches valid variants of GMT timezones.
+     */
+    public static final Pattern GMT_REGEX =
+            Pattern.compile("^GMT([+-](0|00)((:)?00)?)?$",
+                    Pattern.MULTILINE);
```
mike-jumper, line 38: Why multiline?

guacamole-ext/src/test/java/org/apache/guacamole/properties/TimeZoneGuacamolePropertyTest.java
```diff
@@ -0,0 +1,230 @@
...
+    private static final List<String> TZ_CUSTOM_INVALID = Arrays.asList(
+            "GMT-9999",
+            "GMT+2500",
+            "GMT+29:30",
+            "GMT-1:99",
+            "GMT+10:65"
+    );
+    
+    private static final TimeZoneGuacamoleProperty WHERE_IN_WORLD =
+            new TimeZoneGuacamoleProperty() {
```
mike-jumper, line 107: Please document this very important property. ;)

necouchman: Ah, yes.  Done.

guacamole-ext/src/main/java/org/apache/guacamole/properties/TimeZoneGuacamoleProperty.java
necouchman: Oh, I think I was having issues with one of the previous RegExs not working as expected and threw that in to try it out. Forgot to go back and clean it up.

mike-jumper (changes requested):
I like the approach taken here - purely relying on a GMT regex seems simple and clean.

However, since this approach does not reference `getAvailableIDs()`, I think we should additionally test against the values returned by `getAvailableIDs()` within the unit test. It would be good to verify that `TimeZoneGuacamoleProperty` correctly accepts those values.

necouchman:
> However, since this approach does not reference getAvailableIDs(), I think we should additionally test against the values returned by getAvailableIDs() within the unit test. It would be good to verify that TimeZoneGuacamoleProperty correctly accepts those values.

Okay, I've added this test.

mike-jumper (approved):

PR #464, opened 2020-01-23
GUACAMOLE-936: Each LdapNetworkConnection must be closed or threads are leaked, causing Tomcat to fail to shutdown.

mike-jumper:


mike-jumper:
Yes, and it does work. I think I remember the case you describe, and I believe that was a different function like `bindAs()` where the connection is used later.

Here, I think that `ldapConnection` is no longer used once it leaves scope. I'l recheck.

mike-jumper:
Yep, `ldapConnection` is used only directly within the functions that it's passed to. It's not cached or stored for future use. Once it about to go out of scope here, it's a leak unless it's disconnected.

necouchman:
Sounds good.  Merging.

PR #465, opened 2020-01-23
GUACAMOLE-937: Return null on bindAs() failures. Rely on caller to interpret and handle failures.

mike-jumper:
Throwing `GuacamoleInvalidCredentialsException` breaks separation of concerns (`bindAs()` shouldn't assume that it's being used during login and that credentials given are the Guacamole user's credentials), and has unintended side effects (throwing subclasses of `GuacamoleUnauthorizedException` causes implicit session invalidation).


mike-jumper:
I've marked this as a draft as I've intentionally based it off the change from #464, as it would otherwise conflict.

mike-jumper:
I'll rebase to latest to avoid the duplicate commit there.

mike-jumper:
There. Back to having one commit.

necouchman:
Cool.  Merging.

PR #467, opened 2020-01-25
GUACAMOLE-939: Make the guacamole docker integration easier

theggers:
Patch 1

The new start.sh file pre read required varible from guacamole.properties to configure the tomcat web interface.
This make it mutch easier to start the docker container without so many options.

Thomas


mike-jumper:
Ah, whoops - didn't see that you did indeed create a PR. Sorry about that.

I'll close this for the reasons noted in the JIRA (`guacamole.properties` is already a config file and there already exists external tooling for maintaining the variables used by a set of images).

PR #468, opened 2020-01-25
GUACAMOLE-938: Use same bind/connect process for all LDAP connections, including referrals.

mike-jumper:
This change refactors the LDAP connection process such that the same pair of functions are always used for all connections:

* `createLDAPConnection()`, an internal function with several variants that simply constructs a new `LdapNetworkConnection` object having specific parameters set. The resulting object is *not* yet connected.
* `bindAs()`, which binds an `LdapNetworkConnection` using any one of several possible sources of credentials. The Apache Directory API will automatically connect before attempting to bind.

Part of this involved switching over to using `LdapConnectionConfig` internally for specifying the username/password of the bind operation, as the credentials used to bind will not be available for later reuse (such as when following a referral) unless they are associated with the `LdapConnectionConfig`. Constructing and using a `BindRequest` manually does not touch the `LdapConnectionConfig`.

### DIFF
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPConnectionService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPConnectionService.java
@@ -195,6 +195,15 @@ public class LDAPConnectionService {
         if (LdapUrl.LDAPS_SCHEME.equals(ldapUrl.getScheme()))
             encryptionMethod = EncryptionMethod.SSL;
 
+        // Use STARTTLS for otherwise unencrypted ldap:// URLs if the main
+        // LDAP connection requires STARTTLS
+        else if (confService.getEncryptionMethod() == EncryptionMethod.STARTTLS) {
+            logger.debug("Using STARTTLS for LDAP URL \"{}\" as the main LDAP "
+                    + "connection described in guacamole.properties is "
+                    + "configured to use STARTTLS.", url);
+            encryptionMethod = EncryptionMethod.STARTTLS;
+        }
+
         // If no post is specified within the URL, use the default port
         // dictated by the encryption method
         int port = ldapUrl.getPort();
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPConnectionService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPConnectionService.java
@@ -20,12 +20,10 @@
 package org.apache.guacamole.auth.ldap;
 
 import com.google.inject.Inject;
+import org.apache.directory.api.ldap.model.exception.LdapAuthenticationException;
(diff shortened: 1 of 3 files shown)


mike-jumper:
Some of this inherently improves the logging, as there were cases of errors being relegated to the debug level only due to the way things were structured. The restructured and central `bindAs()` will always log errors except those which are purely authentication-related (and thus expected to occur under normal circumstances).

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPConnectionService.java
```diff
@@ -104,151 +117,254 @@ private LdapNetworkConnection createLDAPConnection() throws GuacamoleException {
...
+        // dictated by the encryption method
+        int port = ldapUrl.getPort();
+        if (port < 1)
+            port = encryptionMethod.DEFAULT_PORT;
+
+        return createLDAPConnection(host, port, encryptionMethod);
 
-            // Explicitly start TLS if requested
-            if (confService.getEncryptionMethod() == EncryptionMethod.STARTTLS)
-                ldapConnection.startTls();
```
necouchman: Why are we not looking at the TLS encryption method, here?  Since there's not really a way that I know of within the LDAP URL to specify that we want TLS encryption, and since we assume no encryption if not set to LDAPS, it seems like we'd want to allow the user to use the URL, but override the encryption with TLS?  Or at least try?

mike-jumper: As far as normal binds go, we are still using TLS when requested - it's just set up via the `LdapConnectionConfig` and thus automatically applied when the Apache Directory API does its internal `connect()`:

https://github.com/apache/guacamole-client/blob/cb53b17afb30d5cbd1c4f744344dd2f32c6fe1b4/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPConnectionService.java#L107-L111

My understanding of the previous iteration of this code was that TLS was not applied to referrals, as the code being removed here (and replaced with `setUseTls(true)`) is specific to the old version of `bindAs()`, which was not called for referral URLs.

We could apply TLS for referrals, as well. It would be a simply matter of choosing that `EncryptionMethod` when delegating the creation of the LDAP connection for an LDAP referral:

https://github.com/apache/guacamole-client/blob/cb53b17afb30d5cbd1c4f744344dd2f32c6fe1b4/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/LDAPConnectionService.java#L193-L196

I agree this makes sense (a deployment which requires STARTTLS for the main connection should generally be expected to also require it for `ldap://` connections encountered in referrals). I'll add this.

mike-jumper:
OK - I've added automatic STARTTLS for LDAP URLs (when the main connection is configured for STARTTLS) and corrected the typos and copypasta.

necouchman:
Cool, looks good to me.

PR #469, opened 2020-01-30
GUACAMOLE-890: Security: Allow image to run as non-root user

benrubson:
Hi,

This solves https://issues.apache.org/jira/browse/GUACAMOLE-890, allowing Guacamole Docker image to run as non-root user.
~~For this, we symlink the `guacamole.war` file during the build, not at run time.
We also world-chmod the `temp` directory which strangely enough is not by default.
Finally, we prevent application from being unpacked, as non-root user can't write to the `webapps` dir.~~
For this, we change the `webapps` directory location to one created by the user itself.
We also create a `temp` directory which is then user-writable.

While here, as another security enhancement, I enabled the remote IP valve, as it's let's say impossible to do it properly at run time, once image is built. IMO not an issue as Docker image is intended to be used [behind a proxy](https://guacamole.apache.org/doc/gug/proxying-guacamole.html)...
it then also solves https://issues.apache.org/jira/browse/GUACAMOLE-1005.

As it's an important **security concern**, would be nice to have this in 1.1.0.
Would really be glad to test a new build an confirm everything's OK (it is on my locally-modified image).

Many thanks for your support 👍

### DIFF
--- a/guacamole-docker/bin/start.sh
+++ b/guacamole-docker/bin/start.sh
@@ -30,7 +30,7 @@
 
 GUACAMOLE_HOME_TEMPLATE="$GUACAMOLE_HOME"
 
-GUACAMOLE_HOME="$HOME/.guacamole"
+GUACAMOLE_HOME="/tmp/guacamole"
 GUACAMOLE_EXT="$GUACAMOLE_HOME/extensions"
 GUACAMOLE_LIB="$GUACAMOLE_HOME/lib"
 GUACAMOLE_PROPERTIES="$GUACAMOLE_HOME/guacamole.properties"
@@ -582,9 +582,18 @@ END
 ##
 start_guacamole() {
 
+    # User-only writable CATALINA_BASE
+    export CATALINA_BASE=$GUACAMOLE_HOME/tomcat
+    for dir in logs temp webapps work; do
+        mkdir -p $CATALINA_BASE/$dir
+    done
+    cp -R /usr/local/tomcat/conf $CATALINA_BASE
+
+    # Turn on the Remote IP Valve
+    sed -i 's|^\(\(\s\)\+\)</Host>|\1\2\2<Valve className="org.apache.catalina.valves.RemoteIpValve" />\n\n\1</Host>|' $CATALINA_BASE/conf/server.xml
+
     # Install webapp
-    rm -Rf /usr/local/tomcat/webapps/${WEBAPP_CONTEXT:-guacamole}
-    ln -sf /opt/guacamole/guacamole.war /usr/local/tomcat/webapps/${WEBAPP_CONTEXT:-guacamole}.war
+    ln -sf /opt/guacamole/guacamole.war $CATALINA_BASE/webapps/${WEBAPP_CONTEXT:-guacamole}.war
 
     # Start tomcat
     cd /usr/local/tomcat


necouchman:
Thanks @benrubson for the contribution.  I will try to review and maybe we can get this merged in.

Off the bat, though, your commit messages need to all have the GUACAMOLE-890: prefix, and the messages should be useful and descriptive.  I'd suggest squashing some of them together - four commits for 12 lines of code is probably not necessary - and one of them, the typo fix, can definitely be squashed into another commit.

guacamole-docker/bin/start.sh
```diff
@@ -583,8 +583,11 @@ END
 start_guacamole() {
 
     # Install webapp
-    rm -Rf /usr/local/tomcat/webapps/${WEBAPP_CONTEXT:-guacamole}
-    ln -sf /opt/guacamole/guacamole.war /usr/local/tomcat/webapps/${WEBAPP_CONTEXT:-guacamole}.war
+    if [ -n "$WEBAPP_CONTEXT" ]
+    then
+        rm /usr/local/tomcat/webapps/guacamole.war
+        ln -sf /opt/guacamole/guacamole.war /usr/local/tomcat/webapps/${WEBAPP_CONTEXT}.war
+    fi
```
necouchman: I'm not sure I understand how this both resolves the non-root user issue and also maintains the `WEBAPP_CONTEXT` capability?  It seems like, if linking at start time rather than build time causes non-root startup to fail, then anyone who is using `WEBAPP_CONTEXT` rather than the default `guacamole.war` is going to still have an issue running under a non-root account, correct??

benrubson:
Really good news, thank you very much for your feedback @necouchman 👍
Will you squash at merge time (using GitHub squash-merge function), or you you prefer me to do it now ?

guacamole-docker/bin/start.sh
benrubson: We can't do both, at least I did not find a way to do so.
We could `chmod 777 webapps` directory, but sounds like a security breach to me...
IMO there's no way to support `WEBAPP_CONTEXT` as non-root user...

Dockerfile
```diff
@@ -56,6 +56,14 @@ WORKDIR /opt/guacamole
 # Copy artifacts from builder image into this image
 COPY --from=builder /opt/guacamole/ .
 
+# Turn on the Remote IP Valve
+RUN sed -i 's+^\(      </Host>\)+        <Valve className="org.apache.catalina.valves.RemoteIpValve" />\n\1+' /usr/local/tomcat/conf/server.xml
+
+# Allow image to run as non-root user
+RUN sed -i 's/unpackWARs="true"/unpackWARs="false"/' /usr/local/tomcat/conf/server.xml
+RUN chmod 777 /usr/local/tomcat/temp
```
mike-jumper: This looks dangerous. Reducing privileges is a good thing, but removing existing restrictions from Tomcat's internals in order to achieve that seems a step in the wrong direction. That would be effectively granting privileges.

Assuming it is possible to configure the Tomcat portion of the image such that it can be run as a lower-privilege user, we should find a way to do that which doesn't lessen the security restrictions already in place.

Dockerfile
```diff
@@ -56,6 +56,14 @@ WORKDIR /opt/guacamole
 # Copy artifacts from builder image into this image
 COPY --from=builder /opt/guacamole/ .
 
+# Turn on the Remote IP Valve
+RUN sed -i 's+^\(      </Host>\)+        <Valve className="org.apache.catalina.valves.RemoteIpValve" />\n\1+' /usr/local/tomcat/conf/server.xml
```
mike-jumper: This looks like it might prove fragile, being both a regex parsing XML and a regex which is sensitive to changes in indentation. Perhaps there is a way to do this which is either aware of XML structure or aware of some degree of context? Applying a patch to `server.xml` might be an easy and more rigorous solution, if there isn't a good alternative that is XML-aware.

Dockerfile
benrubson: Thank you @mike-jumper for your great review.
I think you talk here about the temp directory permissions.
I though rather strange that, by default, `temp` directory was not world-writable.
For sure, being non-writable by non-root user prevents the non-unpacked app to start.
We could workaround this by setting the runtime `CATALINA_TMPDIR` env var to another writable dir, `/tmp` for instance. Let me then know if you think we could get rid of this chmod command.

mike-jumper: There's definitely a difference in purpose between `/tmp` (the system-wide temp directory) and `CATALINA_TMPDIR` (the temp directory dedicated to Tomcat). The former needs to be world-writable, yes, but I don't know what the security implications would be of making the Tomcat-specific temp directory world-writable.

The fact that:

* A system-wide, generic temp directory isn't already used for `CATALINA_TMPDIR`
* The Tomcat-specific directory isn't world-writable by default, but rather locked down for Tomcat's use.

indicates to me that this change should be made with caution. I think we would need to firmly establish that this directory is safe to be world-writable.

One alternative could be to create a Tomcat-specific subdirectory within `/tmp` or `/var/tmp` and point `CATALINA_TMPDIR` to that. As long as that subdirectory has the same permissions as the default `CATALINA_TMPDIR` normally would, this should be safe. Pointing `CATALINA_TMPDIR` directly to `/tmp` itself would not necessarily be safe unless it can be established that what you're already doing (`chmod 777`) is safe.

benrubson: I agree.
I then just pushed a much better way 👍

Dockerfile
benrubson: I've hardened the `sed` expression to be more robust to potential indentation changes.
It's an easy one-line update, sounds now robust IMO.
Of course feel free to tell me if I should rather go with a `patch` command.
Thank you !

manolan1:
> As this image is based off Docker's official image for Tomcat, and that image doesn't appear to support running as a different user, it may be worth looking into whether this has already been requested of that image and whether there are known issues preventing it.

Please see my comment on the issue. I believe this is unnecessary because the image *does* support running as non-root, but we are not using the right image.

Dockerfile
```diff
@@ -56,6 +56,13 @@ WORKDIR /opt/guacamole
 # Copy artifacts from builder image into this image
 COPY --from=builder /opt/guacamole/ .
 
+# Turn on the Remote IP Valve
+RUN sed -i 's+^\(\( *\)</Host>\)+\2  <Valve className="org.apache.catalina.valves.RemoteIpValve" />\n\n\1+' /usr/local/tomcat/conf/server.xml
```
manolan1: Why does the regex care about spaces before the closing tag? And, if it does, shouldn't it use a generic whitespace pattern rather than a space?

What is the impact of configuring the Remote IP Valve with default options? I have only ever seen it with proxy filtering enabled.

benrubson: Thank you for your review @manolan1 👍

> Why does the regex care about spaces before the closing tag?

To be sure, in addition to the `^` anchor, to select the real closing tag, and not for example a comment.

> And, if it does, shouldn't it use a generic whitespace pattern rather than a space?

I used space because the .xml file uses spaces, but I've switched to `\s`, even stronger. Thx.

benrubson: > What is the impact of configuring the Remote IP Valve with default options? I have only ever seen it with proxy filtering enabled.

If you use (as you should) your Docker instance behind a (Apache) reverse proxy :
- without the Remote IP Valve, you'll get the Docker interface IP in the Guacamole logs ;
- with the Remote IP Valve, you'll get the correct client (public) IP in the Guacamole logs.

If you use your Docker instance directly (strange as you won't provide SSL support etc...) :
- with or without the Remote IP Valve, you'll get the correct client (public) IP in the Guacamole logs. The only drawback is that a client with a **local** IP should be able to replace its IP by a fake one in Guacamole logs sending the correct headers (normally sent by the proxy) in his requests (`X-Forwarded-For`).

benrubson:
> > As this image is based off Docker's official image for Tomcat, and that image doesn't appear to support running as a different user, it may be worth looking into whether this has already been requested of that image and whether there are known issues preventing it.
> 
> Please see my comment on the issue. I believe this is unnecessary because the image _does_ support running as non-root, but we are not using the right image.

They do the `chmod 777 /usr/local/tomcat/temp` I initially proposed :)
We would then be able to get rid of the _User writable CATALINA_TMPDIR_ creation done in `start.sh` by this PR, but I think we would still need the few other modifications brought by this patch.
So perhaps we can keep the jre image for now, but up to the repo-maintainers :)

manolan1:
> > > As this image is based off Docker's official image for Tomcat, and that image doesn't appear to support running as a different user, it may be worth looking into whether this has already been requested of that image and whether there are known issues preventing it.
> > 
> > 
> > Please see my comment on the issue. I believe this is unnecessary because the image _does_ support running as non-root, but we are not using the right image.
> 
> They do the `chmod 777 /usr/local/tomcat/temp` I initially proposed :)
> We would then be able to get rid of the _User writable CATALINA_TMPDIR_ creation done in `start.sh` by this PR, but I think we would still need the few other modifications brought by this patch.
> So perhaps we can keep the jre image for now, but up to the repo-maintainers :)

Personally, I would prefer to switch to the supported image tag. Better to resolve this issue in the simplest way possible. *And* the jre8 image is not listed in the supported tags, so we should move off it anyway.

I am also slightly concerned about the unpack change unless someone has profiled the performance, but even if there is no performance impact, it isn't the simplest solution.

benrubson:
> I am also slightly concerned about the unpack change unless someone has profiled the performance, but even if there is no performance impact, it isn't the simplest solution.

Application is a little bit longer to start, with a  slightly higher memory footprint.
But then it runs flawlessly, I did not face any difference with the unpacked version.

benrubson:
I found a better way, let me test and commit :)

benrubson:
> I am also slightly concerned about the unpack change unless someone has profiled the performance, but even if there is no performance impact, it isn't the simplest solution.

No more an issue with my last commit, which is IMO a much better solution, it solves the various topics we discussed above 👍

guacamole-docker/bin/start.sh
benrubson: Finally solved with my last commit 👍

manolan1:
> > I am also slightly concerned about the unpack change unless someone has profiled the performance, but even if there is no performance impact, it isn't the simplest solution.
> 
> No more an issue with my last commit, which is IMO a much better solution, it solves the various topics we discussed above 👍


But I just don't see the point! I think this is, indeed, a much better solution and it preserves the WEBAPP_CONTEXT behaviour.

However, it is solving a problem that *we* shouldn't be solving. If we move to the supported tomcat image, then it runs with a non-root user without any of these changes. Surely that is *much* better than anything else.

By all means add the valve, but I don't think that is solving this same issue, so I believe it is a change that should be made under a different Jira.

benrubson:
> If we move to the supported tomcat image, then it runs with a non-root user without any of these changes. Surely that is much better than anything else.

I downloaded `tomcat:8.5-jdk8` and `webapps` dir is still non-world writable :
```
drwxr-sr-x 2 root staff  4096 Jan 23 05:12 webapps
```
So I'm not sure how the `.war` file can be linked there by a non-root user, we would for sure need to link it at build time.
But then I'm not sure how it will be extracted there in-place by a non-root process 😕

benrubson:
@manolan1, I just switched to the supported Tomcat image, worth it anyway.

benrubson:
I just squashed the various commits and pushed a unique one.

benrubson:
I just squashed the various commits and pushed a unique one.

If you can build a dev-image, I would be glad to test and confirm that everything works as expected.
Perhaps we could then go with a new (1.1.1 ?) Docker image containing these security patches / enhancements, if you accept them / if we agree on them.

Many thanks again 👍

Dockerfile
```diff
@@ -25,7 +25,7 @@
 # such as `--build-arg TOMCAT_JRE=jre8-alpine`
 #
 ARG TOMCAT_VERSION=8.5
-ARG TOMCAT_JRE=jre8
+ARG TOMCAT_JRE=jdk8
```
necouchman: Why?

benrubson: As discussed above with @manolan1, and also based on his comment in the [issue itself](https://issues.apache.org/jira/browse/GUACAMOLE-890), this is a more uptodate / supported Tomcat image, which also helps with non-root user, as it contains some of the fixes we need (especially the permissions on the `temp` folder).

necouchman: Hmmm...okay...this is going to increase the size of this Docker image pretty substantially, I would imagine, so I'm a little skeptical that this is the right thing to do.  Sorry I missed it in the other conversation...

necouchman: I also don't think a world-writable directory is the right way to go - the directory should have the correct ownership and permissions, not just the ones that work because we've blown everything open.  If we're trying to improve security with this issue, making something world-writable seems contradictory to that effort.

benrubson: I've then reverted to original `jre` image, and created a user-writable `temp` dir.
This was what I initially pushed before switching to `sdk` image.

manolan1:
You all seem to have missed the point! ;)

Regardless of this change, we should be switching to the jdk8 suffix. It is
not a question of size, it is a question of support. I don't see that there
is an option.

The jre8 suffix is NOT an officially supported tomcat container. See here
for supported tags: https://hub.docker.com/_/tomcat?tab=description or
https://github.com/docker-library/official-images/blob/master/library/tomcat

The jre8 tag *is* available. It was last updated 8 months ago vs 23 Dec for
the jdk8 image. Unsurprisingly, the jre8 image is not part of the tomcat
docker build process, as you can see from any recent build reports.

Here is the pull request for that change (
https://github.com/docker-library/tomcat/pull/158). I am sure the
requirement for the jdk does not affect us, but I really think we should be
using a supported image and all jre images have been removed.

M.
.

benrubson:
Feel free to agree all together on the best solution and let me know so that I can finally tune this PR accordingly :)
Whatever the final choice, I'll be glad to test :+1:

benrubson:
OK, I've then switched again to the `jdk` image.
And let the creation of the `temp` dir, which is then writable by the user only.
This whole PR should then suit all thoughts / needs :)
Many thanks again 👍

benrubson:
Some news regarding this @mike-jumper @necouchman @manolan1 ?
Thank you 👍

manolan1:
My view is that this must be a common tomcat problem and should be solved
in the tomcat container. However, what they have done doesn't seem to work
for us. I haven't had a chance to look at why not since I thought that was
exactly what they were trying to achieve. I don't need this functionality
right now, so my selfish view would have been to leave it until it is
sorted upstream. But I know that won't work for everyone.

I also think we should definitely move to a supported tomcat container,
though whether that should actually be done under this Jira or under a
different one is a moot point.

I realise that is frustratingly vague, but it is where my thoughts are.

M.
.

benrubson:
I just pushed a better approach.
`start.sh` now creates a whole new `CATALINA_BASE`.
It's then user-writable only, not world-writable, better for security as already stated above.
And this method follows Tomcat guidelines.
Seems perfect IMO.

Regarding `jdk` vs `sdk` image, let's discuss this in another Jira / PR.

Let's merge this @mike-jumper @necouchman ?

Thx 👍

guacamole-docker/bin/start.sh
```diff
@@ -30,7 +30,7 @@
 
 GUACAMOLE_HOME_TEMPLATE="$GUACAMOLE_HOME"
 
-GUACAMOLE_HOME="$HOME/.guacamole"
+GUACAMOLE_HOME="/tmp/guacamole"
```
necouchman: I really don't like this part of it - making the `GUACAMOLE_HOME` into a temp directory doesn't seem wise to me.  Maybe I'm thinking of this too much from a non-docker perspective, but what if the user decides to try to map through `/etc/guacamole` to a specific path?

I don't know if there's any other solution for this in the Docker Tomcat world, but this approach just makes me twitch :-).

benrubson: First, this line is due to the fact the user chosen to run the container may not have a home directory into the container, which would then not start.
Note however that the `guacamole` directory, inside the `tmp` directory, is only reachable by the user running the container, as shown below, from a container running this PR :
```
$ ls -l /tmp/
total 12
drwx------ 5 1024 1024 4096 Mar  4 13:02 guacamole
```

> but what if the user decides to try to map through `/etc/guacamole` to a specific path?

I also do this in my `docker-compose.yml` file, not an issue at all :
```
services:
  guacamole:
    image: guacamole/guacamole:1.1.0
    volumes:
      - ./etc:/etc/guacamole:ro
```

necouchman: My issue with the directory being placed in /tmp is not the accessibility of the directory to other users/processes, it's the fact that a configuration directory (or webapp deployment directory) is in /tmp - this doesn't make sense to me.  Configuration directories should be in configuration directory locations - not in any place that happens to work because we have write access to it :-).

I tend to agree with @manolan1, here - Guacamole can't be the only piece of software fighting this issue.  Is what you've done here a commonly-implemented solution among other Tomcat + Docker web applications that run as non-root users?

I'm willing to defer on this one if @mike-jumper or any of the other project members think we should just move forward, but this still just feels a little off to me.

benrubson: > Guacamole can't be the only piece of software fighting this issue. Is what you've done here a commonly-implemented solution among other Tomcat + Docker web applications that run as non-root users?

It is what Tomcat documentation recommends, it's the goal of `CATALINA_BASE` env var we then use in this PR.
IMO, what is done in the Tomcat Docker image, 777-chmoding application / configuration / work / tmp... default directories is not the best way to go.
Do you, on a standard installation, as non-root user, have write access to these directories ? No :-)
However you're free to set your own `CATALINA_BASE` directory, in your home directory for example, and start a new Tomcat instance over it.
As we do here in this PR.

> Configuration directories should be in configuration directory locations - not in any place that happens to work because we have write access to it :-)

We could `chmod 777 /home` at image build time (in `Dockerfile`), and then use `GUACAMOLE_HOME="/home/guacamole"` instead of `GUACAMOLE_HOME="/tmp/guacamole"`, but the whole idea would remain the same.
Keep in mind that whatever is done when Docker container starts will be lost when it will stop, and will have to be done again (this is what `start.sh` does).
So, configuration which is done when container starts, will be lost when it will stop, wherever it stands, in `/tmp`, `/etc`, `/home` etc...

I've just verified, there's nothing in `/home` in the Tomcat Docker image.
We could then safely `chmod 777` it and move `GUACAMOLE_HOME` into it, if you're more confortable with this.

I think this PR is the proper way to go to address this (non-)root security issue.
Of course I run it since the beginning without any issue.

Thank you again :+1:

benrubson: > Guacamole can't be the only piece of software fighting this issue. Is what you've done here a commonly-implemented solution among other Tomcat + Docker web applications that run as non-root users?

It is what Tomcat documentation recommends, it's the goal of `CATALINA_BASE` env var we then use in this PR.
IMO, what is done in the Tomcat Docker image, 777-chmoding application / configuration / work / tmp... default directories is not the best way to go.
Do you, on a standard installation, as non-root user, have write access to these directories ? No :-)
However you're free to set your own `CATALINA_BASE` directory, in your home directory for example, and start a new Tomcat instance over it.
As we do here in this PR.

> Configuration directories should be in configuration directory locations - not in any place that happens to work because we have write access to it :-)

We could `chmod 777 /home` at image build time (in `Dockerfile`), and then use `GUACAMOLE_HOME="/home/guacamole"` instead of `GUACAMOLE_HOME="/tmp/guacamole"`, but the whole idea would remain the same.
Keep in mind that whatever is done when Docker container starts will be lost when it will stop, and will have to be done again (this is what `start.sh` does).
So, configuration which is done when container starts, will be lost when it will stop, wherever it stands, in `/tmp`, `/etc`, `/home` etc...

I've just verified, there's nothing in `/home` in the Tomcat Docker image.
We could then safely `chmod 777` it and move `GUACAMOLE_HOME` into it, if you're more confortable with this.

We could also come back to the original `GUACAMOLE_HOME="$HOME/.guacamole"`, thus getting rid of the hard-coded `/tmp` path.
Everything will run as today, from inside `/root` directory, for those running the container as `root`.
The others, switching to a different user, will have to set the `$HOME` environment variable to a writable directory.
Could then suit everyone's needs here.

I think this PR is the proper way to go to address this (non-)root security issue.
Of course I run it since the beginning without any issue.

Thank you again :+1:

benrubson: So @necouchman ? :)
As per my previous answers, do you think we could go like this (or with one of the 2 modifications proposed above, if you prefer) ?
Many thanks 👍

mike-jumper:
> However, it is solving a problem that _we_ shouldn't be solving. If we move to the supported tomcat image, then it runs with a non-root user without any of these changes. Surely that is _much_ better than anything else.

I wholeheartedly agree.

> By all means add the valve, but I don't think that is solving this same issue, so I believe it is a change that should be made under a different Jira.

Yep.

benrubson:
> > However, it is solving a problem that _we_ shouldn't be solving. If we move to the supported tomcat image, then it runs with a non-root user without any of these changes. Surely that is _much_ better than anything else.
> 
> I wholeheartedly agree.

> Allowing the image to run as non-root is a worthy goal, but one that should be implemented within the Tomcat image. Once that's the case (it sounds like it already is?), (...)

No, as already stated [here](https://github.com/apache/guacamole-client/pull/469#issuecomment-580483759), it's not.
Some details again.
Please read carefully, sorry I'll certainly repeat / reuse some arguments given above.

Here is then the directory structure from the last `tomcat:8.5-jdk8`  image :
_(as a ref base image update PR from `jre` to `jdk` is here : #485)_

```
# ls -l /usr/local/tomcat/
-rw-r--r-- 1 root root 19318 Mar 11 10:06 BUILDING.txt
-rw-r--r-- 1 root root  5408 Mar 11 10:06 CONTRIBUTING.md
-rw-r--r-- 1 root root 57011 Mar 11 10:06 LICENSE
-rw-r--r-- 1 root root  1726 Mar 11 10:06 NOTICE
-rw-r--r-- 1 root root  3255 Mar 11 10:06 README.md
-rw-r--r-- 1 root root  7136 Mar 11 10:06 RELEASE-NOTES
-rw-r--r-- 1 root root 16262 Mar 11 10:06 RUNNING.txt
drwxr-xr-x 2 root root  4096 Mar 17 23:14 bin
drwxr-xr-x 1 root root  4096 Mar 27 21:39 conf
drwxr-xr-x 2 root root  4096 Mar 17 23:14 include
drwxr-xr-x 2 root root  4096 Mar 17 23:13 lib
drwxrwxrwx 1 root root  4096 Mar 27 21:39 logs
drwxr-xr-x 3 root root  4096 Mar 17 23:14 native-jni-lib
drwxrwxrwx 2 root root  4096 Mar 17 23:13 temp
drwxr-xr-x 2 root root  4096 Mar 17 23:13 webapps
drwxr-xr-x 7 root root  4096 Mar 11 10:04 webapps.dist
drwxrwxrwx 2 root root  4096 Mar 11 10:03 work
```

As you can see, `logs`, `temp` and `work` directories of the default `CATALINA_BASE` have been made world-writable.
But still, the `webapps` directory is not world-writable, so a non-root user won't be able to link `guacamole.war` into it. So the Guacamole `start.sh` script will [fail](https://github.com/apache/guacamole-client/blob/1.1.0/guacamole-docker/bin/start.sh#L579).
So, there's still something missing.

You could then propose to ask the Tomcat image maintainers to `chmod 777 webapps`.
But, and as you already stated [here](https://github.com/apache/guacamole-client/pull/469#discussion_r373092391) @mike-jumper, the goal (of running as non-root user) should not be achieved removing restrictions.
In other words, from a security point of view, the method used by the image maintainers does not seem to be the good one.

To confirm this, here's now the default directory structure from a Debian 10 Tomcat installation :
```
# ls -l /var/lib/tomcat9/
lrwxrwxrwx 1 root   root     12 Jun 13  2019 conf -> /etc/tomcat9
drwxr-xr-x 2 tomcat tomcat 4096 Jun 13  2019 lib
lrwxrwxrwx 1 root   root     17 Jun 13  2019 logs -> ../../log/tomcat9
drwxr-xr-x 2 root   root   4096 Mar 27 22:51 policy
drwxrwxr-x 3 tomcat tomcat 4096 Mar 27 22:51 webapps
lrwxrwxrwx 1 root   root     19 Jun 13  2019 work -> ../../cache/tomcat9
# ls -ld /var/log/tomcat9/ /var/cache/tomcat9/
drwxr-x--- 3 tomcat tomcat 4096 Mar 27 22:51 /var/cache/tomcat9/
drwxr-s--- 2 tomcat adm    4096 Mar 27 22:51 /var/log/tomcat9/
```

As you can see here, `logs`, `work` and `webapps` directories are only writable by the `tomcat` user.
It then does not makes sense to make them world-writable so that another user can start its application... Would be a messy / crappy configuration...

On the contrary, as explained [above](https://github.com/apache/guacamole-client/pull/469#discussion_r392684827), and as explained in the [Tomcat documentation](http://tomcat.apache.org/tomcat-8.5-doc/introduction.html#CATALINA_HOME_and_CATALINA_BASE), we should, in this case, create and use a dedicated proper `CATALINA_BASE` directory structure.

This is what this PR does...
It creates the 5 runtime directories which are then readable / writable only by the user running the application. Link the application into it, and start from it...

Sincerely, it really looks like the way to go.
We follow the Tomcat documentation, the resulting PR is really tiny, it has no side effect, no runtime or performance impact, it allows to run as non-root user, and enforces directory structure security...
I don't see why we wouldn't go for it...

Thank you again if you could then reconsider this 👍

benrubson:
Any thoughts regarding my detailed explanation [above](https://github.com/apache/guacamole-client/pull/469#issuecomment-605352752) ?
Thank you !

benrubson:
5 months have passed since the beginning of this PR...
Any thoughts then regarding my explanation [above](https://github.com/apache/guacamole-client/pull/469#issuecomment-605352752), which gives all the details to understand why we came to this solution ?
Thank you again for your support !

PR #470, opened 2020-02-01
GUACAMOLE-124: Add full-screen action

madmath03:
[GUACAMOLE-124](https://issues.apache.org/jira/browse/GUACAMOLE-124): Add full-screen action

Signed-off-by: mathieu.brunot <mathieu.brunot@monogramm.io>


guacamole/src/main/webapp/app/navigation/directives/guacUserMenu.js
```diff
@@ -134,6 +134,45 @@ angular.module('navigation').directive('guacUserMenu', [function guacUserMenu()
...
+                    if (elem.exitFullscreen) {
+                        elem.exitFullscreen();
+                    } else if (elem.mozCancelFullScreen) { /* Firefox */
+                        elem.mozCancelFullScreen();
+                    } else if (elem.webkitExitFullscreen) { /* Chrome, Safari and Opera */
+                        elem.webkitExitFullscreen();
+                    } else if (elem.msExitFullscreen) { /* IE/Edge */
+                        elem.msExitFullscreen();
+                    }
+                }
```
necouchman: Wow, this is fantastic - no reflection on you, but apparently no standard way to enter full-screen mode.  :rofl:

mike-jumper: There may be a cleaner way to shim this. For example:

https://github.com/apache/guacamole-client/blob/b51d2a88b8415738f6fff19cc99bc2b3e4690da2/guacamole-common-js/src/main/webapp/modules/AudioRecorder.js#L189-L197

madmath03: I have not done much testing yet, which is why I am keeping this as a draft for now.The few tests i have done so far were pretty good, ie. VNC/SSH would just resize as you would expect.
I need to test RDP, and also do tests on more browsers / devices.

And yeah, there is a standard way to enter / exit fullscreen but, as usual, not all  browsers on the market moved to it yet :disappointed:

madmath03: Interesting! I will test if I can shim this like you proposed.Thanks.

madmath03: @mike-jumper That should be good now

madmath03:
> Overall this looks pretty solid to me. My only question is how much testing you've done with scaling/resizing with this? Presumably when it goes to full-screen the screen resolution changes, which, depending on the protocol in use for the connection, will result in either an on-the-fly resize (RDP, maybe SSH/Telnet), a disconnect/reconnect (RDP), or some extra black space around the screen area. What sort of behavior have you seen?

I did some more tests on different devices / browsers and connections. Here are the results:

| Device / Browser | :checkered_flag: Windows RDP | :penguin: Linux SSH | :penguin: Linux VNC (TigerVNC) |
| ------------------- | --------------- | ----------- | ------------------------ |
| :checkered_flag: Windows / Chromium | :warning: | :white_check_mark: | :white_check_mark: |
| :checkered_flag: Windows / Firefox | :warning: | :white_check_mark: | :white_check_mark: |
| :checkered_flag: Windows / Edge | :warning: | :white_check_mark: | :white_check_mark: |
| :checkered_flag: Windows / IE 11 | :no_entry_sign: | :no_entry_sign: | :no_entry_sign: |
| :robot: Android Tablet / Chromium | :warning: | :white_check_mark: | :white_check_mark: |
| :robot: Android Tablet / Firefox | :warning: | :white_check_mark: | :white_check_mark: |
| :green_apple: iPhone / Safari | :x: | :x: | :x: |

-    :white_check_mark: : Fullscreen works properly and connection is immediately resized
-    :warning: : Fullscreen button works but Windows RDP connection displays empty space and needs to be restarted to be resized (disconnect / reconnect, no need to go back to home)
-    :no_entry_sign: : Fullscreen button works but only displays a black screen in connection, whatever the connection type
-    :x: : Fullscreen button does not work

So basically:
* IE 11 and Safari do not work
* RDP connection needs reconnect (seems to be the normal behavior of RDP)
* everything else is good

@necouchman & @mike-jumper : I do not have a Mac to debug behavior on Safari, and I do not believe IE 11 is relevant anymore, so I think that's good enough to be used, or at least to be reviewed.

necouchman:
@madmath03: Please note that all your commits need to have the "GAUCAMOLE-124" tag on them.  Also, might want to consider squashing the commits into a smaller number - five is a lot of commits for the small amount of code that is being changed.

madmath03:
@necouchman done

madmath03:
Hi @mike-jumper ,

Do you think this could be part of the 1.2.0 release ?

guacamole/src/main/webapp/app/navigation/directives/guacUserMenu.js
```diff
@@ -134,6 +134,48 @@ angular.module('navigation').directive('guacUserMenu', [function guacUserMenu()
...
 
+            /**
+             * Requests fullscreen for the Guacamole page.
+             */
+            $scope.fullscreen = function fullscreen() {
+                var elem = document.documentElement;
+                var fullscreenElem = !(!document.fullscreenElement &&
+                    !document.msFullscreenElement &&
+                    !document.mozFullScreenElement &&
+                    !document.webkitFullscreenElement);
```
mike-jumper: `!!(foo || bar || baz)` might be more readable.

guacamole/src/main/webapp/app/navigation/directives/guacUserMenu.js
```diff
@@ -134,6 +134,48 @@ angular.module('navigation').directive('guacUserMenu', [function guacUserMenu()
...
+                    if (!elem.requestFullscreen) {
+                        elem.requestFullscreen = (elem.mozRequestFullScreen
+                                || elem.webkitRequestFullscreen
+                                || elem.msRequestFullscreen).bind(elem);
+                    }
+
+                    if (elem.requestFullscreen) {
+                        elem.requestFullscreen();
+                    }
+                } else {
```
mike-jumper: Please don't cuddle the `else`: http://guacamole.apache.org/guac-style/#braces

guacamole/src/main/webapp/app/navigation/directives/guacUserMenu.js
```diff
@@ -134,6 +134,48 @@ angular.module('navigation').directive('guacUserMenu', [function guacUserMenu()
...
+                    !document.msFullscreenElement &&
+                    !document.mozFullScreenElement &&
+                    !document.webkitFullscreenElement);
+
+                if (!fullscreenElem) {
+                    if (!elem.requestFullscreen) {
+                        elem.requestFullscreen = (elem.mozRequestFullScreen
+                                || elem.webkitRequestFullscreen
+                                || elem.msRequestFullscreen).bind(elem);
+                    }
```
mike-jumper: It would be better to retrieving `document.documentElement` and ensure its `requestFullScreen()` and `exitFullscreen()` are defined just once, rather than each time `$scope.fullscreen()` is invoked. If these checks were moved into the body of the controller rather than the body of this function, I think this would become more readable (and would happen to also do less work).

guacamole/src/main/webapp/app/navigation/directives/guacUserMenu.js
```diff
@@ -134,6 +134,48 @@ angular.module('navigation').directive('guacUserMenu', [function guacUserMenu()
...
+
+                    if (elem.requestFullscreen) {
+                        elem.requestFullscreen();
+                    }
```
mike-jumper: This appears to anticipate that there may not be any `requestFullscreen()` implementation, even browser-specific prefix versions. That's good, but such a case would also cause the previous call to `bind()` to fail, as `elem.mozRequestFullScreen || elem.webkitRequestFullscreen || elem.msRequestFullscreen` would not evaluate to something that has `bind()`.

I'd recommend avoiding both of these issues entirely, adding an anonymous stub function at the end of that chain of `||`, thus ensuring there is always a version of `requestFullscreen()` (even if it does nothing).

guacamole/src/main/webapp/app/navigation/directives/guacUserMenu.js
```diff
@@ -134,6 +134,48 @@ angular.module('navigation').directive('guacUserMenu', [function guacUserMenu()
                 return authenticationService.isAnonymous();
             };
 
+            /**
+             * Requests fullscreen for the Guacamole page.
```
mike-jumper: This function doesn't just request that the Guacamole page enter fullscreen mode, but toggles fullscreen.

guacamole/src/main/webapp/app/navigation/directives/guacUserMenu.js
madmath03: Ok.

Do you prefer the stub function to really do nothing or is it okay if it logs a warning ?

guacamole/src/main/webapp/app/navigation/directives/guacUserMenu.js
madmath03: @mike-jumper If we move the tests for `requestFullScreen()`  and `exitFullscreen()` into the body, then we could detect if the current browser has support for fullscreen before adding the function to the menu.

Should we still display the fullscreen menu or is it better to hide it ?

mike-jumper: Definitely better to hide it.

guacamole/src/main/webapp/app/navigation/directives/guacUserMenu.js
mike-jumper: Assuming this is still relevant given what you mentioned regarding hiding the fullscreen option if it can't be used, I have no objection to logging a warning.

madmath03: Given the fact that we will hide the fullscreen option if not supported, I will as well log a warning.

madmath03:
hello @mike-jumper 

Let me know if the corrections applied are valid for you.

guacamole/src/main/webapp/app/navigation/directives/guacUserMenu.js
```diff
@@ -134,6 +135,64 @@ angular.module('navigation').directive('guacUserMenu', [function guacUserMenu()
...
+                var FULLSCREEN_ACTION = {
+                    name      : 'USER_MENU.ACTION_FULLSCREEN',
+                    className : 'fullscreen',
+                    callback  : $scope.fullscreen
+                };
+    
+                /**
+                 * Add fullscreen action for the current user.
+                 */
+                $scope.actions = $scope.actions.concat( FULLSCREEN_ACTION );
```
mike-jumper: Running this only within `if (!!requestFullscreen && !!exitFullscreen) { ... }` will result in the fullscreen option being available only if there is a vendor-prefixed version of those functions available. If a browser implements the standard but no vendor-prefixed version, the option will not be present.

guacamole/src/main/webapp/app/navigation/directives/guacUserMenu.js
```diff
@@ -160,9 +219,9 @@ angular.module('navigation').directive('guacUserMenu', [function guacUserMenu()
             };
 
             /**
-             * All available actions for the current user.
+             * Add logout action for the current user.
              */
-            $scope.actions = [ LOGOUT_ACTION ];
+            $scope.actions = $scope.actions.concat( LOGOUT_ACTION );
```
mike-jumper: Rather than repeatedly recreating the `$scope.actions` array through calling `concat()`, perhaps it would be better to keep the original initialization to `[ LOGOUT_ACTION ]` and simply `unshift(FULLSCREEN_ACTION)` when needed?

guacamole/src/main/webapp/app/navigation/directives/guacUserMenu.js
```diff
@@ -134,6 +135,64 @@ angular.module('navigation').directive('guacUserMenu', [function guacUserMenu()
...
+                };
+    
+                /**
+                 * Action which requests fullscreen for the Guacamole page.
+                 */
+                var FULLSCREEN_ACTION = {
+                    name      : 'USER_MENU.ACTION_FULLSCREEN',
+                    className : 'fullscreen',
+                    callback  : $scope.fullscreen
+                };
```
mike-jumper: While I agree there is no need for these to be declared except when fullscreen is supported, this `if` block and the controller as a whole might be more readable were `FULLSCREEN_ACTION` and `$scope.fullscreen()` declared elsewhere, ideally grouped logically with `LOGOUT_ACTION` and its callback.

guacamole/src/main/webapp/app/navigation/directives/guacUserMenu.js
madmath03: You're right. I should have added the standard version in `requestFullscreen` and `exitFullscreen`.
Thanks.

madmath03:
hello @mike-jumper

I've applied the modifications you requested (not a 100% sure about the last one though).
Let me know if there is anything else.

madmath03:
@mike-jumper I've tested this on the same devices I mentioned initially and it all seems to work as expected.

Let me know if you need more changes.

echu2013:
@madmath03 : I've tested it on `Version 80.0.3987.149 (Official Build) (64-bit)` under `Windows 10 Pro 64-bit` and it's working OK!
IMO there should be some sort of toggle switch or visual change when full-screen toggled on, like changing "Full Screen" value to "Turn off Full Screen".

madmath03:
> @madmath03 : I've tested it on `Version 80.0.3987.149 (Official Build) (64-bit)` under `Windows 10 Pro 64-bit` and it's working OK!

Glad to hear that 😃 

> IMO there should be some sort of toggle switch or visual change when full-screen toggled on, like changing "Full Screen" value to "Turn off Full Screen".

That might be nice indeed, but since the action name and CSS classes are constants, I'm not sure if there is a way to do that easily though...
@necouchman & @mike-jumper is that something possible and if so how ?

madmath03:
@mike-jumper anything blocking this to be merged ?

madmath03:
@mike-jumper & @necouchman anything blocking this to be merged ?

echu2013:
@madmath03 I can say that for me just worked fine! Just a minor addition I've made to your code, don't show toggle is running Internet explorer as it renders a black screen (known bug, but it's IE..)
```diff
--- a/guacamole/src/main/webapp/app/navigation/directives/guacUserMenu.js
+++ b/guacamole/src/main/webapp/app/navigation/directives/guacUserMenu.js
@@ -191,6 +191,17 @@ angular.module('navigation').directive('guacUserMenu', [function guacUserMenu()
              */
             $scope.actions = [ LOGOUT_ACTION ];
 
+
+            // Check if Internet Explorer UserAgent
+            var isIE = false;
+            var ua = window.navigator.userAgent;
+            var old_ie = ua.indexOf('MSIE ');
+            var new_ie = ua.indexOf('Trident/');
+
+            if ((old_ie > -1) || (new_ie > -1)) {
+                isIE = true;
+}
+
             // Initialize fullscreen functions
             var docElem = document.documentElement;
             var requestFullscreen = docElem.requestFullscreen
@@ -201,8 +212,8 @@ angular.module('navigation').directive('guacUserMenu', [function guacUserMenu()
                 || document.mozCancelFullScreen
                 || document.webkitExitFullscreen
                 || document.msExitFullscreen;
-
-            if (!!requestFullscreen && !!exitFullscreen) {
+            // If running Internet Explorer DONT include FullScreen toggle (Black Screen is produced when toggled)
+            if (!!requestFullscreen && !!exitFullscreen && !isIE ) {
                 // Bind browser-specific fullscreen functions
                 if (!docElem.requestFullscreen) {
                     docElem.requestFullscreen = (requestFullscreen).bind(docElem);

```

madmath03:
> Just a minor addition I've made to your code, don't show toggle is running Internet explorer as it renders a black screen (known bug, but it's IE..)

Hi @echu2013 

I added your suggestion to this PR.

@necouchman & @mike-jumper let me know what you guys think about this.

echu2013:
@madmath03 great job! Thanks let's hope they approve it

guacamole/src/main/webapp/app/navigation/directives/guacUserMenu.js
```diff
@@ -164,6 +191,40 @@ angular.module('navigation').directive('guacUserMenu', [function guacUserMenu()
              */
             $scope.actions = [ LOGOUT_ACTION ];
 
+            // Check if Internet Explorer UserAgent
+            var ua = window.navigator.userAgent;
+            var isIE = (ua.indexOf('MSIE ') > -1) || (ua.indexOf('Trident/') > -1);
```
mike-jumper, line 196: > ... don't show toggle is running Internet explorer as it renders a black screen (known bug, but it's IE..)

I'm wary of browser sniffing like this unless absolutely necessary. @echu2013, can you point to where the bug is documented?

madmath03: I found several messages on forums regarding black screen with Internet Explorer but no official documentation by Microsoft of the issue.

@mike-jumper would you prefer the latest commit to be removed ?

mike-jumper: Sure, that would be OK with me.

madmath03: @mike-jumper done. Added a commit to revert the latest modification.

madmath03:
@mike-jumper can you review this PR ?

necouchman:
>  Despite Guacamole otherwise having full control of the Esc key, that control is lost while in full screen. This is a bit annoying when using an application which leverages Esc, though I'm also not sure anything can be done about this.

We could try to roll https://issues.apache.org/jira/browse/GUACAMOLE-989 into this??

madmath03:
@mike-jumper 

Regarding point 3, as you've mentioned, I'm not sure there is anything that can be done about this _here_.

For the point 1 though, I do not agree with you. It's actually not related to the current page since it switches the whole browser to fullscreen mode.
Also, if we did make it the way you say, we could not enter fullscreen before opening a connection. From my experience, some connections, like RDP, work better if you're already in fullscreen when opening it. Switching on/off fullscreen during connection will sometimes "_mess up_" the resolution and resize.
So I would think it's better to keep as it is.

what do you think @necouchman ?

I can take a look at point 2 though and see how it's done.

madmath03:
@necouchman & @mike-jumper just pinging to get some news on this old PR

depawlur:
I hope this is not dead. I did test this before 1.2.0 release. It's 1.3.0 already. 
This also sounds amazing https://issues.apache.org/jira/browse/GUACAMOLE-989

necouchman:
@depawlur: I don't think it's dead, just didn't get finished up in time to make the 1.3.0 release. Hopefully we can add it to the next one...

necouchman:
@madmath03:

> For the point 1 though, I do not agree with you. It's actually not related to the current page since it switches the whole browser to fullscreen mode.

I agree with @mike-jumper on the placement of this item - while it is true it puts the entire browser into fullscreen mode, the context in which the user executes the action is a specific connection, and so I think it makes more sense to have it grouped with connection-specific options.

> Also, if we did make it the way you say, we could not enter fullscreen before opening a connection.

Ah, I think I see what you're getting at - you've added it to the Guacamole menu where it displayed either here in the hidden menu or in the menu on the home screen, I'm guessing? So that someone can go to the menu on the home screen and put the entire screen in fullscreen mode, and then start a connection?  I guess I'm not opposed to having this functionality available on the Home screen, assuming we can agree on the correct place for it. However, a couple of points, here:
* It _is_ still possible to put the browser into fullscreen mode without this option in the menu on the Home screen - most browsers have both a menu item and a shortcut key (F11 or F12, I think) that trigger fullscreen mode. A menu item on the home screen is not required to enable this - indeed, you can do it today with no modifications to the Guacamole interface.
* Most modern versions of the protocols supported by Guacamole deal with the screen resizing properly when it happens after connection time. RDP has had the functionality for several years to do a dynamic resize without terminating the connection. I cannot remember what version of Windows added this support - maybe 8? - but it does work. Other RDP servers may not support it (doubt xrdp does), but at least modern Windows versions do. SSH and Kubernetes should also be fine  - VNC currently does not support any resolution negotiation between client and server, let alone dynamic, and I'm kind of doubt most Telnet servers deal with resolution changes all that seamlessly.

madmath03:
@necouchman thanks for the feedback. I was really expecting this before doing any further modifications to this PR 😃 

> you've added it to the Guacamole menu where it displayed either here in the hidden menu or in the menu on the home screen, I'm guessing? So that someone can go to the menu on the home screen and put the entire screen in fullscreen mode, and then start a connection?

Yes, that's exactly what I had in mind and the reason why I'm not too much of a fan of making it connection-specific.

> I guess I'm not opposed to having this functionality available on the Home screen, assuming we can agree on the correct place for it.

In that case, how and where should it be placed ?

> * It _is_ still possible to put the browser into fullscreen mode without this option in the menu on the Home screen - most browsers have both a menu item and a shortcut key (F11 or F12, I think) that trigger fullscreen mode.

That's only true for desktop browsers. I developped this PR when I was testing Guacamole on an Android tablet: it worked fine, but there is no way on tablets or smartphones to switch to Fullscreen making it not really practical...
Since the fullscreen API still works on mobile, it's just that there is no menu or shortcut for it, this PR would be really nice.

BTW: another solution but much more complicated would be to make Guacamole into a [PWA](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps) (basically requires HTTPS, manifest.json and a Service Worker) so that we could install Guacamole and run it in fullscreen by default.

necouchman:
@mike-jumper Any further thoughts or guidance on placement of this item?

@madmath03 Fair point about the non-desktop browsers, this makes sense. Let's see if we can reach consensus on where such an option would be placed.

necouchman:
@mike-jumper This one has languished a bit - any further feedback, here?

@madmath03 You still around and willing to continue to work on this if we can get you some guidance?

mike-jumper:
@necouchman I think this may be effectively superseded by #695, which was merged for [GUACAMOLE-1525](https://issues.apache.org/jira/browse/GUACAMOLE-1525), a duplicate of the [GUACAMOLE-124](https://issues.apache.org/jira/browse/GUACAMOLE-124).

necouchman:
Thanks @mike-jumper - didn't even think to check that. I'll verify and close this and the Jira issue out.

necouchman:
Closed by #695.

PR #471, opened 2020-02-06
GUACAMOLE-753: Add TOTP auth method to Docker image

benrubson:
Hi,

This PR adds TOTP authentication method to the Docker image.
It's really helpful, making TOTP easy to use with Docker.

Let's merge this low-impact PR with #469 in a 1.1.1 Docker image ?

Many thanks 👍

### DIFF
--- a/guacamole-docker/bin/build-guacamole.sh
+++ b/guacamole-docker/bin/build-guacamole.sh
@@ -127,12 +127,22 @@ if [ -f extensions/guacamole-auth-radius/target/guacamole-auth-radius*.jar ]; th
     cp extensions/guacamole-auth-radius/target/guacamole-auth-radius*.jar "$DESTINATION/radius"
 fi
 
+#
 # Copy OPENID auth extension and schema modifications
 #
 
 if [ -f extensions/guacamole-auth-openid/target/guacamole-auth-openid*.jar ]; then
     mkdir -p "$DESTINATION/openid"
     cp extensions/guacamole-auth-openid/target/guacamole-auth-openid*.jar "$DESTINATION/openid"
+fi
+
+#
+# Copy TOTP auth extension if it was built
+#
+
+if [ -f extensions/guacamole-auth-totp/target/guacamole-auth-totp*.jar ]; then
+    mkdir -p "$DESTINATION/totp"
+    cp extensions/guacamole-auth-totp/target/guacamole-auth-totp*.jar "$DESTINATION/totp"
 fi
 
 #
--- a/guacamole-docker/bin/start.sh
+++ b/guacamole-docker/bin/start.sh
@@ -537,6 +537,21 @@ END
 }
 
 ##
+## Adds properties to guacamole.properties which configure the TOTP two-factor
+## authentication mechanism.
+##
+associate_totp() {
+    # Update config file
+    set_optional_property "totp-issuer"    "$TOTP_ISSUER"
+    set_optional_property "totp-digits"    "$TOTP_DIGITS"
+    set_optional_property "totp-period"    "$TOTP_PERIOD"
+    set_optional_property "totp-mode"      "$TOTP_MODE"
+
+    # Add required .jar files to GUACAMOLE_EXT
+    ln -s /opt/guacamole/totp/guacamole-auth-*.jar   "$GUACAMOLE_EXT"
+}
+
+##
 ## Adds properties to guacamole.properties which configure the Duo two-factor
 ## authentication service. Checks to see if all variables are defined and makes sure
 ## DUO_APPLICATION_KEY is >= 40 characters.
@@ -705,6 +720,11 @@ POSTGRES_DATABASE environment variables, or check Guacamole's Docker
 documentation regarding configuring LDAP and/or custom extensions.
 END
     exit 1;
+fi
+
+# Use TOTP if specified.
+if [ "$TOTP_ENABLED" = "true" ]; then
+    associate_totp
 fi
 
 # Use Duo if specified.


guacamole-docker/bin/start.sh
```diff
@@ -707,6 +722,11 @@ END
     exit 1;
 fi
 
+# Use TOTP if specified.
+if [ -n "$TOTP_ISSUER" ]; then
```
necouchman: Is this the best way to do this?  The `totp-issuer` has a default value, per the manual, so should setting the issuer be what triggers activation of this module?  Or should we go with something like `TOTP_ENABLED`?

benrubson: I did not want to add another specific var to keep it as simple as possible, but I can for sure :+1:
Your new var / varname sounds good to me.
Should I ?

benrubson: Done, as it really seems proper 👍

guacamole-docker/bin/build-guacamole.sh
```diff
@@ -135,6 +136,15 @@ if [ -f extensions/guacamole-auth-openid/target/guacamole-auth-openid*.jar ]; th
     cp extensions/guacamole-auth-openid/target/guacamole-auth-openid*.jar "$DESTINATION/openid"
 fi
 
+#
+# Copy TOTP auth extension if it was built
```
mike-jumper: Why would it not have been built? From what I recall, the only extension that's conditional within the build is RADIUS support.

guacamole-docker/bin/start.sh
```diff
@@ -707,6 +722,11 @@ END
     exit 1;
 fi
 
+# Use TOTP if specified.
+if [ -n "$TOTP_ENABLED" ]; then
```
mike-jumper: I'm not a huge fan of having to add an environment variable that doesn't map to a property, as it deviates from established patterns, but since the TOTP auth doesn't have any required properties I'm not sure we have any choice.

guacamole-docker/bin/start.sh
```diff
@@ -536,6 +536,21 @@ END
 
 }
 
+##
+## Adds properties to guacamole.properties which configure the TOTP two-factor
+## authentication service.
```
mike-jumper: Unlike Duo, TOTP isn't an authentication service. There may be a more correct way to document this.

guacamole-docker/bin/start.sh
necouchman: FWIW, I'm not tied to this - if we want to change it to something else or some other way of enabling it, I'm open to the options.  It just seemed a little weird to me to trigger it by adding a parameter that isn't required.  But, I'm willing to go with whatever.

guacamole-docker/bin/build-guacamole.sh
benrubson: I took exactly what is done above and below for other extensions :)

guacamole-docker/bin/start.sh
benrubson: The only "generic" solution I see is :
`if env | grep -q "^TOTP_"; then`
So that enforcing at least one of the `TOTP` properties would enable it.

guacamole-docker/bin/start.sh
benrubson: Done, thank you for your review 👍

guacamole-docker/bin/start.sh
necouchman: @mike-jumper Thoughts, here?  I'm not tied to my suggestion, so if you have a vote...?

mike-jumper: My preference would be:

* Continue using `TOTP_ENABLED`.
* Actually check the value rather than simply whether it's set. The current implementation would enable TOTP if the user sets `TOTP_ENABLED` to "1", to "false", to "salmon", etc., and that's not exactly rigorous.

I'm not against additionally accepting `TOTP_*` in lieu of `TOTP_ENABLED` for convenience, but if so:

* Only actual, known variables should be accepted (for similar reasons to the above... setting `TOTP_FISH` to "salmon" shouldn't result in TOTP being enabled).
* `TOTP_ENABLED` should still probably be accepted (and checked), as users shouldn't be required to set a TOTP-related variable to its default value just to enable use of TOTP.

benrubson: Thank you for your feedbacks.
I then updated to :
`if [ "$TOTP_ENABLED" = "true" ]`

benrubson:
Let's merge this @mike-jumper @necouchman ?
Many thanks 👍

guacamole-docker/bin/start.sh
```diff
@@ -536,6 +536,21 @@ END
...
+##
+## Adds properties to guacamole.properties which configure the TOTP two-factor
+## authentication mechanism.
+##
+associate_totp() {
+    # Update config file
+    [ -n "$TOTP_ISSUER" ] && set_property "totp-issuer"    "$TOTP_ISSUER"
+    [ -n "$TOTP_DIGITS" ] && set_property "totp-digits"    "$TOTP_DIGITS"
+    [ -n "$TOTP_PERIOD" ] && set_property "totp-period"    "$TOTP_PERIOD"
+    [ -n "$TOTP_MODE" ]   && set_property "totp-mode"      "$TOTP_MODE"
```
necouchman: Looking at the other sections (mostly above this), it looks like instead of doing
```
 [ -n "$TOTP_ISSUER" ] && set_property "totp-issuer"    "$TOTP_ISSUER"
```
you should be able to do:
```
set_optional_property "totp-issuer" "$TOTP_ISSUER"
```

benrubson: I'm not sure because we would then set empty options in the configuration file.
In the other sections, options are tested to always have a value.

necouchman: This is the code for `set_optional_property()`:
```
set_optional_property() {
    NAME="$1"
    VALUE="$2"
    # Set the property only if a value is provided
    if [ -n "$VALUE" ]; then
        set_property "$NAME" "$VALUE"
    fi
}
```

So, it's essentially duplicating what you're doing, no?

benrubson: Tired certainly...
Thank you @necouchman for pointing this, just fixed 👍

guacamole-docker/bin/start.sh
```diff
@@ -536,6 +536,21 @@ END
...
+##
+## Adds properties to guacamole.properties which configure the TOTP two-factor
+## authentication mechanism.
+##
+associate_totp() {
+    # Update config file
+    set_property "totp-issuer"    "$TOTP_ISSUER"
+    set_property "totp-digits"    "$TOTP_DIGITS"
+    set_property "totp-period"    "$TOTP_PERIOD"
+    set_property "totp-mode"      "$TOTP_MODE"
```
necouchman: I think we want `set_option_property`, right, since all of these already have default values?

benrubson: For sure, I just then pushed the required fix.
Thank you @necouchman 👍

benrubson:
Many thanks @jolentes for your positive feedback 👍

benrubson:
Many thanks @necouchman 👍

benrubson:
Perhaps I should have rather pushed to the `staging/1.2.0` branch ?

necouchman:
Sorry, it isn't slated for 1.2.0, it's slated for 1.3.0, so it was merged correctly.

PR #475, opened 2020-02-26
GUACAMOLE-966: Bump client version numbers to 1.2.0

necouchman:
Pull Request to bump versions within the client up to 1.2.0.  I did *NOT* bump the API version (`guacamole-common-js/src/main/webapp/modules/Version.js`) as I do not think the API has changed??  Let me know if that one should still be added, though.

### DIFF
--- a/doc/guacamole-example/pom.xml
+++ b/doc/guacamole-example/pom.xml
@@ -26,7 +26,7 @@
     <groupId>org.apache.guacamole</groupId>
     <artifactId>guacamole-example</artifactId>
     <packaging>war</packaging>
-    <version>1.1.0</version>
+    <version>1.2.0</version>
     <name>guacamole-example</name>
     <url>http://guacamole.apache.org/</url>
 
@@ -114,7 +114,7 @@
         <dependency>
             <groupId>org.apache.guacamole</groupId>
             <artifactId>guacamole-common-js</artifactId>
-            <version>1.1.0</version>
+            <version>1.2.0</version>
             <type>zip</type>
             <scope>runtime</scope>
         </dependency>
--- a/extensions/guacamole-auth-cas/pom.xml
+++ b/extensions/guacamole-auth-cas/pom.xml
@@ -26,7 +26,7 @@
     <groupId>org.apache.guacamole</groupId>
     <artifactId>guacamole-auth-cas</artifactId>
     <packaging>jar</packaging>
-    <version>1.1.0</version>
+    <version>1.2.0</version>
     <name>guacamole-auth-cas</name>
     <url>http://guacamole.apache.org/</url>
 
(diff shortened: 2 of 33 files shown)


mike-jumper:
I think the guacamole-common-js API has actually changed for this release. It's not changed in an incompatible way, but I can find at least the `drawStream()` of `Guacamole.Display` as a new addition (added via commit 4e130d2afd as part of [GUACAMOLE-884](https://issues.apache.org/jira/browse/GUACAMOLE-884)).

guacamole-common/pom.xml
```diff
@@ -26,7 +26,7 @@
     <groupId>org.apache.guacamole</groupId>
     <artifactId>guacamole-common</artifactId>
     <packaging>jar</packaging>
-    <version>1.1.0</version>
+    <version>1.2.0</version>
```
mike-jumper: Has guacamole-common changed? I'm not seeing any changes from 1.1.0 according to git.

extensions/guacamole-auth-cas/pom.xml
```diff
@@ -210,15 +210,15 @@
         <dependency>
             <groupId>org.apache.guacamole</groupId>
             <artifactId>guacamole-common</artifactId>
-            <version>1.1.0</version>
+            <version>1.2.0</version>
```
mike-jumper: It might be easier to just remove this dependency, relying on inheriting it from guacamole-ext. Unlike projects like guacamole-example, use of guacamole-common within an extension will be to service guacamole-ext.

extensions/guacamole-auth-radius/pom.xml
```diff
@@ -178,15 +178,15 @@
         <dependency>
             <groupId>org.apache.guacamole</groupId>
             <artifactId>guacamole-common</artifactId>
-            <version>1.1.0</version>
+            <version>1.2.0</version>
```
mike-jumper: Same here:

It might be easier to just remove this dependency, relying on inheriting it from guacamole-ext. Unlike projects like guacamole-example, use of guacamole-common within an extension will be to service guacamole-ext.

guacamole-common/pom.xml
necouchman: Good point - reverted that.

extensions/guacamole-auth-cas/pom.xml
necouchman: Removed.

extensions/guacamole-auth-radius/pom.xml
necouchman: Removed.

necouchman:
Okay - addressed the issues:
* Bumped the JS API version up to 1.2.0.
* Reverted the common version to 1.1.0
* Removed the explicit dependency on `guacamole-common` from `pom.xml` files where `guacamole-ext` is already being pulled in.

doc/guacamole-example/pom.xml
```diff
@@ -106,15 +106,15 @@
         <dependency>
             <groupId>org.apache.guacamole</groupId>
             <artifactId>guacamole-common</artifactId>
-            <version>1.1.0</version>
+            <version>1.2.0</version>
```
mike-jumper: 1.1.0*

necouchman: Oops, fixed.

PR #479, opened 2020-03-10
GUACAMOLE-986: Add non-nullable modifier to return type of functions never returning null

vrana:
Alternatively use a primitive type instead of nullable boxed type.


necouchman:
So, aside from discussing the merits of doing this, two things need to be done:
1) Create a JIRA issue (https://issues.apache.org/jira/browse/GUACAMOLE) for this effort.
2) Tag pull request and commit messages with the JIRA issue.

Full guide for contributing is here: http://guacamole.apache.org/open-source/

vrana:
Done.

mike-jumper (approved):
I'll look through the various new code added since this was originally opened, as well. There are probably several new locations that this needs to be adopted.

PR #483, opened 2020-03-17
GUACAMOLE-781: Translation updated based on latest en.json and filename fixed to cs.json

Vict0rC:
GUACAMOLE-781: Translation updated based on latest en.json and filename fixed to cs.json


Vict0rC:
Closing in favor of #484

PR #488, opened 2020-04-01
GUACAMOLE-919: Implement PostgreSQL defaultStatementTimeout and socketTimeout

DouglasHeriot:
Pass through environment variables to set `defaultStatementTimeout` and `socketTimeout`. See issue GUACAMOLE-919 for details.

If the TCP connection to the database is silently dropped (not closed), doing a ping query "SELECT 1;" will hang indefinitely. The socketTimeout option will put a timeout on this.

`defaultStatementTimeout` is not relevant to this issue. I thought it might have been at first. I've left this in incase having the option exposed is useful for other reasons.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/PostgreSQLAuthenticationProviderModule.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/PostgreSQLAuthenticationProviderModule.java
@@ -72,7 +72,7 @@ public class PostgreSQLAuthenticationProviderModule implements Module {
 
         // Only set if > 0. Underlying backend does not take 0 as not-set.
         int defaultStatementTimeout = environment.getPostgreSQLDefaultStatementTimeout();
-        if(defaultStatementTimeout > 0) {
+        if (defaultStatementTimeout > 0) {
             myBatisProperties.setProperty("mybatis.configuration.defaultStatementTimeout", String.valueOf(defaultStatementTimeout));
         }
 
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/PostgreSQLAuthenticationProviderModule.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/PostgreSQLAuthenticationProviderModule.java
@@ -69,7 +69,12 @@ public class PostgreSQLAuthenticationProviderModule implements Module {
         myBatisProperties.setProperty("JDBC.autoCommit", "false");
         myBatisProperties.setProperty("mybatis.pooled.pingEnabled", "true");
         myBatisProperties.setProperty("mybatis.pooled.pingQuery", "SELECT 1");
-        myBatisProperties.setProperty("mybatis.configuration.defaultStatementTimeout", environment.getPostgreSQLDefaultStatementTimeout());
+
+        // Only set if > 0. Underlying backend does not take 0 as not-set.
+        int defaultStatementTimeout = environment.getPostgreSQLDefaultStatementTimeout();
+        if(defaultStatementTimeout > 0) {
+            myBatisProperties.setProperty("mybatis.configuration.defaultStatementTimeout", String.valueOf(defaultStatementTimeout));
+        }
 
         // Use UTF-8 in database
         driverProperties.setProperty("characterEncoding", "UTF-8");
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/PostgreSQLAuthenticationProviderModule.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/PostgreSQLAuthenticationProviderModule.java
@@ -111,6 +111,9 @@ public class PostgreSQLAuthenticationProviderModule implements Module {
             
         }
 
+        // Handle case where TCP connection to database is silently dropped
+        driverProperties.setProperty("socketTimeout", String.valueOf(environment.getPostgreSQLSocketTimeout()));
+
     }
 
     @Override
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/PostgreSQLAuthenticationProviderModule.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/PostgreSQLAuthenticationProviderModule.java
@@ -69,6 +69,7 @@ public class PostgreSQLAuthenticationProviderModule implements Module {
         myBatisProperties.setProperty("JDBC.autoCommit", "false");
         myBatisProperties.setProperty("mybatis.pooled.pingEnabled", "true");
         myBatisProperties.setProperty("mybatis.pooled.pingQuery", "SELECT 1");
+        myBatisProperties.setProperty("mybatis.configuration.defaultStatementTimeout", environment.getPostgreSQLDefaultStatementTimeout());
 
         // Use UTF-8 in database
         driverProperties.setProperty("characterEncoding", "UTF-8");
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/PostgreSQLAuthenticationProviderModule.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/PostgreSQLAuthenticationProviderModule.java
@@ -73,7 +73,10 @@ public class PostgreSQLAuthenticationProviderModule implements Module {
         // Only set if > 0. Underlying backend does not take 0 as not-set.
         int defaultStatementTimeout = environment.getPostgreSQLDefaultStatementTimeout();
         if (defaultStatementTimeout > 0) {
-            myBatisProperties.setProperty("mybatis.configuration.defaultStatementTimeout", String.valueOf(defaultStatementTimeout));
+            myBatisProperties.setProperty(
+                "mybatis.configuration.defaultStatementTimeout",
+                String.valueOf(defaultStatementTimeout)
+            );
         }
 
         // Use UTF-8 in database
@@ -117,7 +120,10 @@ public class PostgreSQLAuthenticationProviderModule implements Module {
         }
 
         // Handle case where TCP connection to database is silently dropped
-        driverProperties.setProperty("socketTimeout", String.valueOf(environment.getPostgreSQLSocketTimeout()));
+        driverProperties.setProperty(
+            "socketTimeout",
+            String.valueOf(environment.getPostgreSQLSocketTimeout())
+        );
 
     }
 
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/conf/PostgreSQLEnvironment.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/conf/PostgreSQLEnvironment.java
@@ -49,17 +49,18 @@ public class PostgreSQLEnvironment extends JDBCEnvironment {
     private static final int DEFAULT_PORT = 5432;
 
     /**
-     * The default defaultStatementTimeout (in seconds),
-     * if POSTGRESQL_DEFAULT_STATEMENT_TIMEOUT is not specified.
-     * Default to 0 (no timeout, property won't be set)
-     * https://mybatis.org/mybatis-3/configuration.html
+     * The default number of seconds the driver will wait for a response from
+     * the database. A value of 0 (the default) means the timeout is disabled.
      */
-    private static final int DEFAULT_DEFAULT_STATEMENT_TIMEOUT = 0;
+    private static final int DEFAULT_STATEMENT_TIMEOUT = 0;
 
     /**
-     * The default socketTimeout (in seconds), if POSTGRESQL_SOCKET_TIMEOUT is not specified.
-     * Default to 0 (no timeout)
-     * https://jdbc.postgresql.org/documentation/head/connect.html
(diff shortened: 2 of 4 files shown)


extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/PostgreSQLEnvironment.java
```diff
@@ -47,6 +47,21 @@
...
+     */
+    private static final String DEFAULT_DEFAULT_STATEMENT_TIMEOUT = "null";
+
+    /**
+     * The default socketTimeout (in seconds), if POSTGRESQL_SOCKET_TIMEOUT is not specified.
+     * Default to 0 (no timeout)
+     * https://jdbc.postgresql.org/documentation/head/connect.html
+     */
+    private static final int DEFAULT_SOCKET_TIMEOUT = 0;
+
```
necouchman, line 64: Is there some reason why you've implemented one of these as a `String` and the other as an `int`?  I think I would stick with `int` to be consistent with how other timeouts are implemented.  Also, one uses `null` for no timeout, the other uses 0 - can we align on that, or does the underlying code note allow for that?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/PostgreSQLEnvironment.java
```diff
@@ -47,6 +47,21 @@
      */
     private static final int DEFAULT_PORT = 5432;
 
+    /**
+     * The default defaultStatementTimeout (in seconds),
+     * if POSTGRESQL_DEFAULT_STATEMENT_TIMEOUT is not specified.
+     * Default to null (no timeout)
+     * https://mybatis.org/mybatis-3/configuration.html
+     */
+    private static final String DEFAULT_DEFAULT_STATEMENT_TIMEOUT = "null";
```
necouchman, line 56: `"null"` != `NULL` - does the underlying setting require the String value `null` for this?

DouglasHeriot:
These properties are passed in as string properties. `"null"` means not-set for `defaultStatementTimeout` however `0` is not-set for `socketTimeout`. I initially decided to try and just keep the properties as strings, but `int` did make more sense for `socketTimeout`.

https://mybatis.org/mybatis-3/configuration.html
https://jdbc.postgresql.org/documentation/head/connect.html

I agree int is nicer, so to keep consistency I have now made `defaultStatementTimeout` and `int` as well, and simply don't set the property on the backend if it's `0` to match the behaviour of `socketTimeout`.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/PostgreSQLEnvironment.java
DouglasHeriot: Yes it takes a string value. I thought that just keeping it as a string was neater than making a nullable `Integer` where all other properties are `int`.
I've now changed this to an `int` and don't pass through when it's `0`.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/PostgreSQLEnvironment.java
DouglasHeriot: Yes, I was trying to keep same as what the backend takes. But I see your point - consistency at this level would be good rather than exposing the underlying differences.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/PostgreSQLEnvironment.java
necouchman: I dug into this a little more, and it does not look to me like the string value "null" is what they intend - I think just mean `NULL` in the since of not setting anything.  So, this should not be `"null"`, it should be `NULL`.

DouglasHeriot: @necouchman I agree, it looks like I made a mistake. See my latest commit on this branch for the fix 330b2c3ec0c1fa5ed62f4efcb8585f7c287f86c9

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/PostgreSQLGuacamoleProperties.java
```diff
@@ -92,6 +92,30 @@ private PostgreSQLGuacamoleProperties() {}
 
     };
 
+    /**
+     * Sets the number of seconds the driver will wait for
```
necouchman, line 96: This is a bit of a nitpick, but the below statement doesn't actually "set" anything.  It is a property that can be used to set it.  Take a look at some of the other properties around this - for example, "The password used to authenticate..."  This can probably be simplified to something similar to that.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/PostgreSQLGuacamoleProperties.java
```diff
@@ -92,6 +92,30 @@ private PostgreSQLGuacamoleProperties() {}
...
+    public static final IntegerGuacamoleProperty
+            POSTGRESQL_DEFAULT_STATEMENT_TIMEOUT = new IntegerGuacamoleProperty(){
+
+        @Override
+        public String getName() { return "postgresql-default-statement-timeout"; }
+
+    };
+
+    /**
+     * Sets the number of seconds the driver will wait in a read() call
```
necouchman, line 108: As above, this isn't setting anything, so more direct language would probably be better, here.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/PostgreSQLEnvironment.java
```diff
@@ -47,6 +47,21 @@
      */
     private static final int DEFAULT_PORT = 5432;
 
+    /**
+     * The default defaultStatementTimeout (in seconds),
+     * if POSTGRESQL_DEFAULT_STATEMENT_TIMEOUT is not specified.
+     * Default to 0 (no timeout, property won't be set)
+     * https://mybatis.org/mybatis-3/configuration.html
```
necouchman, line 54: A couple things for this comment block:
* I'd avoid using the actual variable name (`defaultStatementTimeout`) and just make it read like "The default statement timeout..."
* We don't really link to outside documentation for any of the other properties, and I'm not sure it's worth starting for these two, so I'd remove the links to external sites.
* For the explanation for "Default to 0," something more verbose might be appropriate: "Setting this value to 0 (the default) will result in..."

DouglasHeriot:
Ok, sorry @necouchman for the delay.
I've fixed the style issues and improved the comments.

I've also added a corresponding pull request to the manual repo:
https://github.com/apache/guacamole-manual/pull/146

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/conf/PostgreSQLGuacamoleProperties.java
```diff
@@ -94,6 +94,30 @@ private PostgreSQLGuacamoleProperties() {}
 
     };
 
+    /**
+     * The number of seconds the driver will wait for
+     * a response from the database.
+     */
+    public static final IntegerGuacamoleProperty
+            POSTGRESQL_DEFAULT_STATEMENT_TIMEOUT = new IntegerGuacamoleProperty(){
```
necouchman, line 102: Please put a space between `()` and `{`.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/conf/PostgreSQLGuacamoleProperties.java
```diff
@@ -94,6 +94,30 @@ private PostgreSQLGuacamoleProperties() {}
...
+        public String getName() { return "postgresql-default-statement-timeout"; }
+
+    };
+
+    /**
+     * The number of seconds the driver will wait in a read() call
+     * on the TCP connection to the database.
+     */
+    public static final IntegerGuacamoleProperty
+            POSTGRESQL_SOCKET_TIMEOUT = new IntegerGuacamoleProperty(){
```
necouchman, line 114: Space between `()` and `{`, please.

guacamole-docker/bin/start.sh
```diff
@@ -354,10 +354,18 @@ END
         "postgresql-default-max-group-connections-per-user" \
         "$POSTGRES_DEFAULT_MAX_GROUP_CONNECTIONS_PER_USER"
 
+    set_optional_property               \
+        "postgresql-default-statement-timeout" \
```
necouchman, line 358: Please line up the trailing `\`.

guacamole-docker/bin/start.sh
```diff
@@ -354,10 +354,18 @@ END
...
+        "$POSTGRES_DEFAULT_STATEMENT_TIMEOUT"
+
     set_optional_property          \
         "postgresql-user-required" \
         "$POSTGRES_USER_REQUIRED"
 
+    set_optional_property               \
+        "postgresql-socket-timeout" \
```
necouchman, line 366: Please line up the trailing `\`

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/conf/PostgreSQLEnvironment.java
```diff
@@ -249,6 +265,41 @@ public String getPostgreSQLUsername() throws GuacamoleException {
...
+    public int getPostgreSQLDefaultStatementTimeout() throws GuacamoleException {
+        return getProperty(
+            PostgreSQLGuacamoleProperties.POSTGRESQL_DEFAULT_STATEMENT_TIMEOUT,
+            DEFAULT_STATEMENT_TIMEOUT
+        );
+    }
+    
+    /**
+     * Returns the socketTimeout property to set on PostgreSQL connections.
+     * If unspecified, this will be the default to 0 (no timeout)
```
necouchman, line 289: Probably should either be:
> this will default to

or:
> this will be the default of

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/conf/PostgreSQLEnvironment.java
```diff
@@ -48,6 +48,22 @@
...
+     * the database. A value of 0 (the default) means the timeout is disabled.
+     */
+    private static final int DEFAULT_STATEMENT_TIMEOUT = 0;
+
+    /**
+     * The default timeout (in seconds) used for socket read operations.
+     * If reading from the server takes longer than this value, the
+     * connection is closed. This can be used to handle network problems
+     * such as a dropped connection to the database. Similar to 
+     * DEFAULT_STATEMENT_TIMEOUT, it will also abort queries that take too
```
necouchman, line 62: Two comments, here:
* You say "similar to DEFAULT_STATEMENT_TIMEOUT", but there's no mention of aborting queries on the above timeout.
* It looks like most of the documentation around what these two additional parameters do is here in the `DEFAULT_` sections, rather than down in the `PostgreSQLGuacamoleProperties.java` file.  If you look at the documentation of other `DEFAULT_`s around this, you'll see that the documentation is more succinct and references the actual properties.  I suggest doing the same with these two.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/PostgreSQLAuthenticationProviderModule.java
```diff
@@ -70,6 +70,12 @@ public PostgreSQLAuthenticationProviderModule(PostgreSQLEnvironment environment)
         myBatisProperties.setProperty("mybatis.pooled.pingEnabled", "true");
         myBatisProperties.setProperty("mybatis.pooled.pingQuery", "SELECT 1");
 
+        // Only set if > 0. Underlying backend does not take 0 as not-set.
+        int defaultStatementTimeout = environment.getPostgreSQLDefaultStatementTimeout();
+        if (defaultStatementTimeout > 0) {
+            myBatisProperties.setProperty("mybatis.configuration.defaultStatementTimeout", String.valueOf(defaultStatementTimeout));
```
necouchman, line 76: Might be good to break this line somewhere - this is a looooooooooooooong line :-).

DouglasHeriot:
Ok @necouchman , I've addressed the style issues, and moved most of the detail in the comments to `PostgreSQLGuacamoleProperties.java`.

I've also updated apache/guacamole-manual#146 to match the new wording.

guacamole-docker/bin/start.sh
necouchman: These two lines still do not appear to line up the trailing slashes.

guacamole-docker/bin/start.sh
necouchman: Trailing slashes are still not lined up, here.

DouglasHeriot:
@necouchman Oops, sorry. I made the change and didn't add it to my earlier commit. Fixed now.

necouchman:
Okay, @DouglasHeriot, the only remaining issue I see is that your commit messages are not properly formatted.  The style of the commit message should be:
> GUACAMOLE-919: Description of the changes in the commit

The ":" appears to be missing from your commit messages.  Once that's fixed we should be ready to merge.

DouglasHeriot:
@necouchman  Ok, I've fixed these commit messages now.

PR #490, opened 2020-04-02
GUACAMOLE-1006: Provide implementations for List properties.

necouchman:
Provides some basic implementations for `String` and `URI` list properties in `guacamole-ext`, in anticipation of those being required for things like LDAP and RADIUS server redundancy.

Also, note that I switched over to semicolon delimiting for these - that's worth some discussion, but my rationale for that is that some URIs - LDAP ones, for instance - rely on the ability to use commas to separate LDAP DNs within the URI, so, one way or the other we need to address that.


mike-jumper:
What do you think about enhancing [`Environment`](https://guacamole.apache.org/doc/guacamole-ext/org/apache/guacamole/environment/Environment.html) such that it inherently supports retrieving an ordered `Collection` of property values?

It could then be up to the `GuacamoleProperty` implementation to decide how to represent multiple values (if at all).

necouchman:
Sounds good - I will take a look.

necouchman:
@mike-jumper Okay, I've started down path that I think is approximately what you're suggesting, while trying to avoid breaking everything in the process. So, instead of making `getProperty()` and/or `getRequiredProperty()` return a collection, I added a new set of methods, `getPropertyCollection()` and `getRequiredPropertyCollection()` that are intended to parse into a Collection. 

This would rely on each of the properties to implement a new `getValueCollection()` method, which I've provided a default implementation for in `GuacamoleProperty` that basically says "not yet implemented."

Let me know if this looks good, sane, reasonable, etc., or if there's a better solution? I could break everything and make `getValue()`, `getProperty()`, and `getRequiredProperty()` return collections instead of individual values, if that's preferable?

guacamole-ext/src/main/java/org/apache/guacamole/environment/Environment.java
```diff
@@ -118,20 +124,103 @@ public <Type> Type getProperty(GuacamoleProperty<Type> property)
...
+     * @return
+     *     A sorted collection of the the parsed values of the property as read
+     *     from guacamole.properties.
+     * 
+     * @throws GuacamoleException
+     *     If an error occurs while parsing the value for the given property in
+     *     guacamole.properties.
+     */
+    public <Type> Collection<Type> getPropertyCollection(
+            GuacamoleProperty<Type> property) throws GuacamoleException;
```
mike-jumper, line 166: As this is an `interface`, we'll need to provide `default` implementations for sake of backward compatibility. I suggest:

1. Internally pulling the raw value as a `String` leveraging `StringGuacamoleProperty`.
2. Passing that raw value to `parseValueCollection()` of `property` to get the requested, parsed value.

guacamole-ext/src/main/java/org/apache/guacamole/properties/URIListGuacamoleProperty.java
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * A GuacamoleProperty whose value is a List of URIs. The string value
+ * parsed to produce this list is a semicolon-delimited list.
+ * Duplicate values are ignored, as is any whitespace following delimiters.
+ * To maintain compatibility with the behavior of Java properties in general,
+ * only whitespace at the beginning of each value is ignored; trailing
+ * whitespace becomes part of the value.
+ */
+public abstract class URIListGuacamoleProperty
+        implements GuacamoleProperty<List<URI>> {
```
mike-jumper, line 40: Looks good, however in the context of these changes I think this should be implemented via adding collection support to the existing `URIGuacamoleProperty`.

guacamole-ext/src/main/java/org/apache/guacamole/properties/GuacamoleProperty.java
```diff
@@ -49,5 +51,25 @@
      *                            provided value.
      */
     public Type parseValue(String value) throws GuacamoleException;
+    
+    /**
+     * Parses the given string value into a sorted Collection of values of the
+     * type associated with this GuacamoelProperty.
```
mike-jumper, line 57: GuacamoleProperty*

guacamole-ext/src/main/java/org/apache/guacamole/properties/GuacamoleProperty.java
```diff
@@ -49,5 +51,25 @@
      *                            provided value.
      */
     public Type parseValue(String value) throws GuacamoleException;
+    
+    /**
+     * Parses the given string value into a sorted Collection of values of the
```
mike-jumper, line 56: What do you mean by "sorted" here? Is the order of the values not determined by the property implementation?

guacamole-ext/src/main/java/org/apache/guacamole/properties/GuacamoleProperty.java
```diff
@@ -49,5 +51,25 @@
...
+     * @param value
+     *     The string value to parse.
+     * 
+     * @return
+     *     A sorted Collection of the parsed values.
+     * 
+     * @throws GuacamoleException 
+     *     If an error occurs while parsing the provided value.
+     */
+    default public Collection<Type> parseValueCollection(String value)
```
mike-jumper, line 68: We should note the default behavior in the documentation such that implementations that inherit this implementation will have that behavior documented. It may otherwise be surprising to a developer invoking `parseValueCollection()` that the property they're using doesn't support being parsed as a collection.

guacamole-ext/src/main/java/org/apache/guacamole/properties/GuacamoleProperty.java
```diff
@@ -49,5 +51,25 @@
...
+            throws GuacamoleException {
+        
+        throw new GuacamoleServerException("Collection parsing not implemented for this property type.");
```
mike-jumper, line 71: This should be a `GuacamoleUnsupportedException`. Another option could be something like `Collections.singletonList(parseValue(value))`.

guacamole-ext/src/main/java/org/apache/guacamole/environment/Environment.java
```diff
@@ -118,20 +124,103 @@ public <Type> Type getProperty(GuacamoleProperty<Type> property)
...
+     * @throws GuacamoleException
+     *     If an error occurs while parsing the value for the given property in
+     *     guacamole.properties.
      */
     public <Type> Type getProperty(GuacamoleProperty<Type> property,
             Type defaultValue) throws GuacamoleException;
 
+    /**
+     * Given a GuacamoleProperty, parses and returns a sorted Collection of the
+     * value set for that property in guacamole.properties, if any.
```
mike-jumper, line 149: As it's up to the property implementation to support being read as a `Collection`, I think we should document here and in other `getPropertyCollection()` flavors that not all properties support being read as multiple values and that it's up to the property implementation to define how multiple values are parsed, if at all.

mike-jumper (changes requested):

necouchman:
@mike-jumper Okay, I've re-worked the implementation, here, providing support within the existing property classes for parsing and retrieving lists of values.

mike-jumper (changes requested):
`default` implementations for new functions were added to the `GuacamoleProperty` interface, but I think they're missing from `Environment`.

Otherwise, LGTM!

necouchman:
> `default` implementations for new functions were added to the `GuacamoleProperty` interface, but I think they're missing from `Environment`.
> 
> Otherwise, LGTM!

Okay, I think I managed to get these default implemented correctly in the `Environment` interface, but for some reason it was a bit of a mind-bending exercise on the order of the spoon in The Matrix.

mike-jumper (approved):
Slick approach - I see what you mean about the spoon.

PR #492, opened 2020-04-03
GUACAMOLE-728: Add SSL configuration options for MySQL and PostgreSQL

necouchman:
This pull request adds the configuration options for both MySQL and PostgreSQL to enable these drivers to establish SSL connections to the secured MySQL and PostgreSQL instances and configure the various options associated with them.

Have not had a chance to test, yet, but will try to spin that up, soon.  If anyone else is able to test, that'd be great.


echu2013:
I am trying to figure out this..
I use the following setting because of the lack of this feature:
`mysql-database: guacdb?useSSL=false`

Using this branch I've tried to specify the regular way:
`mysql-database: guacdb`
And:
`mysql-ssl-mode: disabled`

> Also tried the values: 0, false. But logs say these are invalid ssl modes.

But, the following keeps showing up in logs:
> Mon Apr 06 12:23:26 ART 2020 WARN: Establishing SSL connection without server's identity verification is not recommended. According to MySQL 5.5.45+, 5.6.26+ and 5.7.6+ requirements SSL connection must be established by default if explicit option isn't set. For compliance with existing applications not using SSL the verifyServerCertificate property is set to 'false'. You need either to explicitly disable SSL by setting useSSL=false, or set useSSL=true and provide truststore for server certificate verification.

**Reverting to my original setting, WARNs vanish.**

necouchman:
@echu2013: I've tweaked the code a bit and added in the legacy (see below) `useSSL` option as part of the parsing of the code.

According to the MySQL 8.0 JDBC documentation (https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-configuration-properties.html), the `useSSL` mode property has been deprecated in favor of the `sslMode` property, which is why I implemented that one, instead.  However, please give the latest commit a try and see if setting that flag allows it to work properly.

echu2013:
> @echu2013: I've tweaked the code a bit and added in the legacy (see below) `useSSL` option as part of the parsing of the code.
> 
> According to the MySQL 8.0 JDBC documentation (https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-configuration-properties.html), the `useSSL` mode property has been deprecated in favor of the `sslMode` property, which is why I implemented that one, instead. However, please give the latest commit a try and see if setting that flag allows it to work properly.

Hi! 
Thanks, I will try it! How should I set it in guacamole.properties?

necouchman:
@echu2013: Parameters are the same as before, I just added the code that sets `useSSL=false` when mysql-ssl-mode=disabled, and `useSSL=true` when `mysql-ssl-mode` is set to anything else that would enable SSL.  There are also options for the trust store (`mysql-ssl-trust-store`) and trust store password.

mike-jumper:
@necouchman, with additional enum-driven `GuacamoleProperty` implementations being added here, I took a stab at implementing an `EnumGuacamoleProperty` that would remove the need to define new `GuacamoleProperty` subclasses every time we want to map a configuration value to an enum.

Do you think that would be worth pulling in? The draft PR is #519.

necouchman:
> Do you think that would be worth pulling in? The draft PR is #519.

Yeah, I'm good with that!

necouchman:
> Do you think that would be worth pulling in? The draft PR is #519.

Okay, going to refactor with fixes you mentioned, above, and the recently-merged support for EnumGuacamoleProperty.

necouchman:
Okay, refactored with your new fancy `EnumGuacamoleProperty` and corrected comment issues.  Going to try to review it myself, here, tomorrow and see what else needs to be fixed.  Let me know what you find.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/conf/PostgreSQLSSLMode.java
```diff
@@ -0,0 +1,104 @@
...
+     * Given the String value, determine the correct enum value that matches
+     * the string, or null if there is no match.
+     * 
+     * @param value
+     *     The String value to test to find a match.
+     * 
+     * @return 
+     *     The enum value matching the given String.
+     */
+    public static PostgreSQLSSLMode getValue(String value) {
```
mike-jumper, line 96: Is this still used?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/MySQLAuthenticationProviderModule.java
```diff
@@ -79,6 +84,37 @@ public MySQLAuthenticationProviderModule(MySQLEnvironment environment)
         // Allow use of multiple statements within a single query
         driverProperties.setProperty("allowMultiQueries", "true");
         
+        // Set the SSL mode to use when conncting
+        MySQLSSLMode sslMode = environment.getMySQLSSLMode();
+        driverProperties.setProperty("sslMode", sslMode.toString());
```
mike-jumper, line 89: With `MySQLSSLMode` now refactored to rely on `@PropertyValue`, this may not return the value you expect. The `MySQLSSLMode` enum should expose a means of retrieving the value required by the JDBC driver.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/MySQLAuthenticationProviderModule.java
```diff
@@ -79,6 +84,37 @@ public MySQLAuthenticationProviderModule(MySQLEnvironment environment)
...
+        
+        // Set legacy properties
+        if (sslMode == MySQLSSLMode.DISABLED)
+            driverProperties.setProperty("useSSL", "false");
```
mike-jumper, line 93: If `MySQLSSLMode.DISABLED` represents `useSSL=false`, is the value set for `sslMode` above correct in this case? Is specifying `sslMode` legal if `useSSL=false` is also specified?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/conf/PostgreSQLSSLMode.java
```diff
@@ -0,0 +1,104 @@
...
+     *     The value to use when configuring the JDBC driver.
+     */
+    PostgreSQLSSLMode(String configValue) {
+        this.configValue = configValue;
+    }
+    
+    @Override
+    public String toString() {
+        return configValue;
+    }
```
mike-jumper, line 84: I suggest providing access to this value in a way that is specifically tied to the fact that it is the configuration value for the JDBC driver (and documented as such), rather than overriding `toString()`.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/MySQLAuthenticationProviderModule.java
necouchman: I think so - basically, the only case where `useSSL` will be set to `false` will be where `sslMode` is already set to `disabled`.  In every other case of `sslMode`, `useSSL` will be set to `true`, which, based on some feedback and documentation, is required for different (older) versions of MySQL and/or MariaDB.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/conf/MySQLSSLMode.java
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * Possible values for enabling SSL within the MySQL Driver.
+ */
+public enum MySQLSSLMode {
+    
+    /**
+     * Do not use SSL at all.
+     */
+    @PropertyValue("disabled")
+    DISABLED("disabled"),
```
mike-jumper, line 33: According to https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-configuration-properties.html:

> ... The following values are allowed: "DISABLED" - Establish unencrypted connections; "PREFERRED" - (default) Establish encrypted connections if the server enabled them, otherwise fall back to unencrypted connections; "REQUIRED" - Establish secure connections if the server enabled them, fail otherwise; "VERIFY_CA" - Like "REQUIRED" but additionally verify the server TLS certificate against the configured Certificate Authority (CA) certificates; "VERIFY_IDENTITY" - Like "VERIFY_CA", but additionally verify that the server certificate matches the host to which the connection is attempted.

Are the current strings for `driverValue` listed here ("disabled", "verify-ca", etc.) correct?

necouchman: Ah, that explains why I was trying to use `toString()` - just seemed to match up exactly.  They should be correct, now.

PR #495, opened 2020-04-03
GUACAMOLE-770: Support for clearing TOTP data

necouchman:
This implements support for two options:
1) Clear out the TOTP secret.
2) Set the TOTP status back to unconfirmed, prompting the user to re-confirm.


extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/user/TOTPUser.java
```diff
@@ -66,14 +81,8 @@ public User getUndecorated() {
     public Map<String, String> getAttributes() {
 
         // Create independent, mutable copy of attributes
-        Map<String, String> attributes =
-                new HashMap<String, String>(super.getAttributes());
+        Map<String, String> attributes = new HashMap<>(super.getAttributes());
 
-        // Do not expose any TOTP-related attributes outside this extension
-        attributes.remove(TOTP_KEY_SECRET_ATTRIBUTE_NAME);
-        attributes.remove(TOTP_KEY_CONFIRMED_ATTRIBUTE_NAME);
```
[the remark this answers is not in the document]
necouchman: Okay, think I got this done - let me know how it looks!

extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/user/TOTPUser.java
```diff
@@ -82,13 +106,15 @@ public User getUndecorated() {
     public void setAttributes(Map<String, String> attributes) {
 
         // Create independent, mutable copy of attributes
-        attributes = new HashMap<String, String>(attributes);
-
-        // Do not expose any TOTP-related attributes outside this extension
-        attributes.remove(TOTP_KEY_SECRET_ATTRIBUTE_NAME);
```
mike-jumper, line 88: This will still need to be removed in the case that we aren't setting `TOTP_KEY_SECRET_ATTRIBUTE_NAME` to anything, or it will be possible for users to manually set the value of their TOTP key.

extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/user/TOTPUser.java
```diff
@@ -82,13 +106,15 @@ public User getUndecorated() {
...
-        attributes.remove(TOTP_KEY_CONFIRMED_ATTRIBUTE_NAME);
```
mike-jumper, line 89: Is there a use case for allowing the user to reconfirm their existing key, rather than requiring that an entirely new key be generated in order to reset?

extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/user/TOTPUser.java
```diff
@@ -82,13 +106,15 @@ public User getUndecorated() {
...
+        attributes = new HashMap<>(attributes);
+        
+        // Pull off the boolean reset field
+        String reset = attributes.remove(TOTP_KEY_SECRET_RESET_FIELD);
+        
+        // If reset has been set to true, clear the secret.
+        if (reset != null && reset.equals("true"))
+            attributes.put(TOTP_KEY_SECRET_ATTRIBUTE_NAME, null);
```
mike-jumper, line 116: Does `TOTP_KEY_CONFIRMED_ATTRIBUTE_NAME` need to be reset here, as well?

extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/user/UserVerificationService.java
```diff
@@ -108,7 +108,7 @@ private UserTOTPKey getKey(UserContext context,
 
         // If no key is defined, attempt to generate a new key
         String secret = attributes.get(TOTPUser.TOTP_KEY_SECRET_ATTRIBUTE_NAME);
-        if (secret == null) {
+        if (secret == null || secret.isEmpty()) {
```
mike-jumper, line 111: Why expand the representation of no defined key to include the empty string?

extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/user/TOTPUser.java
necouchman: That's what I do below this point, right?  It's just that, instead of using the `remove()` method and sending the value nowhere, I'm assigning it to a `String` for the purpose of checking it.  But it never gets added back or left on.  Correct?  Or am I missing something?

extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/user/TOTPUser.java
necouchman: I think this was mentioned in the mailing list somewhere along the way.  I don't have any particular affinity for it - I think *most* of the situations where you'd want to un-confirm you it you'd also want to reset it, but I left the option in there.  Happy to go either direction.

extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/user/TOTPUser.java
necouchman: Yeah, probably a good idea, although it doesn't matter all that much, because clearing out the secret results in a new QR code and the value being reset, anyway.  But probably good for the sake of cleanliness and consistency as much as anything.

extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/user/UserVerificationService.java
necouchman: I ran into a few situations testing out my changes where the DB value for the secret attribute was an empty string, and it causes problems - basically generates an invalid QR code that caused an error when you try to add it in to something like Google Authenticator.  I can leave it as just `null` if you like, but it seemed like a reasonable check to have in there.

extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/user/TOTPUser.java
necouchman: Oh, I'm missing it...I see it, fixing it.

red-erik:
Hello,
I really would like to use this feature. Is it possible ? Is there something in configuration files to enable it?
Regards,

Red.

necouchman:
@red-erik: It has not been merged, yet, so it isn't available in the main code. If you want to use it you'd have to grab the diff of this pull request (https://github.com/apache/guacamole-client/pull/495.diff) and apply it to the code base, re-build, and re-deploy.  That said, it is in development and could change before it is accepted/merged into the main code, so no guarantees it'll be the same in the future.

red-erik:
Hello,
thank you for your answer but I'm not able to do something like re-build and re-deploy. I solved temporary by editing the MySql backend and setting "guac-totp-key-confirmed" to false for the needed user but I really believe this is an important update and feature that should be available to administrative accounts.
Regards,

Red.

necouchman:
@red-erik : Understood - it'll likely be a while before this is released.  It is not slated to go in 1.3.0, which we're working on finalizing, now, so it'll likely have to wait until the next release.

red-erik:
Thank you, I'll wait but really is common to have to change Google authenticator device, loosing previous tokens so being able to have a TTOP Qrcode again is required.
Regards,

Red.

gripfastistech:
> editing the MySql backend and setting "guac-totp-key-confirmed" to false for the needed user

I needed to do that after I accidentally deleted the profile from my phone app by mistake.

Didn't have any mysql login working in the terminal or otherwise (localhost denied for all users, etc...) so that was annoying.

I installed on Ubuntu 20.04, but with guac itself installed under tomcat9, from this: [script](https://github.com/MysticRyuujin/guac-install) from MysticRyuujin. 

I was really wary to force a mysql password change directly since I had just entered in a few client connections and they were working on a complicated internal network.

So I installed Adminer and PHP in the OS, manually edited the mysql.cf file, under mysqld heading inside to include "--skip-grant-tables" and restarted the mysql service to gain mysql login via Adminer. This was a temp action.

User (missing the QR) had id of "2", set flag to "false" as Red suggested, commented the appended line in the mysql.cf file, restarted mysql, restarted ufw and was able login with the user in question via Guacamole admin. 

The QR code prompt appeared instantly after password login as expected for the user. Scanned the QR, entered into phone app, logged into Guacamole like a champ, with no risk to the initial mysql setup.

thanks @red-erik ! I hope this helps someone in the future...at least until the feature is considered for inclusion as a feature request.

mike-jumper (approved):

PR #496, opened 2020-04-03
GUACAMOLE-996: Add support for configuring group filter.

echu2013:
This provides ldap-group-search-filter functionallity. Tested and working OK!

### DIFF
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/ConfigurationService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/ConfigurationService.java
@@ -25,6 +25,7 @@ import java.util.List;
 import org.apache.directory.api.ldap.model.filter.ExprNode;
 import org.apache.directory.api.ldap.model.filter.PresenceNode;
 import org.apache.directory.api.ldap.model.message.AliasDerefMode;
+import org.apache.directory.api.ldap.model.filter.EqualityNode;
 import org.apache.directory.api.ldap.model.name.Dn;
 import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.environment.Environment;
@@ -318,6 +319,24 @@ public class ConfigurationService {
         return environment.getProperty(
             LDAPGuacamoleProperties.LDAP_USER_SEARCH_FILTER,
             new PresenceNode("objectClass")
+
+    /**
+     * Returns the search filter that should be used when querying the
+     * LDAP server for Guacamole groups.  If no filter is specified,
+     * a default of "(objectClass=group)" is returned.
+     *
+     * @return
+     *     The search filter that should be used when querying the
+     *     LDAP server for groups that are valid in Guacamole, or
+     *     "(objectClass=group)" if not specified.
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed.
+     */
+    public ExprNode getGroupSearchFilter() throws GuacamoleException {
+        return environment.getProperty(
+            LDAPGuacamoleProperties.LDAP_GROUP_SEARCH_FILTER,
+            new EqualityNode("objectClass","group")
         );
     }
 
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
@@ -87,9 +87,9 @@ public class UserGroupService {
         if (confService.getConfigurationBaseDN() != null)
             return new NotNode(new EqualityNode("objectClass","guacConfigGroup"));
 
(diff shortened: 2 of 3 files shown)


extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/ConfigurationService.java
```diff
@@ -317,7 +318,27 @@ public int getMaxReferralHops() throws GuacamoleException {
     public ExprNode getUserSearchFilter() throws GuacamoleException {
         return environment.getProperty(
             LDAPGuacamoleProperties.LDAP_USER_SEARCH_FILTER,
-            new PresenceNode("objectClass")
+            new EqualityNode("objectClass","user")
```
necouchman, line 321: This seems out-of-scope for this particular pull request.  We might discuss the merits of doing this at some point, but I don't think it should be included in an issue to implement the group filter.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
```diff
@@ -79,17 +79,17 @@
...
+    private ExprNode buildGroupSearchFilter() throws GuacamoleException {
 
         // Explicitly exclude guacConfigGroup object class only if it should
         // be assumed to be defined (query may fail due to no such object
         // class existing otherwise)
         if (confService.getConfigurationBaseDN() != null)
             return new NotNode(new EqualityNode("objectClass","guacConfigGroup"));
 
-        // Read any object as a group if LDAP is not being used for connection
+        // Read objects from LDAP with filter defined by "ldap-group-search-filter" as a group if LDAP is not being used for connection
```
necouchman, line 90: This should probably just be made a multi-line comment - it's pretty long.

echu2013: @necouchman sorry, i am pretty newie with contributing online and I make some mistakes. You are right about The out of scope stuff.. And I will rewrite the "pretty long" comment, how could I update this pull request?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/ConfigurationService.java
mike-jumper: +1

Adding the option to narrow by `objectClass` would be a separate feature, the value that is used should not be hard-coded to `user`, and we should avoid breaking compatibility with existing deployments which have been relying on the former `*` filter for ages.

Same for the group filter - if the deployment in question needs to narrow things, then specifying custom filters will allow that. Beyond the filters, the different base DNs should be the only restrictions applied unless the administrator configures otherwise.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
```diff
@@ -79,17 +79,17 @@
      * @throws GuacamoleException
      *     If guacamole.properties cannot be parsed.
      */
-    private ExprNode getGroupSearchFilter() throws GuacamoleException {
+    private ExprNode buildGroupSearchFilter() throws GuacamoleException {
```
mike-jumper, line 82: Why is this being renamed?

echu2013: I though it was more descriptive this name rather than get, also to avoid confusion with 
`        return confService.getGroupSearchFilter();
`

echu2013:
Closing in favor of #497 , sorry for the mess.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/ConfigurationService.java
echu2013: > +1
> 
> Adding the option to narrow by `objectClass` would be a separate feature, the value that is used should not be hard-coded to `user`, and we should avoid breaking compatibility with existing deployments which have been relying on the former `*` filter for ages.
> 
> Same for the group filter - if the deployment in question needs to narrow things, then specifying custom filters will allow that. Beyond the filters, the different base DNs should be the only restrictions applied unless the administrator configures otherwise.

Please note that description inside code states the following:
`    /**
     * Returns the search filter that should be used when querying the
     * LDAP server for Guacamole users.  If no filter is specified,
     * a default of "(objectClass=user)" is returned.
     *
     * @return
     *     The search filter that should be used when querying the
     *     LDAP server for users that are valid in Guacamole, or
     *     "(objectClass=user)" if not specified.
     *
     * @throws GuacamoleException
     *     If guacamole.properties cannot be parsed.
     */`
**If no filter is specified, a default of "(objectClass=user)" is returned.**
So, that's why I corrected the code.

PR #497, opened 2020-04-04
GUACAMOLE-996: Add support for configuring group filter.

echu2013:
This provides ldap-group-search-filter functionallity. Tested and working OK!


extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/ConfigurationService.java
```diff
@@ -321,6 +322,26 @@ public ExprNode getUserSearchFilter() throws GuacamoleException {
...
+     *     LDAP server for groups that are valid in Guacamole, or
+     *     "(objectClass=group)" if not specified.
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed.
+     */
+    public ExprNode getGroupSearchFilter() throws GuacamoleException {
+        return environment.getProperty(
+            LDAPGuacamoleProperties.LDAP_GROUP_SEARCH_FILTER,
+            new EqualityNode("objectClass","group")
```
necouchman, line 341: I think @mike-jumper's vote here was to make the default the `PresenceNode("objectClass")`, similar to how users are done.  I'm good with either one, but it does seem more consistent at present to stick with that.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
```diff
@@ -87,9 +87,9 @@ private ExprNode getGroupSearchFilter() throws GuacamoleException {
         if (confService.getConfigurationBaseDN() != null)
             return new NotNode(new EqualityNode("objectClass","guacConfigGroup"));
```
necouchman, line 88: The filter should also be factored in right here - that is, if the user has configured the configuration DN, the configured search filter should, at the very least, be `AndNode`d together here with this `NotNode` to remove `guacConfigGroup` objects from the results.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/ConfigurationService.java
echu2013: I think he observed the change to pre-existing ldap-user-search-filter function, that it's out of the scope of this GUACAMOLE-996

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
echu2013: You say... Retrieve everything under `configurationbaseDN ` **NOT** containing `guacConfigGroup `right?

necouchman: Retrieve everything under `groupBaseDN` NOT containing `guacConfigGroup`.  Basically, the previous code made the assumption that, if you've configured a DN for configuration items you've also extended your LDAP schema to include the `guacConfigGroup` schema item, and that you do not want those being returned by the groups filter, you want them being returned in the config search.  That is still the case in the line mentioned above, but the new group filter needs to be and-ed with that logic.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/ConfigurationService.java
necouchman: Yes, modifying the user one is certainly out-of-scope - I'm not just not sure if we should be changing the default for user searches at this point vs. providing a way for people to change it if they want to.  In the end the question becomes, "How many confused admins are we going to have out there when they upgrade from 1.2.0 to 1.3.0 and all of the sudden their group searches don't work as expected?"  Changing defaults should be done with care.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
echu2013: @necouchman do you find more convinient to replace "default" value of my proposed configuration item **ldap-group-search-filter** from `EqualityNode("objectClass","group")` to the original behaviour` PresenceNode("objectClass")`? 
In this way, if admin does not specify such filter then original behavior will be inplace. Otherwise, they will have the possibility of fine-grained control of which Groups shall be retrieved.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/ConfigurationService.java
necouchman: @mike-jumper: Thoughts on this?  The above change will essentially change the default group filter from `(objectClass=*)` to `(objectClass=group)`.  While I suspect that's probably okay and what we want in the long-run, I worry about catching unsuspecting admins off-guard...

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
necouchman: My point, here, is that, as you have this implemented, when `ldap-config-base-dn` is set, the new `ldap-group-search-filter` property will be completely ignored, and you'll always end up with the following group filter:
`(!(objectClass=guacConfigGroup)))`

So, the lines above should be something like:
```
if (confService.getConfigurationBaseDN() != null)
    return new AndNode(new NotNode(new EqualityNode("objectClass","guacConfigGroup"), confService.getGroupSearchFilter());
```
Assuming the default group filter, this will result in an LDAP filter like this:
`(&(!(objectClass=guacConfigGroup))(objectClass=group))`

necouchman:
@echu2013 Any progress on my most recent feedback? Also, you've got a merge master commit in here, now, so you'll need to untangle that and rebase the PR on top of the current master.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/ConfigurationService.java
mike-jumper: I think we should lean toward Principle of Least Surprise and maintain the existing `(objectClass=*)` behavior by default. The default behavior is intended to rely on LDAP directory structure to determine which objects should be returned and classified as users or groups, and I think it makes sense to continue to use that as the default behavior.

Filters like `ldap-user-search-filter` and soon-to-be-implemented `ldap-group-search-filter` exist to allow for more complex and intermingled cases, but default filters would involve too many assumptions and could break existing deployments.

msilveirabr:
I'm by no means an experienced programmer, but I managed to write a 'monkey see, monkey do' style patch against release 1.3.0. Sorry I didn't make it a pull request ( no free time to master git as a developer ATM ).
I've tested it with and without ldap-group-search-filter set, both resulted in expected behavior.
I didn't test it with guacConfigGroup Schema change.

Since there's 'postgresql-auto-create-accounts', the next logical step would be to implement (take postgresql as an example):
postgresql-auto-create-groups
postgresql-auto-create-groups-assignments


Here's a patch generated using 'diff -udpr' : [guacamole-client-1.3.0-ldap-group-filter.txt](https://github.com/apache/guacamole-client/files/6864850/guacamole-client-1.3.0-ldap-group-filter.txt)

mike-jumper:
> Since there's 'postgresql-auto-create-accounts', the next logical step would be to implement (take postgresql as an example):
> postgresql-auto-create-groups
> postgresql-auto-create-groups-assignments

@msilveirabr, what do you mean? The `*-auto-create-accounts` property exists for the sake of things like the TOTP extension that expect storage to be available for users once the user logs in. Except in those circumstances, even the user account doesn't need to exist in the database. Under what circumstances would a user's groups need to be automatically created?

necouchman:
This is closed by #629

echu2013:
> This is closed by #629

Thanks!

PR #498, opened 2020-04-06
GUACAMOLE-1007: Revert due to regresion 

echu2013:
"GUACAMOLE-955: Merge provide default translation string for all untranslated messages."

This commit triggers [GUACAMOLE-1007](https://issues.apache.org/jira/projects/GUACAMOLE/issues/GUACAMOLE-1007)

### DIFF
--- a/guacamole/src/main/java/org/apache/guacamole/rest/APIError.java
+++ b/guacamole/src/main/java/org/apache/guacamole/rest/APIError.java
@@ -20,7 +20,6 @@
 package org.apache.guacamole.rest;
 
 import java.util.Collection;
-import java.util.Collections;
 import org.apache.guacamole.GuacamoleClientException;
 import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.GuacamoleResourceNotFoundException;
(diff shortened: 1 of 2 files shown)


necouchman:
@echu2013: I'm not sure the answer is to completely revert these changes; rather, to figure out why these changes have triggered this behavior and correct that.

echu2013:
> @echu2013: I'm not sure the answer is to completely revert these changes; rather, to figure out why these changes have triggered this behavior and correct that.

@necouchman You are right, right now I am not able to circumvent the issue in this code. Do you have any clue?

mike-jumper:
The issue here is not that commit bc83918fb3df60a35279741a3dbe90a6092990da is faulty, but that additional changes are necessary for the intent of that change to be achieved. The proper way to throw an exception such that it is rendered as translatable text is to use an exception which implements `Translatable`. Commit bc83918fb3df60a35279741a3dbe90a6092990da solidified that, but code which relied on the older, incorrect behavior still needs to be updated to leverage `Translatable`.

I have a set of changes which I believe should address this. I will open a pull request shortly.

mike-jumper:
Closing in favor of #499 (see above).

PR #505, opened 2020-04-30
GUACAMOLE-989: Implement keyboard lock menu item.

necouchman:
One potential solution for attempting to use Keyboard Lock within Guacamole Client.  I'm not sure how much it actually helps - Guacamole already intercepts just about all of the keys it can, so I'm not certain this is actually going to provide any additional functionality, but maybe it's worth a shot.  Anyw testing that anyone wants to do would be welcome.

Also, if this menu isn't the right place for it we can find some other place.  Just seemed like a convenient place to stash it :-).

### DIFF
--- a/guacamole/src/main/webapp/app/client/controllers/clientController.js
+++ b/guacamole/src/main/webapp/app/client/controllers/clientController.js
@@ -33,6 +33,7 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
 
     // Required services
     var $location              = $injector.get('$location');
+    var $log                   = $injector.get('$log');
     var authenticationService  = $injector.get('authenticationService');
     var connectionGroupService = $injector.get('connectionGroupService');
     var clipboardService       = $injector.get('clipboardService');
@@ -895,9 +896,24 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
         className : 'danger disconnect',
         callback  : $scope.disconnect
     };
+    
+    /**
+     * Action that locks the keyboard using the Keyboard Lock API within the
+     * current client, and then closes the menu.
+     */
+    var KEYBOARD_LOCK_MENU_ACTION = {
+        name      : 'CLIENT.ACTION_KEYBOARD_LOCK',
+        callback  : function keyboardLock() {
+            if (navigator.keyboard) {
+                navigator.keyboard.lock();
+                $scope.menu.shown = false;
+            }
+        }
+    };
 
     // Set client-specific menu actions
-    $scope.clientMenuActions = [ DISCONNECT_MENU_ACTION ];
+    $scope.clientMenuActions = [ DISCONNECT_MENU_ACTION,
+        KEYBOARD_LOCK_MENU_ACTION ];
 
     /**
      * @borrows Protocol.getNamespace
--- a/guacamole/src/main/webapp/translations/en.json
+++ b/guacamole/src/main/webapp/translations/en.json
@@ -56,6 +56,7 @@
         "ACTION_ACKNOWLEDGE"               : "@:APP.ACTION_ACKNOWLEDGE",
         "ACTION_CLEAR_COMPLETED_TRANSFERS" : "Clear",
         "ACTION_DISCONNECT"                : "Disconnect",
+        "ACTION_KEYBOARD_LOCK"             : "Lock Keyboard",
         "ACTION_LOGOUT"                    : "@:APP.ACTION_LOGOUT",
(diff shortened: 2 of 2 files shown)


echu2013:
Hi! Did this allowed you to capture Alt+Tab key sequence? I am testing a script using ViolentMonkey chrome extension, script actually calls keyboard lock because in order to exit this mode I need to keep pressed ESC key a few seconds, but Alt+Tab is being still captured by underlying OS (Windows 10).

necouchman:
@echu2013: It never worked for me, but it's possible I was just doing something incorrectly.

echu2013:
@necouchman It's not you or me! Check [this Twitter post](https://twitter.com/KyleDavidE/status/1288545200353292289) out, precisely **Kyle Ehrlich's** statement:

> "It looks like it doesn't seem to work with `alt+tab` or any system shortcuts..."

necouchman:
@echu2013: I've received reports from various people with varying degrees of success. I think part of it is how you launch it - that is, I think if you use Chrome's capability to create a desktop shortcut that launches in its own window, and then use the keyboard lock, some additional key combos may work. Of course, ultimately it is up to the client O/S and browser what key combinations get intercepted and what are allowed to pass through, so, in the end, there are probably some that just will not work.

echu2013:
> @echu2013: I've received reports from various people with varying degrees of success. I think part of it is how you launch it - that is, I think if you use Chrome's capability to create a desktop shortcut that launches in its own window, and then use the keyboard lock, some additional key combos may work. Of course, ultimately it is up to the client O/S and browser what key combinations get intercepted and what are allowed to pass through, so, in the end, there are probably some that just will not work.

@necouchman : In my scenario, in which I have converted Guacamole into a "WebApp", launching it in App mode does not actually alter the behaviour in any way for me at least.

necouchman:
@echu2013: Good to know - I think that anything operating in a web browser is going to be limited in its ability to capture all key combos. Thanks for providing this detail.

mike-jumper:
So, I've been plowing through the PR backlog and came across these changes. Pulling them to test things out locally, I see the new menu option:

image

but clicking "Lock keyboard" seems to have no effect, at least when testing from my Chromebook. The Guacamole menu closes as a result of clicking the menu item, but that's it. Is there something I'm missing?

Regarding the functionality itself:

* We should probable use a different icon.
* What if the user wishes to unlock the keyboard? Should this be a checkbox?
* Would locating this elsewhere within Guacamole menu be better?

necouchman:
> but clicking "Lock keyboard" seems to have no effect, at least when testing from my Chromebook.

Yeah, that was my experience, too, and I'm not sure what I'm doing wrong. I put it out here for folks to test out, but clearly something isn't quite right.

Doing some additional reading on the API, it sounds like it only works if you first go into full-screen mode, so that may be part of the issue. It's been a while since I played around with it, and I cannot remember if that was part of my testing.

> We should probable use a different icon.

Sure - I'm not much of a graphics person, and I'm not sure where the existing graphics came from, but open to suggestions.

> What if the user wishes to unlock the keyboard? Should this be a checkbox?

Possibly so - it seems like a 2-5 second hold on the Escape key, at least in Chrome, is what is supposed to forcibly unlock the keyboard. At least one of the considerations should probably be whether enabling keyboard lock (in a way that actually functions) would prevent the Guacamole menu from being accessed at all?

> Would locating this elsewhere within Guacamole menu be better?

I am totally open to suggestions - no serious ties to any of the current way its implemented, where it's located, etc.

https://web.dev/keyboard-lock/
https://developer.mozilla.org/en-US/docs/Web/API/Keyboard/lock
https://wicg.github.io/keyboard-lock/

necouchman:
Okay, made one tweak, and I think I have it working. But, there are some caveats....

Based on my testing of this feature (via https://keyboard-lock.glitch.me/), I see the following behavior:
* Keyboard Lock only works when the browser is in full-screen mode.
* Keyboard lock only captures keys that are fully passed through to the browser and not intercepted by the O/S. For example, in my testing on the above page:
** After activating keyboard lock, Alt-Tab was still intercepted by the O/S
** As mentioned on the page, Control-N and Control-T were not intercepted by the browser
* Additionally, since the Escape key is used to exit Full Screen mode, it a) is not intercepted by Keyboard Lock, and b) results in the browser exiting Full Screen mode and turning off keyboard lock.

So, I'm not really sure how valuable this feature is going to be for Guacamole Client. Maybe there are some key combinations that it will help with, but I find that most of these are already taken care of, at least in Chrome, by making the Guacamole page into a Chrome Application.

PR #506, opened 2020-05-02
GUACAMOLE-944: Restore ability to use UPN-style usernames in search bind credentials

necouchman:
This restores the ability to use UPN-style usernames in guacamole.properties for the search bind user, which was broken when we moved over to the Apache Directory API.  I went ahead and based this against staging/1.2.0, but that's up for discussion as to whether this is truly a regression or not.  It certainly can be worked around even if it's less convenient - either way is fine with me.


mike-jumper:
I was confused for a moment seeing this closed by "asfgit", but I think this just happened automatically because of the deletion of the `staging/1.2.0` branch as part of the release procedures.

necouchman:
Ha - I've no idea why I would have opened that against `staging/1.2.0`.  Opened a new one against `master`.

PR #511, opened 2020-05-22
GUACAMOLE-1025: Allow QuickConnect module to configure allowed and/or denied parameters

necouchman:
This implements the necessary bits to allow someone to specify in the guacamole.properties file a list of either explicitly allowed parameters or a list of explicitly denied parameters for the QuickConnect module.  In specifying either or both of these options, an administrator can control what functionality or access end users have access to while still allowing the flexibility of ad-hoc connections.

### DIFF
--- a/extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectDirectory.java
+++ b/extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectDirectory.java
@@ -19,11 +19,13 @@
 
 package org.apache.guacamole.auth.quickconnect;
 
+import com.google.inject.Inject;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.auth.quickconnect.utility.QCParser;
+import org.apache.guacamole.auth.quickconnect.conf.ConfigurationService;
 import org.apache.guacamole.net.auth.ConnectionGroup;
 import org.apache.guacamole.net.auth.simple.SimpleConnection;
 import org.apache.guacamole.net.auth.simple.SimpleDirectory;
@@ -35,40 +37,46 @@ import org.apache.guacamole.protocol.GuacamoleConfiguration;
  * completely in memory.
  */
 public class QuickConnectDirectory extends SimpleDirectory<Connection> {
-
+    
+    /**
+     * The configuration service for the QuickConnect module.
+     */
+    @Inject
+    private ConfigurationService confService;
+    
     /**
      * The connections to store.
      */
     private final Map<String, Connection> connections =
-            new ConcurrentHashMap<String, Connection>();
+            new ConcurrentHashMap<>();
 
     /**
      * The root connection group for this directory.
      */
-    private final QuickConnectionGroup rootGroup;
+    private QuickConnectionGroup rootGroup;
 
     /**
      * The internal counter for connection IDs.
      */
-    private final AtomicInteger connectionId;
+    private AtomicInteger connectionId;
 
     /**
-     * Creates a new QuickConnectDirectory with the default
-     * empty Map for Connection objects, and the specified
-     * ConnectionGroup at the root of the directory.
+     * Initialize the QuickConnectDirectory with the default empty Map for
+     * Connection objects, and the specified ConnectionGroup at the root of
+     * the directory.
      *
      * @param rootGroup
      *     A group that should be at the root of this directory.
      */
-    public QuickConnectDirectory(ConnectionGroup rootGroup) {
+    public void init(ConnectionGroup rootGroup) {
         this.rootGroup = (QuickConnectionGroup)rootGroup;
         this.connectionId = new AtomicInteger();
         super.setObjects(this.connections);
+        
     }
 
     /**
-     * Returns the current connection identifier counter and
-     * then increments it.
+     * Returns the current connection identifier counter and then increments it.
      *
      * @return
      *     An int representing the next available connection
@@ -84,13 +92,14 @@ public class QuickConnectDirectory extends SimpleDirectory<Connection> {
     }
 
     /**
-     * Create a SimpleConnection object from a GuacamoleConfiguration,
-     * obtain an identifier, and place it on the tree, returning the
-     * identifier value of the new connection.
+     * Taking a URI, parse the URI into a GuacamoleConfiguration object and
+     * then use the configuration to create a SimpleConnection object, obtain
+     * an identifier, and place it on the tree, returning the identifier value
+     * of the new connection.
      *
-     * @param config
-     *     The GuacamoleConfiguration object to use to create the
-     *     SimpleConnection object.
+     * @param uri
+     *     The URI to parse into a GuacamoleConfiguration, which will then be
+     *     used to generate the SimpleConnection.
      *
      * @return
      *     The identifier of the connection created in the directory.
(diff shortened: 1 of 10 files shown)


extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/utility/QCParser.java
```diff
@@ -184,26 +178,39 @@ public static GuacamoleConfiguration getConfiguration(String uri)
...
         if (userinfoMatcher.matches()) {
             String username = userinfoMatcher.group(USERNAME_GROUP);
             String password = userinfoMatcher.group(PASSWORD_GROUP);
 
-            if (username != null && !username.isEmpty())
+            if (username != null && !username.isEmpty()) {
+                try {
                 config.setParameter("username",
                         URLDecoder.decode(username, "UTF-8"));
+                }
```
mike-jumper, line 199: Indentation here somehow became wonky.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/conf/StringListProperty.java
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * A GuacamoleProperty whose value is a List of Strings. The string value
+ * parsed to produce this list is a comma-delimited list. Duplicate values are
+ * ignored, as is any whitespace following delimiters. To maintain
+ * compatibility with the behavior of Java properties in general, only
+ * whitespace at the beginning of each value is ignored; trailing whitespace
+ * becomes part of the value.
+ */
+public abstract class StringListProperty implements GuacamoleProperty<List<String>> {
```
mike-jumper, line 36: I believe this is now part of guacamole-ext.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/conf/ConfigurationService.java
```diff
@@ -0,0 +1,93 @@
...
+        return environment.getProperty(QUICKCONNECT_ALLOWED_PARAMETERS);
+    }
+    
+    /**
+     * Return the list of denied parameters for connections created using the
+     * QuickConnect module, or null if none are defined (thereby allowing all
+     * parameters to be set).
+     * 
+     * @return
+     *     THe list of parameters that cannot be set by connections created
```
mike-jumper, line 83: The*

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectDirectory.java
```diff
@@ -118,5 +136,54 @@ public String create(GuacamoleConfiguration config) throws GuacamoleException {
...
+            if (deniedParams != null
+                    && !deniedParams.isEmpty()
+                    && deniedParams.contains(key)) {
+                
+                logger.debug("Parameter \"{}\" has been denied and will be removed.",
+                        key);
+                config.unsetParameter(key);
+                
+            }
+        }
```
mike-jumper, line 185: `config.getParameters().keySet().retainAll()` and `config.getParameters().keySet().removeAll()` would be easier than manually looping through the map.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectDirectory.java
```diff
@@ -118,5 +136,54 @@ public String create(GuacamoleConfiguration config) throws GuacamoleException {
...
+     * are allowed.
+     * 
+     * @param config
+     *     The GuacamoleConfiguration containing parameters to check against
+     *     configured allowed and denied values.
+     * 
+     * @throws GuacamoleException 
+     *     If guacamole.properties cannot be parsed.
+     */
+    private void checkConfigParameters(GuacamoleConfiguration config) throws GuacamoleException {
```
mike-jumper, line 156: Should this instead be part of `QCParser` (and tested via `QCParserTest`)? Comments within `QCParser` look like this may have originally been intended, and it feels like ensuring `QCParser` always abides by the set of allowed/denied parameters would be a safer approach

mike-jumper (changes requested):

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectDirectory.java
necouchman: Wow. Yes.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectDirectory.java
necouchman: Yeah, probably so - I think I was being lazy and trying to avoid having to pull the `ConfigurationService` class/injection into the `QCParser` class, but it's probably worth it.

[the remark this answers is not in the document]
necouchman: Renamed it to `purgeConfigParameters()` - if you have other ideas, I'm game, I was having a mental block coming up with something more apropos.

necouchman: Okay, moved it, with a couple of extra parameters to avoid having to pull `ConfigurationService` into the `QCParser` class.

necouchman: Also added unit tests...

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectDirectory.java
```diff
@@ -106,6 +114,12 @@ public String create(GuacamoleConfiguration config) throws GuacamoleException {
         // Generate a name for the configuration.
         String name = QCParser.getName(config);
 
+        // Purge configuration parameters that are either not allowed or are
+        // explicitly denied.
+        QCParser.purgeConfigParameters(config,
+                confService.getAllowedParameters(),
+                confService.getDeniedParameters());
```
mike-jumper, line 121: Manually purging parameters after the `GuacamoleConfiguration` has been created makes me uncomfortable. If this is ever accidentally omitted, then a security issue appears. Can this be done at the time that the `GuacamoleConfiguration` is built from the URL?

For example, if `QCParser` were concrete, with a constructor requiring a set of allowed and denied parameter, that would ensure that all parsing of URLs has to explicitly take the allow/deny bits into account, and any usage that (unsafely) _ignores_ those bits would need to be pretty purposefully written to ignore them.

mike-jumper (changes requested):

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/QuickConnectDirectory.java
necouchman: Sure, makes sense. I've got the changes almost done - just need to clean up the unit tests.

necouchman: Okay, things have been re-worked as suggested.

mike-jumper (approved):

PR #518, opened 2020-06-09
GUACAMOLE-903: Improved Chinese internationalization support

XYUU:
https://github.com/apache/guacamole-client/pull/453


extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/resources/guac-manifest.json
```diff
@@ -24,7 +24,8 @@
         "translations/es.json",
         "translations/fr.json",
         "translations/ja.json",
-        "translations/ru.json"
+        "translations/ru.json",
+        "translations/zh.json"
```
necouchman, line 28: Shouldn't this also be added to the `guac-manifest.json` file for PostgreSQL and SQL Server?

XYUU: okay，I add。

PR #519, opened 2020-06-11
GUACAMOLE-728: Implement EnumGuacamoleProperty.

mike-jumper:
This change implements a new property type, `EnumGuacamoleProperty`, intended to ease creation of properties that are driven by `enum` constants. Such properties formerly required defining a custom `GuacamoleProperty` subclass.

`EnumGuacamoleProperty` allows string/enum mapping through either of the following:

* Annotating the constants of an enum with their corresponding string values using `@PropertyValue`.
* Manually listing the constants and corresponding strings of an enum (typically necessary if the definition of the enum cannot be modified).


extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/conf/RadiusAuthenticationProtocol.java
```diff
@@ -19,100 +19,54 @@
...
-    private final String strValue;
-    
-    /**
-     * Create a new RadiusAuthenticationProtocol object having the
-     * given string value.
-     * 
-     * @param strValue
-     *     The value of the protocol to store as a string, which will be used
-     *     in specifying the protocol within the guacamole.properties file, and
-     *     will also be used by the JRadius library for its configuration.
```
mike-jumper, line 77: Eep - will deleting this break RADIUS? I don't recall the string value being used for JRadius, but if that's the case, I should definitely put some of this back.

guacamole-ext/src/test/java/org/apache/guacamole/properties/EnumGuacamolePropertyTest.java
```diff
@@ -0,0 +1,372 @@
...
+    /**
+     * Verifies that GuacamoleException is thrown when attempting to parse an
+     * invalid value, and that the error message contains a sorted list of all
+     * allowed values.
+     */
+    @Test
+    public void testParseInvalidValue() {
+        try {
+            FAVORITE_FISH.parseValue("anchovy");
+            fail("Invalid EnumGuacamoleProperty values should fail to parse with an exception.");
```
necouchman, line 163: Suggested change:
`fail("It isn't possible for anchovies to be anyone's favorite fish.");

;-)

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/conf/RadiusAuthenticationProtocol.java
necouchman: Okay, so, as long as it properly converts to a `String` using `toString()`, it should be okay:
```
        RadiusAuthenticator radAuth = radiusClient.getAuthProtocol(
                confService.getRadiusAuthProtocol().toString());
        
        if (radAuth == null)
            throw new GuacamoleException("Could not get a valid RadiusAuthenticator for specified protocol: " + confService.getRadiusAuthProtocol());
```
(https://github.com/apache/guacamole-client/blob/4e4b6f24f2d34a91f5381da955680f85c2ce56c3/extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/RadiusConnectionService.java#L129-L133)

The Inner Protocol may have to be re-worked, a bit:

```
        // If we're using EAP-TTLS, we need to define tunneled protocol
        if (radAuth instanceof EAPTTLSAuthenticator) {
            RadiusAuthenticationProtocol innerProtocol =
                    confService.getRadiusEAPTTLSInnerProtocol();
            
            if (innerProtocol == null)
                throw new GuacamoleException("Missing or invalid inner protocol for EAP-TTLS.");


            ((EAPTTLSAuthenticator)radAuth).setInnerProtocol(innerProtocol.toString());
```
(https://github.com/apache/guacamole-client/blob/4e4b6f24f2d34a91f5381da955680f85c2ce56c3/extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/RadiusConnectionService.java#L161-L170)

guacamole-ext/src/test/java/org/apache/guacamole/properties/EnumGuacamolePropertyTest.java
mike-jumper: True.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/LDAPGuacamoleProperties.java
```diff
@@ -181,8 +183,13 @@ private LDAPGuacamoleProperties() {}
...
      * (dereferences) aliases as it searches the tree.
      */
-    public static final DereferenceAliasesProperty LDAP_DEREFERENCE_ALIASES =
-            new DereferenceAliasesProperty() {
+    public static final EnumGuacamoleProperty<AliasDerefMode> LDAP_DEREFERENCE_ALIASES =
+            new EnumGuacamoleProperty<AliasDerefMode>(
+                "never",     AliasDerefMode.NEVER_DEREF_ALIASES,
+                "searching", AliasDerefMode.DEREF_IN_SEARCHING,
+                "finding",   AliasDerefMode.DEREF_FINDING_BASE_OBJ,
+                "always",    AliasDerefMode.DEREF_ALWAYS
```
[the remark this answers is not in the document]
mike-jumper: Yep, because `AliasDerefMode` is part of an API outside of this extension and Guacamole. It's not within our power to add `@PropertyValue` annotations there.

extensions/guacamole-auth-radius/src/main/java/org/apache/guacamole/auth/radius/conf/RadiusAuthenticationProtocol.java
mike-jumper: Ah, but "properly" there means into whatever value is required by JRadius.

I'll restore a means of getting the authenticator name specific to JRadius from the protocol.

necouchman: Okay - I basically implemented it such that the value set in `guacamole.properties` is the same value that `JRadius` uses, so it just need to be able to retrieve that value and pass it through.  If there's a way to do that mapping back to the `@PropertyValue` annotation I'm not opposed to that.

mike-jumper: OK - I've added:

* A constructor parameter for the JRadius protocol name (referencing the specific `NAME` constant) and static property, `JRADIUS_PROTOCOL_NAME`, that exposes this.
* A convenience function, `getAuthenticator()`, which returns the `RadiusAuthenticator` instance for the protocol in question.

and carried that refactor through.

PR #520, opened 2020-06-11
GUACAMOLE-1097: Spanish translations for authentication extensions

mmezog:
What the title says. Just Spanish translations for authentication plugins

### DIFF
--- a/extensions/guacamole-auth-cas/src/main/resources/translations/es.json
+++ b/extensions/guacamole-auth-cas/src/main/resources/translations/es.json
@@ -1,0 +1,12 @@
+{
+
+    "DATA_SOURCE_CAS" : {
+        "NAME" : "CAS SSO Backend"
+    },
+
+    "LOGIN" : {
+        "FIELD_HEADER_TICKET"        : "",
+        "INFO_CAS_REDIRECT_PENDING"  : "Por favor espere. Redirigiendo a la autenticación CAS..."
+    }
+
+}
--- a/extensions/guacamole-auth-duo/src/main/resources/translations/es.json
+++ b/extensions/guacamole-auth-duo/src/main/resources/translations/es.json
@@ -1,0 +1,13 @@
+{
+
+    "DATA_SOURCE_DUO" : {
+        "NAME" : "Duo TFA Backend"
+    },
+
+    "LOGIN" : {
+        "FIELD_HEADER_GUAC_DUO_SIGNED_RESPONSE" : "",
+        "INFO_DUO_VALIDATION_CODE_INCORRECT"    : "Código de validación Duo incorrecto.",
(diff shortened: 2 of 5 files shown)


extensions/guacamole-auth-cas/src/main/resources/translations/es.json
```diff
@@ -0,0 +1,12 @@
+{
+
+    "DATA_SOURCE_CAS" : {
+        "NAME" : "CAS SSO Backend"
```
mike-jumper, line 4: Is this correct? The Spanish translation for "CAS SSO Backend" is "CAS SSO Backend"? If yes, that's fine, but I wouldn't have expected the Spanish versions of the names of each of these backends to be identical to English.

necouchman:
@mmezog See Mike's feedback. If you can respond and/or update, I'd like to try to include these updates in the next release.

mmezog:
Sorry for the delay, I was not notified about the new comments or did not see them.

As for the question in itself, "CAS SSO Backend" is surely not Spanish, but it is used like that in computing jargon, at least in Spain (I don't know about other countries in South America). 

As a note, we use many English words in IT, even if we have corresponding Spanish equivalents. For example switch vs "conmutador", hub vs "concentrador", or plotter vs "trazador"... Backend and SSO are two examples of the same. I've never met anyone using the Spanish versions professionally. (I wouldn't even know how to call them).

necouchman:
@mmezog I know it's been quite a while, but if you'd like to get his rebased against staging/1.6.1 we can probably get this into the 1.6.1 release?

adb014:
@necouchman It seems stupid to let PR like this one die because the original author is not available for the rebase. What I'd suggest is pulling this PR to a local git rebasing and generating a new PR so as to keep the original author as the owner of the commit... I'd be willing to help out to do this one this and other PR facing the same issue to try and clear out the backlog of PR... Would this be acceptable ? Can you point me at the PR that need this treatment ?

necouchman:
@adb014 If you're willing to grab the commits and put them into a new pull request and take it through the process, you're certainly welcome to do that. Perfectly acceptable/encouraged - I think the only request is to try to maintain the original commits so that the original author gets some credit for it.

Thanks!

adb014:
> @adb014 If you're willing to grab the commits and put them into a new pull request and take it through the process, you're certainly welcome to do that. Perfectly acceptable/encouraged - I think the only request is to try to maintain the original commits so that the original author gets some credit for it.
> 
> Thanks!

Done. See #1240

adb014:
For PR that you basically accept, this process is trivial, as you can see by the time it took me to do it... What other PR could I update for inclusion in 1.6.1 ?

necouchman:
Closed by #1240.

PR #521, opened 2020-06-11
GUACAMOLE-1099: add Docker secrets support for LDAP properties

ss10sb:
Added _FILE for:

- LDAP_SEARCH_BIND_DN
- LDAP_SEARCH_BIND_PASSWORD
- LDAP_USER_BASE_DN
- LDAP_CONFIG_BASE_DN
- LDAP_GROUP_BASE_DN

Added SSL certificate support via LDAP_SSL_CERT_FILE (eg /run/secrets/ldap_cert) with keytool import for the file.  It could probably use an exists check but it works :)

### DIFF
--- a/guacamole-docker/bin/start.sh
+++ b/guacamole-docker/bin/start.sh
@@ -363,24 +363,47 @@ END
         exit 1;
     fi
 
+    # Handle Docker secrets for LDAP
+    if [ -n "$LDAP_SEARCH_BIND_DN_FILE" ]; then
+        set_optional_property "ldap-search-bind-dn" "$(cat $LDAP_SEARCH_BIND_DN_FILE)"
+    elif [ -n "$LDAP_SEARCH_BIND_DN" ]; then
+        set_optional_property "ldap-search-bind-dn" "$LDAP_SEARCH_BIND_DN"
+    fi
+
+    if [ -n "$LDAP_SEARCH_BIND_PASSWORD_FILE" ]; then
+        set_optional_property "ldap-search-bind-password" "$(cat $LDAP_SEARCH_BIND_PASSWORD_FILE)"
+    elif [ -n "$LDAP_SEARCH_BIND_PASSWORD" ]; then
+        set_optional_property "ldap-search-bind-password" "$LDAP_SEARCH_BIND_PASSWORD"
+    fi
+
+    if [ -n "$LDAP_USER_BASE_DN_FILE" ]; then
+        set_property "ldap-user-base-dn" "$(cat $LDAP_USER_BASE_DN_FILE)"
+    elif [ -n "$LDAP_USER_BASE_DN" ]; then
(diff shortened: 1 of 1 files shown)


necouchman:
@ss10sb: Thank you for the contribution, appreciate it.  Per our standard procedures, all pull requests need to be associated with a JIRA issue, and tagged with that.  Please see:

ss10sb:
[GUACAMOLE-1099](https://issues.apache.org/jira/browse/GUACAMOLE-1099)

guacamole-docker/bin/start.sh
```diff
@@ -363,24 +363,47 @@ END
...
+        set_optional_property "ldap-search-bind-dn" "$LDAP_SEARCH_BIND_DN"
+    fi
+
+    if [ -n "$LDAP_SEARCH_BIND_PASSWORD_FILE" ]; then
+        set_optional_property "ldap-search-bind-password" "$(cat $LDAP_SEARCH_BIND_PASSWORD_FILE)"
+    elif [ -n "$LDAP_SEARCH_BIND_PASSWORD" ]; then
+        set_optional_property "ldap-search-bind-password" "$LDAP_SEARCH_BIND_PASSWORD"
+    fi
+
+    if [ -n "$LDAP_USER_BASE_DN_FILE" ]; then
```
necouchman, line 379: Is there a particular reason to include the various DNs in the Docker secrets support? I understand for password, and possibly for username, but it doesn't seem like an LDAP DN is all that sensitive a value that it needs to be protected??

guacamole-docker/bin/start.sh
```diff
@@ -394,6 +417,10 @@ END
     set_optional_property "ldap-max-referral-hops"  "$LDAP_MAX_REFERRAL_HOPS"
     set_optional_property "ldap-operation-timeout"  "$LDAP_OPERATION_TIMEOUT"
 
+    if [ -n "$LDAP_SSL_CERT_FILE" ]; then
```
necouchman, line 420: I'm on the fence about this - I certainly don't have any problem supporting this functionality, but I'm not sure its entirely in scope of this particular change??

necouchman (changes requested):

mike-jumper (commented):
Reading the environment variable `FOO` from the file `FOO_FILE` feels like the sort of thing that would be best accomplished through changes to `set_optional_property` and `set_property`, rather than repeated `if` checks.

necouchman:
@ss10sb A handful of requests for changes, here - any chance you can polish this up so we can move forward with this? Also, there's a conflict, now, that needs to be resolved.

ss10sb:
> @ss10sb A handful of requests for changes, here - any chance you can polish this up so we can move forward with this? Also, there's a conflict, now, that needs to be resolved.

We're no longer using Guacamole + Docker. Sorry!

PR #530, opened 2020-06-21
GUACAMOLE-103: Implement common redirect form field.

necouchman:
As part of the implementation of the SAML extension in #254, this implements a common redirect-type field and changes over the existing CAS and OpenID extensions to leverage that field instead of implementing their own.


guacamole-ext/src/main/java/org/apache/guacamole/form/RedirectField.java
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * Returns the message that will be displayed for the user while the
+     * redirect takes place.
+     * 
+     * @return
+     *     The message to display for the user.
+     */
+    public String getRedirectMsg() {
+        return redirectMsg;
+    }
```
mike-jumper, line 84: I think:

* As this will be a part of the public extension API, it would be better to not abbreviate "message" as "msg".
* It would be better to leverage `TranslatableMessage` for the translatable message rather than just `String`, as the semantics surrounding `TranslatableMessage` are more explicit and variable substitution is inherently supported.

If you agree with both of the above, I suggest simply implementing `Translatable` and its `getTranslatableMessage()` function.

guacamole-ext/src/main/java/org/apache/guacamole/form/RedirectField.java
```diff
@@ -0,0 +1,86 @@
...
+
+import java.net.URI;
+
+/**
+ * A Guacamole field that redirects a user to another page.
+ */
+public class RedirectField extends Field {
+
+    /**
+     * The encoded URL of the redirect.
```
mike-jumper, line 30: What do you mean by encoded?

guacamole-ext/src/main/java/org/apache/guacamole/form/RedirectField.java
```diff
@@ -0,0 +1,86 @@
...
+        super(name, Field.Type.REDIRECT);
+
+        // Store the URL to which the user will be redirected.
+        this.redirectUrl = redirectUrl;
+        this.redirectMsg = redirectMsg;
+
+    }
+
+    /**
+     * Returns the URL of the redirect.
```
mike-jumper, line 66: "The URL of the redirect" reads a little unclear to me.

Earlier, in the constructor, you use the wording "the URL to which the user should be redirected", which is quite clear. Perhaps that would be better here?

guacamole-ext/src/main/java/org/apache/guacamole/form/RedirectField.java
necouchman: Totally good with this, just might need a little guidance on the best way to leverage this in a `Field` situation.  Based on the above note, the `redirectMsg` (now `redirectMessage`) variable is directly available in the scope.  But, if I move over to `Translatable` and `getTranslatableMessage()`, what's the best way to rework the AngularJS side such that this message will be properly displayed and run through the translation system?  I'm trying to follow examples in the existing code, so maybe I'll figure it out...

necouchman: I think I figured it out...

guacamole/src/main/webapp/app/form/templates/redirectField.html
```diff
@@ -0,0 +1,8 @@
+<div class="redirect-field-container">
+    <div class="redirect-field">
+        <p ng-show="redirectMessage"
+           translate="{{redirectMessage.key}}"
+           translate-values="{{redirectMessage.variables}}">
```
mike-jumper, line 5: 1. Does `redirectMessage` here need to be `field.redirectMessage`?
2. The getter on the Java side for this is `getTranslatableMessage()`, which would be serialized within the JSON as `translatableMessage` unless overridden by an annotation. I think this will need to either be referenced as-is (`translatableMessage`) or annotated such that it's exposed as `redirectMessage`.

guacamole-ext/src/main/java/org/apache/guacamole/form/Field.java
```diff
@@ -122,6 +122,12 @@
          * and protocols which leverage it.
          */
         public static String TERMINAL_COLOR_SCHEME = "TERMINAL_COLOR_SCHEME";
+        
+        /**
+         * A redirect field whose value is an encoded URL to which the user
+         * will be redirected.
+         */
+        public static String REDIRECT = "REDIRECT";
```
mike-jumper, line 130: I think we messed up in the past within this class, but this should be `final`.

extensions/guacamole-auth-openid/src/main/resources/transformToken.js
```diff
@@ -1,34 +0,0 @@
...
- * forward the "id_token" value for server-side validation.
- * 
- * Note that not all OpenID identity providers will include the "id_token"
- * parameter in the first position; it may occur after several other parameters
- * within the fragment.
- */
-(function guacOpenIDTransformToken() {
-    if (/^#(?![?\/])(.*&)?id_token=/.test(location.hash))
-        location.hash = '/?' + location.hash.substring(1);
-})();
```
mike-jumper, line 34: Is this no longer needed?

necouchman: For some reason I remember that you worked on a way to do this in the base code.  I cannot remember where, but I'll try to dig up the JIRA issue for it...

necouchman: Okay, I think I was confusing it with something else (more generic handling of URL parameters in Guacamole Client), so I've restored this file along with the relevant pieces of `pom.xml` and `guac-manifest.json` for the OpenID extension.

extensions/guacamole-auth-openid/src/main/resources/guac-manifest.json
```diff
@@ -17,14 +17,6 @@
...
-    ],
-
-    "css" : [
-        "openid.min.css"
-    ],
-
-    "resources" : {
-        "templates/openidTokenField.html" : "text/html"
-    }
+    ]
```
mike-jumper, line 20: With the template and CSS going away, are similar changes needed for the OpenID `pom.xml`?

The CAS `pom.xml` shows the JS/CSS minification and AngularJS template caching removed. For OpenID, it makes sense that JS minification would stick around, but what about the rest?

necouchman: Ah, I guess template caching can go away, and I can strip out the CSS minification bits.

PR #531, opened 2020-06-22
GUACAMOLE-728: Add Docker support for SSL configuration for MySQL and Postgres

necouchman:
Looking at the JIRA issue, it was originally entered as a Docker problem, so figured it might be nice to provide the Docker bits consistent with existing Docker functionality.

### DIFF
--- a/guacamole-docker/bin/start.sh
+++ b/guacamole-docker/bin/start.sh
@@ -151,7 +151,7 @@ END
 
     # Verify that the required Docker secrets are present, else, default to their normal environment variables
     if [ -n "$MYSQL_USER_FILE" ]; then
-        set_property "mysql-username" `cat $MYSQL_USER_FILE`
+        set_property "mysql-username" "`cat "$MYSQL_USER_FILE"`"
     elif [ -n "$MYSQL_USER" ]; then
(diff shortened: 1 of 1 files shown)


guacamole-docker/bin/start.sh
```diff
@@ -205,6 +205,32 @@ END
...
+        "mysql-ssl-mode"  \
+        "$MYSQL_SSL_MODE"
+
+    set_optional_property        \
+        "mysql-ssl-trust-store"  \
+        "$MYSQL_SSL_TRUST_STORE"
+
+    # For SSL trust store password, check secrets, first, then standard env variable
+    if [ -n "$MYSQL_SSL_TRUST_PASSWORD_FILE" ]; then
+        set_property "mysql-ssl-trust-password" `cat $MYSQL_SSL_TRUST_PASSWORD_FILE`
```
mike-jumper, line 218: The `$MYSQL_SSL_TRUST_PASSWORD_FILE` variable will need to be quoted in case it expands to a value containing whitespace, as does backtick-quoted command (in case the command outputs a value containing whitespace). For example:

    "`cat "$FOO"`"

Same for the other invocations.

PR #532, opened 2020-06-22
GUACAMOLE-1081: Fix issues with upper-case modifiers.

necouchman:


mike-jumper:
I'll think we should close this PR for now - [per the discussion in JIRA](https://issues.apache.org/jira/browse/GUACAMOLE-1081?focusedCommentId=17146623&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17146623), I don't think these changes should be necessary.

Canonicalization is sometimes used to produce a consistent, predictable token name, but it doesn't enter into the processing of token names themselves. It also shouldn't be the case that the storage of connection parameters has any impact - whether that's LDAP or MySQL or something else, all connection parameter storage safely preserves case (a storage mechanism that magically turns parameter values uppercase would break all sorts of things).

I suspect there is no case where something like `${TOKEN_NAME:lower}` is being turned into `${TOKEN_NAME:LOWER}`, and that the token itself may simply have been entered incorrectly. If there _is_ something doing this, then we should determine what is causing that before committing to simply accepting both "lower" and "LOWER" as modifiers.

(That said ... if it was user error resulting in this, and the chosen format is simply easy to mess up because token names are generally always upper case, perhaps we should consider switching to "LOWER", "UPPER", etc. instead?)

PR #534, opened 2020-06-25
GUACAMOLE-103: Adjust checks for entity ID and ACS URL properties.

necouchman:


extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/conf/ConfigurationService.java
```diff
@@ -159,11 +159,10 @@
      *     The URL to send to the SAML IdP as the Client Identifier.
      *
      * @throws GuacamoleException
-     *     If guacamole.properties cannot be parsed, or if the
-     *     property is missing.
+     *     If guacamole.properties cannot be parsed.
      */
     private URI getEntityId() throws GuacamoleException {
-        return environment.getRequiredProperty(SAML_ENTITY_ID);
+        return environment.getProperty(SAML_ENTITY_ID);
```
mike-jumper, line 165: It looks like `URIGuacamoleProperty` was not written to handle `null` values, and a `NullPointerException` is thrown in the case that this property is actually missing.

I think you'll have to add a `null` check within `parseValue()` of `URIGuacamoleProperty`, either as part of this or part of [GUACAMOLE-678](https://issues.apache.org/jira/browse/GUACAMOLE-678).

necouchman: Okay, re-opened GUACAMOLE-678 and did a pull request for that one.

necouchman:
Ahem...I had this set to `master` instead of `staging/1.2.0`.

mike-jumper:
Arg. Well ... if you re-PR the same for `staging/1.2.0`, I'll merge, etc. It's a little wonky, but git should figure it out and what's done is done.

necouchman:
Done...sorry about that...

PR #535, opened 2020-06-25
GUACAMOLE-678: Return null for null property values.

necouchman:


necouchman:
FYI - I see these changes in `staging/1.2.0`, but not in `master`, yet?

PR #543, opened 2020-06-29
GUACAMOLE-944: Allow search bind usernames in non-DN format.

necouchman:


extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/LdapBindGuacamoleProperty.java
```diff
@@ -0,0 +1,61 @@
...
+ * LDAP bind format - either an Active Directory UPN or a LDAP DN.  An exception
+ * is thrown if the provided value is invalid.
+ */
+public abstract class LdapBindGuacamoleProperty implements GuacamoleProperty<String> {
+
+    /**
+     * An expression that matches valid Active Directory UPN values, in one of
+     * two formats - either <DOMAIN>\<USERNAME> or <USERNAME>@<DOMAIN>.
+     */
+    private static final Pattern LDAP_UPN_PATTERN = Pattern.compile("^(?:(?<adusername>[^@]+)@(?<addomain>.+)|(?<nbdomain>[^\\\\]+)\\\\(?<nbusername>.+))$");
```
mike-jumper, line 40: Should a simple domain-less username be accepted?

necouchman: Yeah, I guess we should allow that, too...

necouchman: Okay, I _think_ I've managed to do this...

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/LdapBindGuacamoleProperty.java
```diff
@@ -0,0 +1,61 @@
...
+ * LDAP bind format - either an Active Directory UPN or a LDAP DN.  An exception
+ * is thrown if the provided value is invalid.
+ */
+public abstract class LdapBindGuacamoleProperty implements GuacamoleProperty<String> {
+
+    /**
+     * An expression that matches valid Active Directory UPN values, in one of
+     * three formats - <DOMAIN>\<USERNAME>, <USERNAME>@<DOMAIN> or <USERNAME>
+     */
+    private static final Pattern LDAP_UPN_PATTERN = Pattern.compile("^(?:(?<adusername>)([^@]+(?<addomain>.*))|(?<nbdomain>[^\\\\]+)\\\\(?<nbusername>.+))$");
```
mike-jumper, line 40: I've been playing around with this regex and various test strings, and the regex itself is actually pretty broad. The documentation for the regex sets an honorable goal, but the only strings that will not match this are strings that start with "@" but do not end with "\something".

Checking [Microsoft's documentation for the UPN](https://docs.microsoft.com/en-us/windows/win32/adschema/a-userprincipalname), it's clear that there is an intent for the value to follow a specific format ("based on" RFC 822 and "should" be the user's email address), but there is no apparent hard requirement except that it must be a Unicode string.

Also, while Active Directory may specifically allow username-style binds driven by `userPrincipalName`, I'm not sure this goes for other LDAP server implementations that may have their own definitions for what usernames can be used in lieu of an FQDN.

I think we should probably stick to a plain `StringGuacamoleProperty`...

necouchman: Yeah, that makes sense. I've reverted it to just a `StringGuacamoleProperty`.

mike-jumper (approved):

PR #545, opened 2020-07-03
GUACAMOLE-1125: Fix substituteKeysPressed handler on guacKeyup events

echu2013:
Issue which evidenced starting with guacamole-server version 1.2.0. Send guacSyntheticKeyup (DEL key here) BEFORE deleting it for proper behavior.

### DIFF
--- a/guacamole/src/main/webapp/app/client/controllers/clientController.js
+++ b/guacamole/src/main/webapp/app/client/controllers/clientController.js
@@ -678,8 +678,8 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
         // Deal with substitute key presses
         if (substituteKeysPressed[keysym]) {
             event.preventDefault();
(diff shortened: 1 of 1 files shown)


echu2013:
> @echu2013: Thanks for the quick-fix - I think it looks fine. My only request is that you tweak the commit message and PR title a bit to be less-specific to keys/key sequences and somewhat more generic. While you're seeing this in Ctrl-Alt-End, the issue actually fixes any/all substitute key presses that might be present.

Does now make sense???

necouchman:
@echu2013: Much better - one more request - can you adjust the commit message so that it isn't cut off?  I think a single line for this one will be okay :-).

echu2013:
> @echu2013: Much better - one more request - can you adjust the commit message so that it isn't cut off? I think a single line for this one will be okay :-).

Is it better now?

PR #546, opened 2020-07-04
GUACAMOLE-1123: Standardize REST endpoints for historical data

necouchman:
This standardizes REST endpoints for retrieving history for both connections and users.  There are a handful of items in here I'm unsure of in terms of best practice or not, so let me know what needs to be fixed.  So far I have only tested against PostgreSQL, so probably needs some verification against MySQL/MariaDB and SQL Server.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionService.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionService.java
@@ -213,7 +213,7 @@ public class ConnectionService extends ModeledChildDirectoryObjectService<Modele
         Map<String, String> parameters = connection.getConfiguration().getParameters();
         
         // Convert parameters to model objects
-        Collection<ConnectionParameterModel> parameterModels = new ArrayList<ConnectionParameterModel>(parameters.size());
+        Collection<ConnectionParameterModel> parameterModels = new ArrayList<>(parameters.size());
         for (Map.Entry<String, String> parameterEntry : parameters.entrySet()) {
 
             // Get parameter name and value
@@ -329,7 +329,7 @@ public class ConnectionService extends ModeledChildDirectoryObjectService<Modele
     public Map<String, String> retrieveParameters(ModeledAuthenticatedUser user,
             String identifier) {
 
-        Map<String, String> parameterMap = new HashMap<String, String>();
+        Map<String, String> parameterMap = new HashMap<>();
 
         // Determine whether we have permission to read parameters
         boolean canRetrieveParameters;
@@ -382,7 +382,7 @@ public class ConnectionService extends ModeledChildDirectoryObjectService<Modele
     protected List<ConnectionRecord> getObjectInstances(List<ConnectionRecordModel> models) {
 
         // Create new list of records by manually converting each model
-        List<ConnectionRecord> objects = new ArrayList<ConnectionRecord>(models.size());
+        List<ConnectionRecord> objects = new ArrayList<>(models.size());
         for (ConnectionRecordModel model : models)
             objects.add(getObjectInstance(model));
 
@@ -411,28 +411,16 @@ public class ConnectionService extends ModeledChildDirectoryObjectService<Modele
             ModeledConnection connection) throws GuacamoleException {
 
         String identifier = connection.getIdentifier();
-
-        // Retrieve history only if READ permission is granted
-        if (hasObjectPermission(user, identifier, ObjectPermission.Type.READ)) {
-
-            // Retrieve history
-            List<ConnectionRecordModel> models = connectionRecordMapper.select(identifier);
-
-            // Get currently-active connections
-            List<ConnectionRecord> records = new ArrayList<ConnectionRecord>(tunnelService.getActiveConnections(connection));
-            Collections.reverse(records);
-
-            // Add past connections from model objects
-            for (ConnectionRecordModel model : models)
-                records.add(getObjectInstance(model));
-
-            // Return converted history list
-            return records;
-
-        }
-
-        // The user does not have permission to read the history
-        throw new GuacamoleSecurityException("Permission denied.");
+        
+        // Get current active connections.
+        List<ConnectionRecord> records = new ArrayList<>(tunnelService.getActiveConnections(connection));
+        Collections.reverse(records);
+        
+        // Add in the history records.
+        records.addAll(retrieveHistory(identifier, user, Collections.emptyList(),
+                Collections.emptyList(), Integer.MAX_VALUE));
+        
+        return records;
 
     }
 
@@ -442,6 +430,60 @@ public class ConnectionService extends ModeledChildDirectoryObjectService<Modele
      * the given terms and sorted by the given predicates. Only history records
      * associated with data that the given user can read are returned.
      *
+     * @param identifier
+     *     The optional connection identifier for which history records should
+     *     be retrieved, or null if all readable records should be retrieved.
+     * 
+     * @param user
+     *     The user retrieving the connection history.
+     *
+     * @param requiredContents
+     *     The search terms that must be contained somewhere within each of the
+     *     returned records.
+     *
+     * @param sortPredicates
+     *     A list of predicates to sort the returned records by, in order of
+     *     priority.
+     *
+     * @param limit
+     *     The maximum number of records that should be returned.
+     *
+     * @return
+     *     The connection history of the given connection, including any
+     *     active connections.
+     *
+     * @throws GuacamoleException
+     *     If permission to read the connection history is denied.
+     */
+    public List<ConnectionRecord> retrieveHistory(String identifier,
+            ModeledAuthenticatedUser user,
+            Collection<ActivityRecordSearchTerm> requiredContents,
+            List<ActivityRecordSortPredicate> sortPredicates, int limit)
+            throws GuacamoleException {
+
+        List<ConnectionRecordModel> searchResults;
+
+        // Bypass permission checks if the user is privileged
+        if (user.isPrivileged())
+            searchResults = connectionRecordMapper.search(identifier, requiredContents,
+                    sortPredicates, limit);
+
+        // Otherwise only return explicitly readable history records
+        else
+            searchResults = connectionRecordMapper.searchReadable(identifier,
+                    user.getUser().getModel(), requiredContents, sortPredicates,
+                    limit, user.getEffectiveUserGroups());
+
+        return getObjectInstances(searchResults);
+
+    }
+    
+    /**
+     * Retrieves the connection history records matching the given criteria.
+     * Retrieves up to <code>limit</code> connection history records matching
+     * the given terms and sorted by the given predicates. Only history records
+     * associated with data that the given user can read are returned.
+     * 
      * @param user
      *     The user retrieving the connection history.
      *
@@ -467,22 +509,9 @@ public class ConnectionService extends ModeledChildDirectoryObjectService<Modele
             Collection<ActivityRecordSearchTerm> requiredContents,
             List<ActivityRecordSortPredicate> sortPredicates, int limit)
             throws GuacamoleException {
-
-        List<ConnectionRecordModel> searchResults;
-
-        // Bypass permission checks if the user is privileged
-        if (user.isPrivileged())
-            searchResults = connectionRecordMapper.search(requiredContents,
-                    sortPredicates, limit);
-
-        // Otherwise only return explicitly readable history records
-        else
-            searchResults = connectionRecordMapper.searchReadable(
-                    user.getUser().getModel(), requiredContents, sortPredicates,
-                    limit, user.getEffectiveUserGroups());
-
-        return getObjectInstances(searchResults);
-
+        
+        return retrieveHistory(null, user, requiredContents, sortPredicates, limit);
+        
     }
 
     /**
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/UserService.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/UserService.java
@@ -32,7 +32,6 @@ import org.apache.guacamole.auth.jdbc.base.ModeledDirectoryObjectMapper;
 import org.apache.guacamole.auth.jdbc.base.ModeledDirectoryObjectService;
 import org.apache.guacamole.GuacamoleClientException;
 import org.apache.guacamole.GuacamoleException;
-import org.apache.guacamole.GuacamoleSecurityException;
 import org.apache.guacamole.GuacamoleUnsupportedException;
 import org.apache.guacamole.auth.jdbc.base.ActivityRecordModel;
 import org.apache.guacamole.auth.jdbc.base.ActivityRecordSearchTerm;
@@ -581,13 +580,9 @@ public class UserService extends ModeledDirectoryObjectService<ModeledUser, User
             ModeledUser user) throws GuacamoleException {
 
         String username = user.getIdentifier();
-
-        // Retrieve history only if READ permission is granted
-        if (hasObjectPermission(authenticatedUser, username, ObjectPermission.Type.READ))
-            return getObjectInstances(userRecordMapper.select(username));
-
-        // The user does not have permission to read the history
-        throw new GuacamoleSecurityException("Permission denied.");
+        
+        return retrieveHistory(username, authenticatedUser, Collections.emptyList(),
+                Collections.emptyList(), Integer.MAX_VALUE);
 
     }
 
@@ -597,6 +592,59 @@ public class UserService extends ModeledDirectoryObjectService<ModeledUser, User
      * given terms and sorted by the given predicates. Only history records
      * associated with data that the given user can read are returned.
      *
+     * @param username
+     *     The optional username to which history records should be limited, or
+     *     null if all readable records should be retrieved.
+     * 
+     * @param user
+     *     The user retrieving the login history.
+     *
+     * @param requiredContents
+     *     The search terms that must be contained somewhere within each of the
+     *     returned records.
+     *
+     * @param sortPredicates
+     *     A list of predicates to sort the returned records by, in order of
+     *     priority.
+     *
+     * @param limit
+     *     The maximum number of records that should be returned.
+     *
+     * @return
+     *     The login history of the given user, including any active sessions.
+     *
+     * @throws GuacamoleException
+     *     If permission to read the user login history is denied.
+     */
+    public List<ActivityRecord> retrieveHistory(String username,
+            ModeledAuthenticatedUser user,
+            Collection<ActivityRecordSearchTerm> requiredContents,
+            List<ActivityRecordSortPredicate> sortPredicates, int limit)
+            throws GuacamoleException {
+
+        List<ActivityRecordModel> searchResults;
+
+        // Bypass permission checks if the user is privileged
+        if (user.isPrivileged())
+            searchResults = userRecordMapper.search(username, requiredContents,
+                    sortPredicates, limit);
+
+        // Otherwise only return explicitly readable history records
+        else
+            searchResults = userRecordMapper.searchReadable(username, 
+                    user.getUser().getModel(),
+                    requiredContents, sortPredicates, limit, user.getEffectiveUserGroups());
+
+        return getObjectInstances(searchResults);
+
+    }
+    
+    /**
+     * Retrieves user login history records matching the given criteria.
+     * Retrieves up to <code>limit</code> user history records matching the
+     * given terms and sorted by the given predicates. Only history records
+     * associated with data that the given user can read are returned.
+     * 
      * @param user
      *     The user retrieving the login history.
      *
@@ -621,21 +669,9 @@ public class UserService extends ModeledDirectoryObjectService<ModeledUser, User
             Collection<ActivityRecordSearchTerm> requiredContents,
             List<ActivityRecordSortPredicate> sortPredicates, int limit)
             throws GuacamoleException {
-
-        List<ActivityRecordModel> searchResults;
-
-        // Bypass permission checks if the user is privileged
-        if (user.isPrivileged())
-            searchResults = userRecordMapper.search(requiredContents,
-                    sortPredicates, limit);
-
-        // Otherwise only return explicitly readable history records
-        else
-            searchResults = userRecordMapper.searchReadable(user.getUser().getModel(),
-                    requiredContents, sortPredicates, limit, user.getEffectiveUserGroups());
-
-        return getObjectInstances(searchResults);
-
+        
+        return retrieveHistory(null, user, requiredContents, sortPredicates, limit);
+        
     }
 
 }
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/resources/org/apache/guacamole/auth/jdbc/connection/ConnectionRecordMapper.xml
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/resources/org/apache/guacamole/auth/jdbc/connection/ConnectionRecordMapper.xml
@@ -108,28 +108,35 @@
         LEFT JOIN guacamole_user       ON guacamole_connection_history.user_id       = guacamole_user.user_id
 
         <!-- Search terms -->
-        <foreach collection="terms" item="term"
-                 open="WHERE " separator=" AND ">
-            (
+        <where>
+            
+            <if test="identifier != null">
+                guacamole_connection_history.connection_id = #{identifier,jdbcType=VARCHAR}
+            </if>
+            
+            <foreach collection="terms" item="term" open=" AND " separator=" AND ">
+                (
 
-                guacamole_connection_history.user_id IN (
-                    SELECT user_id
-                    FROM guacamole_user
-                    WHERE POSITION(#{term.term,jdbcType=VARCHAR} IN username) > 0
+                    guacamole_connection_history.user_id IN (
+                        SELECT user_id
+                        FROM guacamole_user
+                        WHERE POSITION(#{term.term,jdbcType=VARCHAR} IN username) > 0
+                    )
+
+                    OR guacamole_connection_history.connection_id IN (
+                        SELECT connection_id
+                        FROM guacamole_connection
+                        WHERE POSITION(#{term.term,jdbcType=VARCHAR} IN connection_name) > 0
+                    )
+
+                    <if test="term.startDate != null and term.endDate != null">
+                        OR start_date BETWEEN #{term.startDate,jdbcType=TIMESTAMP} AND #{term.endDate,jdbcType=TIMESTAMP}
+                    </if>
+
                 )
-
-                OR guacamole_connection_history.connection_id IN (
-                    SELECT connection_id
-                    FROM guacamole_connection
-                    WHERE POSITION(#{term.term,jdbcType=VARCHAR} IN connection_name) > 0
-                )
-
-                <if test="term.startDate != null and term.endDate != null">
-                    OR start_date BETWEEN #{term.startDate,jdbcType=TIMESTAMP} AND #{term.endDate,jdbcType=TIMESTAMP}
-                </if>
-
-            )
-        </foreach>
+            </foreach>
+            
+        </where>
 
         <!-- Bind sort property enum values for sake of readability -->
         <bind name="START_DATE" value="@org.apache.guacamole.net.auth.ActivityRecordSet$SortableProperty@START_DATE"/>
@@ -186,31 +193,38 @@
             AND guacamole_user_permission.permission = 'READ'
 
         <!-- Search terms -->
-        <foreach collection="terms" item="term"
-                 open="WHERE " separator=" AND ">
-            (
+        <where>
+            
+            <if test="identifier != null">
+                guacamole_connection_history.connection_id = #{identifier,jdbcType=VARCHAR}
+            </if>
+            
+            <foreach collection="terms" item="term" open=" AND " separator=" AND ">
+                (
 
-                guacamole_connection_history.user_id IN (
-                    SELECT user_id
-                    FROM guacamole_user
-                    JOIN guacamole_entity ON guacamole_user.entity_id = guacamole_entity.entity_id
-                    WHERE
-                            POSITION(#{term.term,jdbcType=VARCHAR} IN guacamole_entity.name) > 0
-                        AND guacamole_entity.type = 'USER'
+                    guacamole_connection_history.user_id IN (
+                        SELECT user_id
+                        FROM guacamole_user
+                        JOIN guacamole_entity ON guacamole_user.entity_id = guacamole_entity.entity_id
+                        WHERE
+                                POSITION(#{term.term,jdbcType=VARCHAR} IN guacamole_entity.name) > 0
+                            AND guacamole_entity.type = 'USER'
+                    )
+
+                    OR guacamole_connection_history.connection_id IN (
+                        SELECT connection_id
+                        FROM guacamole_connection
+                        WHERE POSITION(#{term.term,jdbcType=VARCHAR} IN connection_name) > 0
+                    )
+
+                    <if test="term.startDate != null and term.endDate != null">
+                        OR start_date BETWEEN #{term.startDate,jdbcType=TIMESTAMP} AND #{term.endDate,jdbcType=TIMESTAMP}
+                    </if>
+
                 )
-
-                OR guacamole_connection_history.connection_id IN (
-                    SELECT connection_id
-                    FROM guacamole_connection
-                    WHERE POSITION(#{term.term,jdbcType=VARCHAR} IN connection_name) > 0
-                )
-
-                <if test="term.startDate != null and term.endDate != null">
-                    OR start_date BETWEEN #{term.startDate,jdbcType=TIMESTAMP} AND #{term.endDate,jdbcType=TIMESTAMP}
-                </if>
-
-            )
-        </foreach>
+            </foreach>
+        
+        </where>
 
         <!-- Bind sort property enum values for sake of readability -->
         <bind name="START_DATE" value="@org.apache.guacamole.net.auth.ActivityRecordSet$SortableProperty@START_DATE"/>
--- a/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Connection.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Connection.java
@@ -110,6 +110,24 @@ public interface Connection extends Identifiable, Connectable, Attributes {
     public List<? extends ConnectionRecord> getHistory() throws GuacamoleException;
 
     /**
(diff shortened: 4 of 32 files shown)


necouchman:
Okay, tested with MariaDB and SQL Server and amended a few things - seems to be working across the board.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/ActivityRecordSet.java
```diff
@@ -56,6 +56,24 @@
...
+     * 
+     * @return
+     *     A collection containing records within this set for the specified
+     *     identifier.
+     *
+     * @throws GuacamoleException
+     *     If an error occurs while retrieving the records within this set.
+     */
+    Collection<RecordType> asCollection(String identifier)
+            throws GuacamoleException;
```
mike-jumper, line 76: A couple things:

* To continue maintaining backward compatibility, default implementations need to be provided for new functions added to existing interfaces within guacamole-ext.
* As defined, this will be ambiguous as to which type of object the identifier refers to. The records within the collection may point to multiple types of objects. For example, a connection record points to both a user and a connection, and may also point to a sharing profile.
* The general pattern for `ActivityRecordSet` is a set of functions which can be layered on top of each other by virtue of those functions each returning `ActivityRecordSet`, similar to the various "builder" objects we see in JAX-RS. Keeping with that pattern, I think this would be best implemented not as new version of `asCollection()`, but as some other function that allows the set to be narrowed to records containing a particular object having a particular identifier.

   I could imagine a new version of `contains()` which accepts an `Identifiable` rather than a `String`, or perhaps multiple `contains()` for `Connection`, `User`, `SharingProfile`, etc.

guacamole/src/main/java/org/apache/guacamole/rest/history/HistoryResource.java
```diff
@@ -86,8 +93,9 @@ public HistoryResource(UserContext userContext) {
      *     If an error occurs while retrieving the connection history.
      */
     @GET
-    @Path("connections")
+    @Path("connections{p:/?}{identifier : ([0-9]+)?}")
     public List<APIConnectionRecord> getConnectionHistory(
+            @PathParam("identifier") String identifier,
```
mike-jumper, line 98: Beware that not all identifiers are numeric. That said, I don't think this is the best place to add this. I think it would be better to continue using the existing per-connection endpoint for connection-specific history searches, but attempt to refactor `HistoryResource` such that it can be used both at the global level and per-connection / per-user.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/ActivityRecordSet.java
mike-jumper: Another possibility could be to add something like `getConnectionHistory()` and `getUserHistory()` (returning `ActivityRecordSet` like their `UserContext` versions) to `Connection` and `User` respectively, deprecating the old `getHistory()`.

That seems clean but would also lend itself well to default implementations (could just wrap `getHistory()` in some sort of basic `ActivityRecordSet` implementation).

necouchman: ok

guacamole/src/main/java/org/apache/guacamole/rest/history/HistoryResource.java
necouchman: Ah, okay.  I think I misunderstood your response to my question in Slack - I thought you were saying the existing API endpoint should be removed in favor of the common `history` endpoint that is already filter-able.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/ActivityRecordSet.java
necouchman: Okay, I have taken a wild stab at this - let me know how it looks, now.

necouchman:
Okay, @mike-jumper, I've significantly refactored this, but sure it still needs some work.

I've also got an error with the current code I cannot quite figure out - I switched the `connection/<identifier>/history` and `user/<identifier>/history` resources to use the common ones you implemented, but this is yielding an error:

HTTP Status 500 - No serializer found for class org.apache.guacamole.rest.history.ConnectionHistoryResource and no properties discovered to create BeanSerializer (to avoid exception, disable SerializationConfig.Feature.FAIL_ON_EMPTY_BEANS) )

I've done a little Google Research and tried a couple of things but not exactly sure why it isn't working, there.  If you have any hints I'll be happy to try things out...

mike-jumper:
Odd. I'll try checking out this PR locally and see if I can reproduce the error.

mike-jumper:
For the functions which now return `ConnectionHistoryResource` (or `UserHistoryResource`), the `@GET` annotation just needs to be removed. It's no longer needed as it's the superclass of those resources that provides the function invoked for `@GET`.

Without the annotation, the JAX-RS implementation will continue searching through the returned resource for the function to invoke for the GET request, finally serializing the result of _that_ function as its response. This is the behavior we want. The presence of that annotation is causing the JAX-RS implementation to not continue its search and instead attempt to serialize `ConnectionHistoryResource` for the body of the response.

necouchman:
Thanks, Mike.  Will remove them.

necouchman:
Okay, should be cleaned up.

I think the only other thing I'm unsure about at this point is whether passing around the `null` identifier is the best way to go.  It keeps the code as lean as possible by leveraging the existing MyBatis classes and queries without the need for additional queries, and still leverages the database to filter the data (which should be the fastest/most efficient way to accomplish that), but passing around `null` feels a little strange.  Is that okay, or should I take a different approach, there - like just pulling all the records from the DB and doing some filtering within a Java class?

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/User.java
```diff
@@ -103,7 +103,26 @@
      *     If an error occurs while reading the history of this user, or if
      *     permission is denied.
      */
+    @Deprecated
```
mike-jumper, line 106: We'll have to be careful here:

* As written, deprecated or not, an implementation of `User` will still be required to implement the older `getHistory()` function.
* Providing a default implementation of `getHistory()` which leverages `getUserHistory()` would allow things to behave as expected, but then older extensions which implement only `getHistory()` will fail.
* Providing default implementations of _both_ which each point at the other would would technically work, but would make implementing `User` dangerous. An extension that fails to implement either function would compile without any warning or error, yet invoking either function would result in infinite recursion.

The way we worked around this for similar changes to `Connectable` was to provide a default implementation only for the deprecated function at the library level, but override that internally with an identical interface specific to the web application that provides default implementations of both.

Finding some way to achieve this purely through enhancements to `ActivityRecordSet` could avoid the above.

mike-jumper (changes requested):

mike-jumper:
> I think the only other thing I'm unsure about at this point is whether passing around the `null` identifier is the best way to go. It keeps the code as lean as possible by leveraging the existing MyBatis classes and queries without the need for additional queries, and still leverages the database to filter the data (which should be the fastest/most efficient way to accomplish that), but passing around `null` feels a little strange. Is that okay, or should I take a different approach, there - like just pulling all the records from the DB and doing some filtering within a Java class?

So long as the meaning of `null` is well documented, I think this is fine. For functions which accept multiple possible criteria to narrow a result, the pattern of leveraging `null` as a wildcard is pretty common, and I'm all for leveraging a feature of MyBatis to keep things clean.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/User.java
necouchman: Well, I've tried to take care of this, but not 100% certain if I succeeded or not. Could you review and let me know if there is something, here, that still needs to be addressed?

mike-jumper: Sure thing.

mike-jumper: OK - I haven't reviewed the meat of this, but the issue with the interface is still there. I think I can explain a bit better and in more depth.

The API issue is that we're adding a new function to a Java interface, in this case `User`. Consider a Java class defined within a Guacamole extension like:

```java
public class ExampleUser implements User {

    @Override
    public ActivityRecordSet<ActivityRecord> getUserHistory() {
        // STUB
    }

}
```

The above class will not compile because it does not implement the deprecated `getHistory()` function. This is problematic, because it seems unreasonable to require developers to implement deprecated functions that should no longer be used. Ideally, they should implement only the current functions and be OK.

Java 8 provides for this with default implementations. If the interface provides a default implementation for `getHistory()`, then implementations need not provide that function:

```java
public interface User {

    ...

    default List<? extends ActivityRecord> getHistory() throws GuacamoleException {
        return new ArrayList<>(getUserHistory().toCollection());
    }

    ...

}
```

So ... problem solved? Nope, because we need to maintain backward compatibility, as well. Unless additional steps are taken, an attempt to use an extension built for 1.2.0, 1.1.0, etc. may fail against 1.3.0 because a class implementing `User` does not provide `getUserHistory()` - the new function.

Ah! But then we can just provide defaults for both, right?

```java
public interface User {

    ...

    default List<? extends ActivityRecord> getHistory() throws GuacamoleException {
        return new ArrayList<>(getUserHistory().toCollection());
    }

    ActivityRecordSet<ActivityRecord> getUserHistory() throws GuacamoleException {
        return new SomeSortOfSimpleActivityRecordSet<>(getHistory());
    }

    ...

}
```

_Now_ problem solved? Almost. The above would work, and would solve all compatibility issues, but we lose the benefit of having a compiler yell at us if a needed function is unimplemented. Worse still: failing to implement at least one of those functions would result in a stack overflow when the webapp invokes `getUserHistory()`.

It is possible to solve this, but it involves:

* Providing a default only for the deprecated function within guacamole-ext. (See [the current `Connectable` interface within guacamole-ext](https://github.com/apache/guacamole-client/blob/0091bb1aea14c567c8166f0ed8eadf7c31b6bd6e/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Connectable.java))
* Within the webapp, copying `User` entirely and providing a default _for both_. (See [the current, internal `Connectable` interface within the webapp](https://github.com/apache/guacamole-client/blob/0091bb1aea14c567c8166f0ed8eadf7c31b6bd6e/guacamole/src/main/java/org/apache/guacamole/net/auth/Connectable.java))

That would allow old extensions to continue to work, with new extensions intending to build against the new API getting the expected compile-time errors, but makes things difficult to maintain until the deprecated things are removed.

Overall, the type of change that results in this is replacing one function with another. For `Connectable`, there was no choice at all. If a sensible set of changes can accomplish what we need here (without replacing one function of an interface with another), then the complexity can be avoided.

necouchman: Okay, so what if the `default` implementations of these methods throw a `GuacamoleUnsupportedException()` exception, which is what several of the implementations of `User` do, anyway? Then, any classes wishing to implement have to provide actual implementations of those methods, and can choose which they choose to implement?

That seems too simple, so there must be some case I'm not thinking of that would not be covered by this?

mike-jumper: At first, my impression was that there are a couple issues with that approach:

1. If an implementation of `User` or `Connection` does not support history tracking, it should return an empty `List` or empty `ActivityRecordSet` (depending on which history function we're talking about). Throwing an exception will cause any associated REST API call to fail with an error.
2. If the `default` version of the _new_ function (`getUserHistory()`) does not point back to the old function (`getHistory()`), then an extension for an older version of Guacamole that implemented the old function will mysteriously cease to function.

*BUT:*

Perhaps this can work well if the REST API itself handles `GuacamoleUnsupportedException`, relying on that to indicate whether the older function should be invoked?

necouchman: Okay, so I could either do that, or just write `default` implementations in `User` that return `List<ActivityRecord>` (`getHistory()`) and `ActivityRecordSet<ActivityRecord>` (`getUserHistory()`). Is there a situations where the default implementation returning empty sets would cause issues?

mike-jumper: If we have something like:

```java
public interface User {

    ...

    @Deprecated
    default List<? extends ActivityRecord> getHistory() throws GuacamoleException {
        return Collections.emptyList();
    }

    ActivityRecordSet<ActivityRecord> getUserHistory() throws GuacamoleException {
         return new SomeSortOfSimpleActivityRecordSet<>(getHistory());
    }

    ...

}
```

Then we end up with:

| Extension implements... | Result |
| -------------- | ------ |
| Only `getHistory()` (older extension) | Webapp invokes `getUserHistory()` which calls `getHistory()` by default and works as expected. |
| Only `getUserHistory()` (newer extension) | Webapp invokes `getUserHistory()` which works as expected. If an older extension decorates `User` objects and attempts to do something with `getHistory()`, it will not have any effect and may behave as if history is unimplemented. |
| Both (masochistic extension) | All calls to history-related functions work as expected. |
| Neither | Calls to either `getUserHistory()` or `getHistory()` have the same result: empty set of records |

That seems OK except for the case of inter-extension calling of the older `getHistory()` function.

mike-jumper: > Okay, so I could either do that, or just write `default` implementations in `User` that return `List<ActivityRecord>` (`getHistory()`) and `ActivityRecordSet<ActivityRecord>` (`getUserHistory()`). Is there a situations where the default implementation returning empty sets would cause issues?

For something like:

```java
public interface User {

    ...

    @Deprecated
    default List<? extends ActivityRecord> getHistory() throws GuacamoleException {
        return Collections.emptyList();
    }

    ActivityRecordSet<ActivityRecord> getUserHistory() throws GuacamoleException {
         return new SimpleActivityRecordSet<>();
    }

    ...

}
```

We end up with:

| Extension implements... | Result |
| -------------- | ------ |
| Only `getHistory()` (older extension) | Webapp invokes `getUserHistory()` which returns an empty set, despite `getHistory()` doing otherwise. |
| Only `getUserHistory()` (newer extension) | Webapp invokes `getUserHistory()` which works as expected. If an older extension decorates `User` objects and attempts to do something with `getHistory()`, it will not have any effect and may behave as if history is unimplemented. |
| Both (masochistic extension) | All calls to history-related functions work as expected. |
| Neither | Calls to either `getUserHistory()` or `getHistory()` have the same result: empty set of records |

That ends up not working well because older extensions would cease being compatible (they would function differently with 1.3.0 than the version of Guacamole they were written for).

guacamole/src/main/java/org/apache/guacamole/rest/connection/ConnectionResource.java
```diff
@@ -150,18 +149,16 @@ public ConnectionResource(@Assisted UserContext userContext,
...
-            apiRecords.add(new APIConnectionRecord(record));
-
-        // Return the converted history
-        return apiRecords;
+        try {
+            return new ConnectionHistoryResource(connection.getConnectionHistory());
+        }
+        catch (GuacamoleUnsupportedException e) {
+            return new ConnectionHistoryResource(new SimpleActivityRecordSet<>());
+        }
```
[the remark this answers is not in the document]
necouchman: Yeah, I think I'm finally starting to get my brain around handling all of the compatibility across implementations. I see what you're saying, and I'll take a shot at tweaking it to your suggestion, here, and punching up the documentation to cover it.

I _think_ that covers all the scenarios, but I'll also try to go over possibilities in my mind and make sure there aren't any corner cases that haven't been considered that would fail in such a scenario.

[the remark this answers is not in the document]
necouchman: Okay, I've worked through this - not entirely sure how well I've done. I hit some issues with the `List<? extends ConnectionRecord>` and converting that to `List<ConnectionRecord>`. I ended up with the loops you see, here, now - not sure if there's a more elegant way to do that translation from `? extends RecordType` to just `RecordType`? I tried various adjustments to the types of methods and such in the classes, and I couldn't really do much without breaking compatibility.

Also, the `HistoryResource` class in `rest/history` pulls history from `UserContext`, whereas the `ConnectionResource` and `UserResource` classes in `rest/connection` and `rest/user` pull history from `Connection` and `User`, and I'm not sure if there are any modifications needed to the `HistoryResource` class or anything in the `UserContext` interfaces or classes that need to be tweaked to handle these changes?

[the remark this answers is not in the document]
mike-jumper: > Okay, I've worked through this - not entirely sure how well I've done. I hit some issues with the `List<? extends ConnectionRecord>` and converting that to `List<ConnectionRecord>`. I ended up with the loops you see, here, now - not sure if there's a more elegant way to do that translation from `? extends RecordType` to just `RecordType`? ...

Can you point to the specific part of the code where you're encountering trouble? I don't think there should be any need to convert.

> Also, the `HistoryResource` class in `rest/history` pulls history from `UserContext`, whereas the `ConnectionResource` and `UserResource` classes in `rest/connection` and `rest/user` pull history from `Connection` and `User`, and I'm not sure if there are any modifications needed to the `HistoryResource` class or anything in the `UserContext` interfaces or classes that need to be tweaked to handle these changes?

Are you referring to whether the code that pulls directly from `UserContext` needs to add special handling for `GuacamoleUnsupportedException`?

guacamole/src/main/java/org/apache/guacamole/rest/connection/ConnectionResource.java
```diff
@@ -150,18 +153,26 @@ public ConnectionResource(@Assisted UserContext userContext,
...
+                List<ConnectionRecord> history = Collections.emptyList();
+                for (ConnectionRecord record : connection.getHistory()) {
+                    history.add(record);
+                }
+                return new ConnectionHistoryResource(new SimpleActivityRecordSet<>(history));
+            }
+            catch (GuacamoleUnsupportedException ex) {
+                return new ConnectionHistoryResource(new SimpleActivityRecordSet<>());
+            }
+        }
```
necouchman, line 175: This is the area that I'm talking about in terms of the conversion.  So, what I initially tried to do was:
```
try {
    return new ConnectionHistoryResource(connection.getConnectionHistory());
}
catch (GuacamoleUnsupportedException e) {
    try {
        return new ConnectionHistoryResource(new SimpleActivityRecordSet<ConnectionRecord>(connection.getHistory()));
    }
    catch (GuacamoleUnsupportedException ex) {
        return new ConnectionHistoryResource(new SimpleActivityRecordSet<>());
    }
}
```
However, when I do that, I get a compile error:

    incompatible types: List<CAP#1> cannot be converted to List<ConnectionRecord>

This appears to be because the return type of `getHistory()` is `List<? extends ConnectionRecord>` whereas the input type of `SimpleActivityRecordSet()` is `ActivityRecord`.  For some reason that `? extends` causes it a bunch of heartache.

guacamole/src/main/java/org/apache/guacamole/rest/connection/ConnectionResource.java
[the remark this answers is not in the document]
necouchman: > Can you point to the specific part of the code where you're encountering trouble? I don't think there should be any need to convert.

See comment below: https://github.com/apache/guacamole-client/pull/546#discussion_r513709095

> Are you referring to whether the code that pulls directly from UserContext needs to add special handling for GuacamoleUnsupportedException?

Correct, or whether `UserContext` itself needs to take that into account at all.

guacamole/src/main/java/org/apache/guacamole/rest/connection/ConnectionResource.java
mike-jumper: Ah, I see what you're saying. It's a rather confusing aspect of Java generics. Consider if this was possible:

```java
List<FancyThing> a = new ArrayList<>();
List<? extends Thing> b = a; // This is completely OK
List<Thing> c = b; // This shouldn't be allowed because ...
c.add(new Thing()); // ... this breaks the contract of List<FancyThing>
```

Suddenly, that `List` that's expected to contain only fancy things contains a non-fancy thing, breaking the contract of the type. Declaring something as `List<Thing>` is saying "a list of `Thing` or any subclass of `Thing`". Declaring something as `List<? extends Thing>` is saying "a list of **an unspecified subclass of `Thing`**".

For the approach here with `SimpleActivityRecordSet`:

* There is a convenient way to turn `List<? extends T>` into `List<T>`, [`Collections.unmodifiableList()`](https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#unmodifiableList-java.util.List-).
* If you're adding a convenience constructor to `SimpleActivityRecordSet()` that accepts a list of records, you should probably use `List<? extends RecordType>` for the parameter so that it can accept any `List` having usable elements. This is what `ArrayList` does, presumably for the same reason: https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html#ArrayList-java.util.Collection-

mike-jumper: ...unless the intent is that `SimpleActivityRecordSet` will be backed by the provided `List`, in which case `List<RecordType>` **is** correct. This is what is done with `SimpleDirectory`: https://guacamole.apache.org/doc/guacamole-ext/org/apache/guacamole/net/auth/simple/SimpleDirectory.html#SimpleDirectory-java.util.Collection-

I would suggest that the type be `Collection<RecordType>`, though, for API's sake.

guacamole/src/main/java/org/apache/guacamole/rest/connection/ConnectionResource.java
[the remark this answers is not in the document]
mike-jumper: I agree that it would make sense to be consistent in the handling of `GuacamoleUnsupportedException` and provide the same functionality there, as well. It could make things easier for developers that may well use the same code to drive both.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Connection.java
```diff
@@ -98,17 +99,55 @@
...
+    }
 
+    /**
+     * Returns an ActivityRecordSet containing ConnectionRecords that
+     * represent the usage history of this Connection, including any active
+     * users. ConnectionRecords in this list will be sorted in descending order
+     * of end time (active connections are first), and then in descending order
+     * of start time (newer connections are first). If connection history
+     * tracking has not been implemented this function should throw
+     * GuacamoleUnsupportExpcetion.
```
mike-jumper, line 135: `GuacamoleUnsupportException`*

mike-jumper, line 135: Point of clarification: `GuacamoleUnsupportedException` would be thrown if history tracking has not been implemented _**or**_ if it has only been implemented using the older, deprecated `getHistory()` function. Callers intending to support extensions using older versions of the API should consider additionally invoking `getHistory()` if `GuacamoleUnsupportedException` is thrown.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/DelegatingUser.java
```diff
@@ -93,11 +93,18 @@ public Date getLastActive() {
         return user.getLastActive();
     }
 
+    @Deprecated
     @Override
     public List<? extends ActivityRecord> getHistory()
```
mike-jumper, line 98: I've seen this applied a few times to implementations of the now-deprecated functions. Is the `@Deprecated` annotation not inherited from the interface that deprecated the overridden function?

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/User.java
```diff
@@ -93,17 +94,49 @@
      * of this user, including any active sessions. ActivityRecords
      * in this list will be sorted in descending order of end time (active
      * sessions are first), and then in descending order of start time
-     * (newer sessions are first).
+     * (newer sessions are first). If user login history is not implemented
+     * this method should throw GuacamoleUnsupportedException.
```
mike-jumper, line 98: Same here - I think we should specifically note that `GuacamoleUnsupportedException` may indicate that history is supported via `getHistory()`.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleActivityRecordSet.java
```diff
@@ -35,10 +36,30 @@
 public class SimpleActivityRecordSet<RecordType extends ActivityRecord>
         implements ActivityRecordSet<RecordType> {
 
+    private final List<RecordType> records;
```
mike-jumper, line 39: Please document.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleActivityRecordSet.java
```diff
@@ -35,10 +36,30 @@
...
+    }
+    
+    /**
+     * Create a new SimpleActivityRecordSet that contains the provided records.
+     * 
+     * @param records 
+     *     The records that this SimpleActivityRecordSet should contain.
+     */
+    public SimpleActivityRecordSet(List<RecordType> records) {
+        this.records = records;
```
mike-jumper, line 56: If the provided `List` will back the resulting `SimpleActivityRecordSet` (such that changes to the `List` affect the content of the set), that should be noted here.

guacamole/src/main/java/org/apache/guacamole/rest/connection/ConnectionResource.java
mike-jumper, line 175: A couple of points here:

1. The logic here is involved enough that I think it deserves some comments.
2. It's worth logging these exceptions with context at the debug level.
3. Because each `try` would `return` from the function entirely, you can avoid nesting the `try`/`catch` deeper and deeper with something like:

   ```java
   // Try to do it
   try {
      return doSomething();
   }
   catch (GuacamoleUnsupportedException e) {
       logger.debug("Useful context.", e);
   }

   // Well that didn't work
   try {
      return doSomethingElse();
   }
   catch (GuacamoleUnsupportedException e) {
       logger.debug("Useful context.", e);
   }

   // Darn
   return doLoseAllHope();
   ```

guacamole/src/main/java/org/apache/guacamole/rest/user/UserResource.java
```diff
@@ -110,18 +112,26 @@ public UserResource(@Assisted UserContext userContext,
...
+                List<ActivityRecord> history = Collections.emptyList();
+                for (ActivityRecord record : user.getHistory()) {
+                    history.add(record);
+                }
+                return new UserHistoryResource(new SimpleActivityRecordSet<>(history));
+            }
+            catch (GuacamoleUnsupportedException ex) {
+                return new UserHistoryResource(new SimpleActivityRecordSet<>());
+            }
+        }
```
mike-jumper, line 134: Same here regarding possible logic cleanup, etc. (see `ConnectionResource`)

mike-jumper (changes requested):
Reviewed! With the SQL changes effectively repeated across the three supported SQL dialects, please consider my comments for one particular dialect as applying to all.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionService.java
```diff
@@ -411,28 +411,16 @@ protected ConnectionRecord getObjectInstance(ConnectionRecordModel model) {
             ModeledConnection connection) throws GuacamoleException {
 
         String identifier = connection.getIdentifier();
-
-        // Retrieve history only if READ permission is granted
-        if (hasObjectPermission(user, identifier, ObjectPermission.Type.READ)) {
```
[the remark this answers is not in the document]
necouchman: Because this version of `retrieveHistory()` now simply brokers the call to the next `retrieveHistory()` (below it), which uses `user.isPrivileged()` and then `connectionRecordMapper.searchReadable()` to gather records that the user should have access to:

https://github.com/apache/guacamole-client/pull/546/files/96af10d29ac0715afad4e70148865d9fb3bd0223#diff-463b715cf1341f9d584036345fa3fed272acee431bffc12d3db3d11a792b9540R458-R479

The only other concern would be `tunnelService.getActiveConnections()`, which does its own checks for permissions.  Thus it seems that checking for readable permissions, here, is just verifying what we already know?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/UserService.java
```diff
@@ -581,13 +581,9 @@ protected ActivityRecord getObjectInstance(ActivityRecordModel model) {
             ModeledUser user) throws GuacamoleException {
 
         String username = user.getIdentifier();
-
-        // Retrieve history only if READ permission is granted
-        if (hasObjectPermission(authenticatedUser, username, ObjectPermission.Type.READ))
-            return getObjectInstances(userRecordMapper.select(username));
```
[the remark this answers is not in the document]
necouchman: Similar to above - call is delegated down to another `retrieveHistory()` call which checks `user.isPrivileged()` and then uses `userMapper.searchReadable()` for unprivileged users to gather readable user records.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/resources/org/apache/guacamole/auth/jdbc/connection/ConnectionRecordMapper.xml
```diff
@@ -108,28 +108,35 @@
...
-                    FROM guacamole_user
-                    WHERE POSITION(#{term.term,jdbcType=VARCHAR} IN username) > 0
-                )
+        <where>
+            
+            <if test="identifier != null">
+                guacamole_connection_history.connection_id = #{identifier,jdbcType=VARCHAR}
+            </if>
+            
+            <foreach collection="terms" item="term" separator=" AND ">
```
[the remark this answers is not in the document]
necouchman: Hmmmm...well, I thought it would, but looking at the documentation, now, I'm not so sure.  I'll dig in and do some testing, and maybe add an extra `AND` above the `<foreach>` just for kicks. The documentation does say that it knows to trim extra ones, so adding an `AND` would probably be the safest route if I'm unable to determine exactly how it is designed to behave.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/DelegatingUser.java
necouchman: Yeah - if I try to compile without the `@Deprecated` annotation on anything that implements a deprecated method, I get a compiler warning, which, in my case, translates to an error, because `error: warnings found and -Werror specified`.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionService.java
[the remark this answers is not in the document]
mike-jumper: Right - `getActiveConnections()` is required to only return connections visible to the current user.

mike-jumper:
It sounds like everything is pretty much there (or about to be there). Let me know when you want me to dive back in. I'm guessing it's no longer draft at this point?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/resources/org/apache/guacamole/auth/jdbc/connection/ConnectionRecordMapper.xml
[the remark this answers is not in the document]
necouchman: I notice you marked this as resolved, but I hadn't done a ton of verification.  Everything appears to work fine with it as it is (no extra `AND`s in there), but I haven't tested externsively...

necouchman:
> It sounds like everything is pretty much there (or about to be there). Let me know when you want me to dive back in. I'm guessing it's no longer draft at this point?

Yeah, I think it's pretty close, now.  Feel free to give it another look and see what else is missing. I'm still not 100% sure about the `ConnectionRecordMapper` and `UserRecordMapper` changes, specifically whether the `AND` is implicit, so that's probably worth a little more legwork on my part to make sure we don't introduce any bugs with that change.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/ActivityRecordSet.java
```diff
@@ -19,7 +19,9 @@
 
 package org.apache.guacamole.net.auth;
 
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.List;
```
mike-jumper, line 24: Same here - unless the build is currently failing, it's unlikely new `import` statements are needed without other code changes within the same file requiring them.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/User.java
```diff
@@ -93,17 +96,50 @@
...
+     *     and use getUserHistory(), instead.
+     * 
      * @return
      *     A list of ActivityRecords representing the login history of this
      *     User.
      *
      * @throws GuacamoleException
-     *     If an error occurs while reading the history of this user, or if
-     *     permission is denied.
+     *     If history tracking is not implement, if an error occurs while
```
mike-jumper, line 114: implemented*

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/User.java
```diff
@@ -93,17 +96,50 @@
...
+     * time (newer sessions are first). If login history tracking is not
+     * implemented, or is only implemented using the deprecated {@link getHistory}
+     * method, this method should throw GuacamoleUnsupportedException.
+     * 
+     * @return
+     *     An ActivityRecordSet containing ActivityRecords representing the
+     *     login history for this user.
+     * 
+     * @throws GuacamoleException
+     *     If history tracking is not implement, if an error occurs while
```
mike-jumper, line 136: implemented*

mike-jumper (changes requested):
OK - I've reviewed everything except the SQL and `<where>` behavior that's being verified.

necouchman:
Okay, I _think_ this is ready to go. I've tweaked the XML code for `ConnectionRecordMapper` and `UserRecordMapper` for all three of the JDBC modules, and I believe it works equally well whether there is a record (connection or user) defined, a search term, or both.

I'll try to go through and review things once more and see if anything looks amiss, but I think it's pretty solid, now.

mike-jumper (approved):

mike-jumper:
> I'll try to go through and review things once more and see if anything looks amiss, but I think it's pretty solid, now.

I've re-reviewed and all looks good to me. If you've not run into anything amiss, I'm happy to click the merge button.

necouchman:
> I've re-reviewed and all looks good to me. If you've not run into anything amiss, I'm happy to click the merge button.

I went through and corrected a few minor things, but overall looks okay to me.

PR #547, opened 2020-07-06
GUACAMOLE-1126: Set a maximum of 4 auto-reconnect attempts

echu2013:
This a base and non-user-configurable change to limit maximum automatic reconnection retries when "reconnectable" client/tunnel error occurs.
Further implementations to do are:
1- set default behaviour as original one
2- add user (per connection?) setting to handle this
3- Show relevant message, in countdown ??

### DIFF
--- a/guacamole/src/main/webapp/app/client/controllers/clientController.js
+++ b/guacamole/src/main/webapp/app/client/controllers/clientController.js
@@ -44,6 +44,16 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
     var requestService         = $injector.get('requestService');
     var tunnelService          = $injector.get('tunnelService');
     var userPageService        = $injector.get('userPageService');
+	
+	/**
+	 * Counter and limiter for maximum auto-reconnect attempts that we should 
+	 * automatically try to establish a connection when CLIENT_AUTO_RECONNECT or 
+	 * TUNNEL_AUTO_RECONNECT error types occur.
+	 *
+	 * @type Number
+	 */
+	var AUTO_RECONNECT_TRY     = 0;
+	var AUTO_RECONNECT_MAXTRY  = 4;
 
     /**
      * The minimum number of pixels a drag gesture must move to result in the
@@ -770,7 +780,13 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
             var errorName = (status in CLIENT_ERRORS) ? status.toString(16).toUpperCase() : "DEFAULT";
 
             // Determine whether the reconnect countdown applies
-            var countdown = (status in CLIENT_AUTO_RECONNECT) ? RECONNECT_COUNTDOWN : null;
+            if (status in CLIENT_AUTO_RECONNECT && AUTO_RECONNECT_TRY < AUTO_RECONNECT_MAXTRY ) {
+                var countdown = RECONNECT_COUNTDOWN;
+                AUTO_RECONNECT_TRY++;
+            } else {
+                var countdown = null;
+                AUTO_RECONNECT_TRY = 0;
+            }
 
             // Show error status
             notifyConnectionClosed({
@@ -792,7 +808,13 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
             var errorName = (status in TUNNEL_ERRORS) ? status.toString(16).toUpperCase() : "DEFAULT";
 
             // Determine whether the reconnect countdown applies
-            var countdown = (status in TUNNEL_AUTO_RECONNECT) ? RECONNECT_COUNTDOWN : null;
+            if (status in TUNNEL_AUTO_RECONNECT && AUTO_RECONNECT_TRY < AUTO_RECONNECT_MAXTRY ) {
+                var countdown = RECONNECT_COUNTDOWN;
+                AUTO_RECONNECT_TRY++;
+            } else {
+                var countdown = null;
+                AUTO_RECONNECT_TRY = 0;
+            }
 
             // Show error status
             notifyConnectionClosed({
@@ -828,6 +850,8 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
 
             // Hide status notification
             guacNotification.showStatus(false);
+            // Reset auto-reconnect attempts counter
+            AUTO_RECONNECT_TRY = 0;
 
         }
 


guacamole/src/main/webapp/app/client/controllers/clientController.js
```diff
@@ -44,6 +44,16 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
...
+	
+	/**
+	 * Counter and limiter for maximum auto-reconnect attempts that we should 
+	 * automatically try to establish a connection when CLIENT_AUTO_RECONNECT or 
+	 * TUNNEL_AUTO_RECONNECT error types occur.
+	 *
+	 * @type Number
+	 */
+	var AUTO_RECONNECT_TRY     = 0;
+	var AUTO_RECONNECT_MAXTRY  = 4;
```
necouchman, line 56: So, rather than setting a constant, here, I would say:
- Add an attribute to the Connection for tracking whether auto-reconnect is disabled or not (AUTO_RECONNECT_DISABLE, for example), which defaults to having auto-reconnect enabled.
- Add an attribute to the Connection for setting the maximum number of retries, which should default to unlimited.
- Might even be worth making the retry interval configurable with a similar attribute, such that it defaults to Guacamole's global default, but could be set per-connection to something other than 15 seconds.

guacamole/src/main/webapp/app/client/controllers/clientController.js
```diff
@@ -792,7 +808,13 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
...
 
             // Determine whether the reconnect countdown applies
-            var countdown = (status in TUNNEL_AUTO_RECONNECT) ? RECONNECT_COUNTDOWN : null;
+            if (status in TUNNEL_AUTO_RECONNECT && AUTO_RECONNECT_TRY < AUTO_RECONNECT_MAXTRY ) {
+                var countdown = RECONNECT_COUNTDOWN;
+                AUTO_RECONNECT_TRY++;
+            } else {
+                var countdown = null;
+                AUTO_RECONNECT_TRY = 0;
+            }
```
necouchman, line 817: Since the countdown is already part of the notification, I would think it would be pretty easy to add the number of tries and max tries pretty easily.

guacamole/src/main/webapp/app/client/controllers/clientController.js
echu2013: How can we get a profile's parameter value within clientController.js?

echu2013: Is there any chance we can get any parameter without modifying guacamole-server @necouchman ??

necouchman: Yes, we definitely won't need to modify guacamole-server, as the connection attributes all stay on the guacamole-client side and are available via the REST API.  However, there will be some changes required to the various Java classes in guacamole-client, along with the JDBC extensions, to support storage of the attributes.

necouchman: At least a couple of places that would need to be changed would be:

https://github.com/apache/guacamole-client/blob/master/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionModel.java

https://github.com/apache/guacamole-client/blob/master/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java

echu2013: > Yes, we definitely won't need to modify guacamole-server, as the connection attributes all stay on the guacamole-client side and are available via the REST API. However, there will be some changes required to the various Java classes in guacamole-client, along with the JDBC extensions, to support storage of the attributes.

I pretend not to modify any schema. Is it OK to use `guacamole_connection_attribute` to store these parameters? What do you think?

necouchman: I would think we would want to add a couple of fields to the DB schema, here, similar to the other standard connection attributes.  the `guacamole_connection_attribute` table is more intended for arbitrary attributes that someone might implement as part of an additional extension that needs to store data, while the main connection attributes tend to be stored as columns within the `guacamole_connection` table itself.

I suppose we could look at making the connection attributes behave more like connection parameters, where they are all stored in the `guacamole_connection_attributes` table, but that would definitely be a change to the way it is implemented today.

@mike-jumper Thoughts?

mike-jumper: My main concerns with schema modifications would be ... that they are schema modifications. We certainly did quite a few of those in the past, but doing that now would require a full version number bump and would break our nice release streak of backward compatibility.

Regarding modifications to support arbitrary attributes, that should already pretty much just work. It may be possible to achieve this without serious modifications to the webapp or database schema. If needed, it might make sense to add AngularJS events that extensions could hook into to manipulate this and other behaviors.

I'm not 100% behind the idea behind adding such a limit. Reading through GUACAMOLE-1126, it seems the primary concern motivating this is that each reconnect counts as user activity and therefore allows the session to continue indefinitely. If that is the case, perhaps there is a way that things could be arranged such that reconnects do not count toward session activity.

necouchman: I was thinking of this similar in light of the VNC `autoretry` parameter, but maybe the goals and implementation details are slightly different.  I guess that parameter does the retries transparently in the background (guacd side), while to the client it only looks like a single connection attempt.  There was mention when implementing WoL support (GUACAMOLE-513) of the possibility of supporting `autoretry` with some configurable timers/limits across the board.

echu2013: @mike-jumper @necouchman I am sorry but I don't understand why would be need to modify guacamole's schema to achieve this.. 
Why can't we just use `guacamole_connection_parameter` which already holds all kind of profile parameters under column named `parameter_name`???
What I've done so far is an addition of 3 parameters (autoreconnect-disable, autoreconnect-maxretries, autoreconnect-wait-seconds) which I need to get in `clientController.js` somehow...

mike-jumper: > Why can't we just use `guacamole_connection_parameter` which already holds all kind of profile parameters under column named `parameter_name`???

Because those aren't connection parameters. Connection parameters are a specific type of name/value pairs and are defined by the protocol plugins used by guacd. Connection parameter storage should not be used for anything that isn't a connection parameter.

There _is_ a built-in facility for arbitrary name/value pair storage for connections (connection attributes), and this would be the appropriate vehicle for these changes. New connection attributes do not generally require schema changes, at least in the case of non-standard attributes.

Thus: connection parameters would be a non-starter, but connection attributes would be (generally) fine.

The awkwardness in this particular change comes from:

* If the web application will have built-in handling for these attributes, they should be standard attributes, perhaps part of the API.
* Standard attributes (those which are standardized by the extension API) really should be stored within properly-typed columns. This is already done for the existing standard connection attributes.
* We desire to not modify the database schema, as doing so breaks backward compatibility ... but _not_ modifying the schema in this case would mean that we deviate from established convention that is grounded in good technical reasoning (attributes that are standard really should be able to be queried directly).

So ... yes, you're not wrong that it could be done as you describe, but it really should not be.

echu2013: Ok @mike-jumper I get your point.
So, you intend to extend this `guacamole_connection` table columns:
image
in order to include "reconnection" parameters?
Or, do you have another idea?

mike-jumper: > So, you intend to extend this `guacamole_connection` table columns:
> image
> in order to include "reconnection" parameters?

Not exactly. This would be the correct approach if standard attributes are going to be used here, but I strongly wish to avoid this.

> Or, do you have another idea?

Yes, [listed above in an earlier comment](https://github.com/apache/guacamole-client/pull/547#discussion_r451290773). Specifically:

1. Modifying the handling of automatic reconnect such that extensions can easily override it. An extension could then add this functionality, and any attributes used would not need to be standardized.
2. Avoiding the issue entirely by ensuring that automated reconnect attempts are not sufficient to prevent a session from expiring.

I would lean toward the latter (ensuring automated reconnect attempts do not prevent a session from expiring). It avoids schema changes, API changes, and any additional configuration, and instead simply leverages existing behavior to solve the issue motivating GUACAMOLE-1126.

ceharris:
Does @echu2013 plan to continue this work?

necouchman:
@ceharris: As we have not heard from @echu2013 in quite a while, I think it's safe to say this is orphaned work. Are you looking to take it up and move it forward?

echu2013:
Hello, where do you prefer to store this configurable setting @necouchman  ? At a connection profile level (this requires altering SQL schema?) or within **guacamole.properties** level?

necouchman:
Sorry @echu2013, I missed your question quite some time ago. I would say that the there should definitely be a per-connection mechanism for storing these, in the guacamole_connection_attributes table, as Mike mentioned previously.

At this same time, it might be nice to add a global setting in guacamole.properties that could be set as a default for the entire system.

PR #550, opened 2020-07-11
GUACAMOLE-1130: Only retrieve LDAP attributes that are strictly necessary

echu2013:
This aims to reduce the amount of attributes handled in code as described in [Jira](https://issues.apache.org/jira/browse/GUACAMOLE-1130)

### DIFF
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ObjectQueryService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ObjectQueryService.java
@@ -197,6 +197,10 @@ public class ObjectQueryService {
      * @param searchHop
      *     The current level of referral depth for this search, used for
      *     limiting the maximum depth to which referrals can go.
+     * 
+     * @param relevantAttributes
+     *     The attribute(s) relevant to return for this search,
+	 *     if all available should be returned pass null as value.
      *
      * @return
      *     A list of all results accessible to the user currently bound under
@@ -208,7 +212,8 @@ public class ObjectQueryService {
      *     guacamole.properties.
      */
     public List<Entry> search(LdapNetworkConnection ldapConnection,
-            Dn baseDN, ExprNode query, int searchHop) throws GuacamoleException {
+            Dn baseDN, ExprNode query, int searchHop,
+            Collection<String> relevantAttributes) throws GuacamoleException {
 
         // Refuse to follow referrals if limit has been reached
         int maxHops = confService.getMaxReferralHops();
@@ -225,12 +230,15 @@ public class ObjectQueryService {
 
         // Search within subtree of given base DN
         SearchRequest request = ldapService.getSearchRequest(baseDN, query);
-            
+        if (relevantAttributes != null) {
+            request.addAttributes(relevantAttributes.toArray(new String[0]));
+        }
+
         // Produce list of all entries in the search result, automatically
         // following referrals if configured to do so
         List<Entry> entries = new ArrayList<>();
-            
         try (SearchCursor results = ldapConnection.search(request)) {
+
             while (results.next()) {
 
                 // Add entry directly if no referral is involved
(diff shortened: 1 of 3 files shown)


extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ObjectQueryService.java
```diff
@@ -197,6 +197,9 @@ public ExprNode generateQuery(ExprNode filter,
      * @param searchHop
      *     The current level of referral depth for this search, used for
      *     limiting the maximum depth to which referrals can go.
+     * 
+     * @param relevantAttributes
+     *     The attribute(s) relevant to return in this search.
```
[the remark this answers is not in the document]
echu2013: You are right, I will add the clarification as you pointed because is relevant!

echu2013:
> I currently only see two issues with this, both related to documentation:
> 
> 1. The intentional use of `null` to represent all attributes should be documented. The existing documentation only describes the meaning of a non-null value.
> 2. The commit message ("GUACAMOLE-1130: Add LDAP attributes filtering handler") is unclear to me. Only through reading the code do I see the intent of these changes (to retrieve only the LDAP attributes that are necessary). I think this should be rephrased such that someone reading through the commit logs will understand the high-level intent of these changes. This is particularly important if the code should prove incorrect sometime in the future.
> 
> There were recent changes related to LDAP attribute handling which allow the LDAP attributes associated with a user to be made available as parameter tokens when specified with [the `ldap-user-attributes` property](http://guacamole.apache.org/doc/gug/ldap-auth.html):
> 
> https://github.com/apache/guacamole-client/blob/3c4c81f0b6b9700abccaefcc695058e515b8b20b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java#L234-L257
> 
> I don't think these changes will affect that behavior, but have you already confirmed this?

`The commit message ("GUACAMOLE-1130: Add LDAP attributes filtering handler") is unclear to me. Only through reading the code do I see the intent of these changes (to retrieve only the LDAP attributes that are necessary). I think this should be rephrased such that someone reading through the commit logs will understand the high-level intent of these changes. This is particularly important if the code should prove incorrect sometime in the future.`
I will definitely rephrase commit log!
`I don't think these changes will affect that behavior, but have you already confirmed this?`
This is indeed confirmed, I use LDAP Tokens and they keep working OK as usual.

echu2013:
@mike-jumper is it better now??

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
```diff
@@ -208,13 +208,21 @@ private ExprNode getGroupSearchFilter() throws GuacamoleException {
             }
         }
 
+        // Gather all attributes relevant for a group
+        ArrayList<String> groupAttributes = new ArrayList<String>();
```
mike-jumper, line 212: 1. The size of the required backing array can be anticipated here. This will work as-is, but unnecessarily relies on automatic array resizing or sufficiently-large defaults.
2. Unless code consuming this relies on the object being an `ArrayList`, the type declared for the variable should be kept minimal, presumably `List<String>` or `Collection<String>`.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
```diff
@@ -208,13 +208,21 @@ private ExprNode getGroupSearchFilter() throws GuacamoleException {
...
+        groupAttributes.add(confService.getMemberAttribute());
+        confService.getGroupNameAttributes().forEach(
+                       attribute -> groupAttributes.add(attribute)
+                       );
```
mike-jumper, line 216: Why not `addAll()`?

And what if one of the attribute names within `confService.getGroupNameAttributes()` is equal to `confService.getMemberAttribute()`? Will the duplicate attribute cause any problems for the search?

necouchman:
@echu2013 Any progress on @mike-jumper's most recent change requests?

PR #552, opened 2020-07-16
GUACAMOLE-1135: Return StoreUrls in URI format

rsg98:
The MySQL driver expects locations for TrustStore and ClientStore to be in URI format (file:/path/to/store) - this pushes the URI format into the driverProperties, instead of getAbsolutePath(), which returns without the file:/ schema.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/MySQLAuthenticationProviderModule.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/MySQLAuthenticationProviderModule.java
@@ -95,7 +95,7 @@ public class MySQLAuthenticationProviderModule implements Module {
         File trustStore = environment.getMySQLSSLTrustStore();
         if (trustStore != null)
             driverProperties.setProperty("trustCertificateKeyStoreUrl",
-                    trustStore.getAbsolutePath());
+                    trustStore.toURI().toString());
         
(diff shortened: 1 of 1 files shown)


extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/MySQLAuthenticationProviderModule.java
```diff
@@ -95,7 +95,7 @@ public MySQLAuthenticationProviderModule(MySQLEnvironment environment)
         File trustStore = environment.getMySQLSSLTrustStore();
         if (trustStore != null)
             driverProperties.setProperty("trustCertificateKeyStoreUrl",
-                    trustStore.getAbsolutePath());
+                    String.valueOf(trustStore.toURI()));
```
necouchman, line 98: Instead of `String.valueOf()` it seems like it would be better to use the `URI` class's `toString()` method, here.  It's possible that's already what `String.valueOf()` does under the hood, but seems like a more direct reference to it would be better.

necouchman:
@rsg98 : Okay, changes look fine to me; however, your second commit message lacks the "GUACAMOLE-1135:" prefix on it.  This is a small enough change that I'd suggest just squashing the second commit into the first and doing a force-push so there's only a single commit.

rsg98:
> @rsg98 : Okay, changes look fine to me; however, your second commit message lacks the "GUACAMOLE-1135:" prefix on it. This is a small enough change that I'd suggest just squashing the second commit into the first and doing a force-push so there's only a single commit.

Thanks - should be fixed now

PR #485, opened 2020-03-22
GUACAMOLE-980: used tomcat-jre8 Docker-Image seems to be deprecated

asutosh936:

PR #553, opened 2020-07-20
GUACAMOLE-980: used tomcat-jre8 Docker-Image seems to be deprecated

asutosh936:


asutosh936:
@mike-jumper @benrubson @necouchman - Reference to PR: https://github.com/apache/guacamole-client/pull/485

PR #489, opened 2020-04-02
GUACAMOLE-1005: Docker, configure RemoteIPValve

gabibbo97:
See https://issues.apache.org/jira/browse/GUACAMOLE-1005

Docs PR: https://github.com/apache/guacamole-manual/pull/130

PR #554, opened 2020-07-20
GUACAMOLE-890: Security: Allow image to run as non-root user

benrubson:
Hi,

This is a rework of #469, and solves https://issues.apache.org/jira/browse/GUACAMOLE-890, allowing Guacamole Docker image to run as non-root user.

Below is why, for now, Guacamole image can't be run as non-root user, and how we solve this.

So first, now, thanks to #553, Docker image is now based on `tomcat:8.5-jdk8`.

Here's below the directory structure from the last `tomcat:8.5-jdk8`  image :

```
# ls -l /usr/local/tomcat/
total 152
-rw-r--r-- 1 root root 19318 Jun 30 21:53 BUILDING.txt
-rw-r--r-- 1 root root  5408 Jun 30 21:53 CONTRIBUTING.md
-rw-r--r-- 1 root root 57011 Jun 30 21:53 LICENSE
-rw-r--r-- 1 root root  1726 Jun 30 21:53 NOTICE
-rw-r--r-- 1 root root  3255 Jun 30 21:53 README.md
-rw-r--r-- 1 root root  7136 Jun 30 21:53 RELEASE-NOTES
-rw-r--r-- 1 root root 16262 Jun 30 21:53 RUNNING.txt
drwxr-xr-x 2 root root  4096 Jul 17 00:48 bin
drwxr-xr-x 1 root root  4096 Jul 20 19:45 conf
drwxr-xr-x 2 root root  4096 Jul 17 00:48 lib
drwxrwxrwx 1 root root  4096 Jul 20 19:45 logs
drwxr-xr-x 2 root root  4096 Jul 17 00:48 native-jni-lib
drwxrwxrwx 2 root root  4096 Jul 17 00:48 temp
drwxr-xr-x 2 root root  4096 Jul 17 00:48 webapps
drwxr-xr-x 7 root root  4096 Jun 30 21:50 webapps.dist
drwxrwxrwx 2 root root  4096 Jun 30 21:49 work
```

As we can see, `logs`, `temp` and `work` directories of the default `CATALINA_BASE` have been made world-writable.
But still, the `webapps` directory is not world-writable, so a non-root user won't be able to link `guacamole.war` into it.
So the Guacamole `start.sh` script will fail here below, and this is why it can't run as non-root user :
https://github.com/apache/guacamole-client/blob/21ec5640a560f95725c081391d9943d5d0beb88d/guacamole-docker/bin/start.sh#L634-L636

We could then ask the Tomcat image maintainers to `chmod 777 webapps`.
But the goal of running as non-root user should not be achieved removing restrictions.
In other words, from a security point of view, the method used by the image maintainers does not seem to be the good one.

To confirm this, here's now the default directory structure from a Debian 10 Tomcat installation :
```
# ls -l /var/lib/tomcat9/
lrwxrwxrwx 1 root   root     12 Jun 13  2019 conf -> /etc/tomcat9
drwxr-xr-x 2 tomcat tomcat 4096 Jun 13  2019 lib
lrwxrwxrwx 1 root   root     17 Jun 13  2019 logs -> ../../log/tomcat9
drwxr-xr-x 2 root   root   4096 Mar 27 22:51 policy
drwxrwxr-x 3 tomcat tomcat 4096 Mar 27 22:51 webapps
lrwxrwxrwx 1 root   root     19 Jun 13  2019 work -> ../../cache/tomcat9
# ls -ld /var/log/tomcat9/ /var/cache/tomcat9/
drwxr-x--- 3 tomcat tomcat 4096 Mar 27 22:51 /var/cache/tomcat9/
drwxr-s--- 2 tomcat adm    4096 Mar 27 22:51 /var/log/tomcat9/
```

As we can see here, `logs`, `work` and `webapps` directories are only writable by the `tomcat` user.
It then does not make sense to make them world-writable so that another user can start its application...
Would be a rather messy configuration...

On the contrary, as explained in the [Tomcat documentation](http://tomcat.apache.org/tomcat-8.5-doc/introduction.html#CATALINA_HOME_and_CATALINA_BASE), we should, in this case, create and use a dedicated proper `CATALINA_BASE` directory structure.
There's indeed no reason why we should absolutely run the Guacamole application from the default runtime directories owned by `root`...

This is then what this PR does.
It creates the 5 runtime directories which are then readable / writable only by the user running the application. Links the application into it, and starts from it...
Thus enforcing directory structure and whole application security.

Thank you 👍

### DIFF
--- a/guacamole-docker/bin/start.sh
+++ b/guacamole-docker/bin/start.sh
@@ -659,9 +659,15 @@ associate_header() {
 ##
 start_guacamole() {
 
+    # User-only writable CATALINA_BASE
+    export CATALINA_BASE=$HOME/tomcat
+    for dir in logs temp webapps work; do
+        mkdir -p $CATALINA_BASE/$dir
+    done
+    cp -R /usr/local/tomcat/conf $CATALINA_BASE
+
     # Install webapp
-    rm -Rf /usr/local/tomcat/webapps/${WEBAPP_CONTEXT:-guacamole}
-    ln -sf /opt/guacamole/guacamole.war /usr/local/tomcat/webapps/${WEBAPP_CONTEXT:-guacamole}.war
+    ln -sf /opt/guacamole/guacamole.war $CATALINA_BASE/webapps/${WEBAPP_CONTEXT:-guacamole}.war
 
     # Start tomcat
     cd /usr/local/tomcat
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,6 +56,15 @@ WORKDIR /opt/guacamole
 # Copy artifacts from builder image into this image
 COPY --from=builder /opt/guacamole/ .
 
+# Create a new user guacamole
+ARG UID=1001
+ARG GID=1001
+RUN groupadd --gid $GID guacamole
+RUN useradd --system --create-home --shell /usr/sbin/nologin --uid $UID --gid $GID guacamole
+
+# Run with user guacamole
+USER guacamole
+
 # Start Guacamole under Tomcat, listening on 0.0.0.0:8080
 EXPOSE 8080
 CMD ["/opt/guacamole/bin/start.sh" ]


guacamole-docker/bin/start.sh
```diff
@@ -659,9 +659,15 @@ associate_header() {
 ##
 start_guacamole() {
 
+    # User-only writable CATALINA_BASE
+    export CATALINA_BASE=$GUACAMOLE_HOME/tomcat
```
mike-jumper, line 663: I don't think `CATALINA_BASE` should be a subdirectory of `GUACAMOLE_HOME`. The contents and purpose of `GUACAMOLE_HOME` are dictated by the webapp. It contains specific things in a specific structure and shouldn't be used as a general-purpose storage location.

guacamole-docker/bin/start.sh
```diff
@@ -659,9 +659,15 @@ associate_header() {
...
+    for dir in logs temp webapps work; do
+        mkdir -p $CATALINA_BASE/$dir
+    done
+    cp -R /usr/local/tomcat/conf $CATALINA_BASE
```
mike-jumper, line 667: Will a symbolic link not work in this case?

mike-jumper (changes requested):

guacamole-docker/bin/start.sh
benrubson: Thank you very much for your feedback @mike-jumper 👍
You're right, I modified accordingly.

benrubson:
I also added a dedicated new system user at image build time, as done in https://github.com/apache/guacamole-server/pull/292 (and https://github.com/apache/guacamole-server/pull/314, to merge).
So that both containers run the same way, as non-root user by default 👍

guacamole-docker/bin/start.sh
benrubson: Well, it would work.
But at least one other PR plans to modify the configuration : #489.
Let's then keep it writable 👍

benrubson:
@mike-jumper do you perhaps plan this for `1.3.0` ?
So that we would have both server (thanks to PR referenced [above](https://github.com/apache/guacamole-client/pull/554#issuecomment-736063177)) and client containers running as non-root user by default, starting `1.3.0`.
Thx 👍

mike-jumper (changes requested):
I think this looks OK. Please combine the duplicate commits (or use commit messages that apply to those commits specifically, rather than duplicating the same message) and I'll be happy to merge:

image

benrubson:
Thank you very much for your feedback @mike-jumper 👍
I then just combined the various commits.
Thank you again !

mike-jumper (approved):

PR #562, opened 2020-08-26
GUACAMOLE-641: Expand extension API to allow properties to be retrieved from key vaults.

mike-jumper:
The initial in-progress support for key vaults (#336) includes the ability to pull sensitive data for use within connection parameter tokens only. If sensitive data needs to be pulled for configuration information otherwise stored within `guacamole.properties`, that vault support will not be able to pull that data.

This change augments the existing `Environment` interface such that extensions can serve as arbitrary sources of configuration properties. For any properties not already defined within `guacamole.properties`, additional sources are polled in order until a value is retrieved.

Part of this involved:

* Refactoring `LocalEnvironment` into a singleton implementation of `Environment` (with the old constructor remaining for compatibility but being deprecated).
* Refactoring the old reading of properties from `guacamole.properties` and (optionally) environment variables into their own separate property sources which are added by the web application by default.
* Modifying the way the JDBC auth establishes database connections such that the credentials are not assumed to be constant, but may be provisioned just-in-time by an arbitrary vault service.

These changes do _not_ include the modifications to #336 to pull these properties, however I will look into doing that next.


necouchman (changes requested):
A couple of conflicts, here, that need to be resolved.

There was one thing that wasn't obvious to me that I also wondered about - for the `SystemEnvironmentGuacamoleProperties` class, are these read at startup or in realtime? My guess is that these are still only read at startup - which is fine - just making sure I understand the implications of the changes, there.

mike-jumper:
> A couple of conflicts, here, that need to be resolved.

Ah - I'll rebase.

> There was one thing that wasn't obvious to me that I also wondered about - for the `SystemEnvironmentGuacamoleProperties` class, are these read at startup or in realtime? My guess is that these are still only read at startup - which is fine - just making sure I understand the implications of the changes, there.

For `SystemEnvironmentGuacamoleProperties`, the `getProperty()` implementation will read the current value from `System.getenv()` each time it's invoked without caching the result. That said, I don't believe it's possible to externally change the environment of a running process without dark magic.

necouchman:
> For SystemEnvironmentGuacamoleProperties, the getProperty() implementation will read the current value from System.getenv() each time it's invoked without caching the result. That said, I don't believe it's possible to externally change the environment of a running process without dark magic.

Okay, sounds good. If `getProperty()` calls `getEnv()` each time it runs, are there any performance benefits to caching it, since externally changing the environment requires calling Merlin to conjure up a spell? Or is any performance benefit there outweighed by the amount of code it would take to cache it?

mike-jumper:
> Okay, sounds good. If `getProperty()` calls `getEnv()` each time it runs, are there any performance benefits to caching it, since externally changing the environment requires calling Merlin to conjure up a spell? Or is any performance benefit there outweighed by the amount of code it would take to cache it?

Well, I had not benchmarked things, but my expectation was that the environment is already implemented as a hash table by its own nature, and that the overhead of providing our own cache on top of that would be a net performance loss.

I just now tried a mini benchmark and caching is indeed faster. Reading 10,000,000 environment variables via uncached `System.getenv()` takes ~250ms, whereas the same operations take a total of 80-100ms depending on the ratio of defined to  undefined variables.

IMHO, I think having simpler code here outweighs the performance gain, particularly given that these sorts of configuration reads are infrequent.

mike-jumper:
I've rebased to resolve the conflicts. Also had to update the guacamole-auth-quickconnect and guacamole-auth-json extensions to use `LocalEnvironment.getInstance()` rather than `new LocalEnvironment()`.

necouchman (approved):

PR #563, opened 2020-08-27
GUACAMOLE-793: validateTicket() returns the CASAuthenticatedUser instance

doctorfree:
This is a revision of the pull request by @siacali at https://github.com/apache/guacamole-client/pull/454 which provides a functioning solution to this PR. The revised approach is to modify validateTicket() to return the CASAuthenticatedUser instance rather than just a token so CAS Provider can return Group - like LDAP Provider. This alleviates the need for a new class which was the objection to the previous pull request.

This is my first pull request to this project so please feel free to help me comply with your standards. I've tried to incorporate changes that address all of the comments in the previous pull request but may still need to address other issues I may have overlooked. Thanks for any feedback.

### DIFF
--- a/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/CASGuacamoleProperties.java
+++ b/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/CASGuacamoleProperties.java
@@ -20,6 +20,7 @@
 package org.apache.guacamole.auth.cas.conf;
 
 import org.apache.guacamole.properties.URIGuacamoleProperty;
+import org.apache.guacamole.properties.StringGuacamoleProperty;
 
 /**
  * Provides properties required for use of the CAS authentication provider.
@@ -68,5 +69,28 @@ public class CASGuacamoleProperties {
         public String getName() { return "cas-clearpass-key"; }
 
     };
+  
+    /**
+     * The attribute used for group membership
+     * example:  memberOf  (case sensitive)
+     */
+    public static final StringGuacamoleProperty CAS_GROUP_ATTRIBUTE =
+            new StringGuacamoleProperty() {
 
+        @Override
+        public String getName() { return "cas-group-attribute"; }
+
+    };
+
+    /**
+     * The name of the attribute used for group membership, such as "memberOf".
+     * This attribute is case sensitive.
+     */
+    public static final StringGuacamoleProperty CAS_GROUP_DN_FORMAT =
+            new StringGuacamoleProperty() {
+
+        @Override
+        public String getName() { return "cas-group-dn-format"; }
+
+    };
 }
--- a/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/ConfigurationService.java
+++ b/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/ConfigurationService.java
@@ -85,4 +85,40 @@ public class ConfigurationService {
         return environment.getProperty(CASGuacamoleProperties.CAS_CLEARPASS_KEY);
     }
 
+    /**
+     * Returns the attribute used to determine group memberships
+     * in CAS, or null if not defined.
+     *
+     * @return
+     *     The attribute name used to determine group memberships in CAS,
+     *     null if not defined.
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed.
+     */
+    public String getGroupAttribute() throws GuacamoleException {
+        return environment.getProperty(CASGuacamoleProperties.CAS_GROUP_ATTRIBUTE);
+    }
+
+    /**
+     * Returns the full DN specification used to format group DN's in CAS,
+     * or null if not defined.
+     * If CAS is backed by LDAP, it will return an LDAP DN, such as
+     * CN=foo,OU=bar,DC=example,DC=com.  This DN specification may be set to
+     * CN=%s,OU=bar,DC=example,DC=com and given the example above, would result
+     * in a group called "foo".  CAS backed by something other than LDAP would
+     * likely not need this.
+     *
+     * @return
+     *     The DN format specification.
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed.
+     */
+    public String getGroupDnFormat() throws GuacamoleException {
+        return environment.getProperty(CASGuacamoleProperties.CAS_GROUP_DN_FORMAT);
+    }
+
+
+
 }
--- a/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
+++ b/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
@@ -21,6 +21,7 @@ package org.apache.guacamole.auth.cas.ticket;
 
 import com.google.common.io.BaseEncoding;
 import com.google.inject.Inject;
+import com.google.inject.Provider;
 import java.net.URI;
 import java.security.InvalidKeyException;
 import java.security.NoSuchAlgorithmException;
@@ -33,10 +34,15 @@ import java.nio.charset.Charset;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
+import java.util.HashSet;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.GuacamoleSecurityException;
 import org.apache.guacamole.GuacamoleServerException;
 import org.apache.guacamole.auth.cas.conf.ConfigurationService;
+import org.apache.guacamole.auth.cas.user.CASAuthenticatedUser;
 import org.apache.guacamole.net.auth.Credentials;
 import org.apache.guacamole.token.TokenName;
 import org.jasig.cas.client.authentication.AttributePrincipal;
@@ -69,6 +75,12 @@ public class TicketValidationService {
     private ConfigurationService confService;
 
     /**
+     * Provider for AuthenticatedUser objects.
+     */
+    @Inject
+    private Provider<CASAuthenticatedUser> authenticatedUserProvider;
+
+    /**
      * Validates and parses the given ID ticket, returning a map of all
      * available tokens for the given user based on attributes provided by the
      * CAS server.  If the ticket is invalid an exception is thrown.
@@ -81,14 +93,13 @@ public class TicketValidationService {
      *     password values in.
      *
      * @return
-     *     A Map all of tokens for the user parsed from attributes returned
-     *     by the CAS server.
+     *     The CASAuthenticatedUser instance returned by the CAS server.
      *
      * @throws GuacamoleException
      *     If the ID ticket is not valid or guacamole.properties could
      *     not be parsed.
      */
-    public Map<String, String> validateTicket(String ticket,
+    public CASAuthenticatedUser validateTicket(String ticket,
             Credentials credentials) throws GuacamoleException {
 
         // Retrieve the configured CAS URL, establish a ticket validator,
@@ -100,6 +111,7 @@ public class TicketValidationService {
         validator.setEncoding("UTF-8");
         try {
             Map<String, String> tokens = new HashMap<>();
+            Set<String> effectiveGroups = new HashSet<>();
             URI confRedirectURI = confService.getRedirectURI();
             Assertion a = validator.validate(ticket, confRedirectURI.toString());
             AttributePrincipal principal =  a.getPrincipal();
@@ -122,16 +134,42 @@ public class TicketValidationService {
             }
             
             // Convert remaining attributes that have values to Strings
+            String groupAttribute = confService.getGroupAttribute();
+            // Use cas-member-attribute to retrieve and set group memberships
+            String groupDnFormat = confService.getGroupDnFormat();
+            String groupTemplate = "";
+            if (groupDnFormat != null) {
+                // if CAS is backended to LDAP, groups come in as RFC4514 DN
+                // syntax.  If cas-group-dn-format is set, this strips
+                // an entry such as "CN=Foo,OU=Bar,DC=example,DC=com" to
+                // "Foo"
+                groupTemplate = groupDnFormat.replace("%s","([A-Za-z0-9_\\(\\)\\-\\.\\s+]+)");
+                // the underlying parser aggregates all instances of the same
+                // attribute, so we need to be able to parse them out
+                groupTemplate=groupTemplate+",*\\s*";
+            } else {
+               groupTemplate = "([A-Za-z0-9_\\(\\)\\-\\.\\s+]+,*\\s*)";
+            }
+            Pattern pattern = Pattern.compile(groupTemplate);
+
             for (Entry <String, Object> attr : ticketAttrs.entrySet()) {
                 String tokenName = TokenName.canonicalize(attr.getKey(),
                         CAS_ATTRIBUTE_TOKEN_PREFIX);
                 Object value = attr.getValue();
-                if (value != null)
+                if (value != null) {
                     tokens.put(tokenName, value.toString());
+                    if (attr.getKey().equals(groupAttribute)) {
+                        String valueWithoutBrackets = value.toString().substring(1,value.toString().length()-1);
+                        Matcher matcher = pattern.matcher(valueWithoutBrackets);
+                        while (matcher.find()) {
+                            effectiveGroups.add(matcher.group(1));
+                        }
+                    }
+                }
             }
 
-            return tokens;
-
+            CASAuthenticatedUser authenticatedUser = authenticatedUserProvider.get();
+            return authenticatedUser;
         } 
         catch (TicketValidationException e) {
             throw new GuacamoleException("Ticket validation failed.", e);
--- a/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
+++ b/extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
@@ -157,10 +157,11 @@ public class TicketValidationService {
                         CAS_ATTRIBUTE_TOKEN_PREFIX);
                 Object value = attr.getValue();
                 if (value != null) {
-                    tokens.put(tokenName, value.toString());
+                    String attrValue = value.toString();
+                    tokens.put(tokenName, attrValue);
                     if (attr.getKey().equals(groupAttribute)) {
-                        String valueWithoutBrackets = value.toString().substring(1,value.toString().length()-1);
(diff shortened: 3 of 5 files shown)


extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
```diff
@@ -122,16 +134,42 @@
...
+                if (value != null) {
                     tokens.put(tokenName, value.toString());
+                    if (attr.getKey().equals(groupAttribute)) {
+                        String valueWithoutBrackets = value.toString().substring(1,value.toString().length()-1);
+                        Matcher matcher = pattern.matcher(valueWithoutBrackets);
+                        while (matcher.find()) {
+                            effectiveGroups.add(matcher.group(1));
+                        }
+                    }
+                }
```
necouchman, line 168: I feel like this could be cleaned up a little more.  The `value.toString()` method gets called a couple of times, so might be worth assigning that  to a `String` at the top of this `if()` statement. On the other hand, `valueWithoutBrackets` is only ever used once, within the `Matcher`, so seems like that could be removed and just put the `String.substring()`, etc., into the `pattern.matcher()` call directly.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
```diff
@@ -122,16 +134,42 @@
...
             }
 
-            return tokens;
-
+            CASAuthenticatedUser authenticatedUser = authenticatedUserProvider.get();
+            return authenticatedUser;
```
necouchman, line 172: If this is where the `CASAuthenticatedUser` object is being generated, I think it probably makes sense to call the `init()` method, here, prior to returning it, with all of the relevant data we've just generated through the ticket validation process?

necouchman (changes requested):
Overall I like the direction of these changes. There are a few things that need to be cleaned up, though - I'm actually not certain this will work as intended right now.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/CASGuacamoleProperties.java
```diff
@@ -68,5 +69,30 @@ private CASGuacamoleProperties() {}
         public String getName() { return "cas-clearpass-key"; }
 
     };
+  
+    /**
+     * The attribute used for group membership
+     * example:  memberOf  (case sensitive)
+     */
```
mike-jumper, line 76: This reads unclearly to me. At first glance, it looks like "(case sensitive)" is part of the example attribute value. JavaDoc convention also dictates that periods must be used.

Perhaps something more like:

```java
/**
 * The name of the attribute used for group membership, such as "memberOf".
 * This attribute is case sensitive.
 */

File: extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/CASGuacamoleProperties.java [hunk 4]
```diff
@@ -68,5 +69,30 @@ private CASGuacamoleProperties() {}
...
+        public String getName() { return "cas-group-attribute"; }
+
+    };
 
+    /**
+     * The attribute used for group DN Format
+     * such as CN=%s,OU=myou,DC=example,DC=com  (case sensitive)
+     * used to strip all but %s.  This is only necessary when
+     * CAS backend is LDAP. 
+     */
```
mike-jumper, line 90: I'm not sure this is correct. How is the format an attribute?

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/ConfigurationService.java
```diff
@@ -85,4 +85,40 @@ public PrivateKey getClearpassKey() throws GuacamoleException {
...
+     * Returns the attribute used to format group DN's
+     * in CAS, or null if not defined.  If CAS is backed by LDAP, it will
+     * return an LDAP DN, such as CN=foo,OU=bar,DC=example,DC=com.
+     * This attribute may be set to CN=%s,OU=bar,DC=example,DC=com and given
+     * the example above, would result in a group called "foo."  CAS backed
+     * by something other than LDAP (such as a database) would likely not 
+     * need this.
+     *
+     * @return
+     *     The attribute name
```
mike-jumper, line 113: JavaDoc convention dictates that there needs to be a period at the end of this.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/ConfigurationService.java
```diff
@@ -85,4 +85,40 @@ public PrivateKey getClearpassKey() throws GuacamoleException {
...
+    }
+
+    /**
+     * Returns the attribute used to format group DN's
+     * in CAS, or null if not defined.  If CAS is backed by LDAP, it will
+     * return an LDAP DN, such as CN=foo,OU=bar,DC=example,DC=com.
+     * This attribute may be set to CN=%s,OU=bar,DC=example,DC=com and given
+     * the example above, would result in a group called "foo."  CAS backed
+     * by something other than LDAP (such as a database) would likely not 
+     * need this.
```
mike-jumper, line 110: As above, I'm not sure this is correct. How is the format itself an attribute?

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
```diff
@@ -81,14 +93,13 @@
      *     password values in.
      *
      * @return
-     *     A Map all of tokens for the user parsed from attributes returned
-     *     by the CAS server.
+     *     The CASAuthenticatedUser instance returned by the CAS server.
```
mike-jumper, line 96: I think this needs to be clarified. The CAS server is not returning an instance of the `CASAuthenticatedUser` class to us.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
```diff
@@ -122,16 +134,42 @@
...
+                // an entry such as "CN=Foo,OU=Bar,DC=example,DC=com" to
+                // "Foo"
+                groupTemplate = groupDnFormat.replace("%s","([A-Za-z0-9_\\(\\)\\-\\.\\s+]+)");
+                // the underlying parser aggregates all instances of the same
+                // attribute, so we need to be able to parse them out
+                groupTemplate=groupTemplate+",*\\s*";
+            } else {
+               groupTemplate = "([A-Za-z0-9_\\(\\)\\-\\.\\s+]+,*\\s*)";
+            }
+            Pattern pattern = Pattern.compile(groupTemplate);
```
mike-jumper, line 153: What if the DN format contains a character that has special meaning to `Pattern.compile()`?

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/user/CASAuthenticatedUser.java
```diff
@@ -50,6 +51,11 @@
      */
     private Map<String, String> tokens;
 
+    /**
+     * The unique identifiers of all user groups which a user is a member of.
```
mike-jumper, line 55: I think you mean "this user" here.

mike-jumper (changes requested):

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/ConfigurationService.java
mike-jumper: I'd also like to point out that the line wrapping is a bit wonky here. We normally wrap our comments at 80 columns, but here the pattern of wrapping is inconsistent.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
doctorfree: These two changes are in the latest commits, value.toString() is assigned to a String variable and the substring expression is used directly in the call to pattern.matcher(). Thanks for the suggested code improvements.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/CASGuacamoleProperties.java
doctorfree: Modified the comment in latest commits to clarify this. Thanks for identifying this unclear comment.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/ConfigurationService.java
doctorfree: Added the period

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/conf/ConfigurationService.java
doctorfree: Clarified the comment and fixed line wrapping

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
doctorfree: Latest commits should resolve this and an instance of CASAuthenticatedUser should be returned now

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
doctorfree: This is the only feedback I have not yet addressed in commits. I am not sure if this is a problem or, if it is, how best to address it. I will look into this later this week if I have time. Please feel free to suggest or advise on best approach.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/user/CASAuthenticatedUser.java
doctorfree: Corrected. Yes, it should have been "this user"

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
doctorfree: Thanks for identifying this. I moved the initialization of the CAS authenticated user object into validateTicket as you suggest.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
necouchman: I'm not sure this actually addresses Mike's comment, here.  Mike is pointing out that it is inaccurate to say that the "CASAuthenticatedUser instance returned by the CAS server", because the CAS server is not returning a `CASAuthenticatedUser` instance - in fact, the CAS server does not know anything about the `CASAuthenticatedUser` class.  So, a more accurate comment would be:

> A CASAuthenticatedUser instance containing the ticket data returned by the CAS server.

Since the ticket _is_ what the CAS server returns.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
necouchman: I'm not 100% certain if this will work, but, based on this page:

https://www.baeldung.com/java-regexp-escape-char

It looks like the `\Q` and `\E` special characters can be used to escape anything between them. So it might be possible to use those around the value of the `groupDnFormat` String to automatically escape anything within that ??

necouchman (changes requested):
Getting close - hang in there! ;-)

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
doctorfree: Committed the suggested JavaDoc change to the description of the return value. However, my latest commits are not showing up here after rebasing on staging/1.3.0 so maybe I need to go back to git school.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
doctorfree: Although best practices would prohibit special characters in the DN it does not guarantee that. I believe your suggested use of \Q and \E to escape any special characters in the stripped DN is a good idea. Can we simply use the Pattern.quote() method as follows:

groupTemplate = Pattern.quote(groupDnFormat.replace("%s","([A-Za-z0-9_\\(\\)\\-\\.\\s+]+)"));

Or am I mistaken about where to escape the stripped DN?

necouchman: @mike-jumper Thoughts, here? RegEx still causes my mind to twist in ways I didn't think were possible...

image

mike-jumper: Welp:

> It looks like the `\Q` and `\E` special characters can be used to escape anything between them. So it might be possible to use those around the value of the `groupDnFormat` String to automatically escape anything within that ??

Nope - that would be unsafe. What if the value of `groupDnFormat` were `"\E.*\Q"`?

> Can we simply use the `Pattern.quote()` method as follows:
>
>     groupTemplate = Pattern.quote(groupDnFormat.replace("%s","([A-Za-z0-9_\(\)\-\.\s+]+)"));

Not quite - that would escape the regex used as the replacement, as well. I think you have the right idea. Perhaps:

```java
groupTemplate = Pattern.quote(groupDnFormat).replace("%s","([A-Za-z0-9_\(\)\-\.\s+]+)");
```

BUT:

* Definitely have some [Leaning Toothpick Syndrome](https://en.wikipedia.org/wiki/Leaning_toothpick_syndrome) going on with that regex. It's really tough to validate in the current context. Probably best to move that to a constant that can be separately documented, understood, and verified... that way this magic turns into something like: `Pattern.quote(groupDnFormat).replace("%s", REGEX_THAT_WE_KNOW_IS_CORRECT)`
* I'm not 100% sure the regex _is_ correct ... depending on what it's supposed to match against (which I'm also not 100% sure of). Perhaps any added documentation will clarify that and allow for easier review.

mike-jumper: Also ... what if the DN format needs to contain a literal "%s"?

mike-jumper: Also:

* What if the relevant part of the DN in an actual group contains characters that must be escaped for the sake of LDAP (like a comma)?
* What if the relevant groups are nested in different subtrees of a common base DN, or involve different attributes?

It seems to me that the code transforming a DN received from CAS into a generic group name would need to be LDAP syntax-aware, rather than employing pattern matching.

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
```diff
@@ -122,16 +134,44 @@
...
+                if (value != null) {
+                    String attrValue = value.toString();
+                    tokens.put(tokenName, attrValue);
+                    if (attr.getKey().equals(groupAttribute)) {
+                        Matcher matcher =
+                            pattern.matcher(attrValue.substring(1,attrValue.length()-1));
+                        while (matcher.find()) {
+                            effectiveGroups.add(matcher.group(1));
+                        }
+                    }
```
mike-jumper, line 168: I believe it was mentioned before that the format of the `memberOf` CAS attribute was `['GROUP1', 'GROUP2', ...]`. Do we know:

* Is this _truly_ the case? Or is it the result of `value.toString()` on some type that could be iterated cleanly? Is there any way to know whether the CAS client implementation returns an actual iterable type?
* If it does indeed provide only a string, where is that format dictated? How does it handle the possible presence of meaningful characters like `[`, `,`, or `]`?

lchanouha (commented):
Is this truly the case? 
=> no, the toString() is converting the collection to something displayable, not iterable. This is the CAS java client implementation. instanceof Collection is the way to test if CAS attribute is multivalued.
In the CAS 2.0 protocol, every value of attribute is separated in a xml tag.

This should look like this (not tested):
`
if (attr.getKey().equals(groupAttribute)) {
    if (attr.getValue() instanceof Collection) {
                for (final Object value : (Collection) attr.getValue()) {
                    effectiveGroups.add(value.toString());
                }
     } else {
       effectiveGroups.add(attr.getValue().toString()):
     }
}
`
IMHO

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
lchanouha: Is this truly the case?
=> no, the toString() is converting the collection to something displayable, not iterable. This is the CAS java client implementation. instanceof Collection is the way to test if CAS attribute is multivalued.
In the CAS 2.0 protocol, every value of attribute is separated in a xml tag.

This should look like this (not tested):
```
if (attr.getKey().equals(groupAttribute)) {
  if (attr.getValue() instanceof Collection) {
    for (final Object value : (Collection) attr.getValue()) { 
        effectiveGroups.add(value.toString()); } 
    } else { 
        effectiveGroups.add(attr.getValue().toString()): 
    }
  }
}
```
(wihout dn strip logic)
IMHO

extensions/guacamole-auth-cas/src/main/java/org/apache/guacamole/auth/cas/ticket/TicketValidationService.java
lchanouha: I think LdapName is the way:
http://cr.openjdk.java.net/~iris/se/12/latestSpec/api/java.naming/javax/naming/ldap/LdapName.html#%3Cinit%3E(java.lang.String)

necouchman:
Closed by #579

PR #566, opened 2020-10-14
add postgresql-auto-create-accounts mysql-auto-create-accounts env va…

michimau:
…riable management in docker stacks

You cannot set either postgresql-auto-create-accounts nor mysql-auto-create-accounts via env variables in docker-compose driven stacks. This very simple pull request adds the possibility.

### DIFF
--- a/guacamole-docker/bin/start.sh
+++ b/guacamole-docker/bin/start.sh
@@ -213,6 +213,10 @@ END
         "mysql-ssl-trust-store"  \
         "$MYSQL_SSL_TRUST_STORE"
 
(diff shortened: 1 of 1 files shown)


necouchman:
This looks pretty straight-forward; however, two things:
* All changes need a JIRA issue, so you'll need to create a JIRA issue for this.
* Once the JIRA issue is created, you'll need to tag both the pull request and any commit messages with the JIRA issue. See existing pull requests for examples.

necouchman:
This has already been merged with #585

PR #570, opened 2020-10-31
GUACAMOLE-1021: Do not duplicate objects when permissions are inherited from multiple entities.

mike-jumper:
Before user group support existed, permissions within the database support could be inherited from one location only: the user. This meant that narrowing results according to permissions could be accomplished with a simple `JOIN`.

This is no longer the case. With user groups now supported, a `JOIN` against permission-related tables may result in relevant objects being duplicated in the result set, with one copy of the object for each entity providing the permission.

These changes alter the permission-driven SQL queries for each implementation such that each readable object is returned only once, regardless of how many entities grant that same read permission. This is accomplished by creating a common SQL fragment, `getReadableIDs`, which other SQL queries may refer to. Various versions of this fragment are now the only SQL queries which explicitly refer to the "READ" permission.


necouchman (approved):
Well, I think I'm going cross-eyed after looking at that much XML, but it looks okay to me.

PR #513, opened 2020-05-25
GUACAMOLE-1078: Add Catalan Language to Guacamole

JordiRoman:
This branch provides the Catalan translation for the guacamole-client project (version 1.1.0).

Thank you very much.

PR #572, opened 2020-11-02
GUACAMOLE-1205: Bump Guacamole Client components to version 1.3.0

necouchman:
This bumps impacted client components to version `1.3.0`.  Of note for this:
* There were changes to both `guacamole-common` and `guacamole-common-js`, so they get bumped up to `1.3.0`
* There were no changes at all to `guacamole-auth-header`, including the pending change requests for the last of the issues.
* There are no committed changes to `guacamole-auth-ldap` or `guacamole-auth-saml`; however, because there are pending pull requests for those for translation fixes, I went ahead and bumped up the version numbers.

Let me know if I missed anything.

### DIFF
--- a/doc/guacamole-example/pom.xml
+++ b/doc/guacamole-example/pom.xml
@@ -26,7 +26,7 @@
     <groupId>org.apache.guacamole</groupId>
     <artifactId>guacamole-example</artifactId>
     <packaging>war</packaging>
-    <version>1.2.0</version>
+    <version>1.3.0</version>
     <name>guacamole-example</name>
     <url>http://guacamole.apache.org/</url>
 
@@ -106,7 +106,7 @@
         <dependency>
             <groupId>org.apache.guacamole</groupId>
             <artifactId>guacamole-common</artifactId>
-            <version>1.1.0</version>
+            <version>1.3.0</version>
             <scope>compile</scope>
         </dependency>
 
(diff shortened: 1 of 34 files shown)


mike-jumper (changes requested):
I think some changes are missing for guacamole-auth-header:

* The main `<version>` within `pom.xml` was not bumped.
* The `guac-manifest.json` was not updated.

mike-jumper:
> I think some changes are missing for guacamole-auth-header: ...

_Almost_ never mind. I was wrong on guacamole-auth-header having been touched (see apache/guacamole-manual#150), but that being the case its guacamole-ext dependency shouldn't have been updated here.

mike-jumper (changes requested):
Per apache/guacamole-manual#150:

* It looks like guacamole-auth-header shouldn't have been bumped
* Depending on whether there are pending changes for guacamole-auth-saml, it looks like that shouldn't have been bumped, either.

necouchman:
> Almost never mind. I was wrong on guacamole-auth-header having been touched (see apache/guacamole-manual#150), but that being the case its guacamole-ext dependency shouldn't have been updated here.

Ah, for some reason I thought we'd still want to rev the guacamole-ext version used by it, but makes sense. I've removed that change from the commit.

necouchman:
> Depending on whether there are pending changes for guacamole-auth-saml, it looks like that shouldn't have been bumped, either.

The Catalan add in #513 impacted the SAML extension.

mike-jumper:
> Ah, for some reason I thought we'd still want to rev the guacamole-ext version used by it, but makes sense.

Probably habit from past releases. In the pre-1.x days, we definitely needed to bump guacamole-ext each release or the extensions would be broken.

mike-jumper (approved):

PR #575, opened 2020-11-16
GUACAMOLE-1207: Add Portuguese translation

rbm0407:
Add translation files for the webapp and the following extensions:
  - guacamole-auth-cas
  - guacamole-auth-duo
  - guacamole-auth-jdbc
  - guacamole-auth-openid
  - guacamole-auth-quickconnect
  - guacamole-auth-saml
  - guacamole-auth-totp

### DIFF
--- a/extensions/guacamole-auth-cas/src/main/resources/guac-manifest.json
+++ b/extensions/guacamole-auth-cas/src/main/resources/guac-manifest.json
@@ -14,6 +14,7 @@
         "translations/de.json",
         "translations/en.json",
         "translations/ja.json",
+        "translations/pt.json",
         "translations/ru.json"
     ]
 
--- a/extensions/guacamole-auth-cas/src/main/resources/translations/pt.json
+++ b/extensions/guacamole-auth-cas/src/main/resources/translations/pt.json
@@ -1,0 +1,12 @@
+{
+
+    "DATA_SOURCE_CAS" : {
+        "NAME" : "CAS SSO Backend"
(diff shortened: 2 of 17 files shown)


necouchman (changes requested):
Overall looks pretty good to me, just one or two things to note:
* If you're not translating a string, leave it out of the translation file, and it will fall back to the English translation. If the translation is identical in Portuguese as in English, it should be included in the translation file.
* In the JDBC module, you need to modify the `guac-manifest.json` files for each of the modules and add the pt.json translation:

https://github.com/apache/guacamole-client/blob/master/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/resources/guac-manifest.json

https://github.com/apache/guacamole-client/blob/master/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/resources/guac-manifest.json

https://github.com/apache/guacamole-client/blob/master/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/resources/guac-manifest.json

necouchman:
@rbm0407: Looking pretty good. However, now you have some extra commits - please squash them all down into a single commit, and make sure the commit message is descriptive. After that I think it'll be ready to go!

rbm0407:
Sorry, this is my first pull request hahaha

Squashed to a single commit

necouchman (approved):

PR #576, opened 2020-11-21
GUACAMOLE-1218: Include "guacamole-auth-json" within Apache Guacamole.

mike-jumper:
This change adds the [guacamole-auth-json extension](https://github.com/glyptodon/guacamole-auth-json) to the main guacamole-client project. It was originally created and maintained by [Glyptodon, Inc.](https://enterprise.glyptodon.com/), but been increasingly cited on the mailing lists and would definitely be a good addition to the set of extensions provided out-of-the-box by Apache Guacamole.

From the original `README.md`:

> guacamole-auth-json is an authentication extension for Apache Guacamole which authenticates users using JSON which has been signed using HMAC/SHA-256 and encrypted with 128-bit AES in CBC mode. This JSON contains all information describing the user being authenticated, as well as any connections they have access to.

The extension allows external applications to authenticate and authorize users by declaring those users' identities and available connections within encrypted and signed JSON. Encryption ensures the contents of the JSON (typically sensitive connection parameters) cannot be read except by Guacamole itself, while signing ensures the JSON is accepted only if it has been generated by a server authorized to generate it.

## Additional changes

The first commit here is a verbatim copy from [the original guacamole-auth-json repository](https://github.com/glyptodon/guacamole-auth-json) at [commit f7b2eaf](https://github.com/glyptodon/guacamole-auth-json/tree/f7b2eaf6a65b7cd25fd73437360e36fe46e0bcb9). From there, I have made the following additional changes:

* Refactored `org.glyptodon.guacamole.*` packages to `org.apache.guacamole.*`.
* Relicensed under Apache License v2.0.
* Removed the `@author` annotations that we do not use in the guacamole-client project.
* Refactored useful property classes out of the extension and into guacamole-ext.
* Documented the licenses of all bundled dependencies.
* Updated to the latest guacamole-ext.
* Updated the Spring Security library to the latest (used here to validate IP addresses).

## Third-party contributions and the Software Grant Agreement

I've already gone ahead and signed an [SGA](https://www.apache.org/licenses/contributor-agreements.html#grants) on behalf of Glyptodon, so things should be good to go. There was a recent contribution from @BojanZelic (glyptodon/guacamole-auth-json#11), which could have made the SGA more complicated. The version of the code contributed here is just prior to that contribution, and @BojanZelic says he's OK with re-contributing things to the guacamole-client repository when it's ready.


extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/AuthenticationProviderService.java
```diff
@@ -0,0 +1,116 @@
...
+     * @throws GuacamoleException
+     *     If the UserContext cannot be created due to an error.
+     */
+    public UserContext getUserContext(org.apache.guacamole.net.auth.AuthenticatedUser authenticatedUser)
+            throws GuacamoleException {
+
+        // The JSONAuthenticationProvider only provides data for users it has
+        // authenticated itself
+        if (!(authenticatedUser instanceof AuthenticatedUser))
+            return null;
```
necouchman, line 106: If I'm reading this code/comment correctly, it is probably worth mentioning in the documentation - either the `README.md` file within this extension, or whatever GUG chapter is generated for this later (or both) - that, by design, this module will essentially not "stack" with other authentication modules.

extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/CryptoService.java
```diff
@@ -0,0 +1,215 @@
...
+        }
+        catch (InvalidKeyException e) {
+            throw new GuacamoleServerException(e);
+        }
+        catch (IllegalBlockSizeException e) {
+            throw new GuacamoleServerException(e);
+        }
+        catch (BadPaddingException e) {
+            throw new GuacamoleServerException(e);
+        }
```
necouchman, line 167: Any reason not to use the multi-catch, here?

extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/CryptoService.java
```diff
@@ -0,0 +1,215 @@
...
+        // Rethrow all signature failures identically
+        catch (NoSuchAlgorithmException e) {
+            throw new GuacamoleServerException(e);
+        }
+        catch (InvalidKeyException e) {
+            throw new GuacamoleServerException(e);
+        }
+        catch (IllegalStateException e) {
+            throw new GuacamoleServerException(e);
+        }
```
necouchman, line 211: And here?

extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/RequestValidationService.java
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * Service for testing the validity of received HTTP requests.
+ */
+public class RequestValidationService {
+
+    /**
+     * Logger for this class.
+     */
+    private final Logger logger = LoggerFactory.getLogger(RequestValidationService.class);
```
necouchman, line 39: I think the `Logger` implementations in most other classes are also `static` - any reason this deviates from that practice?

extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/user/UserDataService.java
```diff
@@ -0,0 +1,373 @@
...
+/**
+ * Service for deriving Guacamole extension API data from UserData objects.
+ */
+@Singleton
+public class UserDataService {
+
+    /**
+     * Logger for this class.
+     */
+    private final Logger logger = LoggerFactory.getLogger(UserDataService.class);
```
necouchman, line 59: As with other class above, any reason this should or should not be `static`, in keeping consistency with other Guacamole code? I have no particular preference, just wondering.

extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/user/UserDataService.java
```diff
@@ -0,0 +1,373 @@
...
+
+    /**
+     * ObjectMapper for deserializing UserData objects.
+     */
+    private static final ObjectMapper mapper = new ObjectMapper();
+
+    /**
+     * Blacklist of single-use user data objects which have already been used.
+     */
+    private final UserDataBlacklist blacklist = new UserDataBlacklist();
```
necouchman, line 69: Any thoughts/feelings about avoiding use of the term "blacklist", both here and as the name of the `UserDataBlacklist` class?

https://thenextweb.com/dd/2020/07/13/linux-kernel-will-no-longer-use-terms-blacklist-and-slave/

I'm not one to get overly worked up about this one way or the other, and it is certainly still broadly-used in the computing realm in ways that don't necessarily have racial meaning or implication; on the other hand, given the history of the term and, as noted in the page above, the fact that there are some movements to try to move away from that, maybe we should do the same?

necouchman (changes requested):
Just a few things popped up during this initial review...

extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/AuthenticationProviderService.java
mike-jumper: It can stack. For example, if you submit JSON like:

```json
{
    "username" : "guacadmin"
}
```

that would be sufficient for the database authentication to accept the user as "guacadmin".

The JSON extension simply won't attempt to provide _data_ (a `UserContext`) for a user that it hasn't authenticated, as its data comes from the JSON. If there isn't any JSON, then the JSON extension has no data to provide.

extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/CryptoService.java
mike-jumper: Sure - I'll switch instances of this over to multi-catch.

extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/RequestValidationService.java
mike-jumper: No reason - older code was all over the place with `static` vs. not for `Logger`. I'll switch this over to `static`.

extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/user/UserDataService.java
mike-jumper: Sure. Regardless of these words' origins, regardless of intent, we should not continue to assign negative connotations to "black".

extensions/guacamole-auth-json/README.md
```diff
@@ -0,0 +1,189 @@
...
+2. Sign the JSON using the secret key (the same 128-bit key stored within
+   `guacamole.properties` with the `json-secret-key` property) with
+   **HMAC/SHA-256**. Prepend the binary result of the signing process to the
+   plaintext JSON that was signed.
+3. Encrypt the result of (2) above using **AES in CBC mode**, with the initial
+   vector (IV) set to all zero bytes.
+4. Encode the encrypted result using base64.
+5. POST the encrypted result to the `/api/tokens` REST endpoint as the value of
+   an HTTP parameter named `data` (or include it in the URL of any Guacamole
+   page as a query parameter named `data`).
```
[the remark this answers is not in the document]
mike-jumper: How large did the parameter need to be to run into that limit?

extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/user/UserDataService.java
mike-jumper: I agree we should maintain coding standards regarding that, and that new code should use the established `private static final` practice for `Logger` instances.

extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/AuthenticationProviderService.java
necouchman: Ah, okay, that makes sense. Thanks for the clarification.

extensions/guacamole-auth-json/README.md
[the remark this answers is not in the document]
necouchman: Hmmm...I'm not sure I ever nailed it down when doing that - just knew that trying to push the SAML metadata back in via a `GET` request didn't work. It seems like it is something that is configurable, and that Tomcat has a default 8K limit for the request URL as a whole, so it's probably fairly large. Maybe it's not a big deal..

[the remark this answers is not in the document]
mike-jumper: Would it be better if I pull the original `README.md` out of these changes and then revisit documenting query parameter limits in the guacamole-manual version of the same?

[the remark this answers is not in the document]
necouchman: Nah, let's just keep the README for now and we can address it if someone stumbles across it, and/or in the documentation.

necouchman (approved):

PR #577, opened 2020-11-21
GUACAMOLE-1219: Allow TOTP to be disabled by group membership

janhf:
This parts changes the code, so that TOTP can be disabled by membership in a special (configurable) group.

### DIFF
--- a/extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/conf/ConfigurationService.java
+++ b/extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/conf/ConfigurationService.java
@@ -53,10 +53,22 @@ public class ConfigurationService {
     };
 
     /**
+     * The group of users which are not enforced to use TOTP. By default,
+     * this will be empty/no one.
+     */
+    private static final StringGuacamoleProperty TOTP_DISABLED_GROUP =
+            new StringGuacamoleProperty() {
+
+        @Override
+        public String getName() { return "totp-disabled-group"; }
+
+    };
+
+    /**
      * The number of digits which should be included in each generated TOTP
      * code. By default, this will be 6.
      */
-    private static final IntegerGuacamoleProperty TOTP_DIGITS=
+    private static final IntegerGuacamoleProperty TOTP_DIGITS =
             new IntegerGuacamoleProperty() {
 
         @Override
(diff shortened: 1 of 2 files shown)


mike-jumper:
Just as the TOTP support currently leverages a set of custom attributes to store a user's TOTP key and whether enrollment has succeeded, I think a better approach would be to use the same part of the extension API (custom attributes) to allow the administrator to check "Disable TOTP" for a user or group.

necouchman:
@janhf Any further progress on doing things in the way that @mike-jumper suggested, using attributes?

JLE-ATS:
Hello,

I hope it's OK to add a comment on a closed PR. I am new to this project. Let me know if I should do things differently.

I am currently in the process of deploying Guacamole in my company and I think I need this specific feature.
Here is why. I am installing Guacamole with Ansible. Some configurations (user groups, connection groups, permissions) are set through the REST API.
If I enable TOTP, I can't use the REST API and thus, configure Guacamole.

If this feature was added, I could :

1. add "totp-disabled-group: XXX" to my guacamole.properties,
2. add the XXX group through the REST API,
3. add my API user to the XXX group through the REST API,
4. enable the TOTP extension.
5. Restart tomcat (then TOTP is working, but disabled for the API's user).

Then I would proceed with the rest of the configuration through the API, as usual.

Is it possible to consider this PR again or is there another solution I am not aware of ?
I have found another PR ([911](https://github.com/apache/guacamole-client/pull/911)) which talk about disabling MFA based on IP addresses. It could do the trick, but it seems less relevant to my use case.

Best regards

necouchman:
@JLE-ATS In general you should not comment on an open PR - if you have a question, use the mailing list.

The PR was abandoned and closed because support for this feature was merged in #808. The feature will be released as part of the 1.6.0 release. The other PR you mentioned, #911, is still in review and will likely go into 1.6.0, but may be in a release after that if it doesn't make the 1.6.0 scope.

EvannG1:
Hi @necouchman I have the same request as @JLE-ATS I use Apache Guacamole 1.5.4 (latest version at the moment) with the TOTP extension.
The problem is that I can't use the REST API for automation purposes with this extension.
If I've understood correctly, the TOTP disabling feature for a user and/or user group will arrive in 1.6.0, but I can't find any information about an approximate release date for this version?

necouchman:
@EvannG1 We do not have a release date planned, yet. We're working to ship 1.5.5, which will hopefully be out in the next week or so. After that we'll turn our attention to 1.6.0. I would say 1.6.0 is several months out, but that's just my rough guess/approximation, and I can't get any more granular than that.

EvannG1:
> @EvannG1 We do not have a release date planned, yet. We're working to ship 1.5.5, which will hopefully be out in the next week or so. After that we'll turn our attention to 1.6.0. I would say 1.6.0 is several months out, but that's just my rough guess/approximation, and I can't get any more granular than that.

I'd like to come back to you now that 1.5.5 has been released. Is there an approximate release date for 1.6.0?

EvannG1:
> @EvannG1 I think we are probably a week or two out from the release of 1.6.0 - we're finishing up the final issues (6 open ones remaining) and moving rapidly toward that release.

Any news about the 1.6.0 release?

PR #578, opened 2020-11-24
GUACAMOLE-221: Expose underlying protocol at tunnel level.

mike-jumper:
Previously, the underlying protocol of an active connection was available only via the connection definition itself. That information is not available if the user is connecting through a connection group or a shared connection, causing handling of the `argv` and `required` instructions to fail in those cases.

These changes establish a REST API endpoint for retrieving underlying protocol unformation from the tunnel, independent of whether a connection, connection group, or shared connection was used. This involved:

* Adding a `onuuid` event to `Guacamole.Tunnel` to allow implementations to reliably wait for assignment of the tunnel UUID.
* Adding a `getProtocol()` function to the `GuacamoleSocket` interface, defaulting to unknown (`null`). This required upgrading the guacamole-common build to Java 1.8.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
@@ -201,87 +201,54 @@ public abstract class AbstractGuacamoleTunnelService implements GuacamoleTunnelS
             ModeledConnectionGroup connectionGroup);
 
     /**
-     * Returns a guacamole configuration containing the protocol and parameters
-     * from the given connection. If the ID of an active connection is
-     * provided, that connection will be joined instead of starting a new
-     * primary connection. If tokens are used in the connection parameter
-     * values, credentials from the given user will be substituted
-     * appropriately.
-     *
-     * @param user
-     *     The user whose credentials should be used if necessary.
+     * Returns a GuacamoleConfiguration which connects to the given connection.
+     * If the ID of an active connection is provided, that active connection
+     * will be joined rather than establishing an entirely new connection. If
+     * a sharing profile is provided, the parameters associated with that
+     * sharing profile will be used to define the access provided to the user
+     * accessing the shared connection. If tokens are used in the connection
+     * parameter values of the sharing profile, credentials from the given user
+     * will be substituted appropriately.
      *
      * @param connection
-     *     The connection whose protocol and parameters should be added to the
-     *     returned configuration.
+     *     The connection that the user is connecting to.
      *
      * @param connectionID
-     *     The ID of the active connection to be joined, as returned by guacd,
-     *     or null if a new primary connection should be established.
-     *
-     * @return
-     *     A GuacamoleConfiguration containing the protocol and parameters from
-     *     the given connection.
-     */
-    private GuacamoleConfiguration getGuacamoleConfiguration(RemoteAuthenticatedUser user,
-            ModeledConnection connection, String connectionID) {
-
-        // Generate configuration from available data
-        GuacamoleConfiguration config = new GuacamoleConfiguration();
-
-        // Join existing active connection, if any
-        if (connectionID != null)
-            config.setConnectionID(connectionID);
-
-        // Set protocol from connection if not joining an active connection
-        else {
-            ConnectionModel model = connection.getModel();
-            config.setProtocol(model.getProtocol());
-        }
-
-        // Set parameters from associated data
-        Collection<ConnectionParameterModel> parameters = connectionParameterMapper.select(connection.getIdentifier());
-        for (ConnectionParameterModel parameter : parameters)
-            config.setParameter(parameter.getName(), parameter.getValue());
-
-        return config;
-        
-    }
-
-    /**
-     * Returns a guacamole configuration which joins the active connection
-     * having the given ID, using the provided sharing profile to restrict the
-     * access provided to the user accessing the shared connection. If tokens
-     * are used in the connection parameter values of the sharing profile,
-     * credentials from the given user will be substituted appropriately.
-     *
-     * @param user
-     *     The user whose credentials should be used if necessary.
+     *     The ID of the active connection being joined, as provided by guacd
+     *     when the original connection was established, or null if a new
+     *     connection should be established instead.
      *
      * @param sharingProfile
      *     The sharing profile whose associated parameters dictate the level
-     *     of access granted to the user joining the connection.
-     *
-     * @param connectionID
-     *     The ID of the connection being joined, as provided by guacd when the
-     *     original connection was established, or null if a new connection
-     *     should be created instead.
+     *     of access granted to the user joining the connection, or null if the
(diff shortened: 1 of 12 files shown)


knacktim:
Cherry-picked these into my staging/1.3.0 testing branch and all seems well.

Just for my own knowledge -- is there a reason that we could not just send the protocol and form information through the websocket-tunnel?  Could that cause race conditions?

mike-jumper:
> Just for my own knowledge -- is there a reason that we could not just send the protocol and form information through the websocket-tunnel? Could that cause race conditions?

No, the reason that forms cannot be exposed automaticlaly as part of the tunnel is separation of concerns. The tunnel exists as a low-level component defined by guacamole-common and guacamole-common-js, whereas the concept of fields and forms are specific to the Guacamole webapp and its extension API (guacamole-ext). Only the protocol name and arbitrary parameter name/value pairs exist at that low a level.

Before deciding on the current implementation, I also considered:

* **Allowing `ConnectionGroup` and `ActiveConnection` to expose a `GuacamoleConfiguration`, possibly by moving the relevant function from `Connection` to `Connectable`.** This would have required more substantial changes, and would added further complexity to the part of the client-side code that pulls this information.
* **Modifying the `argv` and `required` instructions to expose the protocol name.** This feels redundant and would break backward compatibility for the established `argv` instruction.

Adding a new instruction that simply exposes the protocol name could be another approach.

mike-jumper:
Converted this PR to a draft - I'd like to investigate the protocol-driven approach.

mike-jumper:
I think this PR is still the best approach. In the general case of usage of guacamole-common and guacamole-common-js, it should be up to the implementation to decide whether the protocol name should be exposed. 

In the Apache Guacamole web application, we do not consider the underlying protocol name to be secret, and we _do_ need the protocol name to process `argv` and `required`, but we can't assume that to be the case for all implementations.

necouchman (changes requested):
Just one nitpick about the comment section, and some nostalgia for the last of Java 1.6, but looks pretty good.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
```diff
@@ -201,87 +201,54 @@ protected abstract void release(RemoteAuthenticatedUser user,
...
-     * @param user
-     *     The user whose credentials should be used if necessary.
+     * Returns a GuacamoleConfiguration which connects to the given connection.
+     * If the connection ID of an active is provided, that active connection
+     * will be joined rather than establishing an entirely new connection. If
+     * a sharing profile is provided, the parameters associated with that
+     * sharing profile will be used to define the access provided to the user
+     * accessing the shared connection. If tokens are used in the connection
+     * parameter values of the sharing profile, credentials from the given user
+     * will be substituted appropriately.
```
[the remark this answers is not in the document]
mike-jumper: Right - that wording is wrong as of commit 1210d5624c4eb173417cab8358eca4cc3b6c0ebe (which deprecated the `StandardTokens` class). This function used to perform token substitution, but no longer needs to and hasn't since that commit.

I've now removed that "If tokens are used ..." sentence.

guacamole-common/pom.xml
```diff
@@ -57,14 +57,14 @@
     <build>
         <plugins>
 
-            <!-- Written for 1.6 -->
+            <!-- Written for 1.8 -->
```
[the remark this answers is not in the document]
mike-jumper: 😿

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
```diff
@@ -201,87 +201,52 @@ protected abstract void release(RemoteAuthenticatedUser user,
...
-     * from the given connection. If the ID of an active connection is
-     * provided, that connection will be joined instead of starting a new
-     * primary connection. If tokens are used in the connection parameter
-     * values, credentials from the given user will be substituted
-     * appropriately.
-     *
-     * @param user
-     *     The user whose credentials should be used if necessary.
+     * Returns a GuacamoleConfiguration which connects to the given connection.
+     * If the connection ID of an active is provided, that active connection
```
necouchman, line 205: > If the connection ID of an active *connection* is provided

Although, the term "active connection" is used several times in rapid succession, so maybe:

> If the ID of an active connection is provided

would make it a little more succinct.

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleConfiguration.java
```diff
@@ -103,15 +102,34 @@ public void setConnectionID(String connectionID) {
...
 
     /**
-     * Sets the name of the protocol to be used.
-     * @param protocol The name of the protocol to be used.
+     * Sets the name of the protocol to be used. If no connection is being
+     * joined (a new connection is being established), this value must be set.
+     *
+     * <p>If a connection is being joined, <strong>this value should still be
+     * set</strong> to ensure that protocol-specific responses like the
+     * "required" and "argv" instructions  can be understood in their proper
```
necouchman, line 119: Extra space slipped in here...

guacamole/src/main/webapp/app/rest/services/tunnelService.js
```diff
@@ -79,6 +79,36 @@ angular.module('rest').factory('tunnelService', ['$injector',
...
+     *     success.
+     */
+    service.getProtocol = function getProtocol(tunnel) {
+
+        // Build HTTP parameters set
+        var httpParameters = {
+            token : authenticationService.getCurrentToken()
+        };
+
+        // Retrieve all associated sharing profiles
```
necouchman, line 102: Copy pasta? Don't think this one is retrieving sharing profiles...

necouchman (changes requested):
A few more documentation fix-ups and we can merge.

guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleConfiguration.java
mike-jumper: Removed via rebase.

guacamole/src/main/webapp/app/rest/services/tunnelService.js
mike-jumper: Yep. Fixed via rebase.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
mike-jumper: Sure. Change made via rebase.

necouchman (approved):
:shipit:

knacktim:
Thanks for the quick turn-around guys!  Happy Turkey week 🦃

PR #581, opened 2020-12-18
GUACAMOLE-1218: Add auth-json config & extension to docker

BojanZelic:
The Docker file is currently missing the auth-json extension & configuration. 

This allows us to enable the extension by defining the $JSON_SECRET_KEY environmental variable

### DIFF
--- a/guacamole-docker/bin/start.sh
+++ b/guacamole-docker/bin/start.sh
@@ -688,6 +688,19 @@ END
 }
 
 ##
+## Adds properties to guacamole.properties which configure the json
+## authentication provider.
+##
+associate_json() {
+    # Update config file
+    set_property          "json-secret-key"        "$JSON_SECRET_KEY"
+    set_optional_property "json-trusted-networks"  "$JSON_TRUSTED_NETWORKS"
+
+    # Add required .jar files to GUACAMOLE_EXT
+    ln -s /opt/guacamole/json/guacamole-auth-*.jar "$GUACAMOLE_EXT"
+}
+
+##
 ## Starts Guacamole under Tomcat, replacing the current process with the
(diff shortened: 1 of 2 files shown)


guacamole-docker/bin/start.sh
```diff
@@ -687,6 +687,18 @@ END
     ln -s /opt/guacamole/cas/guacamole-auth-*.jar   "$GUACAMOLE_EXT"
 }
 
+##
+## Adds properties to guacamole.properties which configure the json
+## authentication provider.
+##
+associate_json() {
+    # Update config file
+    set_optional_property "json-secret-key"         "$JSON_SECRET_KEY"
```
necouchman, line 696: Is this an optional property? Isn't this required if the JSON extension is used?

necouchman (changes requested):

guacamole-docker/bin/start.sh
BojanZelic: Good catch. should be fixed now.

guacamole-docker/bin/start.sh
```diff
@@ -687,6 +687,18 @@ END
     ln -s /opt/guacamole/cas/guacamole-auth-*.jar   "$GUACAMOLE_EXT"
 }
 
+##
+## Adds properties to guacamole.properties which configure the json
+## authentication provider.
+##
+associate_json() {
+    # Update config file
+    set_property "json-secret-key"                 "$JSON_SECRET_KEY"
```
necouchman, line 696: While we're at it, should we also make the optional property "json-trusted-networks" available, here?

BojanZelic: Good point, that property wasn't documented and I wasn't aware it existed. Should be fixed now

necouchman (approved):

PR #584, opened 2020-12-28
GUACAMOLE-1199: Only update unrestricted JDBC attributes if they have been provided.

necouchman:


extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/user/UserVerificationService.java
```diff
@@ -166,7 +166,7 @@ private boolean setKey(UserContext context, UserTOTPKey key)
 
         // Get mutable set of attributes
         User self = context.self();
-        Map<String, String> attributes = new HashMap<String, String>();
+        Map<String, String> attributes = new HashMap<>(self.getAttributes());
```
mike-jumper, line 169: The [documentation for `setAttributes()`](https://guacamole.apache.org/doc/guacamole-ext/org/apache/guacamole/net/auth/Attributes.html#setAttributes-java.util.Map-) states:

> ... Any attributes not within the given map will be left untouched. ...

If the backend handling storage of these attributes is working correctly, this _shouldn't_ be necessary. Perhaps the database support is not correctly implementing `setAttributes()`?

necouchman: Maybe so - the current implementation does seem to be consistently wiping out other attributes when the TOTP attributes are added. Interestingly it does _not_ touch the restricted attributes - things like login date/time limitations - so it's only the unrestricted ones that are being wiped out.

necouchman: I'm still trying to track this down, but I think that the issue is here:

https://github.com/apache/guacamole-client/blob/e5deb9b6a44ca809ae23921a09fb2f79a2af10f0/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java#L496-L512

and then in the concrete implementations of this:

https://github.com/apache/guacamole-client/blob/e5deb9b6a44ca809ae23921a09fb2f79a2af10f0/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectMapper.java#L144

Since the attributes that are being overwritten are not stored in the `guacamole_user_attributes` table (for arbitrary attributes), and the update query updates the record in the `guacamle_user` table, the query assumes that _all_ of the attribute values will be available and present in that query. For whatever reason the restricted ones are still there, but the unrestricted ones get pulled off somewhere along the way.

alexandru-bagu: The issue lies in https://github.com/apache/guacamole-client/blob/9ad5d4801c2ce9510afdbd347067edcab86cafbc/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledUser.java#L512
Perhaps https://github.com/apache/guacamole-client/blob/9ad5d4801c2ce9510afdbd347067edcab86cafbc/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledUser.java#L518 should be
`setUnrestrictedAttributes(self.getAttributes());`
since we have already merged the old attributes with the new ones.

Likewise for https://github.com/apache/guacamole-client/blob/9ad5d4801c2ce9510afdbd347067edcab86cafbc/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledUser.java#L522

necouchman: @mike-jumper Any other thoughts on this?

mike-jumper: I think the issue boils down to the current implementation of `setUnrestrictedAttributes()` and `setRestrictedAttributes()`. For example, consider `setUnrestrictedAttributes()`:

https://github.com/apache/guacamole-client/blob/754e9649f1fa0ba225ee42b56ded64bc283d17df/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledUser.java#L474-L488

When this is called from within `setAttributes()`, it will clear out any of the modeled attributes that are absent from the `attributes` map, which is incorrect behavior.

In both `setUnrestrictedAttributes()` and `setRestrictedAttributes()`, I think code like:

necouchman: Okay, I've updated as requested and am in the process of testing it to make sure that it works as expected.

necouchman: Well, this resolves the issue of the data getting wiped out during TOTP enrollment; however, the data is no longer displayed, at all, in the admin interface - in fact, when I go to edit a user, I don't see the "Profile" section at all. Guessing something else needs to be updated in addition to these two places...

mike-jumper: This might have something to do with the logic that decides whether an attribute field appears in the edit screens. Essentially all attributes with corresponding form information are shown when creating things, but only fields with corresponding explicit values (even if blank) are shown when editing.

I thought there was logic in place that returned `""` values for these attributes if unset...

necouchman: Okay, I _think_ I've finally managed to get it - pretty much as simple as you said :smile:. Testing looks good - the attributes appear to be maintained across TOTP enrollment, and the fields are present both when editing a user and when creating a new user.

mike-jumper (commented):
This looks real nice and simple.

I notice only `setUnrestrictedAttributes()` is touched by these changes. Do the same changes need to be made to `setRestrictedAttributes()`?

necouchman:
@mike-jumper Yes, that's a good point. Interestingly in my testing I never ran into those being wiped out, but it seems like a sane check, anyway. I've added that.

mike-jumper (approved):

PR #589, opened 2021-02-09
GUACAMOLE-1284: add OPENID_MAX_TOKEN_VALIDITY to start.sh

tschoonj:
Due to the settings of the OIDC provider, I was forced to set openid-max-token-validity to 24 hours. This was not supported by the start.sh script so I was forced to create a build myself. I think it would be justified for this small patch to go into the script though.


mike-jumper (approved):

tschoonj:
Thanks!

PR #590, opened 2021-02-10
GUACAMOLE-1204: Add support for pass-through of multi-touch events.

mike-jumper:
These changes add client-side support for pass-through of multi-touch events via the Guacamole protocol and supporting objects. They correspond to the server-side changes within apache/guacamole-server#331.

This support is implemented through:

* A new abstraction, `Guacamole.Touch`, which provides cross-browser touch event handling (similar in purpose and pattern to `Guacamole.Mouse` and `Guacamole.Keyboard`).
* A new event system, `Guacamole.Event` and `Guacamole.Event.Target`, which allows for multiple event listeners and consistent control over DOM propagation and default behavior.
* A new base class, `Guacamole.Position`, allowing `Guacamole.Mouse.State` and `Guacamole.Touch.State` to share some common code.
* Addition of a unit test framework (!!!) and a set of tests for the new `Guacamole.Event` system.

I've only leveraged the new event system for the new `Guacamole.Touch` object, but will update the various other components in another set of changes if all this looks OK. The system is designed to be backward compatible with the old `oneventname` handlers.

I've also taken this opportunity to:

* Simplify the constructor for `Guacamole.Mouse.State`, which is matched by `Guacamole.Touch.State`.
* Add optional automatic scaling of mouse and touch events relative to the display scale (rather than require implementations to manually scale state before passing to `sendMouseState()` and `sendTouchState()`).


guacamole-common-js/src/main/webapp/modules/Position.js
```diff
@@ -0,0 +1,92 @@
...
+
+var Guacamole = Guacamole || {};
+
+/**
+ * A position in 2-D space.
+ *
+ * @constructor
+ * @param {Guacamole.Position|Object} [template={}]
+ *     The object whose properties should be copied within the new
+ *     Guacamole.Mouse.State.
```
jmuehlner, line 28: Mouse?

guacamole-common-js/src/main/webapp/modules/Position.js
```diff
@@ -0,0 +1,92 @@
...
+     *
+     * @param {Number} clientY
+     *     The viewport-relative Y coordinate to translate.
+     */
+    this.fromClientPosition = function fromClientPosition(element, clientX, clientY) {
+
+        this.x = clientX - element.offsetLeft;
+        this.y = clientY - element.offsetTop;
+
+        // This is all JUST so we can get the mouse position within the element
```
jmuehlner, line 71: mouse?

guacamole-common-js/src/main/webapp/modules/Position.js
mike-jumper: 🖱️ ✔️ 

Fixed via rebase.

guacamole-common-js/src/main/webapp/modules/Position.js
mike-jumper: Also fixed via rebase.

PR #592, opened 2021-02-18
prompt when closing the window/tab to avoid accidents with ctrl-w

phreakocious:
Unintentionally closing a guac tab with ctrl-w is a frequent complaint.  This change prompts the user when closing the tab if there is an active connection.


guacamole/src/main/webapp/app/client/controllers/clientController.js
```diff
@@ -219,6 +220,12 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
         remaining: 15
     };
 
+    // Catch window or tab closing (ctrl-w) and prompt the user
+    var ctrlwlistener = function onBeforeUnload(e) {
+            e.preventDefault();
+            e.returnValue = '';
+     };
```
mike-jumper, line 227: 1. Please correct the indentation here. The body of the function appears to be indented far further than elsewhere (we use 4-space indents for each level), and the closing brace of the function definition is misaligned.
2. Please use JSDoc to document functions, their return values, and their parameters.
3. I think calling this function `ctrlwlistener()` is a bit misleading. Not all browsers will use Ctrl+W for closing a tab or window, nor is this function directly related to that shortcut.

guacamole/src/main/webapp/app/client/controllers/clientController.js
```diff
@@ -776,6 +783,8 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
 
         // Client error
         else if (connectionState === ManagedClientState.ConnectionState.CLIENT_ERROR) {
+            // Stop intercepting window close
+            $window.removeEventListener('beforeunload', ctrlwlistener, false);
```
mike-jumper, line 787: Why repeatedly add, remove, add, remove, etc. the listener vs. checking the relevant state within the listener itself?

What about connections that are running in the background, as will occur if the user navigates back to the home screen or to a settings screen without disconnecting?

mike-jumper (changes requested):
Please also read through our contribution guidelines when you can:

https://github.com/apache/guacamole-client/blob/65895aecdefecba94272065d5820db95e8322d1e/CONTRIBUTING

In particular, each change needs a corresponding JIRA issue, and that JIRA issue needs to be tagged in the commit message(s) following established convention. You can see examples of this in practice in our git history.

guacamole/src/main/webapp/app/client/controllers/clientController.js
phreakocious: I'm not much of a JS programmer.. couldn't find a way to test whether there was currently a connection.  If using connection groups, it would be ideal to be prompted even if at one of the settings screens, as they could lose the connection they had if it closes.

guacamole/src/main/webapp/app/client/controllers/clientController.js
phreakocious: Thanks, will look at it.

guacamole/src/main/webapp/app/client/controllers/clientController.js
phreakocious: The new commit has moved this to a check in the event handler.

phreakocious:
Was there anything else you'd like altered?

guacamole/src/main/webapp/app/client/controllers/clientController.js
```diff
@@ -219,6 +220,26 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
...
+        var managedClient = $scope.client;
+        if (managedClient) {
+
+            // Get current connection state
+            var connectionState = managedClient.clientState.connectionState;
+
+            // If connected, prompt to close
+            if (connectionState === ManagedClientState.ConnectionState.CONNECTED)
+                e.preventDefault();
+                e.returnValue = '';
```
necouchman, line 239: It's possible I'm just missing something here about how JS behaves, but does this actually prompt the user? It looks like all it does is check if the connection is active, block the default action, and clear the return value of the event? Does that somehow generate a browser prompt? Or does this just silently block the attempt to close the window?

necouchman (commented):
One question/concern from me...

guacamole/src/main/webapp/app/client/controllers/clientController.js
phreakocious: This event has special properties and causing the browser to generate a prompt is exactly what it does.  It used to be possible to provide a custom message via the returnValue, but that has been deprecated in pretty much all browsers for security reasons.

necouchman: Got it, thanks for clarifying!

guacamole/src/main/webapp/app/client/controllers/clientController.js
```diff
@@ -831,6 +853,8 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
 
         // Hide status and sync local clipboard once connected
         else if (connectionState === ManagedClientState.ConnectionState.CONNECTED) {
+            // Start intercepting ctrl-w / window close
+            $window.addEventListener('beforeunload', windowCloseListener, false);
```
mike-jumper, line 857: I think this will be problematic. Since the event listener is being added conditionally depending on whether the _current_ client has connected (and removing it upon error), there will be issues if:

* The user navigates away from the client view (the client will remain active and connecting in the background, ready for the user to resume)
* The user has multiple connections open within the same tab (they will fight over handling of `beforeunload`). Depending on whether the event handler is recognized as the same function by the event stack, this would result either in the connections removing each other's handlers or the connections adding multiple copies of the event handler.

I think this instead needs to be at the root level, outside the handling of a connection. If there were a _single_ `beforeunload` event handler that is never removed, that handler could:

1. Check whether any clients are present in the overall set of managed clients (this is exposed by the `guacClientManager` service).
2. If there are clients, prompt. If there aren't, don't.

In fact, `guacClientManager` already handles `unload`:

https://github.com/apache/guacamole-client/blob/0091bb1aea14c567c8166f0ed8eadf7c31b6bd6e/guacamole/src/main/webapp/app/client/services/guacClientManager.js#L164-L165

Perhaps it would make sense for it to handle `beforeunload` as well, and for that to be the sole instance of this event handler?

mike-jumper (changes requested):

mike-jumper:
@phreakocious - FYI: looks like you inadvertently pushed a bunch of commits from `master`. Might need to rebase.

phreakocious:
> @phreakocious - FYI: looks like you inadvertently pushed a bunch of commits from `master`. Might need to rebase.

Thanks for the heads up.

necouchman:
@phreakocious: One of your commits still lacks the `GUACAMOLE-1291:` prefix on it. Aside from that, not sure if @mike-jumper has any other concerns on this one...

phreakocious:
> @phreakocious: One of your commits still lacks the `GUACAMOLE-1292:` prefix on it. Aside from that, not sure if @mike-jumper has any other concerns on this one...

It will likely be awhile before I'll have time to address the concerns raised in the last review.  It did work to achieve the intended goal when I last tested it, but there could certainly be edge cases.

necouchman:
@phreakocious Thanks for letting us know. If it gets addressed prior to 1.4.0 release, great, otherwise we'll push it to the next release.

webtroter:
Hi,
I'm no developer, but I'm looking forward to that safety net. I already accidentally closed guacamole five times in the last hour.

I'm here to see if that issue can be reactivated.

Cyberes:
Just chiming in and voicing my support for this PR.

necouchman:
Thanks @Cyberes and @webtroter - at this point we're just waiting for the contributor to finish fixing up the requested items, or someone else to come along and for the changes and finish up the work.

PR #595, opened 2021-02-24
GUACAMOLE-1298: Automatically limit HTTP request size.

mike-jumper:
This change adds a new configuration property, `api-max-request-size`, which defines the maximum size allowed for REST requests. The property can be set to "0" to remove all limits. By default, the bodies of REST requests will be limited to 2 MB (2097152 bytes).

If a REST method should _never_ be limited (regardless of `api-max-request-size`), it can be annotated with `@RequestSizeFilter.DoNotLimit`. This should only be done for requests that are processed as streams, such as the handler for inbound file transfers (the only location where this annotation is currently used).


guacamole/src/main/java/org/apache/guacamole/GuacamoleServletContextListener.java
```diff
@@ -78,33 +121,47 @@ public void contextInitialized(ServletContextEvent servletContextEvent) {
             throw new RuntimeException(e);
         }
 
+        // NOTE: The superclass implementation of contextInitialized() is
+        // expected invoke getInjector(), hence the need to call AFTER setting
```
necouchman, line 125: *to invoke

guacamole/src/main/java/org/apache/guacamole/rest/RequestSizeFilter.java
```diff
@@ -0,0 +1,108 @@
...
+     */
+    @Retention(RetentionPolicy.RUNTIME)
+    @Target(ElementType.METHOD)
+    public static @interface DoNotLimit {}
+
+    /**
+     * The default maximum number of bytes to accept within the entity body of
+     * any particular REST request.
+     */
+    private final long DEFAULT_MAX_REQUEST_SIZE = 2097152;
```
necouchman, line 58: Is there a reason why this size was chosen? I've not issue with it, just curious as to the rationale.

necouchman (changes requested):
Overall looks fine to me - one nitpick and one question.

guacamole/src/main/java/org/apache/guacamole/rest/RequestSizeFilter.java
mike-jumper: My initial thinking was that something on the order of a few MB would be a reasonable default covering most cases, and my gut suggested 2 MB. I checked against the defaults of similar components, particularly Tomcat, and found that this matched up:

* The default `maxPostSize` of Tomcat's HTTP connector is 2 MB.
* Nginx' default request size limit is 1 MB.

guacamole/src/main/java/org/apache/guacamole/GuacamoleServletContextListener.java
mike-jumper: Oops. Fixed via rebase.

guacamole/src/main/java/org/apache/guacamole/rest/RequestSizeFilter.java
necouchman: Okay. Is this something that is going to be configurable? I think the 1MB limit in Nginx tends to be problematic for large file uploads into Guacamole Client - I suspect this is also going to limit those uploads??

mike-jumper: > Is this something that is going to be configurable?

Yep - with the `api-max-request-size` property that is part of these changes.

> I think the 1MB limit in Nginx tends to be problematic for large file uploads into Guacamole Client - I suspect this is also going to limit those uploads??

Nope, this will specifically not affect uploads, which are annotated with `@RequestSizeFilter.DoNotLimit`. When a user configures their reverse proxy to not limit requests, file uploads of any size will work, while all other requests would still be limited by Guacamole.

(See PR description 😉)

necouchman: Ah, shoot, sorry about that - missed both those things!

necouchman: Wow, just a few lines below this, too - and I even recall reviewing that last time around...

necouchman (approved):

PR #609, opened 2021-05-01
GUACAMOLE-1339: Add missing Spanish translations for client 1.3.0

djrm05:
Adding Spanish translations for new options available from last versions. Also improved the translation for some of the existing text.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/resources/translations/es.json
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/resources/translations/es.json
@@ -3,15 +3,15 @@
     "LOGIN" : {
 
         "ERROR_PASSWORD_BLANK"    : "@:APP.ERROR_PASSWORD_BLANK",
-        "ERROR_PASSWORD_SAME"     : "La nueva contraseña debe ser diferente a la expirada.",
+        "ERROR_PASSWORD_SAME"     : "La nueva contraseña debe ser diferente a la que ha expirado.",
         "ERROR_PASSWORD_MISMATCH" : "@:APP.ERROR_PASSWORD_MISMATCH",
         "ERROR_NOT_VALID"         : "Esta cuenta de usuario no es válida actualmente.",
-        "ERROR_NOT_ACCESSIBLE"    : "Actualmente no se permite el acceso a esta cuenta. Por favor intente de nuevo mas tarde.",
+        "ERROR_NOT_ACCESSIBLE"    : "Actualmente no se permite el acceso a esta cuenta. Por favor, inténtelo de nuevo mas tarde.",
 
-        "INFO_PASSWORD_EXPIRED" : "Su contraseña ha expirado y debe renovarla. Por favor introduzca una nueva contraseña para continuar.",
+        "INFO_PASSWORD_EXPIRED" : "Su contraseña ha expirado y debe renovarla. Por favor, introduzca una nueva contraseña para continuar.",
 
         "FIELD_HEADER_NEW_PASSWORD"         : "Nueva contraseña",
-        "FIELD_HEADER_CONFIRM_NEW_PASSWORD" : "Confirmar nueva contraseña"
+        "FIELD_HEADER_CONFIRM_NEW_PASSWORD" : "Confirmar nuena contraseña"
 
     },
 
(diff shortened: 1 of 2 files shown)


djrm05:
@mike-jumper, question. I have added the section "USER_GROUP_ATTRIBUTES" to the file "extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/resources/translations/es.json" as per this commit, however looking to the New Group screen, I still seeing GROUP RESTRICTIONS and Disabled in English... is this text comming from other file? Im unable to find any other than the above.... thx

mike-jumper:
Nope, that's the correct file. Have you tried clearing browser cache?

djrm05:
thanks @mike-jumper . This was due to a wrong version of guacamole-auth-jdbc-mysql-1.3.0.jar in my test env. So let me know if there is any other thing required on this pull request...

guacamole/src/main/webapp/translations/es.json
```diff
@@ -260,16 +282,16 @@
...
 
-        "DIALOG_HEADER_CONFIRM_DELETE" : "Eliminar perfil de compartir",
+        "DIALOG_HEADER_CONFIRM_DELETE" : "Eliminar perfil compartido",
         "DIALOG_HEADER_ERROR"          : "@:APP.DIALOG_HEADER_ERROR",
 
         "FIELD_HEADER_NAME"               : "Nombre:",
         "FIELD_HEADER_PRIMARY_CONNECTION" : "Conexión primaria:",
 
-        "SECTION_HEADER_EDIT_SHARING_PROFILE" : "Editar perfil de compartir",
+        "SECTION_HEADER_EDIT_SHARING_PROFILE" : "Editar perfil compartido",
```
mike-jumper, line 291: The English name for this Guacamole concept, "sharing profile", refers to a set of configuration parameters that describe how the related connection can be shared. It is a profile (set of parameters) that describes sharing (the type of action performed by the user on the connection).

I'm not a Spanish speaker, but as far as I can tell, "compartido" is an adjective that means "shared". In English, "shared profile" would be incorrect, as it's not the profile itself that is shared.

For example:

* "_Sharing_ profile" - A profile describing how sharing may be performed (correct).
* "_Shared_ profile" - A profile that is shared with others (incorrect).

Does the new text have the correct meaning?

djrm05: Thanks for the explanation of sharing profile. I have reverted this to "perfil de compartir" as "perfil compartido" is not the right one for this meaning...  Already updated

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/resources/translations/es.json
```diff
@@ -72,34 +72,42 @@
...
         "ERROR_REQUIRES_NON_ALNUM"     : "Las contraseñas deben contener al menos un símbolo.",
-        "ERROR_REUSED"                 : "Esta contraseña ya se ha usado. Por favor no vuelva a usar ninguna de las  {HISTORY_SIZE} {HISTORY_SIZE, plural, one{password} other{passwords}} anteriores.",
-        "ERROR_TOO_SHORT"              : "Las contraseñas deben ser al menos {LENGTH} {LENGTH, plural, one{character} other{characters}} de largo.",
-        "ERROR_TOO_YOUNG"              : "La contraseña de esta cuenta ya ha sido renovada. Por favor espere al menos {WAIT} mas {WAIT, plural, one{day} other{days}} antes de cambiarla de nuevo."
+        "ERROR_REUSED"                 : "Esta contraseña ya se ha utilizado. Por favor, no vuelva a usar ninguna de las anteriores {HISTORY_SIZE} {HISTORY_SIZE, plural, one{password} other{passwords}}.",
+        "ERROR_TOO_SHORT"              : "Las contraseñas deben ser como mínimo {LENGTH} {LENGTH, plural, one{character} other{characters}} de largo.",
+        "ERROR_TOO_YOUNG"              : "La contraseña de esta cuenta ya se ha cambiado. Espere como mínimo {WAIT} mas {WAIT, plural, one{day} other{days}} antes de volverla a cambiar."
```
mike-jumper, line 86: The pluralization rules in these translations will say "0 days", "1 day", "2 days", etc., with "day" and "days" being in English. Is this correct? (Same for "character" and "password").

djrm05:
Thanks @mike-jumper . I thought this was just a function that got the translation from others. I was really stuck with the translation of "ERROR REUSED" because in spanish the plural is not really just change 1 word in the sentence but some words need to be changed in order to have a right one. So I have modified it as you can see in the last update and for me looks great in Spanish, but as I sais it is bit different than the standard logic. Pls let me know if this cause any issue from your side.

mike-jumper:
Thanks, @djrm05! There's no issue with a more complex structure. What you've done is exactly what should be done and exactly why a pluralization library is used.

mike-jumper (approved):

PR #611, opened 2021-05-24
GUACAMOLE-1048: Support server control commands during handshake

mike-jumper:
This change builds off #500, contributed by @holograph, refactoring the additional "error" and "disconnect" instruction handling to:

* Leverage a new `toException()` function on `GuacamoleStatus` that converts a status code into its corresponding exception.
* Throw `GuacamoleConnectionClosedException` for disconnects (rather than `GuacamoleServerException`).

No new exception class is defined in this case, as the existing equivalency between status codes and corresponding exceptions is used instead.

@holograph, I understand you're no longer working at the company that needed and proposed the original changes, but feel free to look over this and let us know what you think.


guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
```diff
@@ -59,17 +68,87 @@
...
+        else {
+
+            // Parse the status code from the received error instruction
+            int statusCode;
+            try {
+                statusCode = Integer.parseInt(args.get(1));
+            }
+            catch (NumberFormatException e) {
+                logger.debug("Received \"error\" instruction with non-numeric status code.", e);
+                return;
```
holograph, line 105: Are you sure ignoring the instruction is the right thing to do here? This arguably represents a completely corrupt connection that should be aborted outright. Just my opinion, of course...

holograph (approved):
One concern around illegal `error` instructions (with non-numeric error codes), otherwise LGTM

guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
holograph: This should probably be included in the release notes though (or added explicitly to the exception signature) to reflect the API refinement.

mike-jumper: Good catch! That's copy-pasta. That `return` definitely should not be there.

mike-jumper: Corrected via refactor and rebase.

guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
```diff
@@ -59,17 +68,83 @@
...
+        // to be server errors
+        GuacamoleStatus status = GuacamoleStatus.SERVER_ERROR;
+
+        // Parse human-readable message from "error" instruction, warning if no
+        // message was given
+        List<String> args = instruction.getArgs();
+        if (args.size() >= 1)
+            message = args.get(0);
+        else
+            logger.debug("Received \"error\" instruction no corresponding message.");
```
necouchman, line 100: instruction _with_ no corresponding message?

necouchman (changes requested):
Minor nitpicks, but otherwise all is good with me.

@holograph Any further comments from you?

holograph (approved):
Nope, LGTM (except the nit, which is, well, a nit :-))

guacamole-common/src/main/java/org/apache/guacamole/protocol/ConfiguredGuacamoleSocket.java
mike-jumper: Whoops. I've updated the message to be less caveman and rebased.

PR #615, opened 2021-05-31
GUACAMOLE-1322: Include SAML Plugin in Docker Image

petzsch:
Addressing the issue of missing SAML Plugin in Docker Image.

Please include in next release. Thank You!

### DIFF
--- a/guacamole-docker/bin/start.sh
+++ b/guacamole-docker/bin/start.sh
@@ -774,7 +774,7 @@ END
     set_optional_property "saml-debug"                       "$SAML_DEBUG"
     set_optional_property "saml-compress-request"            "$SAML_COMPRESS_REQUEST"
     set_optional_property "saml-compress-response"           "$SAML_COMPRESS_RESPONSE"
-    set_optional_property "saml-group-attribute"             "SAML_GROUP_ATTRIBUTE"
+    set_optional_property "saml-group-attribute"             "$SAML_GROUP_ATTRIBUTE"
 
     # Add required .jar files to GUACAMOLE_EXT
     # "1-{}" make it sorted as a first provider (only authentication)
@@ -1101,4 +1101,3 @@ fi
 #
 
 start_guacamole
-
--- a/guacamole-docker/bin/start.sh
+++ b/guacamole-docker/bin/start.sh
@@ -744,6 +744,7 @@ END
 
 }
 
+##
 ## Adds properties to guacamole.properties which select the SAML
 ## authentication provider, and configure it to connect to the specified SAML
 ## provider.
@@ -756,7 +757,7 @@ associate_saml() {
         cat <<END
 FATAL: Missing required environment variables
 -------------------------------------------------------------------------------
-If using an openid authentication, you must provide each of the following
+If using a SAML authentication, you must provide each of the following
 environment variables:
 
     SAML_IDP_METADATA_URL           The URI of the XML metadata file that from the SAML Identity
--- a/guacamole-docker/bin/start.sh
+++ b/guacamole-docker/bin/start.sh
@@ -752,22 +752,33 @@ END
 associate_saml() {
 
     # Verify required parameters are present
-    if [ -z "$SAML_IDP_METADATA_URL" ]
+    if [ -z "$SAML_IDP_METADATA_URL" ] && \
+       [ -z "$SAML_ENTITY_ID" -o -z "$SAML_CALLBACK_URL" ]
     then
         cat <<END
 FATAL: Missing required environment variables
 -------------------------------------------------------------------------------
-If using a SAML authentication, you must provide each of the following
-environment variables:
+If using a SAML authentication, you must provide either SAML_IDP_METADATA_URL
+or both SAML_ENTITY_ID and SAML_CALLBACK_URL environment variables:
 
-    SAML_IDP_METADATA_URL           The URI of the XML metadata file that from the SAML Identity
-                                    Provider
+    SAML_IDP_METADATA_URL   The URI of the XML metadata file that from the SAML Identity
+                            Provider that contains all of the information the SAML
+                            extension needs in order to know how to authenticate with
+                            the IdP. This URI can either be a remote server (e.g. https://)
+                            or a local file on the filesystem (e.g. file://).
+
+    SAML_ENTITY_ID          The entity ID of the Guacamole SAML client, which is
+                            generally the URL of the Guacamole server
+
+    SAML_CALLBACK_URL       The URL that the IdP will use once authentication has
+                            succeeded to return to the Guacamole web application and
+                            provide the authentication details to the SAML extension.
 END
         exit 1;
     fi
 
     # Update config file
-    set_property          "saml-idp-metadata-url"            "$SAML_IDP_METADATA_URL"
+    set_optional_property "saml-idp-metadata-url"            "$SAML_IDP_METADATA_URL"
     set_optional_property "saml-idp-url"                     "$SAML_IDP_URL"
     set_optional_property "saml-entity-id"                   "$SAML_ENTITY_ID"
     set_optional_property "saml-callback-url"                "$SAML_CALLBACK_URL"
--- a/guacamole-docker/bin/start.sh
+++ b/guacamole-docker/bin/start.sh
@@ -769,12 +769,12 @@ END
     set_property          "saml-idp-metadata-url"            "$SAML_IDP_METADATA_URL"
     set_optional_property "saml-idp-url"                     "$SAML_IDP_URL"
     set_optional_property "saml-entity-id"                   "$SAML_ENTITY_ID"
-	set_optional_property "saml-callback-url"                "$SAML_CALLBACK_URL"
-	set_optional_property "saml-strict"                      "$SAML_STRICT"
-	set_optional_property "saml-debug"                       "$SAML_DEBUG"
-	set_optional_property "saml-compress-request"            "$SAML_COMPRESS_REQUEST"
-	set_optional_property "saml-compress-response"           "$SAML_COMPRESS_RESPONSE"
-	set_optional_property "saml-group-attribute"             "SAML_GROUP_ATTRIBUTE"
+    set_optional_property "saml-callback-url"                "$SAML_CALLBACK_URL"
+    set_optional_property "saml-strict"                      "$SAML_STRICT"
+    set_optional_property "saml-debug"                       "$SAML_DEBUG"
+    set_optional_property "saml-compress-request"            "$SAML_COMPRESS_REQUEST"
+    set_optional_property "saml-compress-response"           "$SAML_COMPRESS_RESPONSE"
+    set_optional_property "saml-group-attribute"             "SAML_GROUP_ATTRIBUTE"
 
     # Add required .jar files to GUACAMOLE_EXT
     # "1-{}" make it sorted as a first provider (only authentication)
--- a/guacamole-docker/bin/start.sh
+++ b/guacamole-docker/bin/start.sh
@@ -744,6 +744,46 @@ END
 
 }
 
+## Adds properties to guacamole.properties which select the SAML
+## authentication provider, and configure it to connect to the specified SAML
+## provider.
+##
+associate_saml() {
+
+    # Verify required parameters are present
+    if [ -z "$SAML_IDP_METADATA_URL" ]
+    then
+        cat <<END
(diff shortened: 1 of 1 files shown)


guacamole-docker/bin/start.sh
```diff
@@ -594,6 +594,46 @@ END
...
+                                    Provider
+END
+        exit 1;
+    fi
+
+    # Update config file
+    set_property          "saml-idp-metadata-url"            "$SAML_IDP_METADATA_URL"
+    set_optional_property "saml-idp-url"                     "$SAML_IDP_URL"
+    set_optional_property "saml-entity-id"                   "$SAML_ENTITY_ID"
+	set_optional_property "saml-callback-url"                "$SAML_CALLBACK_URL"
```
knacktim, line 622: Spacing is slightly off from here down.

petzsch:
added another commit to fix the whitespace issue hopefully.

knacktim:
Code looks good!  You need to re-do all of your commits to have the issue tag at the front, like this: `GUACAMOLE-1322: fixed whitespaces`

petzsch:
I wasn't aware of that. I assume the commits would get squashed on merge anyway.
I've done a rebase of the last 3 commits. Hopefully that was the intended solution.

mike-jumper:
> You need to re-do all of your commits to have the issue tag at the front, like this: `GUACAMOLE-1322: fixed whitespaces`

Also, _please_ do not use a commit message like "fixed whitespaces". There just isn't any context there. Like a code comment, a good commit message should describe the nature of the change, not the literal change:

For example, after seeing your commit, I still had no idea what whitespace issue you were referring to or why anything needed to be fixed. I looked at the content of the commit itself and was able to see for myself, but that sort of information should already be apparent from the message.

With something like:

> GUACAMOLE-1322: fixed whitespaces

A developer looking through the git history is left wondering what happened. If something is wrong with the content of your change, the developer goes to the relevant commit message to understand your reasoning for that change, and here that would be absent.

On the other hand, something like:

> GUACAMOLE-1322: Correct indentation of SAML property assignments to match established code style.

is very clear. If (somehow) the change didn't achieve what it aimed at achieving, that would be apparent even weeks to months down the line.

No commit or change is too trivial to warrant a good message.

petzsch:
@mike-jumper I agree with your arguments and am always willing to improve. Before I do another rebase and alter my "fixed whitespaces" commit message, are the other ones I used somewhat OK or are they still within an acceptable range?

mike-jumper:
> @mike-jumper I agree with your arguments and am always willing to improve. Before I do another rebase and alter my "fixed whitespaces" commit message, are the other ones I used somewhat OK (are they still within an acceptable range)?

The other messages are better, for sure. If you'd like to improve those messages:

* Avoid shorthand like "ext" and "env" for "extension" and "environment variable". It's unnecessarily obfuscated for something that's intended to be documentation.
* Avoid describing what you're modifying by its filename alone - git already tells us this, and someone reading through the git history may not be as familiar with the purpose of each file. Describing things at a high level is better than saying "Update something.foo include BAR cfg"

For example, consider:

> GUACAMOLE-1322: Update build-guacamole.sh - include saml ext.
> GUACAMOLE-1322: Update start.sh include SAML env

vs.

> GUACAMOLE-1322: Build and include SAML authentication extension within Docker image.
> GUACAMOLE-1322: Add Docker environment variables for configuring SAML.

guacamole-docker/bin/start.sh
```diff
@@ -594,6 +594,46 @@ END
...
+    # Update config file
+    set_property          "saml-idp-metadata-url"            "$SAML_IDP_METADATA_URL"
+    set_optional_property "saml-idp-url"                     "$SAML_IDP_URL"
+    set_optional_property "saml-entity-id"                   "$SAML_ENTITY_ID"
+    set_optional_property "saml-callback-url"                "$SAML_CALLBACK_URL"
+    set_optional_property "saml-strict"                      "$SAML_STRICT"
+    set_optional_property "saml-debug"                       "$SAML_DEBUG"
+    set_optional_property "saml-compress-request"            "$SAML_COMPRESS_REQUEST"
+    set_optional_property "saml-compress-response"           "$SAML_COMPRESS_RESPONSE"
+    set_optional_property "saml-group-attribute"             "SAML_GROUP_ATTRIBUTE"
```
petzsch, line 627: should be:

```suggestion
    set_optional_property "saml-group-attribute"             "$SAML_GROUP_ATTRIBUTE"
```

petzsch (commented):
stupid typo: sorry for the hassle.

guacamole-docker/bin/start.sh
knacktim: Nice catch!

petzsch:
I've just been testing the changes and for some reason the .jar file for saml is not present in the docker image:

> guacamole@aaf35270b1e9:/opt/guacamole/saml$ ls -al
total 8
drwxr-xr-x 2 root root 4096 Jun  3 09:17 .
drwxr-xr-x 1 root root 4096 Jun  3 09:17 ..

If someone spots the problem, any help would be appreciated.
Will update, if I find it myself.

**Until that's solved: This pull request is not ready to be merged!**

guacamole-docker/bin/build-guacamole.sh
```diff
@@ -165,6 +165,21 @@ if [ -f extensions/guacamole-auth-duo/target/*.tar.gz ]; then
...
 fi
 
+#
+# Copy Saml auth extension if it was built
+#
+
+if [ -f extensions/guacamole-auth-saml/target/*.tar.gz ]; then
+    mkdir -p "$DESTINATION/saml"
+    tar -xzf extensions/guacamole-auth-saml/target/*.tar.gz \
+        -C "$DESTINATION/duo/"                             \
```
knacktim, line 175: > I've just been testing the changes and for some reason the .jar file for saml is not present in the docker image:

You are copying it to the wrong directory.  Copy paste error from the one above it ;)

petzsch:
```suggestion
        -C "$DESTINATION/saml/"                             \
```

petzsch (commented):
this should fix it.

petzsch:
Commit Messages have been corrected as requested. Please continue with the review. I'll remove the Draft status.

The docker image I've build for my usecase seems to run. Still working on getting the SAML integration with my coportate IdP to work.

Hopefully the [documentation](https://guacamole.apache.org/doc/gug/saml-auth.html) is right about this:

> Often the metadata file contains most of the required properties for SAML authentication and the other parameters are not required.

petzsch:
From my side the changes worked for my deployment together with the IdP that my customer uses. If any more changes are needed, please let me know. Otherwise a merge before the next release would be appreciated. 🥇

guacamole-docker/bin/start.sh
```diff
@@ -594,6 +594,46 @@ END
...
+##
+associate_saml() {
+
+    # Verify required parameters are present
+    if [ -z "$SAML_IDP_METADATA_URL" ]
+    then
+        cat <<END
+FATAL: Missing required environment variables
+-------------------------------------------------------------------------------
+If using an openid authentication, you must provide each of the following
```
mike-jumper, line 609: Oops: "SAML", not "OpenID".

guacamole-docker/bin/start.sh
```diff
@@ -594,6 +594,46 @@ END
...
+environment variables:
+
+    SAML_IDP_METADATA_URL           The URI of the XML metadata file that from the SAML Identity
+                                    Provider
+END
+        exit 1;
+    fi
+
+    # Update config file
+    set_property          "saml-idp-metadata-url"            "$SAML_IDP_METADATA_URL"
```
mike-jumper, line 619: I think this has to be optional, as well, unfortunately. Only the SAML support itself is going to be able to accurately sanity check these variables.

It's possible for the `saml-idp-metadata-url` property to be omitted, and the required values instead provided through individual properties (like `saml-entity-id` and `saml-callback-url`), and it's also possible to specify `saml-idp-metadata-url` _and_ individual properties to make up for any information not included in the XML pointed to by that URL.

guacamole-docker/bin/start.sh
```diff
@@ -594,6 +594,46 @@ END
 
 }
 
+## Adds properties to guacamole.properties which select the SAML
+## authentication provider, and configure it to connect to the specified SAML
+## provider.
+##
```
mike-jumper, line 600: Please correct the style of this comment to match:

```
##
## This somewhat JavaDoc-like format that we've adopted for documenting shell
## script functions.
##
```

rather than:

```
## This, which is missing the leading JavaDoc-like "opening" blank line
##
```

You're just missing that initial empty `##` line.

guacamole-docker/bin/start.sh
```diff
@@ -830,6 +870,12 @@ if [ -n "$OPENID_AUTHORIZATION_ENDPOINT" ]; then
     INSTALLED_AUTH="$INSTALLED_AUTH openid"
 fi
 
+# Use SAML if specified
+if [ -n "$SAML_IDP_METADATA_URL" ]; then
```
mike-jumper, line 874: This check will need to be more complicated, since the `saml-idp-metadata-url` property can be omitted entirely so long as the required information is given with other properties.

Overall, the user needs either `saml-idp-metadata-url`, or `saml-entity-id` and `saml-callback-url`, or a combination of these if the metadata XML is incomplete.

mike-jumper (changes requested):

guacamole-docker/bin/start.sh
petzsch: > [...]if the metadata XML is incomplete

That will be hard to check from within a bash script as I would need to parse the content of saml-idp-metadata-url to check for what is missing and if it is present in the other variables.

Can't really say when I will get to this. If someone else wants to make the needed changes and needs permissions on my fork, simply let me know.

mike-jumper: I wouldn't recommend making things that complex or attempting to duplicate the logic that the extension itself will be using.

I think it would be far better to just relax the logic a bit and consider the presence of any of those environment variables sufficient to enable SAML, rather than actually parse the XML.

petzsch:
> I still can have a look if you give perms 😅

I just added you to the project.

jbpaux:
I've implemented the requested changes :)

guacamole-docker/bin/start.sh
jbpaux: fixed :)

guacamole-docker/bin/start.sh
jbpaux: done

guacamole-docker/bin/start.sh
jbpaux: I've implemented the logic.

petzsch:
is this still being monitored?

mike-jumper:
Yep, I just got real busy toward the end of last year. I'll try to look over this later today.

guacamole-docker/bin/build-guacamole.sh
```diff
@@ -165,6 +165,21 @@ if [ -f extensions/guacamole-auth-duo/target/*.tar.gz ]; then
...
+    mkdir -p "$DESTINATION/saml"
+    tar -xzf extensions/guacamole-auth-saml/target/*.tar.gz \
+        -C "$DESTINATION/saml/"                             \
+        --wildcards                                        \
+        --no-anchored                                      \
+        --no-wildcards-match-slash                         \
+        --strip-components=1                               \
+        "*.jar"
+fi
+
```
mike-jumper, line 182: I believe this may no longer be necessary due to the restructuring of the SSO extensions (CAS, OpenID, and SAML) into a single .tar.gz as of 1.4.0:

https://github.com/apache/guacamole-client/blob/aa99b4bc8a511a0f96466d16de2c88af9befea83/guacamole-docker/bin/build-guacamole.sh#L113-L122

Can you rebase and recheck against the latest?

JackWolfard:
What's the status of this PR? Will it be included as a feature in 1.5.0?

jbpaux:
I've rebased on master and removed commits dealing with adding the saml jar since it's already included with all other sso plugins. should be good @mike-jumper 🤞

mike-jumper (approved):

trident25:
I've found a bug in this pull request - captured here: https://issues.apache.org/jira/browse/GUACAMOLE-1570

I think the logic when deciding to load the SAML extension is incomplete. @petzsch

PR #616, opened 2021-06-02
GUACAMOLE-773: Update all JavaScript dependencies to latest versions.

mike-jumper:
This change updates absolutely all JavaScript dependencies to the latest. Unfortunately, with additional transitive dependencies, doing this required migrating to npm and webpack for the frontend part of the building/bundling process. WebJars was just too eager to include transitive dependencies that aren't truly used, whereas webpack will magically determine such things.

For example, the following dependency tree is pulled in for the updated WebJars:

```
+- org.webjars:webjars-servlet-2.x:jar:1.5:compile
+- org.webjars.npm:blob-polyfill:jar:5.0.20210201:runtime
+- org.webjars.npm:file-saver:jar:2.0.5:runtime
+- org.webjars.npm:lodash:jar:4.17.21:runtime
+- org.webjars.npm:jquery:jar:3.6.0:runtime
+- org.webjars.npm:angular:jar:1.8.2:runtime
+- org.webjars.npm:angular-route:jar:1.8.2:runtime
+- org.webjars.npm:angular-touch:jar:1.8.2:runtime
+- org.webjars.bower:angular-module-shim:jar:0.0.4:runtime
+- org.webjars.npm:angular-translate:jar:2.18.4:runtime
+- org.webjars.npm:angular-translate-interpolation-messageformat:jar:2.18.4:runtime
|  \- org.webjars.npm:messageformat:jar:1.0.2:runtime (version selected from constraint [1.0.2,1.1))
|     +- org.webjars.npm:nopt:jar:3.0.6:runtime (version selected from constraint [3.0.6,3.1))
|     |  \- org.webjars.npm:abbrev:jar:1.1.1:runtime (version selected from constraint [1,2)) 
|     +- org.webjars.npm:make-plural:jar:3.0.6:runtime (version selected from constraint [3.0.6,3.1))
|     |  \- org.webjars.npm:minimist:jar:1.2.5:runtime (version selected from constraint [1.2.0,2))
|     +- org.webjars.npm:reserved-words:jar:0.1.2:runtime (version selected from constraint [0.1.1,0.2))
|     +- org.webjars.npm:glob:jar:7.0.6:runtime (version selected from constraint [7.0.6,7.1))
|     |  +- org.webjars.npm:once:jar:1.4.0:runtime (version selected from constraint [1.3.0,2))
|     |  |  \- org.webjars.npm:wrappy:jar:1.0.2:runtime (version selected from constraint [1,2)) 
|     |  +- org.webjars.npm:inflight:jar:1.0.6:runtime (version selected from constraint [1.0.4,2))
|     |  +- org.webjars.npm:inherits:jar:2.0.4:runtime (version selected from constraint [2,3)) 
|     |  +- org.webjars.npm:minimatch:jar:3.0.4:runtime (version selected from constraint [3.0.2,4))
|     |  |  \- org.webjars.npm:brace-expansion:jar:1.1.11:runtime (version selected from constraint [1.1.7,2))
|     |  |     +- org.webjars.npm:balanced-match:jar:1.0.2:runtime (version selected from constraint [1.0.0,2))
|     |  |     \- org.webjars.npm:concat-map:jar:0.0.1:runtime
|     |  +- org.webjars.npm:fs.realpath:jar:1.0.0:runtime (version selected from constraint [1.0.0,2))
|     |  \- org.webjars.npm:path-is-absolute:jar:1.0.1:runtime (version selected from constraint [1.0.0,2))
|     \- org.webjars.npm:messageformat-parser:jar:1.1.0:runtime (version selected from constraint [1.0.0,2))
+- org.webjars.npm:angular-translate-loader-static-files:jar:2.18.4:runtime
+- org.webjars.npm:datalist-polyfill:jar:1.24.4:runtime
+- org.webjars.npm:jstz:jar:2.1.1:compile
\- org.webjars.npm:simonwep__pickr:jar:1.8.0:compile
   +- org.webjars.npm:nanopop:jar:2.1.0:compile (version selected from constraint [2.1.0,3))
   \- org.webjars.npm:core-js:jar:3.13.1:compile (version selected from constraint [3.8.0,4))
```

From npm and webpack, the list of dependencies makes much more sense:

```
@simonwep/pickr:1.8.1
angular-route:1.8.2
angular-touch:1.8.2
angular-translate-interpolation-messageformat:2.18.4
angular-translate-loader-static-files:2.18.4
angular-translate:2.18.4
angular:1.8.2
assert:1.5.0
blob-polyfill:5.0.20210201
css-loader:5.2.4
datalist-polyfill:1.24.4
file-saver:2.0.5
inherits:2.0.3
jquery:3.6.0
jstz:2.1.1
lodash:4.17.21
make-plural:3.0.6
messageformat-parser:1.1.0
messageformat:1.0.2
object-assign:4.1.1
process:0.11.10
reserved-words:0.1.2
util:0.11.1
webpack:4.46.0
```

With JavaScript dependencies no longer 100% coming from Maven, I had to write a webpack plugin to walk through the contents of the bundle, determine which dependencies are _actually still present in the bundle_, and dump a list of those dependencies to a file for automated license processing.

It's not as huge a change as it initially appears. Most of the modified files are renames (moving from `src/main/webapp/` to `src/main/frontend/src/`), and most of the lines are `package-lock.json`. Beware: just like the basilisk, `package-lock.json` is controlled by magic and looking upon it with your own eyes can be fatal.


mike-jumper:
So far, I'm seeing the following regressions:

* The pagination directive changed from this:

  Working pagination directive

  to this:

  Broken pagination directive

* The "Pickr" color picker is no longer detected as unavailable on IE10 and IE11. The service handling initialization of the Pickr library should be recognizing that things will not work and mark Pickr as unavailable, but this isn't happening.

We should also review and check against the 1.6.x to 1.7.x migration guide for AngularJS:

https://docs.angularjs.org/guide/migration#migrating-from-1-6-to-1-7

There are no applicable breaking changes for 1.7.x to 1.8.x, which affected only jqLite (we use full jQuery instead of this).

mike-jumper:
The issue with the pagination directive was due to a change in AngularJS 1.8.x and jQuery 3.5.0+ where the automatic XHTML -> HTML filtering has been removed. A self-closing tag like `<div/>` previously would have been translated to `<div></div>`, but now is instead given to the browser verbatim. The browser then interprets it as an unclosed `<div>`, leading to unexpected nesting of elements.

mike-jumper:
The issue with Pickr and IE10/IE11 was due to `Pickr.create()` no longer throwing an exception outright if browser support is missing. Instead, we must rely on the lack of an "init" event from the created Pickr object to detect whether Pickr can be used.

mike-jumper:
OK - I've gone through the [AngularJX 1.6 -> 1.7 migration guide](https://docs.angularjs.org/guide/migration#migrating-from-1-6-to-1-7) and the only applicable change is with the ngTouch module (angular-touch). We used to use this module to remove the ~300ms click delay on touch devices, but the new version ngTouch no longer does this.

This should still be OK - as noted in the migration guide, modern browsers and even IE10+ no longer apply that delay. With that no longer being an issue, and with support for working around the issue gone, there's no need to include ngTouch as a dependency.

mike-jumper:
I think this is now ready!

guacamole/src/main/frontend/package.json
```diff
@@ -0,0 +1,35 @@
+{
+    "private": true,
+    "scripts": {
+        "build": "webpack --progress"
+    },
+    "dependencies": {
+        "@simonwep/pickr": "1.8.1",
+        "angular": "1.8.2",
+        "angular-route": "1.8.2",
+        "angular-templatecache-webpack-plugin": "^1.0.1",
```
jmuehlner, line 10: Do we plan on having any policy on whether these dependencies should use the `^` operator or not? The `package-lock.json` ensures that builds will be repeatable, but I usually like to lock down the packages to specific versions by removing the `^` and other operators out of `package.json` as well.

I'm not really sure what the right call is here.

mike-jumper: Reading up on this a bit, it seems the best approach is to use `^` except where there's a specific reason not to:

* It's the default for npm and therefore the least surprising to anyone already well-steeped in that ecosystem.
* It avoids duplicating the duty of `package-lock.json`, which we are including in version control here following the _other_ established best practice for npm.

I definitely prefer the Maven approach of absolute reproducibility by default (with the Maven equivalent to `^` and `~` being considered bad practice and almost never used), but we're stuck with the npm approach here.

I'll update to 🥕s everywhere.

guacamole/src/main/frontend/src/app/index/indexModule.js
```diff
@@ -17,22 +17,37 @@
...
     'login',
     'manage',
     'navigation',
-    'ngRoute',
-    'ngTouch',
     'notification',
-    'pascalprecht.translate',
     'rest',
     'settings',
+
```
jmuehlner, line 44: Why is there a space before this last one?

guacamole/src/main/frontend/src/app/index/indexModule.js
```diff
@@ -17,22 +17,37 @@
  * under the License.
  */
 
+require('angular-module-shim.js');
```
jmuehlner, line 20: What are these imports doing up at the top here? I'm guessing these are just global runtime requirements?

guacamole/src/main/frontend/src/app/index/indexModule.js
```diff
@@ -17,22 +17,37 @@
...
+
+require('angular-translate-interpolation-messageformat');
+require('angular-translate-loader-static-files');
+
 /**
  * The module for the root of the application.
  */
 angular.module('index', [
+
+    require('angular-route'),
```
jmuehlner, line 31: Why are these here?

guacamole/src/main/frontend/src/app/index/indexModule.js
mike-jumper: I thought it made sense to visually separate things into three groups:

* AngularJS modules from external libraries
* AngularJS modules we define ourselves with actual code
* AngularJS modules that are generated at build time (just `templates-main`)

guacamole/src/main/frontend/src/app/index/indexModule.js
mike-jumper: Yep, exactly.

guacamole/src/main/frontend/plugins/dependency-list-plugin.js
```diff
@@ -0,0 +1,146 @@
...
+            const moduleCoords = {};
+
+            // Map each file used within any bundle built by the compiler to
+            // its corresponding NPM package, ignoring files that have no such
+            // package
+            stats.compilation.fileDependencies.forEach(file => {
+
+                // Locate NPM package corresponding to file dependency (there
+                // may not be one)
+                const moduleFinder = finder(file)
```
jmuehlner, line 117: No semicolon?

guacamole/src/main/frontend/src/app/index/indexModule.js
jmuehlner: Makes sense to me.

guacamole/src/main/frontend/src/app/index/indexModule.js
mike-jumper: These are the external AngularJS libraries that we depend on. It's a convention with AngularJS libraries within NPM to simply export the name of the relevant AngularJS module, so `require('angular-route')` here returns a JavaScript string that is the name of the AngularJS module providing routing.

guacamole/src/main/frontend/plugins/dependency-list-plugin.js
mike-jumper: 🙀 Oops

mike-jumper: Fixed via rebase.

jmuehlner:
Well, I think this all looks pretty reasonable to me. Let's do it.

PR #619, opened 2021-06-06
GUACAMOLE-538: Add support for system audit permission

necouchman:
This pull requests implements support for an "AUDIT" permission that allows a user to be assigned read-only permissions to active sessions and connection history. Marking it as a draft for the moment, and welcome any comments on it. I was going to try to figure out a way to allow the audit permission to get access to the user login history, but haven't quite figured that part out, yet.


necouchman:
@mike-jumper or @jmuehlner Thoughts on this - not necessarily details, but the approach, etc.? I don't want to spend a lot of time cleaning this up if there is resistance to the overall idea or if there are significant issues with the approach.

MariasStory:
Please, implement this feature. I did not check if this pull request works. Still the feature is needed. We need to know who is logged to which system.

necouchman:
@MariasStory : You don't need this pull request in order to be able to see who is logging in to the system or who is using the connections. This information is already available and logged in the GUI, and these changes do not adjust at all the amount of information logged or where it is logged.

The only thing that this does is set up a separate role that you can assign to specific users to be able to see that data without granting them admin access to the system.

MariasStory:
@necouchman : Thanks for the comment. I should have explained the issue in details.
In our team, we work on several systems using guacamole. As a normal user, you can see a number of users  logged in to specific system in the systems list, but there are no names, just a number. In settings you can see active sessions, but only your own.
Administrator can see all active sessions in the settings.
To collaborate properly, a regular user needs to know who is logged into the systems. I expected to be able to enable this permission for each role, but it seems not to be the case.
Having the role may solve the problem. Unless there is another way to do it.

necouchman:
@MariasStory: Yes, with the "AUDIT" permission added you would be able to assign that role to users or groups and they would be able to see details of who is using connections outside of just their own.

MariasStory:
@jmuehlner and @mike-jumper  Can you please review this pull request and get the feature in?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/schema/upgrade/upgrade-pre-1.6.0.sql
```diff
@@ -0,0 +1,26 @@
...
+-- under the License.
+--
+
+--
+-- Add new system-level audit permission
+--
+
+ALTER TYPE guacamole_system_permission_type
+    ADD VALUE 'AUDIT'
+    BEFORE 'ADMINISTER';
```
jmuehlner, line 26: Why before `ADMINISTER`? Alphabetically it should go after...

necouchman: Yeah, that's a fair point - I was probably thinking privilege level vs. alphabetic, but I can re-organize it to make sure it's consistent.

necouchman: Actually, looking at the other code, it's because that's how those enums tend to be organized:

https://github.com/apache/guacamole-client/blob/6493a2313bd976440fea9479742715040662ed74/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/schema/001-create-schema.sql#L53-L60

`ADMINISTER` is the last item in that list, and, when I added `AUDIT` to the creation script, I left it that way, so when I do the `ALTER TYPE`, here, in the upgrade script, I insert it in the same way that it would be for the creation script.

jmuehlner: Fair

necouchman: Now, that said, what I'm not sure is if the order it's added in has any implications for how the database stores and/or (re)indexes the existing values. So, if I add it before, does Postgres (in this case) have to go through and update references to it in tables that have that permission? Whereas if I just add it after, it becomes the next value and no updates need to be done?

MariasStory: Hi @jmuehlner and @mike-jumper, thanks for moving this topic.
Is there something that I can help with?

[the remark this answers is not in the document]
necouchman: Added commit that puts newlines at the end of the schema files.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/schema/upgrade/upgrade-pre-1.6.0.sql
```diff
@@ -0,0 +1,42 @@
...
+    'CREATE_USER_GROUP',
+    'AUDIT',
+    'ADMINISTER'
+);
+GO
+
+EXEC sp_bindrule
+    'guacamole_system_permission_list',
+    'guacamole_system_permission';
+GO
```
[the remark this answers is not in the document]
necouchman: Added commit that puts newlines at the end of the schema files.

mike-jumper (approved):

PR #620, opened 2021-06-14
GUACAMOLE-680: Do not immediately re-authenticate after logout.

mike-jumper:
As discussed in #573, this change adds an intermediate "you have been logged out" message that displays after the user manually logs out, allowing the user to _choose_ to re-authenticate when ready:

you-have-been-logged-out

This prevents the user from being immediately signed back in when SSO is involved.


knacktim:
Looks good to me, one question -- Does this mean for things like [GUACAMOLE-519](https://issues.apache.org/jira/browse/GUACAMOLE-519) that we would just use the following handler hook to do the necessary operations?
```javascript
$scope.on('guacLogout', <function stuff>)
```

guacamole/src/main/frontend/src/app/auth/service/authenticationService.js
```diff
@@ -252,6 +252,24 @@ angular.module('auth').factory('authenticationService', ['$injector',
 
     };
 
+    /**
+     * Makes a request to revoke an authentication token using the token REST
+     * API endpoint, returning a promise succeeds only if the token was
```
necouchman, line 257: _that_ succeeds?

necouchman (changes requested):
One minor nitpick, but overall LGTM.

mike-jumper:
> Looks good to me, one question -- Does this mean for things like [GUACAMOLE-519](https://issues.apache.org/jira/browse/GUACAMOLE-519) that we would just use the following handler hook to do the necessary operations?
> 
> ```js
> $scope.on('guacLogout', <function stuff>)
> ```

For anything that absolutely must be done client-side, yes. Otherwise, it would be better to rely on [the server-side `invalidate()` function of `UserContext`](http://guacamole.apache.org/doc/1.3.0/guacamole-ext/org/apache/guacamole/net/auth/UserContext.html#invalidate--).

guacamole/src/main/frontend/src/app/auth/service/authenticationService.js
mike-jumper: Oops.

mike-jumper: There's a second occurrence of this in the documentation of the function that this was copied from. I've corrected both in a fresh commit.

necouchman (approved):

PR #621, opened 2021-06-14
GUACAMOLE-1364: Add support for overriding extension priority without renaming files.

mike-jumper:
Currently, Guacamole extensions are always loaded in order of their filenames, and changing that order requires renaming those files. These changes add a new property, `extension-priority`, that allows the order of extensions to be overridden independently of filenames.

The format of the property is a comma-separated list of extension namespaces, with the special value "*" used to denote any extensions not explicitly listed. This allows the order to be overridden for only those extensions that the administrator decides _must_ be loaded first or last when present. It's not an error for an extension to be listed that isn't actually installed (the property just declares the desired relative order when any of the extensions listed are present), but a warning will be logged if an extension is listed more than once.

### Examples

Load SAML before all others:

```properties
extension-priority: saml
```

Load SAML and LDAP before all others:

```properties
extension-priority: saml, ldap
```

Load LDAP last:

```properties
extension-priority: *, ldap
```

Load MySQL first, LDAP last, and any others in order of filename:

```properties
extension-priority: mysql, *, ldap
```

### DIFF
--- a/extensions/guacamole-auth-cas/src/main/resources/guac-manifest.json
+++ b/extensions/guacamole-auth-cas/src/main/resources/guac-manifest.json
@@ -3,7 +3,7 @@
     "guacamoleVersion" : "1.3.0",
 
     "name"      : "CAS Authentication Extension",
-    "namespace" : "guac-cas",
+    "namespace" : "cas",
 
     "authProviders" : [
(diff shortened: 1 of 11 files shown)


necouchman (approved):
:shipit:

PR #622, opened 2021-06-21
GUACAMOLE-641: Correct regressions in custom pooled datasource behavior.

mike-jumper:
The `DynamicallyAuthenticatedDataSource` introduced via [GUACAMOLE-641](https://issues.apache.org/jira/browse/GUACAMOLE-641) does not correctly set driver-specific JDBC properties like `allowMultiQueries`, resulting in failures for queries that depend on those properties. Additionally, the `PooledDataSource` constructor in use does not initialize internal values used by `PooledDataSource` to verify connections being returned to the pool, resulting in an eternally empty pool and new connections being created for each query.

This change corrects the above by:

1. Adding support for the same named injectable values as mybatis-guice's `PooledDataSourceProvider`.
2. Manually invoking `forceCloseAll()` within the constructor, which will initialize the required internals for `PooledDataSource` sanity checks.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/DynamicallyAuthenticatedDataSource.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/DynamicallyAuthenticatedDataSource.java
@@ -24,6 +24,7 @@ import com.google.inject.Singleton;
 import com.google.inject.name.Named;
 import java.sql.Connection;
 import java.sql.SQLException;
+import java.util.Properties;
 import org.apache.guacamole.GuacamoleException;
 import org.apache.ibatis.datasource.pooled.PooledDataSource;
 import org.apache.ibatis.datasource.unpooled.UnpooledDataSource;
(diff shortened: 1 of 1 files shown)


extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/DynamicallyAuthenticatedDataSource.java
```diff
@@ -71,6 +80,77 @@ public Connection getConnection() throws SQLException {
...
+    @Override
+    @Inject(optional=true)
+    public void setPoolMaximumActiveConnections(
+            @Named("mybatis.pooled.maximumActiveConnections") int poolMaximumActiveConnections) {
+        super.setPoolMaximumActiveConnections(poolMaximumActiveConnections);
+    }
+
+    @Override
+    @Inject(optional=true)
+    public void setDriverProperties(@Named("JDBC.driverProperties") Properties driverProps) {
```
necouchman, line 140: Are all of the other properties - like time zone, SSL options, etc. - handled, here? Or do they also need to be specified in this list?

mike-jumper: They're all handled here, yes. The key things that need explicit injection are those mentioned in the `PooledDataSourceProvider` class from mybatis-guice:

https://github.com/mybatis/guice/blob/mybatis-guice-3.10/src/main/java/org/mybatis/guice/datasource/builtin/PooledDataSourceProvider.java

necouchman (approved):

PR #623, opened 2021-07-05
GUACAMOLE-724: Add support for displaying multiple connections as tiles.

mike-jumper:
This change adds a new grid layout of multiple connections that allows those connections to be visible within the same tab. The user can also select multiple visible connections to interact with each using the keyboard simultaneously.

The main entrypoint into these changes is the connection dropdown in the Guacamole menu, which now contains checkboxes:

Screenshot 2021-07-04 8 50 11 PM

If multiple boxes are checked, those connections will automatically be opened and tiled evenly:

Screenshot 2021-07-04 8 50 40 PM

Keyboard focus is assigned by clicking on a connection. By holding Shift or Ctrl while clicking, multiple connections can be focused, and each will then receive keyboard input simultaneously:

Screenshot 2021-07-04 8 51 14 PM

Focus also dictates the contents of the Guacamole menu, which will generally only show options specific to the currently-focused connection. If multiple connections are focused, only options that are applicable to all connections will be visible.

Clicking the "X" in the corner of any connection closes the connection, with others rearranging to fill the new space:

Screenshot 2021-07-04 8 51 37 PM

Part of these changes involved moving the client panel to a higher, common level, allowing connections to be monitored and switched while viewing other parts other Guacamole webapp:

Screenshot 2021-07-04 8 51 54 PM

As I also needed to add some new variations of existing graphics, I went ahead and migrated our old PNGs to SVG so that we can benefit from scaling and so that they can be an actual part of the source. All browsers currently supported by Guacamole support SVG, even IE10.


guacamole/src/main/frontend/src/app/client/directives/guacClientNotification.js
```diff
@@ -0,0 +1,431 @@
...
+         * @type {Notification|Object|Boolean}
+         */
+        $scope.status = false;
+
+        /**
+         * All client error codes handled and passed off for translation. Any error
+         * code not present in this list will be represented by the "DEFAULT"
+         * translation.
+         */
+        var CLIENT_ERRORS = {
```
jmuehlner, line 73: Have you considered the `Set` type? It's a lot less awkward than this.

guacamole/src/main/frontend/src/app/client/controllers/clientController.js
```diff
@@ -26,6 +26,7 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
     // Required types
     var ConnectionGroup    = $injector.get('ConnectionGroup');
     var ManagedClient      = $injector.get('ManagedClient');
+    var ManagedClientGroup = $injector.get('ManagedClientGroup');
```
jmuehlner, line 29: What do you think of the new `let` and `const` keywords? I'm certainly not asking you to rewrite this beast, but I find them to be clearer than using `var`. 

Plus we'll get compile-time checking for things that shouldn't be mutated.

guacamole/src/main/frontend/src/app/client/controllers/clientController.js
```diff
@@ -370,134 +384,58 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
...
-     */
-    var clipboardDataFromKey = {};
-
-    /*
-     * Check to see if all currently pressed keys are in the set of menu keys.
+     * @param {Guacamole.Keyboard} keyboard
+     *     The Guacamole.Keyboard object tracking the local keyboard state.
+     *
+     * @returns {boolean}
+     *     true of Ctrl+Alt+Shift has been pressed, false otherwise.
```
jmuehlner, line 394: Typo: "true of ..."

guacamole/src/main/frontend/src/app/client/types/ManagedClientGroup.js
```diff
@@ -0,0 +1,359 @@
...
+        if (!_.isArray(group))
+            group = _.map(group.clients, client => client.id);
+
+        return group.join('.');
+
+    };
+
+    /**
+     * Returns an array of client identifiers for all clients contained within
+     * the given ManagedClientGroup. Orcer of the identifiers is preserved
```
jmuehlner, line 134: Orcer?

guacamole/src/main/frontend/src/app/client/controllers/clientController.js
mike-jumper: Sure. I'll at least update the files that have been touched here.

guacamole/src/main/frontend/src/app/client/types/ManagedClientGroup.js
```diff
@@ -0,0 +1,359 @@
...
+        var relevantClients = _.filter(group.clients, client => client.clientProperties.focused);
+        if (!relevantClients.length)
+            relevantClients = group.clients;
+
+        return _.filter(relevantClients, (client => !!client.name)).map(client => client.name).join(', ') || '...';
+
+    };
+
+    /**
+     * A callback that is invoked for a ManagedClient within a ManagedClientGroup.
```
jmuehlner, line 234: What callback is this referring to?

jmuehlner: I guess it's referring to the callback invoked by `forEach` below?

guacamole/src/main/frontend/src/app/client/types/ManagedClientGroup.js
```diff
@@ -0,0 +1,359 @@
...
+
+        // Operate on cached copy of grid
+        var clientGrid = group._grid || (group._grid = []);
+
+        // Delete any rows in excess of the required size
+        clientGrid.splice(group.rows);
+
+        for (var row = 0; row < group.rows; row++) {
+
+            // Prefer to use existing column arrays, deleting any columnd in
```
jmuehlner, line 325: typo: 'columnd'

guacamole/src/main/frontend/src/app/client/types/ManagedClientGroup.js
mike-jumper: Yes. It's type documentation for the callback used by `forEach`.

guacamole/src/main/frontend/src/app/client/directives/guacClientNotification.js
mike-jumper: I haven't, no. Is that something that will get magically compiled down to essentially this? I know things like the arrow notation, `const`, etc. are compiled away during the build, allowing us to maintain IE10+ compatibility, but I'm not sure about full-fledged objects/types like `Set`.

mike-jumper: I'll just do a quick test refactoring this one instance to `Set` and see what code is emitted.

mike-jumper: No dice:

Screenshot 2021-07-07 9 38 21 PM

If we wanted to use `Set`, we'd need a polyfill.

mike-jumper: OK - migrated all directly touched code from `var` to `const`/`let` via rebase. I avoided making sweeping changes, updating only the code that is already within this diff.

guacamole/src/main/frontend/src/app/client/controllers/clientController.js
mike-jumper: OK - migrated all directly touched code from `var` to `const`/`let` via rebase. I avoided making sweeping changes, updating only the code that is already within this diff.

guacamole/src/main/frontend/src/app/client/controllers/clientController.js
mike-jumper: Corrected via rebase.

guacamole/src/main/frontend/src/app/client/types/ManagedClientGroup.js
mike-jumper: Corrected via rebase.

guacamole/src/main/frontend/src/app/client/types/ManagedClientGroup.js
mike-jumper: Corrected via rebase.

necouchman:
Thanks @jmuehlner! I was trying to find time to review it and haven't had a lot, unfortunately.

jmuehlner:
Well, I think all my concerns are now addressed. :shipit:

PR #626, opened 2021-07-21
GUACAMOLE-275: Automatically refresh browser cache after extension/application updates.

mike-jumper:
To avoid mismatches between cached and current resources, and to ensure extension JavaScript is automatically kept up-to-date, this change:

* Adds `Cache-Control` and `Pragma` headers to the `app.js`, `app.css`, and `index.html` reponses to ensure browsers will revalidate their cached copies of these resources going forward.
* Adds an automated JavaScript sanity check that compares the version of `app.js` against `index.html` and attempts to refresh the browser's cache if `index.html` is older.

### DIFF
--- a/guacamole/pom.xml
+++ b/guacamole/pom.xml
@@ -154,6 +154,7 @@
                             <excludes>
                                 <exclude>translations/*.json</exclude>
                                 <exclude>index.html</exclude>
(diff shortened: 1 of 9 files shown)


jmuehlner (approved):
LGTM

PR #627, opened 2021-07-21
GUACAMOLE-1384: Ensure all $http error objects are wrapped in Guacamole error objects, even if the relevant request could not be sent.

mike-jumper:


guacamole/src/main/frontend/src/app/rest/services/requestService.js
```diff
@@ -56,6 +56,11 @@ angular.module('rest').factory('requestService', ['$injector',
                 if (response.data)
                     throw new Error(response.data);
 
+                // Fall back to a generic internal error if the request couldn't
+                // even be issued (webapp is down, etc.)
+                else if ('data' in response)
```
necouchman, line 61: Is this somehow different from the above `if (response.data)`? Or is it intended to replace that??

mike-jumper: Yep - definitely different. The original `if (response.data)` test will only be true if `response.data` is truthy. For other values, such as if `response.data` is `null`, that test will fail, whereas `'data' in response` would succeed. This particular change provides a fallback `Error` representation for requests that have failed because the webapp cannot be reached at all.

necouchman (approved):
:shipit:

PR #629, opened 2021-07-26
GUACAMOLE-996: Add support for specifying an LDAP group filter.

mike-jumper:
This change builds off the proposed changes submitted @echu2013 via #497, additionally addressing the feedback on that PR:

* The default value for the new `ldap-group-search-filter` property should produce backward-compatible behavior when not specified (it should default to `(objectClass=*)`).
* The group search filter should always be taken into account when retrieving user groups (it was erroneously ignored if connections were also being stored in LDAP).

I also went ahead and added a corresponding environment variable for the Docker image.


necouchman (approved):

necouchman:
This closes #479

PR #630, opened 2021-07-27
GUACAMOLE-1386: Correct definitions of "Meta" and "Super" keys.

mike-jumper:
Guacamole's on-screen keyboard and part of its general keyboard handling incorrectly define "Super" as the "Windows" or "Command" key. This is incorrect. The correct key for the "Windows" key (or the Mac "Command" key) is "Meta".

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Keyboard.js
+++ b/guacamole-common-js/src/main/webapp/modules/Keyboard.js
@@ -439,7 +439,7 @@ Guacamole.Keyboard = function Keyboard(element) {
         46:  [0xFFFF, 0xFFFF, 0xFFFF, 0xFFAE], // delete      / KP decimal
         91:  [0xFFE7], // left windows/command key (meta_l)
         92:  [0xFFE8], // right window/command key (meta_r)
-        93:  null,     // select key
+        93:  [0xFF67], // menu key
         96:  [0xFFB0], // KP 0
         97:  [0xFFB1], // KP 1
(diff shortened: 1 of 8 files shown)


guacamole-common-js/src/main/webapp/modules/Keyboard.js
```diff
@@ -583,7 +583,7 @@ Guacamole.Keyboard = function Keyboard(element) {
         "UIKeyInputUpArrow": [0xFF52],
         "Up": [0xFF52],
         "Undo": [0xFF65],
-        "Win": [0xFFEB],
+        "Win": [0xFFE7, 0xFFE7, 0xFFE8],
```
necouchman, line 586: Is there a reason that `0xFFE7` needs to be listed twice? I see this done for other keys that have multiple identifiers, just wondering why?

mike-jumper: Yep - it's due to the way key location is handled by JavaScript keyboard events and for compatibility. If a key has left/right variations, like Left Shift vs. Right Shift, then JavaScript may represent that with a key location value where `1` indicates left , `2` indicates right, and `3` indicates the numpad:

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent#keyboard_locations

`0` is used for keys that do not vary by location or if the location is unknown, thus listing `0xFFE7` twice here is effectively a fallback.

necouchman (approved):

PR #632, opened 2021-07-29
Add support for hashing passowrds with SHA256 in user-mapping.xml

mire3212:
While it is understood that `user-mapping.xml` is not necessarily the best method for user management, it is still something used for small deployments (like home users, or even small businesses with limited IT teams). 

PLAIN_TEXT and MD5 are not cryptographically secure options for hashing a password and should not be used. Adding SHA256 is at least a better way to keep the plain test password from being viewed when using the `user-mapping.xml` file.

This change adds the required code to support SHA256 as well as including an updated example in the docs folder.

### DIFF
--- a/guacamole/doc/example/user-mapping.xml
+++ b/guacamole/doc/example/user-mapping.xml
@@ -35,7 +35,7 @@
             encoding="md5">
 
         <!-- First authorized connection -->
-		<connection name="localhost">
+    	<connection name="localhost">
             <protocol>vnc</protocol>
             <param name="hostname">localhost</param>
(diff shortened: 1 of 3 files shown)


necouchman:
@mire3212: Thanks for the contribution. I'm not opposed to adding this, though I do think the `user-mapping.xml` file should remain more of a proof-that-the-system-is-working mechanism than a production one.

That said, you'll need to open a Jira issue for this (https://issues.apache.org/jira/projects/GUACAMOLE/issues) and tag both the PR and the commits with the JIRA issue. See other open pull requests, and the following guide:

PR #633, opened 2021-07-29
GUACAMOLE-1391: Add support for hashing passwords with SHA-256 in user-mapping.xml

mire3212:
Currently, the `user-mapping.xml` file only supports plain-text or MD5. While it is typically expected that this authentication mechanism is only used for small proof-of-concept deployments, it may still be widely used for smaller deployments, like in a home-lab or small business with a limited IT team.

 

The changes I'm proposing would add support for the SHA-256 hashing algorithm  to bring the authentication mechanism into the current cryptographic era. MD5 has been broken since near the turn of the century and SHA1 just a few years ago.

https://issues.apache.org/jira/browse/GUACAMOLE-1391

### DIFF
--- a/guacamole/doc/example/user-mapping.xml
+++ b/guacamole/doc/example/user-mapping.xml
@@ -35,7 +35,7 @@
             encoding="md5">
 
         <!-- First authorized connection -->
-		<connection name="localhost">
+    	<connection name="localhost">
             <protocol>vnc</protocol>
             <param name="hostname">localhost</param>
             <param name="port">5901</param>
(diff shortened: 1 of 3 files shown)


guacamole/doc/example/user-mapping.xml
```diff
@@ -35,21 +35,36 @@
...
+    	<connection name="localhost">
             <protocol>vnc</protocol>
             <param name="hostname">localhost</param>
             <param name="port">5901</param>
             <param name="password">VNCPASS</param>
         </connection>
 
         <!-- Second authorized connection -->
-   		<connection name="otherhost">
+		<connection name="otherhost">
```
necouchman, line 46: Tab crept in, here...

necouchman (approved):
Looks good to me. @mike-jumper any concerns with adding this?

mike-jumper:
> @mike-jumper any concerns with adding this?

Nope, except that the whitespace issues should be corrected. Looks straightforward and welcome.

mire3212:
Silly tabs! I've updated the file and it should now just be spaces.

necouchman (approved):

PR #634, opened 2021-07-30
GUACAMOLE-1391: Correct typo when referencing SHA_256 enum

mire3212:
There was a typo in referencing the Enum value.


necouchman:
I'm okay with this change, but could you update the commit message to be a little more appropriate ;-).

mire3212:
That was an interesting process. I've never had to do that in GitHub...

necouchman (approved):

PR #637, opened 2021-08-13
GUACAMOLE-1386: Defer handling of "Meta" key until its identity is verified.

mike-jumper:
On ChromeOS (and possibly other platforms), the "Meta" key is used as an alternative means of pressing keys that aren't present on the hardware keyboard, like "End" (Meta+Right) or "Home" (Meta+Left). In these cases, the flag representing the Meta key is cleared once it ceases to be a modifier of the key pressed, but if Guacamole has already sent the Meta keypress to the remote desktop, it's too late to undo that keypress. The remote environment incorrectly believes Meta was pressed, resulting in counterintuitive behavior like the Windows "Start" menu opening up.

These changes correct the above behavior by deferring handling of "Meta" keypresses until after its identity has been verified through either (1) an explicit keyup event releasing "Meta" or (2) another key event that acknowledges that Meta is functioning as a modifier within its own modifier flags.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Keyboard.js
+++ b/guacamole-common-js/src/main/webapp/modules/Keyboard.js
@@ -136,13 +136,24 @@ Guacamole.Keyboard = function Keyboard(element) {
      *
      * @private
      * @constructor
+     * @param {Guacamole.Keyboard.ModifierState} [modifiers]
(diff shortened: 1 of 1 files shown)


jmuehlner (approved):
Looks reasonable to me.

PR #640, opened 2021-08-22
GUACAMOLE-1130: Limit LDAP attribute retrieval to only those required or requested

necouchman:
This builds on #550 by @echu2013, incorporating a couple of cleanup changes to the code that limits the retrieval of LDAP attributes to ones that are specifically required by the code or requested by the Guacamole configuration, making LDAP searches more efficient.

This closes #550.


extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
```diff
@@ -192,7 +192,8 @@ private ExprNode getGroupSearchFilter() throws GuacamoleException {
                 ldapConnection,
                 userDN,
                 confService.getUserSearchFilter(),
-                0);
+                0,
+                null);
```
mike-jumper, line 196: Why `null` here vs. a more restricted set of attributes?

mike-jumper (commented):
Are there changes that should be applied to retrieval of users, as well? I see the base changes within the generic `ObjectQueryService`, and within `UserGroupService`, but I don't see changes restricting the contents of retrieved users.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
necouchman: I'll take a look and see - probably can be restricted.

necouchman:
Yeah, I'll see about applying this to users and `guacConfigGroup`s, as well.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
```diff
@@ -214,13 +215,17 @@ private ExprNode getGroupSearchFilter() throws GuacamoleException {
...
+        groupAttributes.add(confService.getMemberAttribute());
+
         // Get all groups the user is a member of starting at the groupBaseDN,
         // excluding guacConfigGroups
         return queryService.search(
             ldapConnection,
             groupBaseDN,
             getGroupSearchFilter(),
-            Collections.singleton(confService.getMemberAttribute()),
+            groupAttributes,
```
necouchman, line 228: Ugh, now that I look at this, I think this is not going to result in intended behavior at all. This change, here, will make it so that the query that's run to match membership gets executed against all of the group attributes that are identified as "relevant" attributes, via this code, here:

https://github.com/necouchman/guacamole-client/blob/d6c161c1d67c349a1796fb346a5c33cb744073a0/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/ObjectQueryService.java#L336-L341

I'm pretty sure that's not what we want to happen - we want certain attributes to be retrieved, not searched against. So, this is going to take a little more reworking...

necouchman:
Okay, I've reworked this a bit, but ***have not tested, yet*** - I'll give it a go when I get a chance. I may have gone a bit overboard reworking the `ConnectionService` class, in particular - I can back out some of those constants and such if that's just too complicating to the code.

necouchman:
Okay, fixed a couple of issues and did some testing, looks clean to me.

PR #643, opened 2021-09-16
GUACAMOLE-1418: Add support of SQL Server JDBC plugin in Docker Image

jbpaux:
Implement SQL Server support in Docker Image.
According to JIRA issue, need to check if it's legally allowed (saw other ASF software are using it so I guess it could be)
Have a good review :)

### DIFF
--- a/guacamole-docker/README.md
+++ b/guacamole-docker/README.md
@@ -166,16 +166,6 @@ documented in
 Deploying Guacamole with SQLServer authentication
 --------------------------------------------------
 
-    docker run --name some-guacamole --link some-guacd:guacd \
-        --link some-sqlserver:sqlserver      \
-        -e SQLSERVER_DATABASE=guacamole_db  \
-        -e SQLSERVER_USER=guacamole_user    \
-        -e SQLSERVER_PASSWORD=some_password \
-        -e SQLSERVER_DATABASE_FILE=/run/secrets/<secret_name> \
-        -e SQLSERVER_USER_FILE=/run/secrets/<secret_name> \
-        -e SQLSERVER_PASSWORD_FILE=/run/secrets/<secret_name> \
-        -d -p 8080:8080 guacamole/guacamole
-
 Linking Guacamole to SQLServer requires three environment variables. If any of
 these environment variables are omitted, you will receive an error message, and
 the image will stop:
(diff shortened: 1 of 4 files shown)


necouchman:
> saw other ASF software are using it so I guess it could be

Can you provide these examples?

jbpaux:
Here for example: https://github.com/apache/skywalking-java/blob/e6447d517558031f49cc4450467e41b644320071/apm-sniffer/apm-sdk-plugin/mssql-jdbc-plugin/pom.xml (pom dependency)

necouchman:
@jbpaux In the case you mentioned, the pom.xml file is actually _not_ pulling in the download - it lists it as a dependency, but the `<scope>provided</scope>` tag on that dependency means that Maven will proceed building without actually pulling in those files, and will rely on them being provided by the JDK/JRE at runtime. See the `provided` description in the following section:

https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scope

jbpaux:
Ok, no pb, here then : https://github.com/apache/juddi/blob/5988d077548d54da86c982295e48df3e771ae53e/juddi-tomcat/pom.xml#L88 no provided I think (I'm not java expert so maybe I guess wrong 😄 )

mike-jumper:
The JDBC driver provided by Microsoft for SQL Server is MIT-licensed and should be fine to bundle in the Docker image: https://github.com/microsoft/mssql-jdbc

jbpaux:
Any feedback about this PR @mike-jumper @necouchman ? 😅

jbpaux:
Hi, do you plan to merge it or it requires some updates  @necouchman  ?

guacamole-docker/README.md
```diff
@@ -163,6 +163,63 @@ The process for doing this via the `mysql` utility included with MySQL is
...
+2. `SQLSERVER_USER` - The user that Guacamole will use to connect to SQLServer.
+3. `SQLSERVER_PASSWORD` - The password that Guacamole will provide when
+   connecting to SQLServer as `SQLSERVER_USER`.
+4. `SQLSERVER_DATABASE_FILE` - The path of the docker secret containing the name
+   of database to use for Guacamole authentication.
+5. `SQLSERVER_USER_FILE` - The path of the docker secret containing the name of
+   the user that Guacamole will use to connect to SQLServer.
+6. `SQLSERVER_PASSWORD_FILE` - The path of the docker secret containing the
+   password that Guacamole will provide when connecting to SQLServer as
+   `SQLSERVER_USER.
```
necouchman, line 194: I think this section is a bit confusing, as it says that SQLServer authentication "requires three variable" and goes on to list 6. I know you copied the sections above, for MySQL and PostgreSQL, and those need to be separately reviewed and reworked, but it seems like it may make more sense to leave lines 179-187 as-is, and then put a break, and say something like, "Alternatively, if you want to store database credentials using Docker secrets, the following three variables are required and replace the previous three:", and then list the three _FILE variables.

guacamole-docker/bin/start.sh
```diff
@@ -403,6 +403,142 @@ END
...
+corresponding secret:
+
+    SQLSERVER_USER     The user to authenticate as when connecting to
+                       SQLServer.
+
+    SQLSERVER_PASSWORD The password to use when authenticating with SQLServer
+                       as SQLSERVER_USER.
+
+    SQLSERVER_DATABASE The name of the SQLServer database to use for Guacamole
+                       authentication.
```
necouchman, line 423: Similar to the documentation section, above, this should also probably be reworked to take into account using either these three variables or the `_FILE` versions for Docker secret support. Again, I know it matches the Postgres and MySQL versions, so those should be reworked later, but since we're adding these, here, I think it makes sense to do it correctly.

necouchman (changes requested):
Okay, just a couple requests for fixing things up, here, mainly related to Docker secrets support.

guacamole-docker/README.md
jbpaux: Yes you're right, I was just following the other examples. I've updated the text with your instructions.

guacamole-docker/bin/start.sh
jbpaux: Updated too, hopefully it's ok.

guacamole-docker/bin/start.sh
```diff
@@ -23,7 +23,7 @@
 ##
 ## Automatically configures and starts Guacamole under Tomcat. Guacamole's
 ## guacamole.properties file will be automatically generated based on the
-## linked database container (either MySQL or PostgreSQL) and the linked guacd
+## linked database container (either MySQL,PostgreSQL or SQLServer) and the linked guacd
```
necouchman, line 26: #NitPick, but missing a space, here, after the comma...

necouchman (changes requested):
One more slight correction, and I think it's ready after that.

guacamole-docker/bin/start.sh
jbpaux: done :)

necouchman (approved):

PR #649, opened 2021-10-24
GUACAMOLE-956: Leverage HTTP header instead of query parameter for auth/session tokens.

mike-jumper:
This change refactors the REST API to alternatively accept the authentication token via a `Guacamole-Token` header, and updates the JavaScript side of things to use that header instead of the old `token` parameter. The `token` parameter remains usable as an alternative means of submitting the token.

With the HTTP tunnel using the tunnel UUID as its own sort of session token (to allow the communication for the tunnel to span multiple HTTP requests), these changes also refactor the HTTP tunnel to decouple its internal concept of a session from the tunnel UUID, effectively removing the HTTP tunnel's token from the URL, as well.

The WebSocket tunnel has not been touched here. Part of the reason for this is that WebSocket does not provide for any means of submitting arbitrary headers along with the handshake, thus we must either continue to use the URL or use WebSocket messages. Arguably, continuing to use the WebSocket URL in this way is perfectly fine:

* https://stackoverflow.com/a/65001506
* https://faqs.ably.com/is-it-secure-to-send-the-access_token-as-part-of-the-websocket-url-query-params

If we decide to change this, as well, I suggest we let that be a separate pull request.

### DIFF
--- a/guacamole/src/main/java/org/apache/guacamole/rest/RESTExceptionMapper.java
+++ b/guacamole/src/main/java/org/apache/guacamole/rest/RESTExceptionMapper.java
@@ -50,10 +50,10 @@ public class RESTExceptionMapper implements ExceptionMapper<Throwable> {
     private final Logger logger = LoggerFactory.getLogger(RESTExceptionMapper.class);
     
     /**
-     * The ContainerRequest for HTTP request that resulted in the Throwable
-     * being intercepted. Despite this class being a Singleton, this object will
-     * always be scoped with the current request for the Throwable that is being
-     * processed by this class.
+     * The ContainerRequest representing the HTTP request that resulted in the
+     * Throwable being intercepted. Despite this class being a Singleton, this
+     * object will always be scoped with the current request for the Throwable
+     * that is being processed by this class.
      */
     @Context
     private ContainerRequest request;


iota-008:
@mike-jumper  how to implement this. i want to open the URL in a new tab with token in headers.

necouchman:
@iota-008: Please do not use merged pull requests as a discussion forum. This code was merged into version 1.4.0 - as long as you're running Guacamole Client 1.4.0 or later there is nothing you need to do to implement this.

iota-008:
@necouchman sorry! but where can I ask a question?

necouchman:
@iota-008 : On the mailing lists:

https://guacamole.apache.org/support/

iota-008:
@necouchman: I asked on the mailing list, but didn't receive any reply.  I just have one question regarding this Guacamole-Token. Currently, I am just changing the URL to {guacServerUrl}/#/client/{connectionId}?{authToken}, passing the token in the query param, which directly connects to the  VM.  How to pass the token in header, I tried XMLHTTP get request but it is not working.

necouchman:
@iota-008 Yes, your post made it to the list, you'll jut have to wait on someone to reply, there.

PR #652, opened 2021-11-14
GUACAMOLE-957: Correct handling of empty/deleted ldap-servers.yml.

mike-jumper:
If `ldap-servers.yml` is entirely empty (contains purely placeholder content), the current implementation will fail with a `NullPointerException` as valid but empty YAML is parsed by Jackson as `null`. Further, if `ldap-servers.yml` was used but is later deleted by the admin, the old cached version of `ldap-servers.yml` will continue being used.

Instead, if `ldap-servers.yml` is empty or deleted, the LDAP support should move forward as if no `ldap-servers.yml` was provided.


jmuehlner (approved):
Makes sense to me.

PR #653, opened 2021-11-22
GUACAMOLE-1364: Clean up overall logic of SAML authentication flow.

mike-jumper:
In preparation for the SSO auth flow changes of [GUACAMOLE-1364](https://issues.apache.org/jira/browse/GUACAMOLE-1364), and in anticipation of future SAML changes like [GUACAMOLE-1372](https://issues.apache.org/jira/browse/GUACAMOLE-1372), this change cleans up the logic of the overall SAML authentication flow.

All handling of the actual SAML process is now abstracted away into several smaller pieces that are more narrow in scope:

* `AuthenticationSession`: An internal representation of the continuing SAML authentication process as it spans multiple requests and redirects.
* `AssertedIdentity`: An internal representation of the SAML assertion.
* `SAMLService`: A service which abstracts away all SAML response/request handling (so the Guacamole-specific pieces don't have to)

The general goal of this is to reduce and clarify the responsibility of each class to the extent that future changes will be more obvious and verifiable, particularly changes that touch the semi-complex flow that was involved with SAML. A happy side-effect of these changes is that logging is also a bit more clear.

As I propagated the refactor through, it became clear that the parameter token functionality added to the SAML support is not actually used anywhere (`getTokens()` is not invoked). I'm not sure if this has always been the case, but rather than simply apply a refactor to code that has no effect (and thus isn't testable), this also addresses that.


extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/AuthenticationProviderService.java
```diff
@@ -21,71 +21,51 @@
...
+    private Provider<SAMLAuthenticatedUser> authenticatedUserProvider;
 
     /**
-     * Provider for AuthenticatedUser objects.
+     * Manager of active SAML authentication attempts.
      */
     @Inject
-    private Provider<SAMLAuthenticatedUser> authenticatedUserProvider;
-    
+    private AuthenticationSessionManager authManager;
```
necouchman, line 62: This is a complete #NitPick, but this is called `authManager`, here, and `sessionManager` in the `SAMLService` class - unless there's a compelling reason for naming them differently, might be best to keep them the same?

necouchman (changes requested):
A couple of minor things, but overall looks like a very nice clean-up...

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/acs/AuthenticationSession.java
```diff
@@ -0,0 +1,113 @@
...
+     * considers also whether the SAML response asserting that identity has
+     * expired.
+     *
+     * @return
+     *     true if this authentication session is still valid, false if it has
+     *     expired.
+     */
+    public boolean isValid() {
+        return System.currentTimeMillis() < expirationTimestamp
+                && (identity == null || identity.isValid());
```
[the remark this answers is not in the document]
mike-jumper: Yep - exactly. If no identity has yet been asserted, the only factor in session validity is whether Guacamole itself still considers the session valid. Once an identity has been asserted, then both Guacamole and the SAML response are taken into account.

extensions/guacamole-auth-saml/src/main/java/org/apache/guacamole/auth/saml/AuthenticationProviderService.java
mike-jumper: That's fair - I think this naming is from an earlier stage in the refactor. I'll update it to be consistent.

mike-jumper: Fixed via rebase.

necouchman (approved):

PR #655, opened 2021-11-23
GUACAMOLE-1433: Update postgresql JDBC driver in Docker image to fix "authentication type 10 not supported" error

scottcwang:


necouchman:
Thanks, @scottcwang,
A couple of things, here:
* Seems like the commit message be need to be tweaked to more appropriately reflect the nature of the change (update JDBC driver in Docker image to fix...)
* Do you know if there are any compatibility issues for this? Any older versions of Postgres that won't work after updating this driver, or defaults that would be changed that might break older versions?

scottcwang:
Hi @necouchman,

- Thanks for your feedback about the commit message. I'll force-push a new commit with your suggested improvement.
- I looked through the [postgresql JDBC driver release notes](https://jdbc.postgresql.org/documentation/changelog.html). Version 42.0.0 drops support for PostgreSQL versions below 8.2. It doesn't look like any configuration defaults changed.

mike-jumper:
I _think_ losing support for PostgreSQL 8.2 should be OK. So far, the oldest I've found supported by any mainstream, non-EOL distro is 9.2 (CentOS / RHEL 7) and 9.5 (Ubuntu 16.04 LTS).

necouchman (approved):
Thanks, @scottcwang

Inrego:
While waiting for a new release of the docker container that fixes this, I made this workaround:
On host machine:
`curl -L "https://jdbc.postgresql.org/download/postgresql-42.2.24.jre7.jar" > "/opt/guacamole/postgresql-42.2.24.jre7.jar"`

In my docker compose:
```
    volumes:
      - /opt/guacamole/postgresql-42.2.24.jre7.jar:/opt/guacamole/postgresql/postgresql-9.4-1201.jdbc41.jar:ro
```
Basically what this does, is download a compatible (newer) version of the postgresql jar, and mounts it into the container as read-only to overwrite the file inside the container.

Idea taken from here, but simplified for my own use:
https://github.com/truecharts/charts/blob/185b9cc76e80d1c7a5566d0b97d7b8f53b9d9757/charts/stable/guacamole-client/values.yaml#L228-L289

necouchman:
@Inrego: The code has been merged - instead of doing this, you could simple download the source from Github and build the Docker image from the master branch.

Inrego:
I prefer to just pull the Docker image. I also use watchtower, so it's auto updated. My change persists the modification across updates.

PR #656, opened 2021-11-25
GUACAMOLE-1465: Allow different bind passwords for multi-LDAP

alubbock:
PR #648 recently added support for querying multiple LDAP servers, but as things stand all servers are required to use the same bind password. This PR enables per-server bind passwords in the new ldap-servers.yml file.

### DIFF
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
@@ -228,7 +228,7 @@ public class AuthenticationProviderService {
             }
 
             // Attempt bind (authentication)
-            LdapNetworkConnection ldapConnection = ldapService.bindAs(config, bindDn.getName(), password);
+            LdapNetworkConnection ldapConnection = ldapService.bindAs(config, bindDn.getName(), (password == null || password.isEmpty()) ? config.getSearchBindPassword() : password);
             if (ldapConnection == null) {
                 logger.info("Unable to bind as user \"{}\" against LDAP "
                         + "server \"{}\". Proceeding with next server...",


extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -228,7 +228,7 @@ private UserLDAPConfiguration getLDAPConfiguration(String username,
             }
 
             // Attempt bind (authentication)
-            LdapNetworkConnection ldapConnection = ldapService.bindAs(config, bindDn.getName(), password);
+            LdapNetworkConnection ldapConnection = ldapService.bindAs(config, bindDn.getName(), (password == null || password.isEmpty()) ? config.getSearchBindPassword() : password);
```
mike-jumper, line 231: In this context, `password` is the password provided by the user during login, not the search bind password. The search bind password should definitely not be used as a fallback for a user's login attempt, which would potentially allow a user to successfully authenticate despite not providing a valid password.

The only place that the search bind password should be pulled is in the context of the search bind DN (within `getUserBindDn()`). This should already be the case:

https://github.com/apache/guacamole-client/blob/262643b2930aad5b6dc31df75cb928577b6a99a8/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java#L131-L133

mike-jumper (changes requested):
Please see also [my recent comment on GUACAMOLE-1465](https://issues.apache.org/jira/browse/GUACAMOLE-1465?focusedCommentId=17449308&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17449308), as I suspect the motivation behind these changes may not apply. There was an issue with the way the search bind password was pulled from YAML vs. properties, but that was addressed via #651.

mike-jumper:
Closing as (1) the motivation behind these changes shouldn't apply (though please reopen the JIRA issue if not the case) and (2) the changes themselves introduce an issue where the search bind password may incorrectly be applied as if the user entered it themselves during login.

alubbock:
As you stated above, this PR is not needed and was introduced based on an LDAP configuration issue on my end. Apologies for wasting your time.

PR #657, opened 2021-11-26
GUACAMOLE-1364: Refactor all SSO extensions beneath common base.

mike-jumper:
Similar to the current layout of the database extensions, this change refactors the CAS, OpenID, and SAML extensions beneath a common "guacamole-auth-sso" base such that they can be maintained and distributed together. The result of the build will now include a `guacamole-auth-sso-VERSION.tar.gz` archive containing one subdirectory for each SSO extension.

The primary motivation behind these changes with respect to [GUACAMOLE-1364](https://issues.apache.org/jira/browse/GUACAMOLE-1364) is that the buttons/links on the login screen pointing to the various available SSO providers will be in a single, common area defined by a shared HTML patch. Reorganizing the SSO extensions beneath a common base avoids having to repeat this HTML patch everywhere and allows this shared expectation to make sense structurally.


mike-jumper:
I've just now also extracted a set of common base classes (commit 3d2e0fb), avoiding the Guice initialization and parameter token boilerplate shared across most of our SSO implementations. This will come in handy when each SSO implementation needs to provide a REST resource allowing manual redirection to the IdP (part of [GUACAMOLE-1364](https://issues.apache.org/jira/browse/GUACAMOLE-1364)).

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-base/src/main/java/org/apache/guacamole/auth/sso/SSOAuthenticationProvider.java
```diff
@@ -0,0 +1,152 @@
...
+     */
+    private final Injector injector;
+
+    /**
+     * Creates a new SSOAuthenticationProvider that authenticates users against
+     * an arbitrary SSO system. Guice dependency injection is automatically
+     * configured, with the resulting injector available to implementations via
+     * {@link #getInjector()}. Core authentication functions are provided by
+     * the given SSOAuthenticationProviderService implementation, and
+     * additional may be provided by specifying additional Guice modules.
```
jmuehlner, line 60: any additional ... functions? A noun might be nice here.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-base/src/main/java/org/apache/guacamole/auth/sso/user/SSOAuthenticatedUser.java
```diff
@@ -113,7 +106,7 @@ public AuthenticationProvider getAuthenticationProvider() {
     public Credentials getCredentials() {
         return credentials;
     }
-
+
```
jmuehlner, line 109: You've got some trailing spaces in this file.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-base/src/main/java/org/apache/guacamole/auth/sso/user/SSOAuthenticatedUser.java
```diff
@@ -47,58 +47,51 @@
...
      * @param tokens
      *     A map of all the name/value pairs that should be available
-     *     as tokens when connections are established with this user.
+     *     as tokens when connections are established by this user.
      */
     public void init(String username, Credentials credentials,
-            Map<String, String> tokens, Set<String> effectiveGroups) {
+            Set<String> effectiveGroups, Map<String, String> tokens) {
         this.credentials = credentials;
-        this.tokens = Collections.unmodifiableMap(tokens);
```
jmuehlner, line 90: Why get rid of these conversions here? Like wrapping the tokens in an unmodifable map, or lowercaseing the username?

mike-jumper: I think this is mostly just git deciding that this file is the most similar to the CAS version of things, not that these have been truly removed, but I'll double-check.

mike-jumper: OK - I've restored the unmodifiable map and lowercase transformation. The latter was indeed accidentally removed from CAS.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-base/src/main/java/org/apache/guacamole/auth/sso/SSOAuthenticationProvider.java
mike-jumper: Oops. Reworded via rebase.

jmuehlner:
Ok, LGTM!

PR #658, opened 2021-11-26
GUACAMOLE-1468: Adds SKIP_IF_UNAVAILABLE in options

kevinimbrechts:
Adds properties to guacamole.process wich configure other options (like skip-if-unavailable)

### DIFF
--- a/guacamole-docker/bin/start.sh
+++ b/guacamole-docker/bin/start.sh
@@ -711,10 +711,11 @@ associate_json() {
 }
 
 ##
-## Adds properties to guacamole.process wich configure other
-## options
+## Adds properties to guacamole.properties which configure general
+## options.
+## Covers any properties that are not tied to a specific extension.
 ##
-associate_other() {
+associate_general_options() {
     # Skip unavailable
     set_optional_property     \
         "skip-if-unavailable" \
@@ -884,8 +885,8 @@ if [ -n "$JSON_SECRET_KEY" ]; then
     associate_json
 fi
 
-# Associate other keys
-associate_other
+# Associate general configuration
+associate_general_options
 
 # Set logback level if specified
 if [ -n "$LOGBACK_LEVEL" ]; then
--- a/guacamole-docker/bin/start.sh
+++ b/guacamole-docker/bin/start.sh
@@ -711,6 +711,17 @@ associate_json() {
 }
 
 ##
+## Adds properties to guacamole.process wich configure other
+## options
+##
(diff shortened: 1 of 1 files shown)


necouchman:
Thanks, @kevinimbrechts. This will need to be associated with a Jira issue and tagged appropriately before we will merge it.

kevinimbrechts:
Hi @necouchman. Yes sorry. Is it ok now ? Thanks.

guacamole-docker/bin/start.sh
```diff
@@ -710,6 +710,17 @@ associate_json() {
     ln -s /opt/guacamole/json/guacamole-auth-*.jar "$GUACAMOLE_EXT"
 }
 
+##
+## Adds properties to guacamole.process wich configure other
```
mike-jumper, line 714: Some typos here:

* "guacamole.properties"
* "which"

guacamole-docker/bin/start.sh
```diff
@@ -710,6 +710,17 @@ associate_json() {
...
+## options
```
mike-jumper, line 715: I think "other options" is a bit too vague here. It's unclear from reading the documentation for `associate_other()` exactly which configuration properties are in scope of the function. If the idea is that `associate_other()` covers any properties that are not tied to a specific extension, then I think that should be specified here.

Alternatively, there is already a section devoted to the few options that do not relate to an extension:

https://github.com/apache/guacamole-client/blob/c143c7cb5c73bcae51a4576c3bbe7d1d76ef7620/guacamole-docker/bin/start.sh#L794-L796

Simply placing the relevant call to `set_optional_property()` there may be better and simpler.

guacamole-docker/bin/start.sh
```diff
@@ -873,6 +884,9 @@ if [ -n "$JSON_SECRET_KEY" ]; then
     associate_json
 fi
 
+# Associate other keys
```
mike-jumper, line 887: What is meant by "keys" here?

mike-jumper (changes requested):

kevinimbrechts:
Hi @mike-jumper.
Thank you for the review.
I corrected typos and change du function name to associate_general_options. I added a better description.

Personally, I prefer to keep the function. Maybe placing guacd-hostname and guacd-port into associate_general_options is a good idea?
Let me know and I will adapt accordingly.
Thanks

necouchman:
Closing this, as this option was added by #794.

PR #659, opened 2021-12-04
GUACAMOLE-1364: Allow both traditional username/password and SSO.

mike-jumper:
This change adds a "Sign in with ..." section to the login screen when one or more SSO extensions are installed, automatically listing all SSO providers available:

> image

The user is automatically redirected to the appropriate IdP upon clicking the link for that IdP, and is otherwise still able to log in with a traditional username/password via the login screen itself.


extensions/guacamole-auth-sso/modules/guacamole-auth-sso-cas/src/main/java/org/apache/guacamole/auth/cas/CASResource.java
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Service for authenticating users using CAS.
+     */
+    @Inject
+    private AuthenticationProviderService authService;
+
+    @Override
+    public Response redirectToIdentityProvider() throws GuacamoleException {
+        return Response.seeOther(authService.getLoginURI()).build();
```
jmuehlner, line 40: Isn't this exactly the same code as `OpenIDResource `? Maybe it could be abstracted out to an abstract base class or something?

mike-jumper: Good catch - I just tried this and it is definitely an improvement.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-base/src/main/java/org/apache/guacamole/auth/sso/SSOAuthenticationProvider.java
```diff
@@ -86,20 +93,32 @@ public SSOAuthenticationProvider(
...
         injector = Guice.createInjector(Iterables.concat(Collections.singletonList(new AbstractModule() {
 
             @Override
             protected void configure() {
+
                 bind(AuthenticationProvider.class).toInstance(SSOAuthenticationProvider.this);
                 bind(Environment.class).toInstance(LocalEnvironment.getInstance());
                 bind(SSOAuthenticationProviderService.class).to(authService);
+
+                LinkedBindingBuilder<SSOResource> resourceBinding = bind(SSOResource.class);
```
jmuehlner, line 118: It's not very obvious why this extra complication is here; a comment might be nice.

mike-jumper: Fair. Added via rebase.

necouchman:
@mike-jumper I know I'm a little late to the party for this one, but any thoughts on making this configurable? I can definitely understand the desire to have it as you do, above, where the default is to go to a login screen and then give the user an option of signing in with SSO. However, I can also foresee circumstances where administrators desire the more seamless experience of having the users automatically redirected to the IdP without having to click another link. It seems like, if there's an reasonably easy way to do it, some sort of guacamole.properties option - `automatic-sso-redirect: true | false` - would allow admins to choose how they want it to behave?

mike-jumper:
> I know I'm a little late to the party for this one, but any thoughts on making this configurable?

Ah, but it is! :) The behavior changes based on which extensions have priority, as dictated by either filename or the new `extension-priority` property.

To provide all users with a login screen, ensure the non-SSO extension has priority. For example, if using PostgreSQL:

    extension-priority: postgresql

Alternatively, ensure that the SSO extension does not have priority. For example, to force SAML to be loaded last:

    extension-priority: *, saml

This can also be used to change the order of the SSO providers listed, if multiple are used:

    extension-priority: postgresql, saml, openid

To automatically redirect all users and _not_ provide a login screen, just like things were before, ensure the desired SSO extension has priority. For example, to automatically sign all users in using SAML:

    extension-priority: saml

mike-jumper:
**Note:** This is because of the way Guacamole handles the login prompt, with the first extension to report an explicit login failure being the extension to dictate the login screen if no other extension authenticates the user successfully:

* If an SSO extension has priority, an unauthenticated user is given a login screen that contains only the redirect field. They are thus always automatically redirected and never see the part of the login screen that would have the manual link to the IdP.
* If a non-SSO extension has priority, an unauthenticated user is given a login screen with some sort of prompt, presumably username/password. Since these changes add HTML patches within the SSO extensions to modify the login screen to include links to the IdP, those links are visible in this case.

necouchman:
Great, thanks for the explanation - makes perfect sense! I had forgotten about that `extension-priority` setting.

PR #660, opened 2021-12-04
GUACAMOLE-1433: Fixes authentication type 10 not supported by postgresql

radove:


radove:
Closing Duplicate, it seems there is already one open

PR #661, opened 2021-12-11
GUACAMOLE-1411: Bump version numbers to 1.4.0.

mike-jumper:


necouchman (changes requested):
If the version of `guacamole-common-js` is changing to 1.4.0, it looks like this will also need to change:

https://github.com/apache/guacamole-client/blob/dd563796e7ace1d35c8e95da41f93a785b5be657/guacamole-common-js/src/main/webapp/modules/Version.js#L22-L30

Based on the comment, there, it sounds like that constant should follow the overall `guacamole-common-js` version?

mike-jumper:
> If the version of `guacamole-common-js` is changing to 1.4.0, it looks like this will also need to change:
> 
> https://github.com/apache/guacamole-client/blob/dd563796e7ace1d35c8e95da41f93a785b5be657/guacamole-common-js/src/main/webapp/modules/Version.js#L22-L30
> 
> Based on the comment, there, it sounds like that constant should follow the overall `guacamole-common-js` version?

Yep, good catch.

mike-jumper:
Fixed via rebase.

necouchman (approved):
:shipit:

PR #665, opened 2021-12-19
GUACAMOLE-986: Update documentation of all parameters/returns to specify non-nullable and/or primitive types where appropriate.

mike-jumper:
Many, many, ... many exclamation points.


necouchman (approved):
As far as I can tell this is good-to-go.

PR #666, opened 2021-12-20
GUACAMOLE-1261: Escape strings used to generate URLs for user, groups, and connections.

necouchman:
This pull request implements an AngularJS filter - `escape` - for escaping special characters in strings that will end up as part of a URL, and then implements the `escape` on the user, group, and connection links so that users and groups that contain special characters will be correctly interpreted, resolving the bug in the associated Jira issue.

### DIFF
--- a/guacamole/src/main/frontend/src/app/index/filters/escapeFilter.js
+++ b/guacamole/src/main/frontend/src/app/index/filters/escapeFilter.js
@@ -1,0 +1,35 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


guacamole/src/main/frontend/src/app/index/filters/escapeFilter.js
```diff
@@ -0,0 +1,38 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * A filter for transforming an object into an array of all non-inherited
```
mike-jumper, line 21: Copypasta from the other filter? ;)

mike-jumper (changes requested):
🍝

guacamole/src/main/frontend/src/app/index/filters/escapeFilter.js
necouchman: yum.

necouchman: Fixed via rebase.

mike-jumper (approved):

mike-jumper:
BTW, I just noticed:

image

😈 

That explains the unsettling screaming I heard as I clicked "merge".

PR #668, opened 2021-12-26
GUACAMOLE-1293: Add client-side support for receiving join and leave notifications.

necouchman:
This is the client-side change for apache/guacamole-server#358, which adds support for notifying the connection owner when a user joins the connection.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Client.js
+++ b/guacamole-common-js/src/main/webapp/modules/Client.js
@@ -1857,7 +1857,7 @@ Guacamole.Client.Message = {
     /**
      * A client message that indicates that a user has joined an existing
      * connection. This message expects a single additional argument - the
-     * username of the user who has joined the connection.
+     * name of the user who has joined the connection.
      * 
      * @type {!number}
      */
@@ -1866,7 +1866,7 @@ Guacamole.Client.Message = {
     /**
      * A client message that indicates that a user has left an existing
      * connection. This message expects a single additional argument - the
-     * username of the user who has left the connection.
+     * name of the user who has left the connection.
      * 
      * @type {!number}
      */
--- a/guacamole-common-js/src/main/webapp/modules/Client.js
+++ b/guacamole-common-js/src/main/webapp/modules/Client.js
@@ -1856,7 +1856,7 @@ Guacamole.Client.Message = {
     
     /**
      * A client message that indicates that a user has joined an existing
-     * connection. This message xpects a single additional argument - the
+     * connection. This message expects a single additional argument - the
      * username of the user who has joined the connection.
      * 
      * @type {!number}
--- a/guacamole-common-js/src/main/webapp/modules/Client.js
+++ b/guacamole-common-js/src/main/webapp/modules/Client.js
@@ -689,13 +689,17 @@ Guacamole.Client = function(tunnel) {
     this.onerror = null;
     
     /**
-     * Fired when a message is received from the remote tunnel and needs to be
-     * displayed to the user.
+     * Fired when an arbitrary message is received from the tunnel that should
+     * be processed by the client.
      * 
      * @event
-     * @param {String} message
-     *     The message sent by the remote end of the tunnel that should be
-     *     displayed for the user.
+     * @param {!number} msgcode
+     *     A status code sent by the remote server that indicates the nature of
+     *     the message that is being sent to the client.
+     *     
+     * @param {string[]} args
+     *     An array of arguments to be processed with the message sent to the
+     *     client.
      */
     this.onmsg = null;
 
@@ -1413,7 +1417,7 @@ Guacamole.Client = function(tunnel) {
         
         "msg" : function(parameters) {
             
-            if (guac_client.onmsg) guac_client.onmsg(parameters[0]);
+            if (guac_client.onmsg) guac_client.onmsg(parseInt(parameters[0]), parameters.slice(1));
             
         },
 
@@ -1840,4 +1844,32 @@ Guacamole.Client.DefaultTransferFunction = {
         dst.blue  =  0xFF & ( src.blue  | ~dst.blue);
     }
 
+};
+
+/**
+ * A list of possible messages that can be sent by the server for processing
+ * by the client.
+ * 
+ * @type {!Object.<string, number>}
+ */
+Guacamole.Client.Message = {
+    
+    /**
+     * A client message that indicates that a user has joined an existing
+     * connection. This message xpects a single additional argument - the
+     * username of the user who has joined the connection.
+     * 
+     * @type {!number}
+     */
+    "USER_JOINED": 0x0001,
+    
+    /**
+     * A client message that indicates that a user has left an existing
+     * connection. This message expects a single additional argument - the
+     * username of the user who has left the connection.
+     * 
+     * @type {!number}
+     */
+    "USER_LEFT": 0x0002
+    
 };
--- a/guacamole-common-js/src/main/webapp/modules/Client.js
+++ b/guacamole-common-js/src/main/webapp/modules/Client.js
@@ -687,6 +687,17 @@ Guacamole.Client = function(tunnel) {
      *     A status object which describes the error.
      */
     this.onerror = null;
+    
+    /**
+     * Fired when a message is received from the remote tunnel and needs to be
+     * displayed to the user.
+     * 
+     * @event
+     * @param {String} message
+     *     The message sent by the remote end of the tunnel that should be
+     *     displayed for the user.
+     */
+    this.onmsg = null;
 
     /**
      * Fired when a audio stream is created. The stream provided to this event
@@ -1398,6 +1409,12 @@ Guacamole.Client = function(tunnel) {
                 display.move(layer, parent, x, y, z);
             }
 
+        },
+        
+        "msg" : function(parameters) {
+            
+            if (guac_client.onmsg) guac_client.onmsg(parameters[0]);
+            
         },
 
         "name": function(parameters) {
--- a/guacamole/src/main/frontend/src/app/client/directives/guacClientMessage.js
+++ b/guacamole/src/main/frontend/src/app/client/directives/guacClientMessage.js
@@ -28,15 +28,54 @@ angular.module('client').directive('guacClientMessage', [function guacClientMess
         scope: {
 
             /**
-             * The message to display.
+             * The message to display to the client.
              * 
-             * @type String
+             * @type {!ManagedClientMessage}
              */
             message : '='
 
         },
 
-        templateUrl: 'app/client/templates/guacClientMessage.html'
+        templateUrl: 'app/client/templates/guacClientMessage.html',
+        
+        controller: ['$scope', '$injector', '$element',
+                function guacClientMessageController($scope, $injector, $element) {
+            
+            const ManagedClientMessage = $injector.get('ManagedClientMessage');
+            
+            /**
+             * Uses the msgcode to retrieve the correct translation key for
+             * the client message.
+             * 
+             * @returns {string}
+             */
+            $scope.getMessageKey = function getMessageKey() {
+                
+                let msgString = "DEFAULT";
+                if (Object.values(Guacamole.Client.Message).includes($scope.message.msgcode))
+                    msgString = Object.keys(Guacamole.Client.Message).find(key => Guacamole.Client.Message[key] === $scope.message.msgcode);
+                
+                return "CLIENT.CLIENT_MESSAGE_" + msgString.toUpperCase();
+            };
+            
+            /**
+             * Returns a set of key/value object pairs that represent the
+             * arguments provided as part of the message in the form
+             * ARGS[0] = value.
+             * 
+             * @returns {Object}
+             */
+            $scope.getMessageArgs = function getMessageArgs() {
+                return $scope.message.args.reduce(
+                    function(acc, value, index) {
+                        acc[`ARGS[${index}]`] = value;
+                        return acc;
+                    },
+                    {}
+                );
+            };
+                
+        }]
 
     };
 }]);
--- a/guacamole/src/main/frontend/src/app/client/directives/guacClientMessage.js
+++ b/guacamole/src/main/frontend/src/app/client/directives/guacClientMessage.js
@@ -41,6 +41,7 @@ angular.module('client').directive('guacClientMessage', [function guacClientMess
         controller: ['$scope', '$injector', '$element',
                 function guacClientMessageController($scope, $injector, $element) {
             
+            // Required types
             const ManagedClientMessage = $injector.get('ManagedClientMessage');
             
             /**
@@ -61,7 +62,9 @@ angular.module('client').directive('guacClientMessage', [function guacClientMess
             /**
              * Returns a set of key/value object pairs that represent the
              * arguments provided as part of the message in the form
-             * ARGS[0] = value.
+             * ARGS[0] = value. Guacamole's translation system relies on
+             * the arguments being available in this format in order to be able
+             * to handle substituting values for an arbitrary list of arguments.
              * 
              * @returns {Object}
              */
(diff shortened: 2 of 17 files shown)


jmuehlner:
Hey Nick! So I've had a more thorough look through this and the [backend changes](https://github.com/apache/guacamole-server/pull/358/files#diff-cec0f31d33a52d11818f03d91a6076415f717f58d61ab09b73009df961d8238fR685-R686) and I have a couple of thoughts.

First off, I really like this idea. Having join notifications will be an awesome feature, something that I would enjoy quite a bit.

The new `msg` instruction would also be really handy for implementing some features that I was thinking about as well, but I think a few tweaks would make it more extensible.

One thing is that we don't want to be generating messages for display directly on the backend - this bypasses the translation system on the frontend, and is something we've avoided in the past. We can still send a human-readable message for debugging purproses (like we do with ack, etc), but it shouldn't be used for display to the user.

If the `msg` instruction contained "msg code" or something like that, the frontend can render something based on the code, as well as any arbitrary arguments that might accomopany any given code.

These arguments could contain, for example, a user identifier that could be used to generate (using the translation system) the same message that you're currently generating on the backend.

Or, using a different event code, the frontend might render something completely different (a session recording notification, or a notification that a user is already in a connection, etc). The frontend could just ignore any event code it doesn't understand.

On the subject of username, I think it would be better to send a machine-readable user identifier (like the user ID perhaps) than the actual text username. The frontend would know what kind of identifier it sent, and would therefore know how to render any notification based on it. This could maybe be part of a generic connection context provided to the backend, which may include other things that just the user ID.

On a side note, having a user ID on the backend would be really cool since it would allow the the user log functions to actually differentiate based on user in the log.

So to summarize, this is what I was thinking:

#### The frontend can provide some context when a connection is established, including some user identifier and possibly other things.

#### The `msg` message type could contain:
* A "msg code" (not sure about the name) that contains a machine-readable message type that the frontend will know how to render based on.
* An arbitrary list of arguments that is specific to the "msg code" - in the case of the user join / leave messages this would be the user identifier previously supplied in the handshake.
* A human readable message for debugging purposes.


Does that seem reasonable? @mike-jumper any thoughts?

necouchman:
@jmuehlner Thanks for the feedback - appreciate it. I like your idea/concept for a "code" with some set of arguments so that translation can continue to happen on the front-end - that makes perfect sense to me. I'll wait to see if @mike-jumper has any additional feedback, but I'll start retooling things a bit based on this feedback.

I'm guessing we'll want a new header file where we can define well-known message codes and their human-readable debugging counter-parts, so I'll add that, and then work on tweaking the instruction to take the arguments.

necouchman:
@jmuehlner Okay, I think I've got things mostly reworked, both guacd and guacamole-client side, to where they should be, though the code does need some clean-up (still a bunch of debugging stuff in there), and I'm struggling with one other thing.

I'm trying to make the code as generic as possible, where a message code can be passed through that will then translate to a key that is looked up in the translation system, and then an arbitrary number of arguments can be passed along, as well, that will be substituted into the translation. The general concept of the translation/translation-values is used throughout the code, but I'm struggling with the "arbitrary number" part of it.

My current method is to try to pass through an array of one or more values to the translation-values, and then be able to reference those values as you would any other array. In the template this looks like this:

```
    <!-- Message text -->
    <p class="client-message-text"
       translate="{{ getClientMessage() }}"
       translate-values="{ ARGS: message.args }"></p>
```

and then in the translation file it looks like this:
```
        "CLIENT_MESSAGE_JOINED"  : "User {ARGS[0]} has joined the connection.",
        "CLIENT_MESSAGE_LEFT"    : "User {ARGS[0]} has left the connection.",
```

But this ends up translating to `User undefined has joined the connection.` I see that the `TranslatableMessage` type has an array called `variables`, but this seems to be key/value pairs (objects), and it also seems to originate from the REST API, and not something generated by guacd. I'm not sure if the right way to go is to retool guacd to send any arguments as key/value pairs (and what that would look like coming out of guacd - just write `key=value`?), or if what I've done above should work and just isn't?

Appreciate any guidance....

jmuehlner:
Hey Nick! I really like the direction this is going. It looks a lot better already. On the topic of the `ARGS` and `User undefined has joined the connection` - this seems to be a limitation on the way that `angular-translate` parses the translation strings - see https://github.com/angular-translate/angular-translate/blob/d727abb0cae045c1f645bb7a92f06f8240875987/src/directive/translate.js#L162 - the translation service would be literally looking for a key called `ARGS[0]` in the `translate-values` array.

I played around with your code and got it working by doing exactly that - the following changes get it working:
```
diff --git a/guacamole/src/main/frontend/src/app/client/directives/guacClientMessage.js b/guacamole/src/main/frontend/src/app/client/directives/guacClientMessage.js
index 019d434e6..ba5ff3339 100644
--- a/guacamole/src/main/frontend/src/app/client/directives/guacClientMessage.js
+++ b/guacamole/src/main/frontend/src/app/client/directives/guacClientMessage.js
@@ -43,6 +43,17 @@ angular.module('client').directive('guacClientMessage', [function guacClientMess
             
             const $log                 = $injector.get('$log');
             const ManagedClientMessage = $injector.get('ManagedClientMessage');
+
+            $scope.giveMeTheArgs = function giveMeTheArgs() {
+                return $scope.message.args.reduce(
+                    function(acc, value, index) {
+                        acc[`ARGS[${index}]`] = value;
+                        return acc;
+                    },
+                    {}
+                );
+            };
+
             
             /**
              * Uses the msgcode to retrieve the correct translation key for
diff --git a/guacamole/src/main/frontend/src/app/client/templates/guacClientMessage.html b/guacamole/src/main/frontend/src/app/client/templates/guacClientMessage.html
index a6fc9f25c..310f9aa3b 100644
--- a/guacamole/src/main/frontend/src/app/client/templates/guacClientMessage.html
+++ b/guacamole/src/main/frontend/src/app/client/templates/guacClientMessage.html
@@ -3,6 +3,6 @@
     <!-- Message text -->
     <p class="client-message-text"
        translate="{{ getClientMessage() }}"
-       translate-values="{ ARGS: message.args }"></p>
+       translate-values="{{ giveMeTheArgs() }}"></p>
 
 </div>
```

necouchman:
Thanks @jmuehlner - appreciate the hint and I'll try something along those lines...

necouchman:
@jmuehlner It's been a while, so I need to review and make sure, but I think it's ready for review, again...

guacamole-common-js/src/main/webapp/modules/Client.js
```diff
@@ -1824,3 +1844,31 @@ Guacamole.Client.DefaultTransferFunction = {
...
+ * A list of possible messages that can be sent by the server and displayed to
+ * the client.
+ * 
+ * @type {!Object.<string, number>}
+ */
+Guacamole.Client.Message = {
+    
+    /**
+     * A client message that indicates that a user has joined an existing
+     * connection. This message xpects a single additional argument - the
```
jmuehlner, line 1858: Typo: `xpects`

guacamole/src/main/frontend/src/app/client/directives/guacClientMessage.js
```diff
@@ -0,0 +1,81 @@
...
+            
+            /**
+             * Returns a set of key/value object pairs that represent the
+             * arguments provided as part of the message in the form
+             * ARGS[0] = value.
+             * 
+             * @returns {Object}
+             */
+            $scope.getMessageArgs = function getMessageArgs() {
+                return $scope.message.args.reduce(
```
jmuehlner, line 69: I know why you're doing this, but it would probably be useful for posterity to add a comment here about _why_ this is here.

guacamole/src/main/frontend/src/app/client/directives/guacClientMessage.js
```diff
@@ -0,0 +1,81 @@
...
+            message : '='
+
+        },
+
+        templateUrl: 'app/client/templates/guacClientMessage.html',
+        
+        controller: ['$scope', '$injector', '$element',
+                function guacClientMessageController($scope, $injector, $element) {
+            
+            const ManagedClientMessage = $injector.get('ManagedClientMessage');
```
jmuehlner, line 44: It's a minor nitpick, but when injecting dependencies like this, we usually have a comment at the top of the block like 

```
// Required types
const HappyGuacType = $injector.get('HappyGuacType');
```

We should keep doing that, for consistency.

jmuehlner:
I have a few nitpicks, but the overall approach looks pretty good to me. @mike-jumper thoughts?

guacamole-common-js/src/main/webapp/modules/Client.js
necouchman: That's the hip (read: lazy SMS) way to spell it.

Fixd.

guacamole/src/main/frontend/src/app/client/directives/guacClientMessage.js
necouchman: Added some more comment to that - hopefully that clarifies it enough?

guacamole/src/main/frontend/src/app/client/directives/guacClientMessage.js
necouchman: Added.

guacamole/src/main/frontend/src/app/client/types/ManagedClientMessage.js
```diff
@@ -0,0 +1,55 @@
...
+         * The message code sent by the server that will be used to locate the
+         * message within the Guacamole translation framework.
+         */
+        this.msgcode = template.msgcode;
+        
+        /**
+         * Any arguments that should be passed through the translation system
+         * and displayed as part of the message.
+         */
+        this.args = template.args;
```
jmuehlner, line 49: Type annotations here would make this easier to understand. (And match other type classes 😄)

guacamole/src/main/frontend/src/app/client/directives/guacClientMessage.js
jmuehlner: That is a bit more clear - though I wonder consider quoting the value `"ARGS[0]"` just to make it extra super clear that it's a string. I know object values in javascript have to be strings anyway, but making it explicitly clear never hurts.

It might be even more clear if you used a format for these keys that doesn't look like an array index, like `ARGS_0` or something, just to make it even more obvious that they're strings.

necouchman: That's a really good idea - I've changed it to be what you said, just ARGS_<INT>, as I agree that makes it easier to identify that we're not actually dealing with an array at that point.

guacamole-common-js/src/main/webapp/modules/Client.js
```diff
@@ -687,6 +687,20 @@ Guacamole.Client = function(tunnel) {
...
+     * displayed to the user.
+     * 
+     * @event
+     * @param {!number} msgcode
+     *     A status code sent by the remote server that indicates the message
+     *     that should be displayed to the client.
+     * @param {string[]} args
+     *     An array of arguments that can be added to the message that will be
+     *     displayed to the client. 
+     */
```
mike-jumper, line 702: I think this should be documented more along the lines of receiving a message indicating some sort of arbitrary event or state change, not necessarily something that _needs_ or _will_ to be displayed to the user.

guacamole-common-js/src/main/webapp/modules/Client.js
```diff
@@ -1824,3 +1844,31 @@ Guacamole.Client.DefaultTransferFunction = {
...
+    "JOINED": 0x0001,
+    
+    /**
+     * A client message that indicates that a user has left an existing
+     * connection. This message expects a single additional argument - the
+     * name of the user who has left the connection.
+     * 
+     * @type {!number}
+     */
+    "LEFT": 0x0002
```
mike-jumper, line 1872: I suggest renaming these to `USER_JOINED` and `USER_LEFT` to (1) align with the `GUAC_MESSAGE_USER_JOINED` and `GUAC_MESSAGE_USER_LEFT` constants and (2) clarify the context of the event.

guacamole/src/main/frontend/src/app/client/directives/guacClientMessage.js
```diff
@@ -0,0 +1,84 @@
...
+             * 
+             * @returns {string}
+             */
+            $scope.getMessageKey = function getMessageKey() {
+                
+                let msgString = "DEFAULT";
+                if (Object.values(Guacamole.Client.Message).includes($scope.message.msgcode))
+                    msgString = Object.keys(Guacamole.Client.Message).find(key => Guacamole.Client.Message[key] === $scope.message.msgcode);
+                
+                return "CLIENT.CLIENT_MESSAGE_" + msgString.toUpperCase();
```
mike-jumper, line 59: This should probably be a call to [`translationStringService.canonicalize()`](https://github.com/apache/guacamole-client/blob/165bd413c0ab9c55b0c0e3614fcb829ef5f8ca95/guacamole/src/main/frontend/src/app/locale/services/translationStringService.js#L28-L42).

mike-jumper (changes requested):

guacamole/src/main/java/org/apache/guacamole/tunnel/TunnelRequestService.java
```diff
@@ -340,6 +340,11 @@ public GuacamoleTunnel createTunnel(TunnelRequest request)
         GuacamoleSession session = authenticationService.getGuacamoleSession(authToken);
         AuthenticatedUser authenticatedUser = session.getAuthenticatedUser();
         UserContext userContext = session.getUserContext(authProviderIdentifier);
+        
+        // Attempt to get the user's name and set it for the tunnel client.
+        String name = authenticatedUser.getCredentials().getName();
```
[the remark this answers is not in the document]
necouchman: Fixed via rebase.

guacamole/src/main/frontend/src/app/client/directives/guacClientMessage.js
necouchman: Added a commit to fix this, plus update the translation strings in the `en.json` file with the updated names.

guacamole/src/main/frontend/src/app/client/directives/guacClientMessage.js
```diff
@@ -0,0 +1,87 @@
...
+             * 
+             * @returns {string}
+             */
+            $scope.getMessageKey = function getMessageKey() {
+                
+                let msgString = "GUAC_MESSAGE_DEFAULT";
+                if (Object.values(Guacamole.Client.Message).includes($scope.message.msgcode))
+                    msgString = Object.keys(Guacamole.Client.Message).find(key => Guacamole.Client.Message[key] === $scope.message.msgcode);
+                
+                return translationStringService.canonicalize('CLIENT.' + msgString);
```
mike-jumper, line 62: I believe this will transform `CLIENT.` into `CLIENT_`. To add the `CLIENT` namespace, you'd need to instead:

```js
'CLIENT.' + translationStringService.canonicalize(msgString)
```

but you'll probably want something more like:

```js
'CLIENT.MESSAGE_' + translationStringService.canonicalize(msgString)
```

assuming the keys within `Guacamole.Client.Message` are `USER_JOINED`, etc.

mike-jumper (changes requested):

guacamole-common-js/src/main/webapp/modules/Client.js
```diff
@@ -1824,3 +1845,31 @@ Guacamole.Client.DefaultTransferFunction = {
...
+Guacamole.Client.Message = {
+    
+    /**
+     * A client message that indicates that a user has joined an existing
+     * connection. This message expects a single additional argument - the
+     * name of the user who has joined the connection.
+     * 
+     * @type {!number}
+     */
+    "GUAC_MESSAGE_USER_JOINED": 0x0001,
```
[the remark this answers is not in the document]
necouchman: Okay - it just makes locating the correct translation key a bit less elegant ;-)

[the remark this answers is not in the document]
necouchman: Updated via rebase.

guacamole/src/main/frontend/src/app/client/directives/guacClientMessage.js
necouchman: Fixed via rebase.

mike-jumper (approved):

PR #669, opened 2021-12-26
GUACAMOLE-1402: Expose Client state enum values.

necouchman:


guacamole-common-js/src/main/webapp/modules/Client.js
```diff
@@ -1699,11 +1692,63 @@ Guacamole.Client = function(tunnel) {
...
+ * All possible Guacamole Client states.
+ * 
+ * @type {!Object.<string, number>}
+ */
+Guacamole.Client.State = {
+    
+    /**
+     * The client is idle, with no active connection.
+     * 
+     * @type number
```
mike-jumper, line 1710: I believe this (and the others) needs to be: `@type {!number}`

See: https://jsdoc.app/tags-type.html

mike-jumper (changes requested):
Are there any changes needed to be made to the webapp to leverage these constants?

guacamole-common-js/src/main/webapp/modules/Client.js
necouchman: Okay, should be fixed.

necouchman:
> Are there any changes needed to be made to the webapp to leverage these constants?

I did not find any references to the old constants:

```
$ find guacamole/src/main/frontend/src/ -type f -iname \*.js -exec grep -H STATE_ {} \;
$ 
```

That said, the frontend does contain the ManageedClientState.js file, which implements the `ManagedClientState.ConnectionState` enum, which seems to implement similar states, although adds a couple of extra ones:

https://github.com/apache/guacamole-client/blob/e64013059dcb68f23bea43fc2a47e01940097b56/guacamole/src/main/frontend/src/app/client/types/ManagedClientState.js#L74-L130

Do you think it makes sense to try to combine these somehow?

mike-jumper:
> I did not find any references to the old constants: ...

I think I found them:

https://github.com/apache/guacamole-client/blob/881d203080009c3a8827c602c1f6d314f30a67b7/guacamole/src/main/frontend/src/app/client/types/ManagedClient.js#L429-L468

>  That said, the frontend does contain the ManageedClientState.js file, which implements the `ManagedClientState.ConnectionState` enum ... Do you think it makes sense to try to combine these somehow?

I think it makes sense for these two to remain decoupled - with the low-level API values directing the slightly higher-level ones specific to the webapp that drive the UI.

necouchman:
Aha! Okay, I've added another commit that updates the `ManagedClient.js` file to use the new enum values.

mike-jumper (approved):

PR #674, opened 2021-12-30
GUACAMOLE-1482:  Italian translation update

smart2128:


necouchman (commented):
Thanks @smart2128 - we always appreciate folks stepping in and helping to provide correct translations!

One question for you - have you looked at any of the extension modules (e.g. LDAP, JDBC, etc.), to see if translations for those values can be added or corrected, as well? There are several translation files aside from the main one you updated. If you're willing to do a little more work, you could search for `en.json` and `it.json` and either update or add those translations, as required?

smart2128:
> One question for you - have you looked at any of the extension modules (e.g. LDAP, JDBC, etc.), to see if translations for those values can be added or corrected, as well? There are several translation files aside from the main one you updated. If you're willing to do a little more work, you could search for `en.json` and `it.json` and either update or add those translations, as required?

Thank you @necouchman. I'll check and review them later or as soon as possible.

PR #677, opened 2022-01-03
GUACAMOLE-61: Add JDBC-specific tokens, including start date and time

necouchman:
This is an attempt at resolving the challenge specified in the Jira ticket, which is that the default `${GAUC_DATE}` and `${GUAC_TIME}` tokens may not match up to the JDBC start date/time. JDBC-specific tokens have been added - including the start date/time, which are pulled from the `ActiveConnectionRecord` for the connection - which should allow the date and time to match up with the database records. Other tokens have been added, as well, which will also resolve some other Jira issues, such as wanting to use the connection name in the token.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
@@ -21,6 +21,7 @@ package org.apache.guacamole.auth.jdbc.tunnel;
 
 import com.google.inject.Inject;
 import com.google.inject.Provider;
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -82,6 +83,51 @@ public abstract class AbstractGuacamoleTunnelService implements GuacamoleTunnelS
     private final Logger logger = LoggerFactory.getLogger(AbstractGuacamoleTunnelService.class);
 
     /**
+     * The token that contains the date the connection was started.
+     */
+    private final String JDBC_DATE_TOKEN = "GUAC_JDBC_STARTDATE";
+    
+    /**
+     * The format of the date in the date token.
+     */
+    private final String JDBC_DATE_TOKEN_FORMAT = "yyyyMMdd";
+    
+    /**
+     * The token that contains the start time of the connection.
+     */
+    private final String JDBC_TIME_TOKEN = "GUAC_JDBC_STARTTIME";
+    
+    /**
+     * The format of the time in the time token.
+     */
+    private final String JDBC_TIME_TOKEN_FORMAT = "HHmmss";
+    
+    /**
+     * The token that contains the connection name.
+     */
+    private final String JDBC_CONNECTION_NAME_TOKEN = "GUAC_JDBC_CONNECTION_NAME";
+    
+    /**
+     * The token that contains the connection identifier.
+     */
+    private final String JDBC_CONNECTION_ID_TOKEN = "GUAC_JDBC_CONNECTION_ID";
+    
+    /**
+     * The token that contains the hostname configured in the connection parameters.
+     */
+    private final String JDBC_CONNECTION_HOSTNAME_TOKEN = "GUAC_JDBC_HOSTNAME";
+    
+    /**
+     * The name of the parameter containing the hostname in the configuration.
+     */
+    private final String JDBC_CONNECTION_HOSTNAME_TOKEN_PARAMETER = "hostname";
+    
+    /**
+     * The token containing the protocol configured in the connection.
+     */
+    private final String JDBC_CONNECTION_PROTOCOL_TOKEN = "GUAC_JDBC_PROTOCOL";
+    
+    /**
      * Mapper for accessing connections.
      */
     @Inject
@@ -121,7 +167,7 @@ public abstract class AbstractGuacamoleTunnelService implements GuacamoleTunnelS
      * All active connections through the tunnel having a given UUID.
      */
     private final Map<String, ActiveConnectionRecord> activeTunnels =
-            new ConcurrentHashMap<String, ActiveConnectionRecord>();
+            new ConcurrentHashMap<>();
     
     /**
      * All active connections to a connection having a given identifier.
@@ -415,7 +461,7 @@ public abstract class AbstractGuacamoleTunnelService implements GuacamoleTunnelS
     private GuacamoleTunnel assignGuacamoleTunnel(ActiveConnectionRecord activeConnection,
             GuacamoleClientInformation info, Map<String, String> tokens,
             boolean interceptErrors) throws GuacamoleException {
-
+        
         // Record new active connection
         Runnable cleanupTask = new ConnectionCleanupTask(activeConnection);
         try {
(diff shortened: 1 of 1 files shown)


extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
```diff
@@ -416,6 +462,21 @@ private GuacamoleTunnel assignGuacamoleTunnel(ActiveConnectionRecord activeConne
...
 
+        // Set up JDBC-specific tokens
+        tokens.put(JDBC_DATE_TOKEN,
+                new SimpleDateFormat(JDBC_DATE_TOKEN_FORMAT)
+                        .format(activeConnection.getStartDate()));
+        tokens.put(JDBC_TIME_TOKEN,
+                new SimpleDateFormat(JDBC_TIME_TOKEN_FORMAT)
+                        .format(activeConnection.getStartDate()));
+        tokens.put(JDBC_CONNECTION_NAME_TOKEN, activeConnection.getConnectionName());
+        tokens.put(JDBC_CONNECTION_ID_TOKEN, activeConnection.getConnectionIdentifier());
```
mike-jumper, line 473: I'm not sure about adding tokens specific to JDBC for the identifier and name. What about instead implementing `GUAC_CONNECTION_NAME` / `GUAC_CONNECTION_ID` at the webapp level via `StandardTokenMap`?

And what is the intended behavior for these if a balancing connection group is used?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
```diff
@@ -416,6 +462,21 @@ private GuacamoleTunnel assignGuacamoleTunnel(ActiveConnectionRecord activeConne
...
+        tokens.put(JDBC_CONNECTION_HOSTNAME_TOKEN,
+                activeConnection.getConnection().getConfiguration().getParameter(JDBC_CONNECTION_HOSTNAME_TOKEN_PARAMTER));
+        tokens.put(JDBC_CONNECTION_PROTOCOL_TOKEN,
+                activeConnection.getConnection().getConfiguration().getProtocol());
```
mike-jumper, line 477: Within `AbstractGuacamoleTunnelService`, I think any additional tokens should be added alongside where we are already adding `HISTORY_UUID`:

https://github.com/apache/guacamole-client/blob/165bd413c0ab9c55b0c0e3614fcb829ef5f8ca95/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java#L463-L465

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
```diff
@@ -416,6 +462,21 @@ private GuacamoleTunnel assignGuacamoleTunnel(ActiveConnectionRecord activeConne
             GuacamoleClientInformation info, Map<String, String> tokens,
             boolean interceptErrors) throws GuacamoleException {
 
+        // Set up JDBC-specific tokens
+        tokens.put(JDBC_DATE_TOKEN,
+                new SimpleDateFormat(JDBC_DATE_TOKEN_FORMAT)
+                        .format(activeConnection.getStartDate()));
+        tokens.put(JDBC_TIME_TOKEN,
+                new SimpleDateFormat(JDBC_TIME_TOKEN_FORMAT)
+                        .format(activeConnection.getStartDate()));
```
mike-jumper, line 471: Did we already go through whether it would be better to overwrite `GUAC_DATE` / `GUAC_TIME` with values that will match the history entry? I remember there being a discussion about that at some point, though I don't recall the consensus.

mike-jumper (changes requested):

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
necouchman: Yeah, I'm fine implementing these at a higher level. I think I originally looked at it, but seemed like there were some challenges with grabbing those items for connections not stored in JDBC. I'll take another look, though.

For a balancing group, it's a good point. Seems like there are two possible options - or perhaps we should do some combination of both of these:
* Use the value of the balancing group name & id.
* Use the value of the connection that is actually selected.

Seems like having both available might be useful - there may be circumstances where it'd be useful - for example, when generating recording names - to have the name of both the balancing group and the underlying connection.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
necouchman: Okay, I can relocate things.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
necouchman: I don't think we've had a discussion about it, no. I went along the assumption that there may be some reason why you wouldn't want to lose the value of `GUAC_DATE` and/or `GUAC_TIME`, but I've no strong attachment to it. I'm happy to just overwrite it, as well, if there aren't any concerns from anyone about that.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
necouchman: Ah, yes, a quick look back at that reminds me why I didn't add it in `StandardTokenMap`. Currently `StandardTokenMap` is only invoked within `TunnelRequestService`, and, when it gets called there, the data provided is somewhat limited. The connection/group name is not provided, and the identifier is a bit abstract. I'm not sure how much re-tooling we want to do of that entire chain of classes just to provide these tokens, rather than just relying on the various `Connection` and `ConnectionGroup` implementations to identify them?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
necouchman: Relocated via rebase.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
necouchman: @mike-jumper Are you good with the current approach, or still want to try to re-work this into the `StandardTokenMap`?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
necouchman: @mike-jumper Further thoughts on keeping JDBC-specific ones or overwriting the `GUAC_DATE`/`GUAC_TIME` tokens?

mike-jumper: I think I'm good with keeping things specific to `GUAC_JDBC_*` here, particularly given the expected complexity of the generalized approach. The `GUAC_JDBC_*` tokens are reasonable as they stand and will scratch the community's itch.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
```diff
@@ -81,6 +82,51 @@ public abstract class AbstractGuacamoleTunnelService implements GuacamoleTunnelS
...
+    
+    /**
+     * The token that contains the hostname configured in the connection parameters.
+     */
+    private final String JDBC_CONNECTION_HOSTNAME_TOKEN = "GUAC_JDBC_HOSTNAME";
+    
+    /**
+     * The name of the parameter containing the hostname in the configuration.
+     */
+    private final String JDBC_CONNECTION_HOSTNAME_TOKEN_PARAMTER = "hostname";
```
mike-jumper, line 123: `JDBC_CONNECTION_HOSTNAME_TOKEN_PARAMETER`*

mike-jumper (changes requested):

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AbstractGuacamoleTunnelService.java
necouchman: Oops. Fixed via rebase.

mike-jumper (approved):

PR #683, opened 2022-01-13
Added support for specifying LDAP UPN in properties file

agreenbhm:
Added support for specifying LDAP UPN in properties file.  Allows for any user in LDAP (with the corresponding UPN) to authenticate. Removes requirement of users being within same OU for large LDAP deployments.

### DIFF
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
@@ -120,17 +120,28 @@ public class AuthenticationProviderService {
      *     If required properties are missing, and thus the user DN cannot be
      *     determined.
      */
-    private Dn getUserBindDN(LDAPConfiguration config, String username)
-            throws GuacamoleException {
+    private Dn getUserBindDN(LDAPConfiguration config, String username, String password) throws GuacamoleException {
 
         // If a search DN is provided, search the LDAP directory for the DN
         // corresponding to the given username
-        String searchBindLogon = config.getSearchBindDN();
+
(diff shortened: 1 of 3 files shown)


agreenbhm:
> @agreenbhm : A couple of notes:
> 
> 1. You need to have a Jira issue opened for a pull request, and the PR and commits need to be tagged with the Jira issue.
> 2. There's already a Jira issue for this: https://issues.apache.org/jira/browse/GUACAMOLE-536
> 3. There's already work being done on this front: [GUACAMOLE-536: Implement additional bind types for LDAP #507](https://github.com/apache/guacamole-client/pull/507)

Thanks for the info.  Looks like that PR is nearly 2 years old.  @necouchman: Given that you are the author of that PR, is there anything more that is needed to be done to get that merged? It looks like the changes I made are pretty similar to the ones you did so I'm not sure why your PR wouldn't be merged yet since it seems functional.

mike-jumper (changes requested):

mike-jumper:
I'm not sure whether these changes (in spirit) nor [GUACAMOLE-1511](https://issues.apache.org/jira/browse/GUACAMOLE-1511) are actually necessary:

* The LDAP support does allow for lookup of users based on their UPN. This would involve using `userPrincipalName` for the username attribute.
* Multiple LDAP servers are supported via the recently-introduced multi-server LDAP support. Users can be mapped to their relevant LDAP server based on username patterns, and then the search account can narrow that to a specific user once the relevant server is determined.

I know that Active Directory allows for binding with UPNs, and so adding a feature that would tell Guacamole to accept usernames matching that pattern and attempt to bind with those usernames directly would make configuration easier, but I'm not sure that's what these changes or [GUACAMOLE-1511](https://issues.apache.org/jira/browse/GUACAMOLE-1511) are intended to be.

PR #684, opened 2022-01-16
GUACAMOLE-1503: Japanese Character input cannot be switched By full-width half-width key.

nw-engineer:
Fixed so that full-width and half-width keys work correctly and Japanese input switching works.
Hope to include it in the next release. Thank you.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/InputSink.js
+++ b/guacamole-common-js/src/main/webapp/modules/InputSink.js
@@ -118,11 +118,11 @@ Guacamole.InputSink = function InputSink() {
             var rect = focused.getBoundingClientRect();
             if (rect.left + rect.width > 0 && rect.top + rect.height > 0)
                 return;
+                sink.focus();
 
         }
 
         // Refocus input sink instead of handling click
-        sink.focus();
 
     }, true);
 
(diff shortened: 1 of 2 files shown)


guacamole-common-js/src/main/webapp/modules/InputSink.js
```diff
@@ -118,11 +118,11 @@ Guacamole.InputSink = function InputSink() {
             var rect = focused.getBoundingClientRect();
             if (rect.left + rect.width > 0 && rect.top + rect.height > 0)
                 return;
+                sink.focus();
 
         }
 
         // Refocus input sink instead of handling click
-        sink.focus();
```
mike-jumper, line 125: What's the reasoning behind this change?

guacamole-common-js/src/main/webapp/modules/Keyboard.js
```diff
@@ -559,8 +561,7 @@ Guacamole.Keyboard = function Keyboard(element) {
         "Up": [0xFF52],
         "Undo": [0xFF65],
         "Win": [0xFFE7, 0xFFE7, 0xFFE8],
-        "Zenkaku": [0xFF28],
-        "ZenkakuHankaku": [0xFF2A]
+        "Zenkaku": [0xFF28]
```
mike-jumper, line 564: Why is the definition for `ZenkakuHankaku` being removed?

nw-engineer:
I'm really sorry. I wasn't able to see the exchanges of past issues.（GUACAMOLE-520,GUACAMOLE-951)

Also, some change requests are mistaken.
I would like to confirm the contents again and send a separate pull request.

I canceled this pull est
What should I do now?

nw-engineer:
I didn't comment in particular, so I will close it. I apologize to you for a fuss over.

PR #687, opened 2022-01-21
GUACAMOLE-1508: Add support for nesting .jar files within extensions.

mike-jumper:
This change adds support for nesting .jar files within extensions, automatically extracting those files to a temporary directory for inclusion within the classpath. The files and directory are deleted automatically upon webapp shutdown or, if that somehow fails, upon JVM exit.

This should also remove the need to exclude metadata from signed .jars, as such libraries will now be able to successfully access and verify themselves.

### DIFF
--- a/extensions/pom.xml
+++ b/extensions/pom.xml
@@ -64,12 +64,11 @@
                         <id>copy-runtime-dependencies</id>
                         <phase>prepare-package</phase>
                         <goals>
-                            <goal>unpack-dependencies</goal>
+                            <goal>copy-dependencies</goal>
                         </goals>
                         <configuration>
                             <includeScope>runtime</includeScope>
                             <outputDirectory>${project.build.directory}/classes</outputDirectory>
-                            <excludes>META-INF/*.SF,META-INF/*.DSA</excludes>
                         </configuration>
                     </execution>
                 </executions>
--- a/guacamole/src/main/java/org/apache/guacamole/extension/Extension.java
+++ b/guacamole/src/main/java/org/apache/guacamole/extension/Extension.java
@@ -27,6 +27,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipException;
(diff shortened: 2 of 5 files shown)


necouchman (commented):
My only concern is if this is going to break backward compatibility with older extension files?

guacamole/src/main/java/org/apache/guacamole/extension/Extension.java
```diff
@@ -355,12 +356,18 @@
...
+     *     application shutdown in reverse order.
+     *
      * @throws GuacamoleException
      *     If the provided file is not a .jar file, does not contain the
      *     guac-manifest.json, or if guac-manifest.json is invalid and cannot
      *     be parsed.
      */
-    public Extension(final ClassLoader parent, final File file) throws GuacamoleException {
+    public Extension(final ClassLoader parent, final File file,
+            final List<File> temporaryFiles) throws GuacamoleException {
```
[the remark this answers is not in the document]
mike-jumper: > Should there be a backward-compatible convenience method?

Nope! The classes touched here are strictly the internals of the webapp, not part of guacamole-ext or any other library.

> Is this going to break older versions of extensions in whatever version of Guacamole (1.5.0?) gets this functionality?

Nope! The previous functionality was:

* `.class` files are loaded directly from the extension `.jar` file

while the new functionality is:

* `.class` files are loaded directly from the extension `.jar` file
* `.class` files are additionally loaded directly from any `.jar` files within the archive root of the extension `.jar` file

So what we have here is a superset of what we had before, with modifications to only internal components.

The only way that an extension might encounter problems is if it already includes one or more `.jar` files in its root and relies on Guacamole _not_ loading them ... which would be pretty odd IMHO.

extensions/pom.xml
```diff
@@ -64,12 +64,11 @@
                         <id>copy-runtime-dependencies</id>
                         <phase>prepare-package</phase>
                         <goals>
-                            <goal>unpack-dependencies</goal>
+                            <goal>copy-dependencies</goal>
                         </goals>
                         <configuration>
                             <includeScope>runtime</includeScope>
                             <outputDirectory>${project.build.directory}/classes</outputDirectory>
-                            <excludes>META-INF/*.SF,META-INF/*.DSA</excludes>
```
necouchman, line 72: Sorry, one other question - is this going to cause any issues with signed JARs? I seem to recall that BouncyCastle is a signed JAR, and some of these items had to be removed to get extensions that use it (RADIUS, maybe?) to run correctly.

mike-jumper: I believe those issues should no longer occur, since those signed .jar files can now be correctly verified by the JVM.

necouchman (approved):

PR #689, opened 2022-01-23
GUACAMOLE-641: Add support for reading secrets from key vaults.

mike-jumper:
This change builds off the changes originally proposed via #336, adding:

1. Dynamic retrieval of secrets based on existing connection parameter contents, where the vault implementation has enough structural information to provide this.
2. Retrieval of `guacamole.properties` values from the vault.
3. Support for the Keeper Secrets Manager vault.

I was recently tasked with implemented support for the Keeper Secrets Manager vault as part of my day job, which conveniently gives us a way to move forward with key vault support while we await the missing license information for Azure Key Vault. I went ahead and temporarily removed the Azure portion of these changes in a single, final commit so that we can simply revert it when ready.

Note that these changes depend on #687 to function. The Keeper library and the Bouncy Castle FIPS library both depend on being able to read their own .jar files, which cannot work correctly with the old method of unpacking and mingling all dependencies directly into the extension .jar file.


extensions/guacamole-vault/modules/guacamole-vault-base/src/main/java/org/apache/guacamole/vault/VaultAuthenticationProviderModule.java
```diff
@@ -0,0 +1,98 @@
...
+import org.apache.guacamole.net.auth.UserContext;
+import org.apache.guacamole.vault.user.VaultUserContext;
+import org.apache.guacamole.vault.user.VaultUserContextFactory;
+
+/**
+ * Guice module which configures injections specific to the base support for
+ * key vaults. When adding support for a key vault provider, a subclass
+ * specific to that vault implementation will need to be created.
+ *
+ * @see AzureKeyVaultAuthenticationProviderModule
```
jmuehlner, line 36: See it ... where? I don't see this class anywhere.

mike-jumper: Thaaaat's ... fair. I'll correct these references to point to things that still exist in the current set of changes.

mike-jumper: OK - fixed.

guacamole-ext/src/main/java/org/apache/guacamole/token/TokenFilter.java
```diff
@@ -247,19 +269,71 @@ public String filter(String input) {
...
+     * be interpreted as literals.
+     *
+     * @param input
+     *     The string to filter.
+     *
+     * @return
+     *     A copy of the input string, with any tokens replaced with their
+     *     corresponding values.
+     */
+    public String filter(String input) {
```
jmuehlner, line 294: Why not roll this functionality into the 2-argiment `filter` method? Having another method with the same name that works _sort of_ but not _exactly like_ `filter(*, false)` is honestly a little confusing.

guacamole-ext/src/main/java/org/apache/guacamole/token/TokenFilter.java
```diff
@@ -273,6 +347,34 @@ public void filterValues(Map<?, String> map) {
...
+     * GuacamoleTokenUndefinedException will be thrown.
+     *
+     * @param map
+     *     The map whose values should be filtered.
+     *
+     * @throws GuacamoleTokenUndefinedException
+     *     If at least one token in at least one string has no corresponding
+     *     value.
+     */
+    public void filterValuesStrict(Map<?, String> map)
```
jmuehlner, line 368: I think this API might be a little more straightforward if there were a 1 and 2 argument version each of `filterValue` and `filterValues`, with the 1-argument version passing `false` for the `strict` parameter. That would also reduce the duplicated iteration code in this function.

guacamole-ext/src/main/java/org/apache/guacamole/token/TokenFilter.java
mike-jumper: Ah - because that would require all callers of `filter()` to handle the possible exception, even when there will not be one in the `false` case. There was some discussion on this in #336 (look for resolved conversations regarding `TokenFilter.java - linking to the conversation directly appears to not work for resolved discussions 🤷).

jmuehlner: Ahh right, gotcha. That makes sense.

guacamole-ext/src/main/java/org/apache/guacamole/token/TokenFilter.java
mike-jumper: (See above - this would end up requiring exception handling on the part of the developer even though it's known the exception would never be thrown).

jmuehlner (approved):
This all looks pretty reasonable to me.

PR #694, opened 2022-01-30
GUACAMOLE-1524: Importing LDAPS certificate into docker container

sirux88:
associated JIRA: https://issues.apache.org/jira/browse/GUACAMOLE-1524

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,6 +62,9 @@ ARG GID=1001
 RUN groupadd --gid $GID guacamole
 RUN useradd --system --create-home --shell /usr/sbin/nologin --uid $UID --gid $GID guacamole
 
+# allow guacamole user to import certificates into default java keystore file cacerts
+run chown guacamole /usr/local/openjdk-8/jre/lib/security/cacerts && chmod +w /usr/local/openjdk-8/jre/lib/security/cacerts 
+
 # Run with user guacamole
 USER guacamole
 
--- a/guacamole-docker/bin/start.sh
+++ b/guacamole-docker/bin/start.sh
@@ -599,6 +599,10 @@ END
     set_optional_property "ldap-max-referral-hops"          "$LDAP_MAX_REFERRAL_HOPS"
     set_optional_property "ldap-operation-timeout"          "$LDAP_OPERATION_TIMEOUT"
 
+    if [ -n "$LDAP_SSL_CERT_FILE" ]; then
+        keytool -importcert -file $LDAP_SSL_CERT_FILE -alias $LDAP_SSL_CERT_FILE -storepass changeit -noprompt -keystore $JAVA_HOME/jre/lib/security/cacerts || true
+    fi
+
     # Add required .jar files to GUACAMOLE_EXT
     ln -s /opt/guacamole/ldap/guacamole-auth-*.jar "$GUACAMOLE_EXT"
 


Dockerfile
```diff
@@ -62,6 +62,9 @@ ARG GID=1001
 RUN groupadd --gid $GID guacamole
 RUN useradd --system --create-home --shell /usr/sbin/nologin --uid $UID --gid $GID guacamole
 
+# allow guacamole user to import certificates into default java keystore file cacerts
+run chown guacamole /usr/local/openjdk-8/jre/lib/security/cacerts && chmod +w /usr/local/openjdk-8/jre/lib/security/cacerts
```
mike-jumper, line 66: I don't think this should be done. Should something go awry in the web application, this would extend the malicious potential of that to installing certificates.

The limited-privilege user that runs the webapp shouldn't be able to alter certs.

mike-jumper (changes requested):
Overall:
* I really think this is not the right approach. If the container needs to be able to perform a privileged action involving system-wide certificates, we need to find a safe way to do that, not just open up the cert store to modification by the webapp.
* Please provide a more meaningful commit message than "commit". Commit messages should capture the high-level purpose of the change being made, and need to reference the JIRA issue (see established git history). A good blog post on this: https://www.codelord.net/2015/03/16/bad-commit-messages-hall-of-shame/

Dockerfile
sirux88: i don't like that solution either because of the same security reasons you mentioned. 

To sum up my finding about java and ceritifcates:
- Only a keystorefile can be used as a certificate source within java. 
- You can't load a single cert file into you application
- Importing into default keystore file for unprivileged users is not possible by default

A solution that only affects the container and provides sufficient security is not possible as far as I can assume: 
- You can't run multiple CMD/ENTRYPOINT-commands within a dockerfile with different users
- Importing certifcates with RUN-Commands is nonsense since this would be done while building the image (and not starting a container)

A possible solution would be:
- create a new keystore file with all the necessary certs 
- make the file read only after creating and importing
- merge it on the fly within the application with the system wide keystore (maybe with [https://github.com/1and1/CompositeJKS](url))

But since this topic only affects containers and the approach above requires some changes to the application itself I don't know if it should be done like that.

sirux88: @mike-jumper:
I reviewed all the stuff once again and came to the conclusion that there's a easier and more overall solution.
It ended up in [allow-jsk-usage](https://github.com/sirux88/guacamole-client/tree/allow-jsk-usage)

I would suggest closing this topic as well as the JIRA issue and creating new ones.
Please give me a short hint if this is fine for you. 
I'll then do all the steps

daftaupe: @sirux88 did you create a new PR for that new allow-jsk-usage ? I'm interested by this feature but cant find anything related to the announcement you made here.

sirux88: Since the maintainers never gave some feedback on this I haven't created a PR.

Since contributing is, at least in my eyes, not that easy within in this project I'm not really interested in doing so. 

If you want to feel free to copy my approach and put a PR on this yourself

mike-jumper: That's fair - sorry for being so unresponsive on this and other PRs.

daftaupe: @sirux88 since there's an answer from @mike-jumper, are you going to keep on working on the PR ?

sirux88: Sorry but no @daftaupe I'll close this PR and will not open a new one. Reasons are the same as above and without any offense to anyone mike's comment is not an answer to the question it is an excuse.

Your options are:
1) You could use the `startup.sh` from my branch [allow-jsk-usage](https://github.com/sirux88/guacamole-client/tree/allow-jsk-usage) and mount it into your container. Then set the appropriate environment variables and use your own keystore file. 
2) Copy my approach and open a new PR on yourself
3) There's another PR that deals with loading custom keystore files #805. The PR's reason is different but the solution may fit for your needs aswell.

sirux88:
For anyone stumbling over this:

add the following to `startup.sh` right before `start_guacamole`. The set the corresponding environment variables and use your custom keystore file

<details><summary>Details</summary>
<p>

```shell
# Set java truststore and truststore password if specified
if [ -n "$JAVA_TRUSTSTORE_FILE" ]; then
    
    if [ -z "$JAVA_TRUSTSTORE_PASSWORD" ]; then
        cat <<END
FATAL: Missing required environment variables
-----------------------------------------------------------------------------------
If using a java truststore file, you must provide each of the
following environment variables:

    JAVA_TRUSTSTORE_FILE        The truststore file to be used.

    JAVA_TRUSTSTORE_PASSWORD    The password for the specified truststore file
END
        exit 1;
    fi

    CATALINA_OPTS="$CATALINA_OPTS -Djavax.net.ssl.trustStore=$JAVA_TRUSTSTORE_FILE"
    CATALINA_OPTS="$CATALINA_OPTS -Djavax.net.ssl.trustStorePassword=$JAVA_TRUSTSTORE_PASSWORD"

    if [ -n "$JAVA_TRUSTSTORE_TYPE" ]; then
        CATALINA_OPTS="$CATALINA_OPTS -Djavax.net.ssl.trustStoreType=$JAVA_TRUSTSTORE_TYPE"
    fi

    export CATALINA_OPTS
fi
``` 

</p>
</details>

Dockerfile
daftaupe: thanks @sirux88 I'll think about what is the best option.

PR #695, opened 2022-01-30
GUACAMOLE-1525: Support for true fullscreen mode and keyboard lock

sirux88:
associated JIRA issue: https://issues.apache.org/jira/browse/GUACAMOLE-1525

see also: GUACAMOLE-989 and GUACAMOLE-124

To start fullscreen mode:
Open gucamole's special menu via Ctrl+Alt+Shift and click fullsceen:
image
("Vollbild" is german for Fullscreen)

### DIFF
--- a/guacamole/src/main/frontend/src/app/client/controllers/clientController.js
+++ b/guacamole/src/main/frontend/src/app/client/controllers/clientController.js
@@ -39,6 +39,7 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
     const clipboardService       = $injector.get('clipboardService');
     const dataSourceService      = $injector.get('dataSourceService');
     const guacClientManager      = $injector.get('guacClientManager');
+    const guacFullscreen         = $injector.get('guacFullscreen');
     const iconService            = $injector.get('iconService');
     const preferenceService      = $injector.get('preferenceService');
     const requestService         = $injector.get('requestService');
@@ -683,8 +684,22 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
         callback  : $scope.disconnect
     };
 
+    /**
+     * Action that toggles fullscreen mode within the
+     * currently-connected client and then closes the menu.
+     */
+    var FULLSCREEN_MENU_ACTION = {
+        name      : 'CLIENT.ACTION_FULLSCREEN',
+        classname : 'fullscreen action',
+        callback  : function fullscreen() {
+            
+            guacFullscreen.toggleFullscreenMode();
+            $scope.menu.shown = false;
+        }
+    };
+
     // Set client-specific menu actions
-    $scope.clientMenuActions = [ DISCONNECT_MENU_ACTION ];
+    $scope.clientMenuActions = [ DISCONNECT_MENU_ACTION,FULLSCREEN_MENU_ACTION ];
 
     /**
      * @borrows Protocol.getNamespace
@@ -833,6 +848,9 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
     // Clean up when view destroyed
     $scope.$on('$destroy', function clientViewDestroyed() {
         setAttachedGroup(null);
+
+        // always unset fullscreen mode to not confuse user 
+        guacFullscreen.setFullscreenMode(false);
     });
 
 }]);
--- a/guacamole/src/main/frontend/src/app/client/services/guacFullscreen.js
+++ b/guacamole/src/main/frontend/src/app/client/services/guacFullscreen.js
@@ -1,0 +1,55 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * A service for providing true fullscreen and keyboard lock support.
+ * Keyboard lock is currently only supported by Chromium based browsers
+ * (Edge >= V79, Chrome >= V68 and Opera >= V55)
+ */
+angular.module('client').factory('guacFullscreen', ['$injector', 
+        function guacFullscreen($injector) {
+
+    var service = {};
+
+    // check is browser in true fullscreen mode
+    service.isInFullscreenMode = function isInFullscreenMode() {
+        return document.fullscreenElement;
+    }
+
+    // set fullscreen mode
+    service.setFullscreenMode = function setFullscreenMode(state) {
+        if (document.fullscreenEnabled) {
+            if (state && !service.isInFullscreenMode())
+                document.documentElement.requestFullscreen().then(navigator.keyboard.lock()); 
+            else if (!state && service.isInFullscreenMode())
+                document.exitFullscreen().then(navigator.keyboard.unlock()); 
+        }
+    }
+
+    // toggles current fullscreen mode (off if on, on if off)
+    service.toggleFullscreenMode = function toggleFullscreenMode() {
+        if (!service.isInFullscreenMode())
+            service.setFullscreenMode(true);
+        else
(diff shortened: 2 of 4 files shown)


necouchman:
@sirux88 I know you've been frustrated by difficulty of getting PRs merged. Are you still willing to work on this one? I have a couple of minor style issues and then I think this one could be ready to go and close out several long-standing issues.

sirux88:
@necouchman:
If there's a chance to get it merged by time I'm willing to work on this for sure.
Just let me know.

Since it is only on the jira issue but affecs functionality:
1) `True fullscreen mode` should work on each maintained browser as stated here => https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API
2) `Keyboard lock` will only work on chromium based browsers => https://developer.mozilla.org/en-US/docs/Web/API/Keyboard/lock

guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -62,6 +62,7 @@
         "ACTION_CANCEL"                    : "@:APP.ACTION_CANCEL",
         "ACTION_CLEAR_CLIENT_MESSAGES"     : "Clear",
         "ACTION_CLEAR_COMPLETED_TRANSFERS" : "Clear",
+        "ACTION_FULLSCREEN"                : "Fullscreen",
```
necouchman, line 65: I think this should be down below `ACTION_DISCONNECT` by Alphabetical order?

guacamole/src/main/frontend/src/app/client/services/guacFullscreen.js
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * A service for providing true fullscreen and keyboard lock support.
+ * Keyboard lock is currently only supported by Chromium based browsers
+ * (Edge >= V79, Chrome >= V68 and Opera >= V55)
+ */
+angular.module('client').factory('guacFullscreen', ['$injector',
+
+    
+    function guacFullscreen($injector) {
```
necouchman, line 28: Looking at other services, from a style perspective I think you can remove the space between the `angular...` line and the `function...` line, and then indent the function line one more. For example, see:

https://github.com/apache/guacamole-client/blob/d1faaa9605c5eef668da8bf84279d7a88cad5af7/guacamole/src/main/frontend/src/app/client/services/guacClientManager.js#L20-L25

guacamole/src/main/frontend/src/app/client/services/guacFullscreen.js
```diff
@@ -0,0 +1,58 @@
...
+        service.isInFullscreenMode=function isInFullscreenMode(){
+            return document.fullscreenElement;
+        }
+
+        // set fullscreen mode
+        service.setFullscreenMode = function setFullscreenMode(state) {
+            if(document.fullscreenEnabled){
+                
+                if(state && !service.isInFullscreenMode()) document.documentElement.requestFullscreen().then(navigator.keyboard.lock()); 
+                else if(!state && service.isInFullscreenMode()) document.exitFullscreen().then(navigator.keyboard.unlock());
```
necouchman, line 51: These lines are long enough that it probably makes sense to not try to do one-liners, here, and, instead, put the `document..` part on its own line.

Also, please put spaces between the `if` and the opening parenthesis.

guacamole/src/main/frontend/src/app/client/services/guacFullscreen.js
```diff
@@ -0,0 +1,58 @@
...
+        var service = {};
+
+        // toggles current fullscreen mode (off if on, on if off)
+        service.toggleFullscreenMode = function toggleFullscreenMode(){
+            if(!service.isInFullscreenMode()){
+                service.setFullscreenMode(true);
+            }else{
+                service.setFullscreenMode(false);
+            }
+        }
```
necouchman, line 39: * Please put spaces between parenthesis and brackets.
* Please put spaces between the `if` and its opening parenthesis, and between the `else` and its bracket.
* Please don't cuddle the `else` clause with the closing bracket above it.

guacamole/src/main/frontend/src/app/client/services/guacFullscreen.js
```diff
@@ -0,0 +1,58 @@
...
+
+        // check is browser in true fullscreen mode
+        service.isInFullscreenMode=function isInFullscreenMode(){
```
necouchman, line 42: Spaces around the equals sign, and between `()` and `{`

guacamole/src/main/frontend/src/app/client/services/guacFullscreen.js
```diff
@@ -0,0 +1,58 @@
...
+            return document.fullscreenElement;
+        }
+
+        // set fullscreen mode
+        service.setFullscreenMode = function setFullscreenMode(state) {
+            if(document.fullscreenEnabled){
```
necouchman, line 48: Spaces between `if` and `(` and between `)` and `{`

necouchman (changes requested):
Thanks, @sirux88 - Mostly some style change requests. It can seem like some of these are trivial or pedantic, but when you scale it up to the amount of code, here, it makes a difference after a while for the overall readability of the code, so we tend to be very particular about the style being consisten.

Other than that, unless @mike-jumper or @jmuehlner have any concerns, I think it's good.

guacamole/src/main/frontend/src/app/client/services/guacFullscreen.js
sirux88: It's not always done the same coding style. I tried to align it to your example

sirux88:
> Thanks, @sirux88 - Mostly some style change requests. It can seem like some of these are trivial or pedantic, but when you scale it up to the amount of code, here, it makes a difference after a while for the overall readability of the code, so we tend to be very particular about the style being consisten.

I understand that and it's ok. I hope it's fne now @necouchman 

To be honest maybe you should add a linter or [codestyle checker](https://maven.apache.org/plugins/maven-checkstyle-plugin/usage.html) to your project and setup your rules. So you haven't to explain it to contributers (which is annoying I know)

necouchman:
> To be honest maybe you should add a linter or [codestyle checker](https://maven.apache.org/plugins/maven-checkstyle-plugin/usage.html) to your project and setup your rules.

Ah, nice, I will have a look.

necouchman (approved):
Looks okay to me, now. I will give @mike-jumper and @jmuehlner a while longer to respond with any issues they have, otherwise I'll go ahead and merge it.

sirux88:
Is the failing docker build a problem @necouchman? 
The problem within the build seems not related to this PR.

sirux88:
I fell over an problem concerning browsers not implementing `keyboard.lock` api (e.g. firefox):
If you set fullscreen mode and use the browsers back functionality (e.g. mouse xbutton1) the screen will stay in fullscreen mode.
This might be confusing to some users.

My approach to solve it would be to unload fullscreen mode if the client is "minimized". Meaning guacamole application is going to its homescreen.

But I haven't found an event to register to.
Is there an event like this @necouchman ?

sirux88:
> I fell over an problem concerning browsers not implementing `keyboard.lock` api (e.g. firefox): If you set fullscreen mode and use the browsers back functionality (e.g. mouse xbutton1) the screen will stay in fullscreen mode. This might be confusing to some users.
> 
> My approach to solve it would be to unload fullscreen mode if the client is "minimized". Meaning guacamole application is going to its homescreen.
> 
> But I haven't found an event to register to. Is there an event like this @necouchman ?

I ended up in adding it to `$scope.$on('$destroy', function clientViewDestroyed()` function. 
=> solved in my eyes

necouchman:
> I ended up in adding it to $scope.$on('$destroy', function clientViewDestroyed() function.
=> solved in my eyes

Seems reasonable to me.

guacamole/src/main/frontend/src/app/client/controllers/clientController.js
```diff
@@ -44,7 +44,7 @@ angular.module('client').controller('clientController', ['$scope', '$routeParams
     const requestService         = $injector.get('requestService');
     const tunnelService          = $injector.get('tunnelService');
     const userPageService        = $injector.get('userPageService');
-
+    const guacFullscreen         = $injector.get('guacFullscreen');
```
necouchman, line 47: Nitpick, but could you put this service in alphabetical order with the ones above it, and then keep the space below this before the next code comment?

necouchman (changes requested):
One more style change request.

guacamole/src/main/frontend/src/app/client/controllers/clientController.js
mike-jumper: This is fairly important, IMHO. Maintaining alphabetical order for things like `#include`, `import`, blocks of retrievals from the AngularJS `$injector`, etc. helps avoid merge conflicts.

sirux88: done

necouchman (approved):

PR #698, opened 2022-02-12
GUACAMOLE-462: Add support for associating connection history with logs/recordings.

mike-jumper:
This change adds API representation of logs and recordings at the history entry level, and allows those resources to be retrieved from history entries via the REST API.

This is achieved by:

1. Adding a deterministic (and optional) UUID for history entries with semantics that allow the UUID to be known prior to the connection being fully established, as well as an optional identifier that allows individual history entries to be retrieved directly.
2. Adding an extension, "guacamole-history-recording-storage", which automatically searches a defined path (`/var/lib/guacamole/recordings/` by default) for files matching history entry UUIDs. If any are found, their type is identified and the file is associated with the history entry via decoration.
3. Adding `${HISTORY_PATH}` and `${HISTORY_UUID}` tokens that allow recordings, etc. to be saved and later reassociated with the history entries that correspond to the connections that created those recordings.

Thus, history entries and recordings may be tied together without requiring database schema changes.

Users are only able to retrieve recordings/logs for a history entry if they have access to that history entry. It's up to the implementation to define whether users may view history entries, however the database restricts this to only users that can view both the relevant connection and relevant user. The storage extension implemented here additionally requires that a user have `UPDATE` permission on the associated connection to be able to retrieve a history entry, as it seems sensible to limit access to logs/recordings to only those that manage the connection. It might make sense for this to be configurable in the future.


guacamole/src/main/java/org/apache/guacamole/rest/history/ActivityLogResource.java
```diff
@@ -0,0 +1,77 @@
...
+public class ActivityLogResource {
+
+    /**
+     * The ActivityLog whose contents are being exposed.
+     */
+    private final ActivityLog log;
+
+    /**
+     * Creates a new ActivityRecordSetResource which exposes the records within
+     * the given ActivityRecordSet.
```
necouchman, line 40: Yum....copy pasta? I think this is an `ActivityLogResource` exposing the given `ActivityLog`?

guacamole/src/main/java/org/apache/guacamole/rest/history/ActivityLogResource.java
```diff
@@ -0,0 +1,77 @@
...
+     *     If an error prevents retrieving the content of the log or its size.
+     */
+    @GET
+    public Response getContents() throws GuacamoleException {
+
+        // Build base response exposing the raw contents of the underlying log
+        ResponseBuilder response = Response.ok(log.getContent(),
+                log.getType().getContentType());
+
+        // Include size, if knopwn
```
necouchman, line 68: Silent "p"?

necouchman (changes requested):
A few initial comments - you'll have to forgive the sarcastic humor, it seems especially strong tonight.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ConnectionRecordMapper.java
```diff
@@ -19,113 +19,9 @@
...
-     */
-    List<ConnectionRecordModel> searchReadable(@Param("identifier") String identifier,
-            @Param("user") UserModel user,
-            @Param("terms") Collection<ActivityRecordSearchTerm> terms,
-            @Param("sortPredicates") List<ActivityRecordSortPredicate> sortPredicates,
-            @Param("limit") int limit,
-            @Param("effectiveGroups") Collection<String> effectiveGroups);
-
-}
+public interface ConnectionRecordMapper extends ActivityRecordMapper<ConnectionRecordModel> {}
```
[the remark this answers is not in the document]
mike-jumper: There would be no downstream impact as it's not a part of the public API (these mappers are internal pieces of the database auth implementation). I don't think we can remove it, though, as MyBatis needs to be able to determine with XML mapping file corresponds to which mapper.

guacamole/src/main/java/org/apache/guacamole/rest/history/APIActivityRecord.java
```diff
@@ -55,6 +58,30 @@
...
     private final boolean active;
 
+    /**
+     * The unique identifier assigned to this record, or null if this record
+     * has no such identifier.
+     */
+    private final String identifier;
+    
+    /**
+     * A UUID uniquely identifies this record, or null if no such unique
```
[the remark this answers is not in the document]
mike-jumper: Oops - I'll add that "that".

I think "uniquely identifies" should remain, as it is specifically this record that is uniquely identified. The value _should_ be unique relative to all possible UUIDs, but the specification for the record UUIDs given by `getUUID()` of `ActivityRecord` allows some wiggle room here so long as it's at least unique relative to other records in the same set.

guacamole/src/main/java/org/apache/guacamole/rest/history/ActivityLogResource.java
mike-jumper: Yup - definitely copypasta.

guacamole/src/main/java/org/apache/guacamole/rest/history/ActivityLogResource.java
mike-jumper: Nah, this one is a loud "p", and it must be spoken with as much volume and enunciation as possible.

Also yes. I'll pfix this...

guacamole/src/main/frontend/src/app/rest/types/ActivityLog.js
```diff
@@ -0,0 +1,89 @@
...
+ * under the License.
+ */
+
+/**
+ * Service which defines the ActivityLog class.
+ */
+angular.module('rest').factory('ActivityLog', [function defineActivityLog() {
+
+    /**
+     * The object returned by REST API calls when representing the a log or
```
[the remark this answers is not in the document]
mike-jumper: Fixed via rebase.

mike-jumper:
No more draft - retrieval of individual records is now implemented.

necouchman:
Okay, will work on reviewing tonight.

guacamole/src/main/java/org/apache/guacamole/rest/history/ActivityLogResource.java
```diff
@@ -0,0 +1,77 @@
...
+public class ActivityLogResource {
+
+    /**
+     * The ActivityLog whose contents are being exposed.
+     */
+    private final ActivityLog log;
+
+    /**
+     * Creates a new ActivityLogSetResource which exposes the records within
+     * the given ActivityLogSet.
```
necouchman, line 40: `ActivityLogSetResource` -> `ActivityLogResource`
`ActivityLogSet` -> `ActivityLog`

necouchman (changes requested):
Sorry for the delay - took a couple of days longer. Just a few minor documentation clean-up things and I think it'll be ready to ship.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ActivityRecordMapper.java
```diff
@@ -0,0 +1,147 @@
...
+     *
+     * @param terms
+     *     The search terms that must match the returned records.
+     *
+     * @param sortPredicates
+     *     A list of predicates to sort the returned records by, in order of
+     *     priority.
+     *
+     * @param limit
+     *     The maximum number of records that should be returned.
```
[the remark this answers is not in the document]
mike-jumper: Nope - a limit value must always be supplied and will be enforced.

guacamole/src/main/java/org/apache/guacamole/rest/history/ActivityLogResource.java
mike-jumper: Fixed via rebase.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ActivityRecordMapper.java
```diff
@@ -0,0 +1,147 @@
...
+     *
+     * @return
+     *     The number of rows updated.
+     */
+    int updateEndDate(@Param("record") ModelType record);
+
+    /**
+     * Searches for up to <code>limit</code> activity records that contain
+     * the given terms, sorted by the given predicates, regardless of whether
+     * the data they are associated with is is readable by any particular user.
```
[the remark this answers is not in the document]
mike-jumper: Fixed via rebase.

necouchman (approved):

PR #700, opened 2022-02-17
GUACAMOLE-1539: fix docker init logic when only auth-json is configured

xens:
This commit fixes an issue where the Docker init-script won't start if only the auth-json extension is present.


xens:
Done!

mike-jumper (approved):

PR #710, opened 2022-03-30
GUACAMOLE-1571: Properly wrap stream errors, and check available translations instead of hardcoding.

jmuehlner:
From https://issues.apache.org/jira/browse/GUACAMOLE-1571:

- JSON-formatted errors received from the server during upload streams are not properly handled, leading to error messages not being displayed to the user
- The list of translatable upload errors is hardcoded in the JS, meaning that translations cannot be supplied for currently untranslated errors by extensions.  (NOTE: the same issue exists for the client notification system)
- In Firefox only, upload streams that are rejected do not contain any response information - the status code is 0 in the XHR, and the response information is blank in the network tab - this is likely related to this issue: https://stackoverflow.com/questions/20210252/status-code-text-and-all-response-headers-are-null-for-an-xhr-that-has-status-c

This PR fixes the first 2 problems. I don't know of any obvious way to fix the 3rd. The fetch API seemed like a possible solution, but it doesn't seem to offer anyway to track the progress of uploads, which would be a dealbreaker.

### DIFF
--- a/guacamole/src/main/frontend/src/app/client/directives/guacClientNotification.js
+++ b/guacamole/src/main/frontend/src/app/client/directives/guacClientNotification.js
@@ -34,16 +34,16 @@ angular.module('client').directive('guacClientNotification', [function guacClien
 
         /**
          * The client whose status should be displayed.
-         * 
+         *
          * @type ManagedClient
          */
         client : '='
-        
+
     };
 
     directive.controller = ['$scope', '$injector', '$element',
         function guacClientNotificationController($scope, $injector, $element) {
-   
+
         // Required types
         const ManagedClient      = $injector.get('ManagedClient');
         const ManagedClientState = $injector.get('ManagedClientState');
@@ -53,6 +53,7 @@ angular.module('client').directive('guacClientNotification', [function guacClien
         const $location              = $injector.get('$location');
         const authenticationService  = $injector.get('authenticationService');
         const guacClientManager      = $injector.get('guacClientManager');
+        const guacTranslate          = $injector.get('guacTranslate');
         const requestService         = $injector.get('requestService');
         const userPageService        = $injector.get('userPageService');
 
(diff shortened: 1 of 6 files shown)


guacamole/src/main/frontend/src/app/client/services/guacTranslate.js
```diff
@@ -0,0 +1,82 @@
...
+     *     The requested translation ID, which may or may not be translated.
+     *
+     * @param {Sting} defaultTranslationId
+     *     The translation ID that will be used if no translation is found for `translationId`.
+     *
+     * @returns {Promise.<TranslationResult>}
+     *     A promise which resolves with a TranslationResult containing the results from
+     *     the translation attempt.
+     */
+    function translateWithFallback(translationId, defaultTranslationId) {
```
necouchman, line 50: I'm a little unsure on this one, but should this line be:
```
var translateWithFallback = function translateWithFallback(translationId, defaultTranslationId) {
```
since you `return translateWithFallback` below? I'm a little unsure what the acceptable/preferred ways are to do that??

guacamole/src/main/frontend/src/app/client/directives/guacClientNotification.js
```diff
@@ -65,26 +66,6 @@ angular.module('client').directive('guacClientNotification', [function guacClien
...
-            0x0207: true,
-            0x0208: true,
-            0x0209: true,
-            0x020A: true,
-            0x020B: true,
-            0x0301: true,
-            0x0303: true,
-            0x0308: true,
-            0x031D: true
-        };
```
necouchman, line 86: Are these error codes not needed at all anymore? I see that they're removed here and in the `guacFileTransfer.js` file, but do they not need to be defined at all?

necouchman (commented):
A couple of questions from me...

guacamole/src/main/frontend/src/app/client/directives/guacClientNotification.js
jmuehlner: Yeah, we don't need these anymore - they were just a hardcoded list of error codes for which translations were assumed to exist. 

Now it's checking the available translations, using a status-code-specific one if it exists, and otherwise just going with the default.

They weren't used for anything else.

guacamole/src/main/frontend/src/app/client/services/guacTranslate.js
jmuehlner: Hmm, I'm not sure it makes any practical difference, but it would match the convention to do it the way you suggested. I'll change it.

necouchman: Sounds good - I think the current code tends to assign variables and then return them, but, aside from trying to be consistent, I have no real preference and just wasn't sure if it would cause any functional problems.

necouchman (approved):

PR #717, opened 2022-04-11
GUACAMOLE-1572: Match environment variables for POSTGRESQL auth plugin doesn't match guacamole.properties variable name

jbpaux:
Fix issue described in https://issues.apache.org/jira/browse/GUACAMOLE-1572 where environment variables for POSTGRESQL auth plugin doesn't match guacamole.properties variable name


jbpaux:
I also remarked the user variable for both 3 auth modules is named user in environment variables and username in guacamole.properties, should be updated too.
I'll update the documentation when we're ok on all changes here.
No sure if it fits in 1.5.0 or not but in case of I've set upstream on 1.5.0.

necouchman:
@jbpaux  This should be against master, it will not go into the 1.5.0 release.

jbpaux:
If you have time to have a look at it @necouchman (and my question about the username environment variable too) ;)

guacamole-docker/README.md
```diff
@@ -55,31 +55,31 @@ Deploying Guacamole with PostgreSQL authentication
...
 
     docker run --name some-guacamole --link some-guacd:guacd \
-        --link some-postgres:postgres      \
-        -e POSTGRES_DATABASE=guacamole_db  \
-        -e POSTGRES_USER=guacamole_user    \
-        -e POSTGRES_PASSWORD=some_password \
-        -e POSTGRES_DATABASE_FILE=/run/secrets/<secret_name> \
-        -e POSTGRES_USER_FILE=/run/secrets/<secret_name> \
-        -e POSTGRES_PASSWORD_FILE=/run/secrets/<secret_name> \
+        --link some-postgresql:postgresql      \
```
necouchman, line 58: This particular line may actually need to not be changed - my guess is that `postgres` is the default name of the container for PostgreSQL databases. Also, this doesn't really match up to any `guacamole.properties` entries.

@mike-jumper Thoughts on this one?

necouchman (changes requested):
Just one comment/concern on the link to the Postgres container. Links are supposed to be deprecated, anyway, so at some point we should probably just get rid of that, but, until then.

guacamole-docker/README.md
jbpaux: Yes, I've hesitated too when doing the changes but I thought it may be less error-prone if everything is named postresql instead of postgres only for the link. I can revert this part if you prefer.
What about the username variable ?

mike-jumper: > This particular line may actually need to not be changed ... Thoughts on this one?

Hm ... I'm not sure off-hand. Most of my Docker deployments have been docker-compose for a while, and I recall someone mentioning that Docker had deprecated `--link`.

I think the "postgresql" part is used by Docker to generate environment variables, and so would need to match whatever the container uses.

mike-jumper: Originally, the `POSTGRES_*` prefix was chosen despite the property names to match the similar environment variables used by Docker's official `postgres` container. Aligning things universally with properties makes sense and would allow some automation for setting properties (via the environment properties feature).

There will be pain for anyone using the old variables. If we do this, it's probably worth adding compatibility for the `POSTGRES_` prefix and `--postgres` option.

jbpaux: `--link` is deprecated according to this [Legacy Container links](https://docs.docker.com/network/links/) but it's still working.  No sure how many people are using it.
If it should be removed, it should be removed in another issue/PR.

The `--link some-postgresql:postgresql ` means that environment variables for the ports are generated using the `<name>_PORT_<port>_<protocol>` format where `<name>` is the name of the last part of the link (`postgresql`) for us but `postgres` before my change.

We can add compatibility for older environment variable names but at some point it will be removed and will break for some people, maybe it's "easier" to add it in compatibility change when the image will be released so people can see it explicitely. They'll need to do this change at some point if we want to align everything.

If we don't want to align everything, we must solve the otherway and rename the few environment variables using `POSTGRESQL_*` to `POSTGRES_*` because it doesn't make sense to have some using one syntax and other using the other.

jbpaux: Yes `--link` is deprecated but can still be used. If you want to remove them I guess it will be done in another issue/PR. See more on [Legacy container links](https://docs.docker.com/network/links/). The environment variables related to the link are generated from the last part of the `--link mypostgres:postgres` which results in `<name>_PORT_<port>_<protocol>` variable and `<name>_NAME` variable. So in our case `postgres`. In the PR I've renamed it too so it's consistent with other variable names but it's just to make it prettier.

For the other variables, renaming them from `POSTGRES_` to `POSTGRESSQL_` will of course break current deployment. If needed we can have both but I can imagine it will be removed one day and it will cause the same issue. It would maybe be easier to point it in the release notes that the variables changed their names and need to be changed in docker compose etc.
How do you want to move forward on it ?

necouchman: @jbpaux I agree that removing link is different in scope from this, so no need to worry about that.

For the renaming, I would think something simple at the beginning of the script that checks for the old variables and assigns the values to the new variables would probably work? That way the remaining checks, here, would only have to consider a single set of variables - the "correctly" named ones.

jbpaux: I did it, it's long but should work as expected 😅

guacamole-docker/bin/start.sh
```diff
@@ -251,40 +251,97 @@ environment variables or their corresponding Docker secrets by appending _FILE
...
+fi
+if [ -n "$POSTGRES_SOCKET_TIMEOUT" ]; then
+    POSTGRESQL_SOCKET_TIMEOUT=$POSTGRES_SOCKET_TIMEOUT;
+fi
+if [ -n "$POSTGRES_SSL_KEY_PASSWORD_FILE" ]; then
+    POSTGRESQL_SSL_KEY_PASSWORD_FILE=$POSTGRES_SSL_KEY_PASSWORD_FILE;
+fi
+if [ -n "$POSTGRES_SSL_KEY_PASSWORD" ]; then
+    POSTGRESQL_SSL_KEY_PASSWORD=$POSTGRES_SSL_KEY_PASSWORD;
+fi
```
mike-jumper, line 321: This is lengthy indeed. I think it would be better here to:

1. Leverage indirect references so we can use a loop and not copy/paste the same `if` block repeatedly.
2. Add a warning message about the deprecation and recommend migration.

For example:

```bash
for VAR_BASE in \
    HOSTNAME PORT \
    DATABASE USER PASSWORD \
    DATABASE_FILE USER_FILE PASSWORD_FILE \
    ...; do

        OLD_VAR="POSTGRES_$VAR_BASE"
        NEW_VAR="POSTGRESQL_$VAR_BASE"

        if [ -n "${!OLD_VAR}" ]; then
            printf -v "$NEW_VAR" "%s" "${!OLD_VAR}"
            echo "Some warning about new vs. old..."
        fi

done

Comment (jbpaux) [reply guacamole-docker/bin/start.sh:321]: I thought of that but do we assume `printf -v` is available ? `/bin/sh` is used and I wasn't sure it was ok.

Comment (mike-jumper) [reply guacamole-docker/bin/start.sh:321]: What happened when you tested against the `/bin/sh` present in the Docker container?

Comment (jbpaux) [reply guacamole-docker/bin/start.sh:321]: Just had a look on it.
`/bin/sh` is a link to `/bin/dash`, not `/bin/bash`
Launching the script as is throw an error `/opt/guacamole/bin/start.sh: 337: Bad substitution` corresponding to this line:
```bash
        if [ -n "${!OLD_VAR}" ]; then
```

It's working fine when using `/bin/bash` instead.
So I'm wondering if we should switch to `/bin/bash` as it's available or find another method.

jbpaux: any comment @mike-jumper ?

jbpaux: up ? @mike-jumper

mike-jumper: Whoops - hey there, @jbpaux.

If `/bin/bash` is already available, then that seems by far the best solution. Really anything that avoids having to manually list each mapping of old/new var would be really nice. To me, the important thing is to avoid having such a large, manually-maintained list of ifs.

guacamole-docker/bin/start.sh
```diff
@@ -251,40 +251,97 @@ environment variables or their corresponding Docker secrets by appending _FILE
...
                        PostgreSQL.
 
-    POSTGRES_PASSWORD  The password to use when authenticating with PostgreSQL
-                       as POSTGRES_USER.
+    POSTGRESQL_PASSWORD  The password to use when authenticating with PostgreSQL
+                       as POSTGRESQL_USER.
 
-    POSTGRES_DATABASE  The name of the PostgreSQL database to use for Guacamole
+    POSTGRESQL_DATABASE  The name of the PostgreSQL database to use for Guacamole
                        authentication.
```
mike-jumper, line 261: Beware that these have become misaligned with the addition of the QL.

guacamole-docker/bin/start.sh
jbpaux: Ok, so here we go I switched to `/bin/bash`. I also think it's better to make the list dynamic.

guacamole-docker/bin/start.sh
jmuehlner: The indented following lines should be aligned with the lines above, but that's no longer the case with the addition of the extra "QL" in the variable name. 

So the following lines, like "authentication.", should all be indented by 2 additional characters so that they line up again.

jbpaux: Oh ok I see and fixed. :) Should be good now. I'll tackle the guacamole-manual after that to reflect the changes.

necouchman:
Overall I think it looks okay to me. @mike-jumper any additional concerns?

necouchman (approved):

PR #722, opened 2022-04-27
GUACAMOLE-641: Ensure empty strings within KSM record fields are handled as if the field value is absent.

mike-jumper:
During testing, it was discovered that the fields of a KSM vault record that are initially set to a value and later cleared will be returned as an empty string, not `null`. This change alters the processing of these fields such that both the empty string and `null` are handled equivalently.

### DIFF
--- a/extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmRecordService.java
+++ b/extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmRecordService.java
@@ -113,6 +113,29 @@ public class KsmRecordService {
     }
 
     /**
+     * Returns the single value stored in the given list of strings. If the
+     * list is empty, contains multiple values, or contains only a single empty
+     * string, null is returned.
+     *
+     * @param values
+     *     The list to retrieve a single value from.
+     *
+     * @return
+     *     The single value stored in the given list, or null if the list is
+     *     empty, contains multiple values, or contains only a single empty
+     *     string.
+     */
+    private String getSingleStringValue(List<String> values) {
+
+        String value = getSingleValue(values);
+        if (value != null && !value.isEmpty())
+            return value;
+
+        return null;
+
+    }
+
+    /**
      * Returns the single value stored in the given list, additionally
      * performing a mapping transformation on the single value. If the list is
      * empty or contains multiple values, null is returned.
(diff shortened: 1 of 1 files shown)


extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmRecordService.java
```diff
@@ -112,6 +112,29 @@ private <T> T getSingleValue(List<T> values) {
 
     }
 
+    /**
+     * Returns the single value stored in the given list of strings. If the
```
jmuehlner, line 116: Shouldn't this comment say that `null` will be returned if the string is `null`? That's the whole point of this fix but it's left unclear in the block comments for these functions.

mike-jumper: To clarify, the point of the fix is that `null` should be returned if the string is the _empty string_, but yes it does say that. See immediately after the sentence in question:

> ... If the list is empty, contains multiple values, or contains only a single empty string, null is returned.

It's also listed in the `@return`:

```
     * @return
     *     The single value stored in the given list, or null if the list is
     *     empty, contains multiple values, or contains only a single empty
     *     string.
     */
```

jmuehlner: Ah, I see I had it backwards, I thought the fix was that it was now returning `null`, not that it is now returning an empty string. Gotcha.

That said, I still don't see where it says it will return `null` if the string is `null`. Your quoted text says it will return `null` in the following cases:

- The list is empty
- The list contains multiple values
- The list contains only a single empty string

My point is that it never says that it will return `null` if the string is `null`.

jmuehlner: I guess the logical assumption would be that it would return `null` in that case even if we don't explicitly state it, but if we're gonna list out all these other possibilities, I don't see why we shouldn't list this one as well.

mike-jumper: My thinking was that the documentation was already clear on that point. If the list contains a single `null`, then that's what's documented as returned ("the single value stored in the given list"), with the only exceptions to that rule being the case listed ("a single empty string").

mike-jumper: Looking at `Map`, where `get()` returns `null` if there is no value stored for a particular key, the Java folks feel it necessary to note this ambiguous case:

> A return value of null does not necessarily indicate that the map contains no mapping for the key; it's also possible that the map explicitly maps the key to null.

How about:

> Returns the single value stored in the given list of strings. If the list is empty, contains multiple values, contains only a single empty string, or contains only a single null value, null is returned.

?

mike-jumper: I've pushed commit 837a036 to add a clarifying note regarding the behavior of `null` values to each of the "getSingle" functions.

jmuehlner: LGTM

PR #723, opened 2022-05-02
GUACAMOLE-1594: Update jackson-databind to 2.13.2.1 and bump other dependencies as needed.

jmuehlner:


jmuehlner:
There were licenses for snakeyaml 1.27 and 1.28, but neither of those versions appear to be used anymore, so I removed them.

necouchman (approved):

mike-jumper:
I'm too late to the party here, but:

* I think this should be targeted against 1.5.0, since we need to do a general update-all-the-things task for 1.5.0.
* I think the relevant JIRA issue should be repurposed as said general update-all-the-things task.

We can probably just make the changes against 1.5.0 and git will be still happy to merge.

necouchman:
Oops, sorry about that!

PR #729, opened 2022-05-24
GUACAMOLE-1614: Upgrade to latest jasmine-maven-plugin version to avoid PhantomJS use entirely.

jmuehlner:
Per https://github.com/ariya/phantomjs/issues/15344, PhantomJS is not being actively maintained, and should no longer be used if possible.

The newest version of `jasmine-maven-plugin` does just that - using headless versions of Chrome and Firefox to run tests instead. Unfortunate that it's still a beta version, but it seems to be the best option for fixing this problem.

### DIFF
--- a/guacamole-common-js/pom.xml
+++ b/guacamole-common-js/pom.xml
@@ -118,7 +118,7 @@
             <plugin>
                 <groupId>com.github.searls</groupId>
                 <artifactId>jasmine-maven-plugin</artifactId>
-                <version>2.2</version>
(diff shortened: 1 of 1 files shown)


jmuehlner:
I've tested this fix on ubuntu 22.04 and RHEL 8.5.

mike-jumper (approved):
Seems reasonable. I'm not against using a beta version of a thing if said thing is specific to unit tests (especially when the build fails without it).

PR #731, opened 2022-06-03
(test) Add test for GuacamoleProtocolVersion.equals

lacinoire:
Hey 😊
I want to contribute the following test:

Test that `GuacamoleProtocolVersion.parseVersion("VERSION_012_102_398").equals(new Object())` is false when equals is called with the parameters `obj = new java.lang.Object()`. 
This tests the methods [`GuacamoleProtocolVersion.equals`](https://github.com/apache/guacamole-client/blob/7f707cdb04eb079b398c1391c8d5b17c9c9bb231/guacamole-common/src/main/java/org/apache/guacamole/protocol/GuacamoleProtocolVersion.java#L200).
The test is based on [`testValidVersionParse`](https://github.com/apache/guacamole-client/blob/7f707cdb04eb079b398c1391c8d5b17c9c9bb231/guacamole-common/src/test/java/org/apache/guacamole/protocol/GuacamoleProtocolVersionTest.java#L35).

Curious to hear what you think!

(I wrote this test as part of a research study at TU Delft. [Find out more](https://github.com/lacinoire/lacinoire/blob/main/README.md))

### DIFF
--- a/guacamole-common/src/test/java/org/apache/guacamole/protocol/GuacamoleProtocolVersionTest.java
+++ b/guacamole-common/src/test/java/org/apache/guacamole/protocol/GuacamoleProtocolVersionTest.java
@@ -126,6 +126,7 @@ public class GuacamoleProtocolVersionTest {
         Assert.assertFalse(version.equals(new GuacamoleProtocolVersion(12, 102, 399)));
         Assert.assertFalse(version.equals(new GuacamoleProtocolVersion(12, 103, 398)));
         Assert.assertFalse(version.equals(new GuacamoleProtocolVersion(11, 102, 398)));
+        Assert.assertFalse(version.equals(new Object()));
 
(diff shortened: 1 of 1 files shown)


necouchman:
@lacinoire: I guess I'm unsure what issue this is targeted at resolving? And why it is in its own method, rather than with all of the other tests in `testVersionEquals()`? Maybe I'm just missing something?

Also, any pull requests need a matching Jira issue: https://issues.apache.org/jira/projects/GUACAMOLE/issues

lacinoire:
@necouchman
I intend to test the path of the equals method checking that the other object is from another class.
Moved it to one assertion in `testVersionEquals()` 👍🏼

As this is just a small improvement to the test, similar to a documentation improvment, I wasn't sure it's worth opening a Jira issue.
Still happy to do it if you think that is better!

lacinoire:
Okay! Thank you for your feedback 🙂
Then I will close this Pull Request for now 👍🏼

PR #736, opened 2022-06-15
GUACAMOLE-1623: Extract domain field directly from the vault, or split out of username.

jmuehlner:
As discussed in https://issues.apache.org/jira/browse/GUACAMOLE-1623, this allows the KSM vault extension to 

* read a domain token out of records in the vault directly, 

* or to split the username field into domain and username tokens if no domain is directly defined, if the username matches the expected formats as seen at https://docs.microsoft.com/en-us/windows/win32/secauthn/user-name-formats#user-principal-name, and if the configuration property is enabled.

### DIFF
--- a/extensions/guacamole-vault/modules/guacamole-vault-base/src/main/java/org/apache/guacamole/vault/secret/WindowsUsername.java
+++ b/extensions/guacamole-vault/modules/guacamole-vault-base/src/main/java/org/apache/guacamole/vault/secret/WindowsUsername.java
@@ -1,0 +1,157 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.guacamole.vault.secret;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import javax.annotation.Nonnull;
+
+/**
+ * A class representing a Windows username, which may optionally also include
+ * a domain. This class can be used to parse the username and domain out of a
+ * username from a vault.
+ */
+public class WindowsUsername {
+
+    /**
+     * A pattern for matching a down-level logon name containing a Windows
+     * domain and username - e.g. domain\\user. For more information, see
+     * https://docs.microsoft.com/en-us/windows/win32/secauthn/user-name-formats#down-level-logon-name
+     */
+    private static final Pattern DOWN_LEVEL_LOGON_NAME_PATTERN = Pattern.compile(
+            "(?<domain>[^@\\\\]+)\\\\(?<username>[^@\\\\]+)");
+
+    /**
+     * A pattern for matching a user principal name containing a Windows
+     * domain and username - e.g. user@domain. For more information, see
+     * https://docs.microsoft.com/en-us/windows/win32/secauthn/user-name-formats#user-principal-name
+     */
+    private static final  Pattern USER_PRINCIPAL_NAME_PATTERN = Pattern.compile(
+            "(?<username>[^@\\\\]+)@(?<domain>[^@\\\\]+)");
+
+    /**
+     * The username associated with the potential Windows domain/username
+     * value. If no domain is found, the username field will contain the
+     * entire value as read from the vault.
+     */
+    private final String username;
+
+    /**
+     * The dinaun associated with the potential Windows domain/username
+     * value. If no domain is found, this will be null.
+     */
+    private final String domain;
+
+    /**
+     * Create a WindowsUsername record with no associated domain.
+     *
+     * @param username
+     *     The username, which should be the entire value as extracted
+     *     from the vault.
+     */
+    private WindowsUsername(@Nonnull String username) {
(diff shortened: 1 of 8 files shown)


extensions/guacamole-vault/modules/guacamole-vault-base/src/main/java/org/apache/guacamole/vault/util/FieldParsingUtil.java
```diff
@@ -0,0 +1,163 @@
...
+        Matcher userPrincipalMatcher = USER_PRINCIPAL_NAME_PATTERN.matcher(vaultField);
+        if (userPrincipalMatcher.matches())
+            return new DomainAndUsername(
+                    userPrincipalMatcher.group("username"),
+                    userPrincipalMatcher.group("domain"));
+
+        // If none of the expected formats matched, return the username with do domain
+        return new DomainAndUsername(vaultField);
+
+    }
```
mike-jumper, line 162: Rather than establish a general parsing utility class for all future parsing needs, I suggest instead just defining `DomainAndUsername` and let it do-one-thing-and-do-it-well. ie:

`DomainAndUsername.split(someFieldValue)`

rather than:

`FieldParsingUtil.DomainAndUsername` and `FieldParsingUtil.splitWindowsUsernameFromDomain(someFieldValue)`.

extensions/guacamole-vault/modules/guacamole-vault-base/src/test/java/org/apache/guacamole/vault/util/FieldParsingUtilTest.java
```diff
@@ -0,0 +1,82 @@
...
+        assertEquals("localhost", usernameAndDomain.getDomain());
+
+        // It should parse user principal name style domains
+        usernameAndDomain = FieldParsingUtil.splitWindowsUsernameFromDomain("bob@localhost");
+        assertEquals("bob", usernameAndDomain.getUsername(), "bob");
+        assertTrue(usernameAndDomain.hasDomain());
+        assertEquals("localhost", usernameAndDomain.getDomain());
+
+        // It should not match if there are an invalid number of seperators
+        List<String> invalidSeperators = List.of(
```
mike-jumper, line 64: separator*

mike-jumper (changes requested):

extensions/guacamole-vault/modules/guacamole-vault-base/src/main/java/org/apache/guacamole/vault/util/FieldParsingUtil.java
jmuehlner: Sure

jmuehlner: It's done.

extensions/guacamole-vault/modules/guacamole-vault-base/src/main/java/org/apache/guacamole/vault/secret/DomainAndUsername.java
```diff
@@ -0,0 +1,157 @@
...
+package org.apache.guacamole.vault.secret;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import javax.annotation.Nonnull;
+
+/**
+ * A class that implements to attempts to split a username from a vault into a
+ * Windows domain and username, and stores the result of this operation.
```
mike-jumper, line 29: > A class that implements to attempts to split a username ...

"Implements to attempts" sounds odd to me, but either way - this class is a representation of a Windows username, no? It does have a convenience method for splitting a username, but the class itself is the representation of a username and optional domain, not specifically a domain+username parser.

extensions/guacamole-vault/modules/guacamole-vault-base/src/main/java/org/apache/guacamole/vault/util/FieldParsingUtil.java
```diff
@@ -0,0 +1,163 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+/**
+ * A set of utility methods for parsing field values for records retrieved
+ * from vaults.
+ */
+public class FieldParsingUtil {
```
mike-jumper, line 31: With `DomainAndUsername` now existing on its own, this class should be deleted.

mike-jumper (changes requested):

extensions/guacamole-vault/modules/guacamole-vault-base/src/main/java/org/apache/guacamole/vault/util/FieldParsingUtil.java
jmuehlner: Ack, sorry. I thought I had done that.

extensions/guacamole-vault/modules/guacamole-vault-base/src/main/java/org/apache/guacamole/vault/secret/DomainAndUsername.java
jmuehlner: Lemme reword the comment (and rename the class) to make things a bit clearer.

mike-jumper: Looks much better.

extensions/guacamole-vault/modules/guacamole-vault-base/src/main/java/org/apache/guacamole/vault/secret/WindowsUsername.java
```diff
@@ -0,0 +1,156 @@
...
+     */
+    public String getDomain() {
+        return domain;
+    }
+
+    /**
+     * Return true if a domain was found in the vault field, false otherwise.
+     *
+     * @return
+     */
```
mike-jumper, line 116: Oops - missing the description here.

jmuehlner: Fixed!

mike-jumper (approved):

PR #739, opened 2022-06-29
GUACAMOLE-1629: Allow vault configurations per connection group

jmuehlner:


extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
```diff
@@ -153,24 +180,82 @@ private void addRecordTokens(Map<String, Future<String>> tokens, String prefix,
...
+        // For connection groups, start searching the group directly
+        String parentIdentifier = (connectable instanceof Connection)
+                ? ((Connection) connectable).getParentIdentifier()
+                : ((ConnectionGroup) connectable).getIdentifier();
+
+        Directory<ConnectionGroup> connectionGroupDirectory = userContext.getConnectionGroupDirectory();
+        while (connectionGroupDirectory.getIdentifiers().contains(parentIdentifier)) {
+
+            // Fetch the parent group
+            ConnectionGroup group = connectionGroupDirectory.get(parentIdentifier);
```
mike-jumper, line 223: `getIdentifiers()` can be a heavy function, as it lists everything. Instead, since `get()` will return `null` if there is no such object, we should just call `get()` directly.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
```diff
@@ -56,280 +45,87 @@
...
+     *     cache map, otherwise the existing cache entry.
      *
-     * @param hostname
-     *     The hostname/address that the given record should be associated
-     *     with. This may be null.
+     * @throws GuacamoleException
+     *     If an error occurs while creating the KSM cache.
      */
-    private void addRecordForHost(KeeperRecord record, String hostname) {
+    private KsmCache createCacheIfNeeded(@Nullable String ksmConfig)
```
mike-jumper, line 83: If abstracting things as a cache: I suggest renaming to `getCache()` and letting the creation aspect of this be an internal concern.

That said, I think perhaps this should instead be abstracted at the `KsmClient` level. If the idea is that a particular connection might retrieve secrets using a different KSM config, I think it would make more sense for each config produce a dedicated, separate `KsmClient` singleton, rather than swapping out the internal cache of that client based on which config is in use.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
```diff
@@ -56,280 +45,87 @@
...
+        // If a cache already exists for the provided config, use it
+        KsmCache ksmCache = ksmCacheMap.get(ksmConfig);
+        if (ksmCache != null)
+            return ksmCache;
 
+        // Create and store a new KSM cache instance for the provided KSM config blob
+        SecretsManagerOptions options = confService.getSecretsManagerOptions(ksmConfig);
+        ksmCache = ksmCacheFactory.create(options);
+        ksmCacheMap.put(ksmConfig, ksmCache);
+        return ksmCache;
```
mike-jumper, line 95: Is this threadsafe?

mike-jumper (changes requested):

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
jmuehlner: Good call - will do.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
jmuehlner: uh, no.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
jmuehlner: That's fair - refactored to work that way instead. How does this look?

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
jmuehlner: Switched to using a concurrent map, which also required some refactoring around the way that properties are handled. Let me know what you think.

mike-jumper: Looks good - even in the case that `get()` returns `null` for two interleaved calls and two `KsmClient` instances get created, only one of those instances will win, everything requesting an instance will end up with the same instance, and the constructor that gets arguably unnecessarily invoked is very low overhead.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
mike-jumper: I think this looks much better.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/conf/KsmAttributeService.java
```diff
@@ -0,0 +1,63 @@
...
+     * associated with a connection group.
+     */
+    public static final String KSM_CONFIGURATION_ATTRIBUTE = "ksm-config";
+
+    /**
+     * All attributes related to configuring the KSM vault on a
+     * per-connection-group basis.
+     */
+    public static final Form KSM_CONFIGURATION_FORM = new Form("ksm-config",
+            Arrays.asList(new MultilineField(KSM_CONFIGURATION_ATTRIBUTE)));
```
jmuehlner, line 50: What do people think: should this be a `MultilineField` or a `TextField`? The base-64 encoded JSON configuration blob won't have any newlines in it, so this is what it looks like in practice:

image

necouchman: If it's going to just expand like that, anyway, maybe just a standard `TextField`?

jmuehlner: Yeah, I think this is probably better.
image

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
```diff
@@ -246,7 +258,7 @@ private void validateCache() throws GuacamoleException {
 
             });
 
-            // Cache has been refreshed
+            // Client has been refreshed
```
mike-jumper, line 261: Cache*

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
```diff
@@ -60,6 +76,53 @@ public class KsmSecretService implements VaultSecretService {
...
 
+    /**
+     * Factory for creating KSM client instances.
+     */
+    @Inject
+    private KsmClientFactory ksmClientFactory;
+
+    /**
+     * A map of base-64 encoded JSON KSM config blobs to associated KSM client instances.
+     * The `null` entry in this Map is associated with the KSM configuration parsed
```
mike-jumper, line 87: I think this is incorrect / from an older version of this code: there can't be a `null` entry in this `Map` as it's a `ConcurrentHashMap`.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
```diff
@@ -60,6 +76,53 @@ public class KsmSecretService implements VaultSecretService {
...
+     * from the guacamole.properties config file. A distinct KSM client will exist for
+     * every KSM config.
+     */
+    private final ConcurrentMap<String, KsmClient> ksmClientMap = new ConcurrentHashMap<>();
+
+    /**
+     * Create and return a KSM cache for the provided KSM config if not already
```
mike-jumper, line 94: client*

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
```diff
@@ -60,6 +76,53 @@ public class KsmSecretService implements VaultSecretService {
...
+     * present in the cache map, otherwise return the existing cache entry.
+     *
+     * @param ksmConfig
+     *     The base-64 encoded JSON KSM config blob associated with the cache entry.
+     *     If an associated entry does not already exist, it will be created using
+     *     this configuration.
+     *
+     * @return
+     *     A KSM cache for the provided KSM config if not already present in the
```
mike-jumper, line 103: client*

mike-jumper (changes requested):

extensions/guacamole-vault/modules/guacamole-vault-base/src/main/java/org/apache/guacamole/vault/user/VaultUserContext.java
```diff
@@ -398,9 +421,19 @@ protected void addTokens(Connection connection, Map<String, String> tokens)
...
+    }
+
+    @Override
+    public Collection<Form> getConnectionGroupAttributes() {
 
+        // Add any custom attributes to any previously defined attributes
+        return Stream.concat(
+                super.getConnectionGroupAttributes().stream(),
+                attributeService.getConnectionGroupAttributes().stream()
+        ).collect(Collectors.toUnmodifiableList());
```
mike-jumper, line 436: `Collectors.toUnmodifiableList()` is only available in JDK 10+.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
```diff
@@ -60,6 +76,53 @@ public class KsmSecretService implements VaultSecretService {
...
+     * A map of base-64 encoded JSON KSM config blobs to associated KSM client instances.
+     * The `null` entry in this Map is associated with the KSM configuration parsed
+     * from the guacamole.properties config file. A distinct KSM client will exist for
+     * every KSM config.
+     */
+    private final ConcurrentMap<String, KsmClient> ksmClientMap = new ConcurrentHashMap<>();
+
+    /**
+     * Create and return a KSM cache for the provided KSM config if not already
+     * present in the cache map, otherwise return the existing cache entry.
```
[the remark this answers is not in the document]
mike-jumper: Shouldn't this be client map and client entry?

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
```diff
@@ -60,6 +76,51 @@ public class KsmSecretService implements VaultSecretService {
...
+     *     A KSM client for the provided KSM config if not already present in the
+     *     cache map, otherwise the existing cache entry.
+     *
+     * @throws GuacamoleException
+     *     If an error occurs while creating the KSM client.
+     */
+    private KsmClient getClient(@Nonnull String ksmConfig)
+            throws GuacamoleException {
+
+        // If a cache already exists for the provided config, use it
```
mike-jumper, line 110: client*

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
```diff
@@ -60,6 +76,51 @@ public class KsmSecretService implements VaultSecretService {
...
+        KsmClient ksmClient = ksmClientMap.get(ksmConfig);
+        if (ksmClient != null)
+            return ksmClient;
+
+        // Create and store a new KSM cache instance for the provided KSM config blob
```
mike-jumper, line 115: client*

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
```diff
@@ -60,6 +76,51 @@ public class KsmSecretService implements VaultSecretService {
...
+     * present in the cache map, otherwise return the existing cache entry.
+     *
+     * @param ksmConfig
+     *     The base-64 encoded JSON KSM config blob associated with the cache entry.
+     *     If an associated entry does not already exist, it will be created using
+     *     this configuration.
+     *
+     * @return
+     *     A KSM client for the provided KSM config if not already present in the
+     *     cache map, otherwise the existing cache entry.
```
mike-jumper, line 102: Should each of these references to the cache map, entry, etc. be s/cache/client/?

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
```diff
@@ -60,6 +76,51 @@ public class KsmSecretService implements VaultSecretService {
...
+        KsmClient ksmClient = ksmClientMap.get(ksmConfig);
+        if (ksmClient != null)
+            return ksmClient;
+
+        // Create and store a new KSM cache instance for the provided KSM config blob
+        SecretsManagerOptions options = confService.getSecretsManagerOptions(ksmConfig);
+        ksmClient = ksmClientFactory.create(options);
+        KsmClient prevClient = ksmClientMap.putIfAbsent(ksmConfig, ksmClient);
+
+        // If the cache was already set before this thread got there, use the existing one
```
mike-jumper, line 120: client*

extensions/guacamole-vault/modules/guacamole-vault-base/src/main/java/org/apache/guacamole/vault/conf/VaultAttributeService.java
```diff
@@ -0,0 +1,43 @@
...
+public interface VaultAttributeService {
+
+    /**
+     * Return all custom connection group attributes to be exposed through the
+     * admin UI for the current vault implementation.
+     *
+     * @return
+     *     All custom connection group attributes to be exposed through the
+     *     admin UI for the current vault implementation.
+     *
```
mike-jumper, line 40: Please remove this trailing blank line.

mike-jumper (changes requested):

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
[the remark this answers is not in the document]
jmuehlner: Ugh, yes. I _think_ I got them all now.

extensions/guacamole-vault/modules/guacamole-vault-base/src/main/java/org/apache/guacamole/vault/user/VaultUserContext.java
jmuehlner: Oops! Good catch. Fixed.

PR #741, opened 2022-07-01
GUACAMOLE-1372: Support signing/encryption keys for SAML auth extension.

jmuehlner:
https://issues.apache.org/jira/browse/GUACAMOLE-1372

### DIFF
--- a/extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/src/main/java/org/apache/guacamole/auth/saml/conf/ConfigurationService.java
+++ b/extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/src/main/java/org/apache/guacamole/auth/saml/conf/ConfigurationService.java
@@ -24,7 +24,12 @@ import com.onelogin.saml2.settings.IdPMetadataParser;
 import com.onelogin.saml2.settings.Saml2Settings;
 import com.onelogin.saml2.settings.SettingsBuilder;
 import com.onelogin.saml2.util.Constants;
+
+import java.io.File;
+import java.io.IOException;
 import java.net.URI;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.util.HashMap;
 import java.util.Map;
 import javax.ws.rs.core.UriBuilder;
@@ -32,15 +37,23 @@ import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.GuacamoleServerException;
 import org.apache.guacamole.environment.Environment;
 import org.apache.guacamole.properties.BooleanGuacamoleProperty;
+import org.apache.guacamole.properties.FileGuacamoleProperty;
 import org.apache.guacamole.properties.IntegerGuacamoleProperty;
 import org.apache.guacamole.properties.StringGuacamoleProperty;
 import org.apache.guacamole.properties.URIGuacamoleProperty;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Service for retrieving configuration information regarding the SAML
  * authentication module.
  */
 public class ConfigurationService {
+
+    /**
+     * Logger for this class.
+     */
+    private static final Logger logger = LoggerFactory.getLogger(ConfigurationService.class);
 
     /**
      * The URI of the file containing the XML Metadata associated with the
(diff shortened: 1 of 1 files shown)


extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/src/main/java/org/apache/guacamole/auth/saml/conf/ConfigurationService.java
```diff
@@ -329,6 +366,69 @@ public int getAuthenticationTimeout() throws GuacamoleException {
...
+     * Returns the file containing the x509 certificate to use when signing
+     * requests to the SAML IdP. If the property is not set, null will be
+     * returned.
+     *
+     * @return
+     *     The file containing the x509 certificate to use when signing
+     *     requests to the SAML IdP, or null if not defined.
+     *
+     * @throws GuacamoleException
+     *     If the authentication timeout cannot be parsed.
```
mike-jumper, line 379: X.509 certificate*

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/src/main/java/org/apache/guacamole/auth/saml/conf/ConfigurationService.java
```diff
@@ -329,6 +366,69 @@ public int getAuthenticationTimeout() throws GuacamoleException {
...
+     * Returns the file containing the private key to use when signing
+     * requests to the SAML IdP. If the property is not set, null will be
+     * returned.
+     *
+     * @return
+     *     The file containing the private key to use when signing
+     *     requests to the SAML IdP, or null if not defined.
+     *
+     * @throws GuacamoleException
+     *     If the authentication timeout cannot be parsed.
```
mike-jumper, line 395: private key file*

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/src/main/java/org/apache/guacamole/auth/saml/conf/ConfigurationService.java
```diff
@@ -329,6 +366,69 @@ public int getAuthenticationTimeout() throws GuacamoleException {
...
+    private String readFileContentsIntoString(File file, String name) {
+
+        // Attempt to read the file directly into a String
+        try {
+            return new String(Files.readAllBytes(file.toPath()), StandardCharsets.UTF_8);
+        }
+
+        // If the file cannot be read, log a warning and treat it as if it does not exist
+        catch (IOException e) {
+            logger.warn("{} \"{}\" could not be read.", name, file);
```
mike-jumper, line 425: We should log the exception message here, as well, so that the admin can see the nature of the failure. For example:

https://github.com/apache/guacamole-client/blob/754e9649f1fa0ba225ee42b56ded64bc283d17df/guacamole/src/main/java/org/apache/guacamole/tunnel/InterceptedStreamMap.java#L71-L74

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/src/main/java/org/apache/guacamole/auth/saml/conf/ConfigurationService.java
```diff
@@ -329,6 +366,69 @@ public int getAuthenticationTimeout() throws GuacamoleException {
...
+            logger.debug("{} \"{}\" could not be read.", name, file, e);
+            return null;
```
mike-jumper, line 427: Should this instead be fatal, ultimately causing the call to `getSamlSettings()` to fail if required files can't be read?

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/src/main/java/org/apache/guacamole/auth/saml/conf/ConfigurationService.java
```diff
@@ -161,6 +174,30 @@ public class ConfigurationService {
...
+    private static final FileGuacamoleProperty SAML_X509_CERT_PATH =
+            new FileGuacamoleProperty() {
+
+        @Override
+        public String getName() { return "saml-x509-cert-path"; }
+
+    };
+
+    /**
+     * The file containing the private to use when signing or encrypting
```
mike-jumper, line 190: private key*

mike-jumper (changes requested):

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/src/main/java/org/apache/guacamole/auth/saml/conf/ConfigurationService.java
jmuehlner: Am I not already doing that?

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/src/main/java/org/apache/guacamole/auth/saml/conf/ConfigurationService.java
jmuehlner: Should it? They're not required properties.

mike-jumper: They're not required, yes, but when the admin _does_ specify them, they're declaring that they want the SAML messages to be signed/encrypted using that key. If the admin explicitly requests that messages be signed/encrypted with a particular key, but that key can't be read, I think we should fail outright rather than continue as if no key was provided.

jmuehlner: Required now.

PR #743, opened 2022-07-20
GUACAMOLE-1643: Allow usage of one-time KSM tokens when editing connection groups

jmuehlner:
This change allows a user to put KSM one-time tokens directly into the connection group edit screen. If the token is invalid, an error will be displayed to the user.

This change will also validate the base64 KSM config blob at save time.

There's still no way to use a one-time-token directly in the `guacamole.properties` config file since the webapp can't rewrite that.

### DIFF
--- a/extensions/guacamole-vault/modules/guacamole-vault-base/src/main/java/org/apache/guacamole/vault/user/VaultDirectoryService.java
+++ b/extensions/guacamole-vault/modules/guacamole-vault-base/src/main/java/org/apache/guacamole/vault/user/VaultDirectoryService.java
@@ -1,0 +1,140 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.guacamole.vault.user;
+
+import org.apache.guacamole.GuacamoleException;
+import org.apache.guacamole.net.auth.ActiveConnection;
+import org.apache.guacamole.net.auth.Connection;
+import org.apache.guacamole.net.auth.ConnectionGroup;
+import org.apache.guacamole.net.auth.Directory;
+import org.apache.guacamole.net.auth.SharingProfile;
+import org.apache.guacamole.net.auth.User;
+import org.apache.guacamole.net.auth.UserGroup;
+
+/**
+ * A service that allows a vault implementation to override the directory
+ * for any entity that a user context may return.
+ */
+public abstract class VaultDirectoryService {
+
+    /**
+     * Given an existing User Directory, return a new Directory for
+     * this vault implementation.
+     *
+     * @return
+     *     A new User Directory based on the provided Directory.
+     *
+     * @throws GuacamoleException
+     *     If an error occurs while creating the Directory.
+     */
+    public Directory<User> getUserDirectory(
+            Directory<User> underlyingDirectory) throws GuacamoleException {
+
+        // By default, the provided directly will be returned unchanged
+        return underlyingDirectory;
+    }
+
+    /**
+     * Given an existing UserGroup Directory, return a new Directory for
+     * this vault implementation.
+     *
+     * @return
+     *     A new UserGroup Directory based on the provided Directory.
+     *
+     * @throws GuacamoleException
+     *     If an error occurs while creating the Directory.
+     */
+    public Directory<UserGroup> getUserGroupDirectory(
+            Directory<UserGroup> underlyingDirectory) throws GuacamoleException {
+
+        // Unless overriden in the vault implementation, the underlying directory
+        // will be returned directly
+        return underlyingDirectory;
+    }
+
+    /**
+     * Given an existing Connection Directory, return a new Directory for
+     * this vault implementation.
+     *
+     * @return
+     *     A new Connection Directory based on the provided Directory.
+     *
+     * @throws GuacamoleException
+     *     If an error occurs while creating the Directory.
+     */
+    public Directory<Connection> getConnectionDirectory(
+            Directory<Connection> underlyingDirectory) throws GuacamoleException {
+
+        // By default, the provided directly will be returned unchanged
+        return underlyingDirectory;
+    }
+
+    /**
+     * Given an existing ConnectionGroup Directory, return a new Directory for
+     * this vault implementation.
+     *
+     * @return
+     *     A new ConnectionGroup Directory based on the provided Directory.
+     *
+     * @throws GuacamoleException
+     *     If an error occurs while creating the Directory.
+     */
+    public Directory<ConnectionGroup> getConnectionGroupDirectory(
+            Directory<ConnectionGroup> underlyingDirectory) throws GuacamoleException {
+
+        // By default, the provided directly will be returned unchanged
+        return underlyingDirectory;
+    }
+
+    /**
+     * Given an existing ActiveConnection Directory, return a new Directory for
+     * this vault implementation.
+     *
+     * @return
+     *     A new ActiveConnection Directory based on the provided Directory.
+     *
+     * @throws GuacamoleException
+     *     If an error occurs while creating the Directory.
+     */
+    public Directory<ActiveConnection> getActiveConnectionDirectory(
+            Directory<ActiveConnection> underlyingDirectory) throws GuacamoleException {
+
+        // By default, the provided directly will be returned unchanged
+        return underlyingDirectory;
+    }
+
+    /**
+     * Given an existing SharingProfile Directory, return a new Directory for
+     * this vault implementation.
+     *
+     * @return
+     *     A new SharingProfile Directory based on the provided Directory.
+     *
+     * @throws GuacamoleException
+     *     If an error occurs while creating the Directory.
+     */
+    public Directory<SharingProfile> getSharingProfileDirectory(
+            Directory<SharingProfile> underlyingDirectory) throws GuacamoleException {
+
+        // By default, the provided directly will be returned unchanged
+        return underlyingDirectory;
+    }
+
+}
--- a/extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
+++ b/extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
@@ -1,0 +1,253 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.guacamole.vault.ksm.user;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.Base64;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.guacamole.GuacamoleException;
+import org.apache.guacamole.language.TranslatableGuacamoleClientException;
+import org.apache.guacamole.net.auth.Attributes;
+import org.apache.guacamole.net.auth.ConnectionGroup;
+import org.apache.guacamole.net.auth.DelegatingDirectory;
+import org.apache.guacamole.net.auth.Directory;
+import org.apache.guacamole.vault.ksm.conf.KsmAttributeService;
+import org.apache.guacamole.vault.ksm.conf.KsmConfig;
+import org.apache.guacamole.vault.ksm.conf.KsmConfigurationService;
+import org.apache.guacamole.vault.user.VaultDirectoryService;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.inject.Inject;
+import com.keepersecurity.secretsManager.core.InMemoryStorage;
+import com.keepersecurity.secretsManager.core.SecretsManager;
+import com.keepersecurity.secretsManager.core.SecretsManagerOptions;
+
+/**
+ * A KSM-specific vault directory service that wraps the connection group directory
+ * to enable automatic translation of KSM one-time tokens into base64-encoded JSON
+ * config bundles.
+ */
+public class KsmDirectoryService extends VaultDirectoryService {
+
+    /**
+     * Service for retrieving KSM configuration details.
+     */
+    @Inject
+    private KsmConfigurationService configurationService;
+
+    /**
+     * A singleton ObjectMapper for converting a Map to a JSON string when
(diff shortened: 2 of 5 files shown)


mike-jumper:
> There's still no way to use a one-time-token directly in the `guacamole.properties` config file since the webapp can't rewrite that.

Makes sense.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/resources/translations/en.json
```diff
@@ -6,7 +6,10 @@
 
     "CONNECTION_GROUP_ATTRIBUTES" : {
         "SECTION_HEADER_KSM_CONFIG" : "Keeper Secrets Manager",
-        "FIELD_HEADER_KSM_CONFIG"   : "KSM Service Configuration "
+        "FIELD_HEADER_KSM_CONFIG"   : "KSM Service Configuration ",
+
+        "ERROR_INVALID_KSM_CONFIG_BLOB": "Invalid KSM config blob provided",
+        "ERROR_INVALID_KSM_ONE_TIME_TOKEN": "Invalid KSM one-time token provided"
```
mike-jumper, line 12: I think we should make these a bit more friendly and helpful, so that the admin has a good idea what has failed and how to resolve. ie: "The provided base64-encoded KSM configuration blob is not valid. Please verify that you have copied the entire blob." or "The provided configuration is not a valid KSM one-time token or base64-encoded configuration blob. Please verify ... etc."

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
```diff
@@ -0,0 +1,241 @@
...
+    /**
+     * Given an attributes-enabled entity, check for the presence of the
+     * KSM_CONFIGURATION_ATTRIBUTE attribute. If it's set, check if it's a valid
+     * KSM one-time token. If so, attempt to translate it to a base-64-encoded
+     * json KSM config blob, and set it back to the provided entity.
+     * If it's already a KSM config blob, validate it as config blob. If either
+     * validation fails, a GuacamoleException will be thrown.
+     *
+     * @param entity
+     *     The attributes-enable entity for which the KSM configuration
```
mike-jumper, line 101: What do you mean by "attributes-enable" entity?

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
```diff
@@ -0,0 +1,241 @@
...
+                    // If the parsing attempt fails, throw a translatable error for display
+                    // on the frontend
+                    throw new TranslatableGuacamoleClientException(
+                            "Invalid base64-encoded JSON KSM config provided for "
+                            + KsmAttributeService.KSM_CONFIGURATION_ATTRIBUTE + " attribute",
+                            "CONNECTION_GROUP_ATTRIBUTES.ERROR_INVALID_KSM_CONFIG_BLOB");
+                }
+            }
+
+            else {
```
mike-jumper, line 144: This `else` is superfluous as the related `if` is terminal. This function can be made a bit more readable by removing this and thus shedding one level of nesting.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
```diff
@@ -0,0 +1,241 @@
...
+                }
+
+                catch (GuacamoleException exception) {
+
+                    // If the parsing attempt fails, throw a translatable error for display
+                    // on the frontend
+                    throw new TranslatableGuacamoleClientException(
+                            "Invalid base64-encoded JSON KSM config provided for "
+                            + KsmAttributeService.KSM_CONFIGURATION_ATTRIBUTE + " attribute",
+                            "CONNECTION_GROUP_ATTRIBUTES.ERROR_INVALID_KSM_CONFIG_BLOB");
```
mike-jumper, line 140: As there's an associated cause (`exception`), we should use the constructor variation that includes the cause:

https://github.com/apache/guacamole-client/blob/8772207a7583e0152581a6edd866e1b4d1783b1e/guacamole-ext/src/main/java/org/apache/guacamole/language/TranslatableGuacamoleClientException.java#L95

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
```diff
@@ -0,0 +1,241 @@
...
+
+            // If it's already base-64-encoded, it's a KSM configuration blob,
+            // so validate it immediately
+            if(isBase64(ksmConfigValue)) {
+
+                // Attempt to validate the config as a base64-econded KSM config blob
+                try {
+                    KsmConfig.parseKsmConfig(ksmConfigValue);
+
+                    // If it validates, the entity can be returned as-is
```
mike-jumper, line 129: This doesn't make sense to me:
1. But we're not returning anything.
2. What entity?

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
```diff
@@ -0,0 +1,241 @@
...
+                        // Only write the value into the new map if non-null
+                        String value = storage.getString(configKey);
+                        if (value != null)
+                            configMap.put(configKey, value);
+
+                    });
+
+                    // JSON-encode the value, and then base64 encode that to get the format
+                    // that KSM would expect
+                    String jsonString = new ObjectMapper().writeValueAsString(configMap);
```
mike-jumper, line 180: Best practices for Jackson suggest using a singleton `ObjectMapper`.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
```diff
@@ -0,0 +1,241 @@
...
+                    String base64EncodedJson = Base64.getEncoder().encodeToString(jsonString.getBytes());
+
+                    // Finally, try to parse the newly generated token as a KSM config. If this
+                    // works, the config should be fully functional
+                    KsmConfig.parseKsmConfig(base64EncodedJson);
+
+                    // Make a copy of the existing attributes, modifying just the value for
+                    // KSM_CONFIGURATION_ATTRIBUTE
+                    attributes = new HashMap<>(entity.getAttributes());
```
mike-jumper, line 189: Why not `new HashMap<>(attributes)`? Re-invoking `getAttributes()` will result in any associated processing taking place again.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
```diff
@@ -0,0 +1,241 @@
...
+
+                // The KSM SDK only throws raw Exceptions, so we can't be more specific
+                catch (Exception exception) {
+
+                    // If the parsing attempt fails, throw a translatable error for display
+                    // on the frontend
+                    throw new TranslatableGuacamoleClientException(
+                            "Invalid one-time KSM token provided for "
+                            + KsmAttributeService.KSM_CONFIGURATION_ATTRIBUTE + " attribute",
+                            "CONNECTION_GROUP_ATTRIBUTES.ERROR_INVALID_KSM_ONE_TIME_TOKEN");
```
mike-jumper, line 206: Here, too - the underlying exception should be included in the constructor as the cause.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
```diff
@@ -0,0 +1,241 @@
...
+                        String value = storage.getString(configKey);
+                        if (value != null)
+                            configMap.put(configKey, value);
+
+                    });
+
+                    // JSON-encode the value, and then base64 encode that to get the format
+                    // that KSM would expect
+                    String jsonString = new ObjectMapper().writeValueAsString(configMap);
+                    String base64EncodedJson = Base64.getEncoder().encodeToString(jsonString.getBytes());
```
mike-jumper, line 181: This will use the platform's default charset, whatever that happens to be. We should instead explicitly specify UTF-8 (assuming that's the intent).

mike-jumper (changes requested):

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
jmuehlner: Oops, was supposed to be `enabled`

extensions/guacamole-vault/modules/guacamole-vault-base/src/main/java/org/apache/guacamole/vault/user/VaultDirectoryService.java
```diff
@@ -0,0 +1,140 @@
...
+     * Given an existing User Directory, return a new Directory for
+     * this vault implementation.
+     *
+     * @return
+     *     A new User Directory based on the provided Directory.
+     *
+     * @throws GuacamoleException
+     *     If an error occurs while creating the Directory.
+     */
+    default Directory<User> getUserDirectory(
```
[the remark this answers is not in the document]
jmuehlner: Sure

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
jmuehlner: Good call.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
jmuehlner: Good catch on it not returning anything - I'll fix that.

The entity is `entity` - the `Attributes` instance that's passed into `processAttributes()`. I thought that was relatively clear?

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/resources/translations/en.json
jmuehlner: Sure.

jmuehlner: How do these look?

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
jmuehlner: Is it still unclear?

mike-jumper: > Good catch on it not returning anything - I'll fix that.
> 
> The entity is `entity` - the `Attributes` instance that's passed into `processAttributes()`. I thought that was relatively clear?

I think I just got hung up on "entity" being how the DB refers to the common base representation of a user / user group.

> Is it still unclear?

I think this looks good now.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/resources/translations/en.json
mike-jumper: These look much better.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/resources/translations/en.json
```diff
@@ -6,7 +6,10 @@
 
     "CONNECTION_GROUP_ATTRIBUTES" : {
         "SECTION_HEADER_KSM_CONFIG" : "Keeper Secrets Manager",
-        "FIELD_HEADER_KSM_CONFIG"   : "KSM Service Configuration "
+        "FIELD_HEADER_KSM_CONFIG"   : "KSM Service Configuration ",
+
+        "ERROR_INVALID_KSM_CONFIG_BLOB": "The provided base64-encoded KSM configuration blob is not valid. Please ensure that you have copied the entire blob.",
+        "ERROR_INVALID_KSM_ONE_TIME_TOKEN": "The provided configuration is not a valid KSM one-time token or base64-encoded configuration blob. Please ensure that you have copied the entire token value."
```
mike-jumper, line 12: Can you update to match the style of the existing JSON? (ie: spaces around `:`)

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
```diff
@@ -0,0 +1,251 @@
...
+
+/**
+ * A KSM-specific vault directory service that wraps the connection group directory
+ * to enable automatic translation of KSM one-time tokens into base64-encoded JSON
+ * config bundles.
+ */
+public class KsmDirectoryService extends VaultDirectoryService {
+
+    /**
+     * Service for retreiving KSM configuration details.
```
mike-jumper, line 54: retrieving*

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
```diff
@@ -0,0 +1,251 @@
...
+     */
+    @Inject
+    private KsmConfigurationService configurationService;
+
+    /**
+     * A singleton Objectmapper for converting a Map to a JSON string when
```
mike-jumper, line 60: ObjectMapper*

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
```diff
@@ -0,0 +1,251 @@
...
+     * generating a base64-encoded JSON KSM config blob.
+     */
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+
+    /**
+     * All expected fields in the KSM configuration JSON blob.
+     */
+    private static final List<String> EXPECTED_KSM_FIELDS = Arrays.asList(
```
mike-jumper, line 68: As this is intended to be immutable, I suggest wrapping this in `Collections.unmodifiableList(...)`.

mike-jumper (changes requested):
A few nits, but otherwise I think this LGTM.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/resources/translations/en.json
jmuehlner: yes

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
jmuehlner: Oops I will fix.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
jmuehlner: Uppercase "M" coming in hot.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
jmuehlner: Sure.

mike-jumper (approved):

PR #746, opened 2022-07-25
GUACAMOLE-1320: Provide chunked file upload mechanism

elijahnicpon:
File uploads are split into chunks so that request size limits imposed by CDNs (such as Cloudflare's 100mb limit) are no longer an issue. 

The GuacamoleWriter responsible for writing to the stream stays open until the end of the upload rather than being closed by an opcode upon completion of a POST request.

### DIFF
--- a/guacamole/src/main/frontend/src/app/rest/services/tunnelService.js
+++ b/guacamole/src/main/frontend/src/app/rest/services/tunnelService.js
@@ -393,7 +393,11 @@ angular.module('rest').factory('tunnelService', ['$injector',
         };
 
         /**
-         * Handler for the upload process. 
+         * Handles the recursive upload process. Each time it is called, a 
+         * chunk is made with createChunk(), starting at the offset parameter.
+         * The chunk is then sent by uploadChunk(), which recursively calls 
+         * this handler until the upload process is either completed and the 
+         * promise is resolved, or fails and the promise is rejected.
          * 
          * @param {Number} offset
          *      The byte at which to begin the chunk.
--- a/guacamole/src/main/frontend/src/app/rest/services/tunnelService.js
+++ b/guacamole/src/main/frontend/src/app/rest/services/tunnelService.js
@@ -55,6 +55,15 @@ angular.module('rest').factory('tunnelService', ['$injector',
     var DOWNLOAD_CLEANUP_WAIT = 5000;
 
     /**
+     * The maximum size a chunk may be during uploadToStream() in bytes.
+     * 
+     * @private
+     * @constant
+     * @type Number
+     */
+    const CHUNK_SIZE = 1024 * 1024 * 4;
+
+    /**
      * Makes a request to the REST API to get the list of all tunnels
      * associated with in-progress connections, returning a promise that
      * provides an array of their UUIDs (strings) if successful.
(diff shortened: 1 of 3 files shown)


necouchman:
@elijahnicpon Please make sure your commit messages are also tagged with the "GUACAMOLE-1320:" prefix.

necouchman:
The last commit is good, but now you need to squash them all :-).

elijahnicpon:
sorry lost in git! fixed.

necouchman:
No worries, easy to do. Git can be a bit of a maze.

Changes look okay to me, but I'd prefer if @mike-jumper or @jmuehlner reviewed them before I merge - they're much more familiar with the nuances of JavaScript.

guacamole/src/main/frontend/src/app/rest/services/tunnelService.js
```diff
@@ -54,6 +54,15 @@ angular.module('rest').factory('tunnelService', ['$injector',
      */
     var DOWNLOAD_CLEANUP_WAIT = 5000;
 
+    /**
+     * The maximum size a chunk may be during uploadToStream()
```
jmuehlner, line 58: Minor nitpick: Block comments should end in a period.

Also it would be really helpful if this specified what units this chunk size is in.

guacamole/src/main/frontend/src/app/rest/services/tunnelService.js
```diff
@@ -301,51 +310,73 @@ angular.module('rest').factory('tunnelService', ['$injector',
...
 
-        var xhr = new XMLHttpRequest();
-
-        // Invoke provided callback if upload tracking is supported
-        if (progressCallback && xhr.upload) {
-            xhr.upload.addEventListener('progress', function updateProgress(e) {
-                progressCallback(e.loaded);
-            });
+        // Creates the chunk to be uploaded
+        const createChunk = (offset) => {
```
jmuehlner, line 314: Even functions that are not exposed on the scope should have block comments just like any other.

Also, just from a readability perspective, it is a little odd to have a function called `createChunk` actually do the uploading as well. I think it'd be easier to understand what's going on here if `createChunk` returned a chunk, and then the caller recursively called `uploadChunk` again.

guacamole/src/main/java/org/apache/guacamole/tunnel/InputStreamInterceptingFilter.java
```diff
@@ -112,9 +112,8 @@ private void readNextBlob(InterceptedStream<InputStream> stream) {
             // End stream if no more data
             if (length == -1) {
 
-                // Close stream, send end if the stream is still valid
-                if (closeInterceptedStream(stream))
-                    sendEnd(stream.getIndex());
+                // Close stream
+                closeInterceptedStream(stream);
```
jmuehlner, line 116: Why did you get rid of the "end" instruction here?

elijahnicpon: The end instruction closes the writer, so without deleting that line, each subsequent chunk would just rewrite the previous. The stream is now closed when the upload is complete by ManagedFileUpload's stream state.

jmuehlner: Ah gotcha. One of the reasons that I brought this up is that the block comment for this function still specifies that the an "end" instruction will automatically be sent if the end of the InputStream is reached, which is no longer correct. The comment should be updated accordingly.

elijahnicpon: Thanks! Just committed the changes you requested.

jmuehlner:
Awesome, the code is all looking pretty good to me. The only remaining thing I have is the commit messages. Both have the same message, which is not going to be very helpful when looking back at the commit history in the future.

I would consider either changing the message of the commits so that each summarizes the changes in that particular commit, or you could squash/rebase them both into a single commit.

guacamole/src/main/java/org/apache/guacamole/tunnel/InputStreamInterceptingFilter.java
mike-jumper: @elijahnicpon I'm not seeing where the `end` instruction is sent now that this `sendEnd()` call has been removed. I do see that `ManagedFileUpload` updates its own understanding of the stream state, but the associated stream will still be active and consuming resources within guacd until `end` is received.

Can you point to where the relevant `end` instruction is sent from, if it will not be here?

elijahnicpon: You're right. I could've sworn I traced that up but I definitely made a mistake there. Thanks!

guacamole/src/main/frontend/src/app/rest/services/tunnelService.js
```diff
@@ -301,51 +310,99 @@ angular.module('rest').factory('tunnelService', ['$injector',
...
+
+            };
+
+            // Perform upload
+            xhr.send(chunk);
+
+        };
 
+        /**
+         * Handler for the upload process.
```
mike-jumper, line 396: Please document the function such that the high-level use/purpose of the function is captured. Saying that `uploadHandler()` is a handler for the upload unfortunately doesn't provide much information.

mike-jumper (changes requested):

guacamole/src/main/frontend/src/app/rest/services/tunnelService.js
elijahnicpon: Gotcha. Updated it to:
Handles the recursive upload process. Each time it is called, a chunk is made with createChunk(), starting at the offset parameter. The chunk is then sent by uploadChunk(), which recursively calls this handler until the upload process is either completed and the promise is resolved, or fails and the promise is rejected.

mike-jumper (approved):

PR #748, opened 2022-07-28
GUACAMOLE-1372: Modify usage of SAML library to allow signed requests.

jmuehlner:
The previous approach didn't actually allow requests to be signed - this change modifies the way we're calling the SAML library to allow signatures to be appended to the request as query parameters.

Draft since more testing is needed.

### DIFF
--- a/extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/src/main/java/org/apache/guacamole/auth/saml/acs/SAMLService.java
+++ b/extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/src/main/java/org/apache/guacamole/auth/saml/acs/SAMLService.java
@@ -21,7 +21,9 @@ package org.apache.guacamole.auth.saml.acs;
 
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
+import com.onelogin.saml2.Auth;
 import com.onelogin.saml2.authn.AuthnRequest;
+import com.onelogin.saml2.authn.AuthnRequestParams;
 import com.onelogin.saml2.authn.SamlResponse;
 import com.onelogin.saml2.exception.SettingsException;
 import com.onelogin.saml2.exception.ValidationError;
@@ -29,7 +31,6 @@ import com.onelogin.saml2.settings.Saml2Settings;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
-import javax.ws.rs.core.UriBuilder;
 import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.xpath.XPathExpressionException;
 import org.apache.guacamole.GuacamoleException;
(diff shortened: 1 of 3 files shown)


jmuehlner:
Ok, this is tested against a SAML provider that verifies signatures, and everything seems to be working.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/src/main/java/org/apache/guacamole/auth/saml/acs/SAMLService.java
```diff
@@ -98,6 +116,9 @@ public URI createRequest() throws GuacamoleException {
             throw new GuacamoleServerException("SAML IdP redirect could not "
                     + "be generated due to an error in the URI syntax: "
                     + e.getMessage());
+        } catch (SettingsException e) {
```
mike-jumper, line 119: Please use the same style as already established here (don't cuddle the `catch`).

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/src/main/java/org/apache/guacamole/auth/saml/acs/SAMLService.java
```diff
@@ -98,6 +116,9 @@ public URI createRequest() throws GuacamoleException {
...
+            throw new GuacamoleServerException("SAML Settings errors: "
```
mike-jumper, line 120: Is there a more explanative contextual message we can provide here beyond "SAML Settings errors: ..."?

mike-jumper (changes requested):

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/src/main/java/org/apache/guacamole/auth/saml/acs/SAMLService.java
jmuehlner: Sure, I dug into the source to find the situation where it would throw this error. Will update.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/src/main/java/org/apache/guacamole/auth/saml/acs/SAMLService.java
jmuehlner: Sure

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/src/main/java/org/apache/guacamole/auth/saml/acs/SAMLService.java
```diff
@@ -99,6 +117,11 @@ public URI createRequest() throws GuacamoleException {
                     + "be generated due to an error in the URI syntax: "
                     + e.getMessage());
         }
+        catch (SettingsException e) {
+            throw new GuacamoleServerException("Error while attempting to sign "
+                    + "request using provided private key / certificate: "
```
mike-jumper, line 122: Muuuuuuuch nicer.

mike-jumper (approved):

PR #751, opened 2022-08-04
GUACAMOLE-1656: Add per-user KSM vault functionality.

jmuehlner:
As mentioned in https://issues.apache.org/jira/browse/GUACAMOLE-1656, this allows users to specify their own KSM configurations, which can fill in any values not already provided by the admin-defined KSM configuration.

I've added a section to the end of the user preferences form for users to edit their own attributes. The Save button only applies to the user attributes immediately above - any existing preferences (mouse mode, language, etc) are still automatically updated whenever the changes are made.

image

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/JDBCAuthenticationProviderService.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/JDBCAuthenticationProviderService.java
@@ -89,31 +89,9 @@ public class JDBCAuthenticationProviderService implements AuthenticationProvider
 
     }
 
-    /**
-     * Gets a user context for the given authentication provider and user. If
-     * forceRefresh is set to true, the user record will be re-fetched even if
-     * it has already been loaded from the database. If not, the existing
-     * user will be used.
-     *
-     * @param authenticationProvider
-     *     The authentication provider to use when loading or refreshing the user.
-     *
-     * @param authenticatedUser
-     *     The user for which the user context is being fetched.
-     *
-     * @param forceRefresh
-     *     A flag that, when set to true, will force the authenticated user to
-     *     refreshed from the database. If false, an existing DB user will be
-     *     reused.
-     *
-     * @return
-     *     The fetched user context.
-     *
-     * @throws GuacamoleException
-     *     If an error occurs while fetching or refreshing the user context.
-     */
-    private ModeledUserContext getUserContext(AuthenticationProvider authenticationProvider,
-            AuthenticatedUser authenticatedUser, boolean forceRefresh) throws GuacamoleException {
+    @Override
+    public ModeledUserContext getUserContext(AuthenticationProvider authenticationProvider,
+            AuthenticatedUser authenticatedUser) throws GuacamoleException {
 
         // Always allow but provide no data for users authenticated via our own
         // connection sharing links
@@ -124,9 +102,8 @@ public class JDBCAuthenticationProviderService implements AuthenticationProvider
         boolean databaseCredentialsUsed = (authenticatedUser instanceof ModeledAuthenticatedUser);
         boolean databaseRestrictionsApplicable = (databaseCredentialsUsed || environment.isUserRequired());
 
-        // Retrieve user account for already-authenticated user, forcing a refresh if requested
-        ModeledUser user = userService.retrieveUser(
-                authenticationProvider, authenticatedUser, forceRefresh);
+        // Retrieve user account for already-authenticated user
+        ModeledUser user = userService.retrieveUser(authenticationProvider, authenticatedUser);
         ModeledUserContext context = userContextProvider.get();
         if (user != null && !user.isDisabled()) {
 
@@ -183,20 +160,12 @@ public class JDBCAuthenticationProviderService implements AuthenticationProvider
     }
 
     @Override
-    public ModeledUserContext getUserContext(AuthenticationProvider authenticationProvider,
-            AuthenticatedUser authenticatedUser) throws GuacamoleException {
-
-        // Do not force refresh unless updateUserContext is explicitly called
-        return getUserContext(authenticationProvider, authenticatedUser, false);
-    }
-
-    @Override
     public UserContext updateUserContext(AuthenticationProvider authenticationProvider,
             UserContext context, AuthenticatedUser authenticatedUser,
             Credentials credentials) throws GuacamoleException {
 
-        // Force-refresh the user context
-        return getUserContext(authenticationProvider, authenticatedUser, true);
+        // Refresh the user context
+        return getUserContext(authenticationProvider, authenticatedUser);
 
     }
 
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/JDBCAuthenticationProviderService.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/JDBCAuthenticationProviderService.java
@@ -89,9 +89,31 @@ public class JDBCAuthenticationProviderService implements AuthenticationProvider
 
     }
 
-    @Override
-    public ModeledUserContext getUserContext(AuthenticationProvider authenticationProvider,
-            AuthenticatedUser authenticatedUser) throws GuacamoleException {
+    /**
+     * Gets a user context for the given authentication provider and user. If
+     * forceRefresh is set to true, the user record will be re-fetched even if
+     * it has already been loaded from the database. If not, the existing
+     * user will be used.
+     *
+     * @param authenticationProvider
+     *     The authentication provider to use when loading or refreshing the user.
+     *
+     * @param authenticatedUser
+     *     The user for which the user context is being fetched.
+     *
+     * @param forceRefresh
+     *     A flag that, when set to true, will force the authenticated user to
+     *     refreshed from the database. If false, an existing DB user will be
+     *     reused.
+     *
+     * @return
+     *     The fetched user context.
+     *
+     * @throws GuacamoleException
+     *     If an error occurs while fetching or refreshing the user context.
+     */
+    private ModeledUserContext getUserContext(AuthenticationProvider authenticationProvider,
+            AuthenticatedUser authenticatedUser, boolean forceRefresh) throws GuacamoleException {
 
         // Always allow but provide no data for users authenticated via our own
         // connection sharing links
@@ -102,8 +124,9 @@ public class JDBCAuthenticationProviderService implements AuthenticationProvider
         boolean databaseCredentialsUsed = (authenticatedUser instanceof ModeledAuthenticatedUser);
         boolean databaseRestrictionsApplicable = (databaseCredentialsUsed || environment.isUserRequired());
 
-        // Retrieve user account for already-authenticated user
-        ModeledUser user = userService.retrieveUser(authenticationProvider, authenticatedUser);
+        // Retrieve user account for already-authenticated user, forcing a refresh if requested
+        ModeledUser user = userService.retrieveUser(
+                authenticationProvider, authenticatedUser, forceRefresh);
         ModeledUserContext context = userContextProvider.get();
         if (user != null && !user.isDisabled()) {
 
@@ -160,12 +183,20 @@ public class JDBCAuthenticationProviderService implements AuthenticationProvider
     }
 
     @Override
+    public ModeledUserContext getUserContext(AuthenticationProvider authenticationProvider,
+            AuthenticatedUser authenticatedUser) throws GuacamoleException {
+
+        // Do not force refresh unless updateUserContext is explicitly called
+        return getUserContext(authenticationProvider, authenticatedUser, false);
+    }
+
+    @Override
     public UserContext updateUserContext(AuthenticationProvider authenticationProvider,
             UserContext context, AuthenticatedUser authenticatedUser,
             Credentials credentials) throws GuacamoleException {
 
-        // No need to update the context
-        return context;
+        // Force-refresh the user context
+        return getUserContext(authenticationProvider, authenticatedUser, true);
 
     }
 
--- a/extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/GuacamoleExceptionSupplier.java
+++ b/extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/GuacamoleExceptionSupplier.java
@@ -39,4 +39,5 @@ public abstract class GuacamoleExceptionSupplier<T> {
      *    If an error occurs while attemping to calculate the return value.
      */
     public abstract T get() throws GuacamoleException;
+
 }
--- a/extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/GuacamoleExceptionSupplier.java
+++ b/extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/GuacamoleExceptionSupplier.java
@@ -1,0 +1,42 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.guacamole.vault.ksm;
+
+import org.apache.guacamole.GuacamoleException;
+
+/**
+ * A class that is basically equivalent to the standard Supplier class in
+ * Java, except that the get() function can throw GuacamoleException, which
+ * is impossible with any of the standard Java lambda type classes, since
+ * none of them can handle checked exceptions
+ */
+public abstract class GuacamoleExceptionSupplier<T> {
+
+    /**
+     * Returns a value of the declared type.
+     *
+     * @return
+     *    A value of the declared type.
+     *
+     * @throws GuacamoleException
+     *    If an error occurs while attemping to calculate the return value.
+     */
+    public abstract T get() throws GuacamoleException;
+}
--- a/extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/GuacamoleExceptionSupplier.java
+++ b/extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/GuacamoleExceptionSupplier.java
@@ -26,8 +26,12 @@ import org.apache.guacamole.GuacamoleException;
  * Java, except that the get() function can throw GuacamoleException, which
  * is impossible with any of the standard Java lambda type classes, since
  * none of them can handle checked exceptions
+ *
+ * @param <T>
+ *     The type of object which will be returned as a result of calling
+ *     get().
  */
-public abstract class GuacamoleExceptionSupplier<T> {
+public interface GuacamoleExceptionSupplier<T> {
 
     /**
      * Returns a value of the declared type.
@@ -38,6 +42,6 @@ public abstract class GuacamoleExceptionSupplier<T> {
      * @throws GuacamoleException
      *    If an error occurs while attemping to calculate the return value.
      */
-    public abstract T get() throws GuacamoleException;
+    public T get() throws GuacamoleException;
 
 }
--- a/extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/conf/KsmAttributeService.java
+++ b/extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/conf/KsmAttributeService.java
@@ -23,10 +23,13 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 
+import org.apache.guacamole.GuacamoleException;
+import org.apache.guacamole.form.BooleanField;
 import org.apache.guacamole.form.Form;
 import org.apache.guacamole.form.TextField;
 import org.apache.guacamole.vault.conf.VaultAttributeService;
 
+import com.google.inject.Inject;
 import com.google.inject.Singleton;
 
 /**
@@ -36,28 +39,81 @@ import com.google.inject.Singleton;
 @Singleton
 public class KsmAttributeService implements VaultAttributeService {
 
+
+    @Inject
+    private KsmConfigurationService configurationService;
+
     /**
      * The name of the attribute which can contain a KSM configuration blob
-     * associated with a connection group.
+     * associated with either a connection group or user.
      */
     public static final String KSM_CONFIGURATION_ATTRIBUTE = "ksm-config";
 
     /**
      * All attributes related to configuring the KSM vault on a
-     * per-connection-group basis.
+     * per-connection-group or per-user basis.
      */
     public static final Form KSM_CONFIGURATION_FORM = new Form("ksm-config",
             Arrays.asList(new TextField(KSM_CONFIGURATION_ATTRIBUTE)));
 
     /**
-     * All KSM-specific connection group attributes, organized by form.
+     * All KSM-specific attributes for users or connection groups, organized by form.
      */
-    public static final Collection<Form> KSM_CONNECTION_GROUP_ATTRIBUTES =
+    public static final Collection<Form> KSM_ATTRIBUTES =
             Collections.unmodifiableCollection(Arrays.asList(KSM_CONFIGURATION_FORM));
+
+    /**
+     * The name of the attribute which can controls whether a KSM user configuration
+     * is enabled on a connection-by-connection basis.
+     */
+    public static final String KSM_USER_CONFIG_ENABLED_ATTRIBUTE = "ksm-user-config-enabled";
+
+    /**
+     * The string value used by KSM attributes to represent the boolean value "true".
+     */
+    public static final String TRUTH_VALUE = "true";
+
+    /**
+     * All attributes related to configuring the KSM vault on a per-connection basis.
+     */
+    public static final Form KSM_CONNECTION_FORM = new Form("ksm-config",
+            Arrays.asList(new BooleanField(KSM_USER_CONFIG_ENABLED_ATTRIBUTE, TRUTH_VALUE)));
+
+    /**
+     * All KSM-specific attributes for connections, organized by form.
+     */
+    public static final Collection<Form> KSM_CONNECTION_ATTRIBUTES =
+            Collections.unmodifiableCollection(Arrays.asList(KSM_CONNECTION_FORM));
+
+    @Override
+    public Collection<Form> getConnectionAttributes() {
+        return KSM_CONNECTION_ATTRIBUTES;
+    }
 
     @Override
     public Collection<Form> getConnectionGroupAttributes() {
-        return KSM_CONNECTION_GROUP_ATTRIBUTES;
+        return KSM_ATTRIBUTES;
     }
+
+    @Override
+    public Collection<Form> getUserAttributes() {
+        return KSM_ATTRIBUTES;
+    }
+
+    @Override
+    public Collection<Form> getUserPreferenceAttributes() {
+
+        try {
+
+            // Expose the user attributes IFF user-level KSM configuration is enabled
+            return configurationService.getAllowUserConfig() ? KSM_ATTRIBUTES : Collections.emptyList();
+
+        } catch (GuacamoleException e) {
+
+            // If the configuration can't be parsed, default to not exposing the attribute
+            return Collections.emptyList();
+        }
+    }
+
 
 }
--- a/extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/conf/KsmAttributeService.java
+++ b/extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/conf/KsmAttributeService.java
@@ -28,6 +28,8 @@ import org.apache.guacamole.form.BooleanField;
 import org.apache.guacamole.form.Form;
 import org.apache.guacamole.form.TextField;
 import org.apache.guacamole.vault.conf.VaultAttributeService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
@@ -39,7 +41,14 @@ import com.google.inject.Singleton;
 @Singleton
 public class KsmAttributeService implements VaultAttributeService {
 
+    /**
+     * Logger for this class.
+     */
+    private static final Logger logger = LoggerFactory.getLogger(KsmAttributeService.class);
 
+    /**
+     * Service for retrieving KSM configuration details.
+     */
     @Inject
     private KsmConfigurationService configurationService;
 
@@ -57,7 +66,7 @@ public class KsmAttributeService implements VaultAttributeService {
             Arrays.asList(new TextField(KSM_CONFIGURATION_ATTRIBUTE)));
 
     /**
-     * All KSM-specific attributes for users or connection groups, organized by form.
+     * All KSM-specific attributes for users, connections, or connection groups, organized by form.
      */
     public static final Collection<Form> KSM_ATTRIBUTES =
             Collections.unmodifiableCollection(Arrays.asList(KSM_CONFIGURATION_FORM));
@@ -108,7 +117,16 @@ public class KsmAttributeService implements VaultAttributeService {
             // Expose the user attributes IFF user-level KSM configuration is enabled
             return configurationService.getAllowUserConfig() ? KSM_ATTRIBUTES : Collections.emptyList();
 
-        } catch (GuacamoleException e) {
+        }
+
+        catch (GuacamoleException e) {
+
+            logger.warn(
+                    "Unable to determine if user preference attributes "
+                    + "should be exposed due to config parsing error: {}.", e.getMessage());
+            logger.debug(
+                    "Config parsing error prevented checking user preference configuration",
+                    e);
 
             // If the configuration can't be parsed, default to not exposing the attribute
             return Collections.emptyList();
--- a/extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/conf/KsmAttributeService.java
+++ b/extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/conf/KsmAttributeService.java
@@ -19,20 +19,30 @@
 
 package org.apache.guacamole.vault.ksm.conf;
 
+import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
+import java.util.Base64;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
 import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.form.BooleanField;
 import org.apache.guacamole.form.Form;
 import org.apache.guacamole.form.TextField;
+import org.apache.guacamole.language.TranslatableGuacamoleClientException;
 import org.apache.guacamole.vault.conf.VaultAttributeService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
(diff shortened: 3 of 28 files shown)


jmuehlner:
Alright, this is all tested and working!

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/GuacamoleExceptionSupplier.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.apache.guacamole.GuacamoleException;
+
+/**
+ * A class that is basically equivalent to the standard Supplier class in
+ * Java, except that the get() function can throw GuacamoleException, which
+ * is impossible with any of the standard Java lambda type classes, since
+ * none of them can handle checked exceptions
+ */
+public abstract class GuacamoleExceptionSupplier<T> {
```
jmuehlner, line 30: I really wish that java would provide some syntactical sugary magic that would do this for me, but alas, it needs to be done.

Not sure if this has the right name, or is in the right place? Thoughts?

jmuehlner, line 30: I really wish that java would provide some syntactical sugary magic that would do this for me, but alas, it needs to be done.

Not sure if this has the right name, or is in the right place? Thoughts?

guacamole/src/main/frontend/src/app/settings/directives/guacSettingsPreferences.js
```diff
@@ -56,6 +58,27 @@ angular.module('settings').directive('guacSettingsPreferences', [function guacSe
...
 
+            /**
+             * An action which closes the current dialog, and refreshes
+             * the user data on dialog close.
+             */
+            const ACKNOWLEDGE_ACTION_RELOAD = {
+                name        : 'SETTINGS_PREFERENCES.ACTION_ACKNOWLEDGE',
+                // Handle action
+                callback    : function acknowledgeCallback() {
+                    userService.getUser(dataSource, username)
```
jmuehlner, line 69: This is to make sure that the updated value gets pulled if, for example, a one-time token gets rewritten into a config blob.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/JDBCAuthenticationProviderService.java
```diff
@@ -159,13 +182,21 @@ else if (!databaseRestrictionsApplicable) {
...
+
     @Override
     public UserContext updateUserContext(AuthenticationProvider authenticationProvider,
             UserContext context, AuthenticatedUser authenticatedUser,
             Credentials credentials) throws GuacamoleException {
 
-        // No need to update the context
-        return context;
+        // Force-refresh the user context
+        return getUserContext(authenticationProvider, authenticatedUser, true);
```
jmuehlner, line 199: This is to ensure that the KSM config value saved on the user preferences screen is up to date. Without this, you'd have to log out and back in again before the correct value would be pulled (when doing token replacement on a connection).

mike-jumper: Is there a need for `forceRefresh`? `getUserContext()` of `AuthenticationProvider` is called at most once, and only during initial auth, with `updateUserContext()` being the function that's called for established sessions.

jmuehlner: Ah, I see. In that case it seems like the `authenticatedUser instanceof ModeledAuthenticatedUser` check in `retrieveUser` isn't really needed, yeah? I'll remove the flag and simplify.

mike-jumper: > Ah, I see. In that case it seems like the `authenticatedUser instanceof ModeledAuthenticatedUser` check in `retrieveUser` isn't really needed, yeah?

Well, _that_ check is needed as `authenticatedUser` is not guaranteed to be an implementation of `AuthenticatedUser` from this extension (`ModeledAuthenticatedUser`). If the user is authenticated by some other extension, such as LDAP or SAML, it'll be the `AuthenticatedUser` implementation from that extension.

jmuehlner: I still don't get why that's needed, though. I mean, getting rid of that check just forces it to reload from the DB every time, which is what we want.

That check is what `forceRefresh` existed to specifically disable, which is what's needed to get the user context to actually refresh when `updateUserContext` is called.

mike-jumper:
Beware a bunch of unrelated commits snuck in here:

image

jmuehlner:
Oops, sorry. Not sure what happened there. They should all be gone now.

> Beware a bunch of unrelated commits snuck in here:
> 
> image

mike-jumper:
Beware also that the merge of the other KSM-related change has resulted in this conflicting. It'll probably need to be rebased.

jmuehlner:
> Beware also that the merge of the other KSM-related change has resulted in this conflicting. It'll probably need to be rebased.

Rebased!

guacamole/src/main/java/org/apache/guacamole/rest/schema/SchemaResource.java
```diff
@@ -77,6 +77,26 @@ public Collection<Form> getUserAttributes() throws GuacamoleException {
...
+     * @return
+     *     A collection of forms which describe the possible preference attributes of a
+     *     user object.
+     *
+     * @throws GuacamoleException
+     *     If an error occurs while retrieving the possible attributes.
+     */
+    @GET
+    @Path("userPreferenceAttributes")
+    public Collection<Form> getUserAttrigetUserPreferenceAttributesbutes()
```
mike-jumper, line 92: Two function names got smooshed together here.

guacamole/src/main/java/org/apache/guacamole/rest/user/UserResource.java
```diff
@@ -145,9 +150,51 @@ public UserHistoryResource getUserHistory()
...
+                                (currentValue == null && newValue != null) ||
+                                !currentValue.equals(newValue)
+                            )
+                    )
+                        throw new GuacamoleSecurityException(
+                            "Permission denied. Only user preference attributes"
+                            + " can be modified for the current user.");
+                }
+            }
+        }
```
mike-jumper, line 197: This is already provided in principle via `filterAttributes()` on `DirectoryObjectTranslator`:

https://github.com/apache/guacamole-client/blob/754e9649f1fa0ba225ee42b56ded64bc283d17df/guacamole/src/main/java/org/apache/guacamole/rest/directory/DirectoryObjectTranslator.java#L119-L135

which is already invoked for every received user object for update:

https://github.com/apache/guacamole-client/blob/754e9649f1fa0ba225ee42b56ded64bc283d17df/guacamole/src/main/java/org/apache/guacamole/rest/user/UserObjectTranslator.java#L58-L66

To that end:

* I think we should leverage the existing `filterAttributes()`.
* Won't the existing `filterAttributes()` call filter out anything that isn't declared as a user attribute here?

I wonder if perhaps the existing call to `filterAttributes()` should (1) consider whether the user is attempting to modify themselves and (2) filter based on user preference attributes instead of user attributes in that case.

mike-jumper (changes requested):

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/GuacamoleExceptionSupplier.java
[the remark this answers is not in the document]
mike-jumper: Also, please add JavaDoc for the type parameter `T`.

guacamole/src/main/frontend/webpack.config.js
```diff
@@ -77,18 +77,6 @@ module.exports = {
...
-            // Minify using Google Closure Compiler
-            new ClosureWebpackPlugin({ mode: 'STANDARD' }, {
-                languageIn: 'ECMASCRIPT_2020',
-                languageOut: 'ECMASCRIPT5',
-                compilationLevel: 'SIMPLE'
-            }),
-
-            new CssMinimizerPlugin()
-
-        ],
```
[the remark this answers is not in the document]
jmuehlner: Gah, sorry. It shouldn't be. I'll put it back.

guacamole/src/main/frontend/src/app/settings/directives/guacSettingsPreferences.js
```diff
@@ -78,6 +101,26 @@ angular.module('settings').directive('guacSettingsPreferences', [function guacSe
...
+            /**
+             * All available user attributes, as a mapping of form name to form
+             * object. The form object contains a name, as well as a Map of fields.
+             *
+             * The Map type is used here to maintain form/name uniqueness, as well as
+             * insertion order, to ensure a consistent UI experience.
+             *
+             * @type Map<String, Object>
+             */
+            $scope.attributeMap = new Map();
```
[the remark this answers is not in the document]
jmuehlner: It does not. Is that a problem? This usage of the `Map` data structure looks pretty well supported, including in IE11.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
```diff
@@ -222,6 +231,36 @@ public void processAttributes(Attributes entity) throws GuacamoleException {
 
     }
 
+    @Override
+    public Directory<Connection> getConnectionDirectory(
+            Directory<Connection> underlyingDirectory) throws GuacamoleException {
+
+        // A Connection directory that will intercept add and update calls to
+        // validate KSM configurations, and translate one-time-tokens, if possible
```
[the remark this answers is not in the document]
jmuehlner: Nope, this documentation is completely wrong. I will update it.

guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -942,6 +943,7 @@
         "HELP_UPDATE_PASSWORD"      : "If you wish to change your password, enter your current password and the desired new password below, and click \"Update Password\". The change will take effect immediately.",
 
         "INFO_PASSWORD_CHANGED" : "Password changed.",
+        "INFO_PREFERENCE_ATTRIBUTES_CHANGED" : "User attributes saved.",
```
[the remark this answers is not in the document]
jmuehlner: Hmm, you're definitely not wrong. I'm not really sure what exactly to put here though. The most straightforward and correct thing for this new interface would be to just say "KSM Configuration Saved" or something, but that will immediately be wrong if we ever add any other preference attributes.

"User preferences saved"? That makes it sound like it applies to the stuff above as well, (language, etc).

guacamole/src/main/java/org/apache/guacamole/rest/schema/SchemaResource.java
jmuehlner: Wow they certainly did. I will fix.

guacamole/src/main/frontend/src/app/settings/directives/guacSettingsPreferences.js
```diff
@@ -197,7 +240,84 @@ angular.module('settings').directive('guacSettingsPreferences', [function guacSe
...
+                                    }, new Map()
+                                )
+
+                            })
+                        }
+
+                    });
+
+                    // Re-generate the attributes array every time
+                    $scope.attributes = Array.of(...$scope.attributeMap.values()).map(function convertFieldsToArray(formObject) {
```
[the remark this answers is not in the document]
jmuehlner: The spread operator does get polyfilled, to this: `Array.of.apply(Array,$jscomp.arrayFromIterable(a.attributeMap.values())).`

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
```diff
@@ -304,6 +325,72 @@ private String getConnectionGroupKsmConfig(
...
+        // attribute is set to true
+        if (connectable instanceof Connection)
+            return KsmAttributeService.TRUTH_VALUE.equals(((Connection) connectable).getAttributes().get(
+                KsmAttributeService.KSM_USER_CONFIG_ENABLED_ATTRIBUTE));
+
+        // KSM token replacement is not enabled for balancing groups, so for
+        // now, user-level KSM configs will be explicitly disabled.
+        // TODO: If token replacement is implemented for balancing groups,
+        // implement this functionality for them as well.
+        return false;
```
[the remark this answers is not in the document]
jmuehlner: Updated to be more explicit about what would need to be implemented.

guacamole/src/main/frontend/src/translations/en.json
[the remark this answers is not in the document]
mike-jumper: How about "User settings saved"?

guacamole/src/main/frontend/src/app/settings/directives/guacSettingsPreferences.js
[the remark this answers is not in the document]
mike-jumper: Yep, because it's not supported in IE10, which we have not dropped support for. That said, I'm surprised that it doesn't get polyfilled during compilation now that Closure Compiler is back.

guacamole/src/main/frontend/src/translations/en.json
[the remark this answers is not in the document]
jmuehlner: Sure

guacamole/src/main/frontend/src/app/settings/directives/guacSettingsPreferences.js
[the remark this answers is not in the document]
jmuehlner: Ah, ok. I'll find another way.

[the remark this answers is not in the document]
jmuehlner: And yeah, it is strange that it's not polyfilled - it sure sounds like it _should_ be, based on the documentation: https://github.com/google/closure-compiler/wiki/Polyfills

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
```diff
@@ -314,77 +401,85 @@ public Map<String, Future<String>> getTokens(UserContext userContext, Connectabl
...
-                filteredDomain = null;
-                filteredGatewayDomain = null;
+                // Retrieve and define user-specific tokens, if any
+                // NOTE that non-RDP connections do not have a domain
+                // field in the connection parameters, so the domain
+                // will always be null
+                String username = parameters.get("username");
+                if (username != null && !username.isEmpty())
+                    addRecordTokens(tokens, "KEEPER_USER_",
+                            ksm.getRecordByLogin(filter.filter(username), null));
```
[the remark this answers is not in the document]
jmuehlner: Done.

guacamole/src/main/java/org/apache/guacamole/rest/user/UserResource.java
jmuehlner: Ah yes, that would be much simpler. I've updated it to leverage `filterAttributes()`. 

As to your question about the existing call filtering out the attributes - it doesn't, because the preference attributes for the KSM extension are also included in the regular user attributes.

guacamole/src/main/frontend/src/app/settings/directives/guacSettingsPreferences.js
```diff
@@ -197,7 +240,84 @@ angular.module('settings').directive('guacSettingsPreferences', [function guacSe
...
+            // Fetch the user record
+            userService.getUser(dataSource, username).then(function saveUserData(user) {
+                $scope.user = user;
+            })
+
+            // Get all datasources that are available for this user
+            authenticationService.getAvailableDataSources().forEach(function loadAttributesForDataSource(dataSource) {
+
+                // Fetch all user attribute forms defined for the datasource
+                schemaService.getUserPreferenceAttributes(dataSource).then(function saveAttributes(attributes) {
```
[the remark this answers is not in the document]
jmuehlner: Yeah, looking closer at this, this is messy and unnecessary. I was thinking that I'd want to save the value of the attributes that are defined on the KSM extension to the JDBC auth extension that provided the user, but I don't need to go fetching the attributes from all extensions anyway - the KSM attributes are exposed on every user context anyway.

I'll get rid of all this logic around fetching attributes from every extension, which luckily has the side effect of obviating the `Map` issue.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
```diff
@@ -304,18 +325,101 @@ private String getConnectionGroupKsmConfig(
...
+            return KsmAttributeService.TRUTH_VALUE.equals(((Connection) connectable).getAttributes().get(
+                KsmAttributeService.KSM_USER_CONFIG_ENABLED_ATTRIBUTE));
 
-        // Attempt to find a KSM config for this connection or group
-        String ksmConfig = getConnectionGroupKsmConfig(userContext, connectable);
+        // KSM token replacement is not enabled for balancing groups, so for
+        // now, user-level KSM configs will be explicitly disabled.
+        // TODO: If token replacement for connection parameters is implemented
+        // for balancing groups, implement this functionality for them as well.
+        return false;
```
mike-jumper, line 352: Token replacement for balancing groups is available, just not based on the contents of underlying connection parameters. Tokens defined via the YAML will work.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
```diff
@@ -304,18 +325,101 @@ private String getConnectionGroupKsmConfig(
...
+     *     Connectable, false otherwise.
+     */
+    private boolean isKsmUserConfigEnabled(Connectable connectable) {
 
-        Map<String, Future<String>> tokens = new HashMap<>();
-        Map<String, String> parameters = config.getParameters();
+        // If it's a connection, user-level config is enabled IFF the appropriate
+        // attribute is set to true
+        if (connectable instanceof Connection)
+            return KsmAttributeService.TRUTH_VALUE.equals(((Connection) connectable).getAttributes().get(
```
mike-jumper, line 345: The `Attributes` interface might be a better choice.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
```diff
@@ -325,13 +429,6 @@ public Map<String, Future<String>> getTokens(UserContext userContext, Connectabl
 
         // Tokens specific to RDP
         if ("rdp".equals(config.getProtocol())) {
-
-            // Retrieve and define gateway server-specific tokens, if any
-            String gatewayHostname = parameters.get("gateway-hostname");
-            if (gatewayHostname != null && !gatewayHostname.isEmpty())
-                addRecordTokens(tokens, "KEEPER_GATEWAY_",
-                        ksm.getRecordByHost(filter.filter(gatewayHostname)));
```
mike-jumper, line 333: Why is this being removed?

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
```diff
@@ -386,6 +482,38 @@ public Map<String, Future<String>> getTokens(UserContext userContext, Connectabl
...
+        // Iterate through the KSM clients, processing using the user-specific
+        // config first (if it exists), to ensure that any admin-defined values
+        // will override the user-specific values
+        Iterator<KsmClient> ksmIterator = ksmClients.iterator();
+        while (ksmIterator.hasNext()) {
+
+            // Add tokens to the Connectable for each KSM client
+            addConnectableTokens(
+                    config, ksmIterator.next(), tokens, parameters, filter);
+        }
```
mike-jumper, line 516: Why build up a temporary list of `KsmClient` vs. call `addConnectableTokens()` twice?

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
```diff
@@ -222,6 +231,37 @@ public void processAttributes(Attributes entity) throws GuacamoleException {
...
+            @Override
+            protected Connection decorate(Connection connection) throws GuacamoleException {
+
+                // Wrap in a KsmConnection class to ensure that all defined KSM fields will be
+                // present
+                return new KsmConnection(
+                        connection,
+                        ksmAttributeService.getConnectionAttributes().stream().flatMap(
+                                form -> form.getFields().stream().map(field -> field.getName())
+                        ).collect(Collectors.toList()));
```
mike-jumper, line 252: Feels somewhat wonky to pull this dynamically when we have `KSM_CONNECTION_ATTRIBUTES` available statically and the stated purpose of `KsmConnection` is to ensure those attributes are defined. Why not have `KsmConnection` leverage `KSM_CONNECTION_ATTRIBUTES` internally such that it need not be fed each attribute name manually?

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
```diff
@@ -269,4 +309,55 @@ protected ConnectionGroup undecorate(ConnectionGroup connectionGroup) throws Gua
...
+            @Override
+            protected User decorate(User user) throws GuacamoleException {
+
+                // Wrap in a KsmUser class to ensure that all defined KSM fields will be
+                // present
+                return new KsmUser(
+                        user,
+                        ksmAttributeService.getUserAttributes().stream().flatMap(
+                                form -> form.getFields().stream().map(field -> field.getName())
+                        ).collect(Collectors.toList()));
```
mike-jumper, line 351: Same here - would it make more sense to let `KsmUser` do this itself internally with `KSM_ATTRIBUTES`?

guacamole/src/main/java/org/apache/guacamole/rest/user/UserObjectTranslator.java
```diff
@@ -59,9 +59,20 @@ public void applyExternalChanges(User existingObject,
...
-        object.setAttributes(filterAttributes(userContext.getUserAttributes(),
-                object.getAttributes()));
+        // If a user is editing themselves ...
+        if (object.getUsername().equals(userContext.self().getIdentifier())) {
+
+            // ... they may only edit preference attributes
+            object.setAttributes(filterAttributes(userContext.getUserPreferenceAttributes(),
+                    object.getAttributes()));
+
+        } else {
```
mike-jumper, line 69: Please don't cuddle the `else`: https://guacamole.apache.org/guac-style/#braces

mike-jumper (changes requested):

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/GuacamoleExceptionSupplier.java
mike-jumper: Eh - seems fine to me, particularly since it's just part of the internals of an extension. If this were part of the public API, we'd need to agonize over this more, but this seems perfectly clear in purpose within the scope of the KSM support.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/JDBCAuthenticationProviderService.java
mike-jumper: You're right - I was thinking about acceptance of `AuthenticatedUser` objects from other extensions, which has no impact here except if possibly reading the cached user object, which we no longer need to do.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
```diff
@@ -144,7 +148,24 @@ public Future<String> getValue(UserContext userContext, Connectable connectable,
...
+            @Override
+            public Future<String> get() throws GuacamoleException {
+
+                // Get the user-supplied KSM config, if allowed by config and
+                // set by the user
+                String userKsmConfig = getUserKSMConfig(userContext, connectable);
+
+                // If the user config happens to be the same as admin-defined one,
+                // don't bother trying again
+                if (userKsmConfig != ksmConfig)
```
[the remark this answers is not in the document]
jmuehlner: Oops, no!

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
jmuehlner: For the check on line 344? I guess it probably wouldn't do much harm to do this even for ConnectionGroups.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
jmuehlner: Yeah, that would make more sense.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
jmuehlner: Yes

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
jmuehlner: Oops - it should not have been

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
jmuehlner: No good reason - this was a holdover from when there was a lot more stuff in this loop. I'll remove the list.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
mike-jumper: The check and the typecast, yep.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/conf/KsmAttributeService.java
```diff
@@ -36,28 +41,94 @@
...
+            // Expose the user attributes IFF user-level KSM configuration is enabled
+            return configurationService.getAllowUserConfig() ? KSM_ATTRIBUTES : Collections.emptyList();
+
+        }
+
+        catch (GuacamoleException e) {
+
+            logger.warn(
+                    "Unable to determine if user preference attributes "
+                    + "should be exposed due to config parsing error.", e);
```
mike-jumper, line 126: We should only log the full stack trace at the debug level. For example:

https://github.com/apache/guacamole-client/blob/754e9649f1fa0ba225ee42b56ded64bc283d17df/extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/RequestValidationService.java#L69-L70

The idea is that only the details needed by an administrator perusing the logs should be presented at levels above debug, while noisier details needed for development (like stack traces) should be kept at the debug level.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
```diff
@@ -144,7 +149,24 @@ public Future<String> getValue(UserContext userContext, Connectable connectable,
...
+            @Override
+            public Future<String> get() throws GuacamoleException {
+
+                // Get the user-supplied KSM config, if allowed by config and
+                // set by the user
+                String userKsmConfig = getUserKSMConfig(userContext, connectable);
+
+                // If the user config happens to be the same as admin-defined one,
+                // don't bother trying again
+                if (!userKsmConfig.equals(ksmConfig))
```
mike-jumper, line 163: This will throw an NPE if `getUserKSMConfig()` returned `null`.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
```diff
@@ -304,18 +326,97 @@ private String getConnectionGroupKsmConfig(
...
-        Map<String, Future<String>> tokens = new HashMap<>();
-        Map<String, String> parameters = config.getParameters();
+        // User-level config is enabled IFF the appropriate attribute is set to true
+        if (connectable instanceof Attributes)
+            return KsmAttributeService.TRUTH_VALUE.equals(((Attributes) connectable).getAttributes().get(
+                KsmAttributeService.KSM_USER_CONFIG_ENABLED_ATTRIBUTE));
 
-        // Attempt to find a KSM config for this connection or group
-        String ksmConfig = getConnectionGroupKsmConfig(userContext, connectable);
+        // If there's no attributes to check, the use config cannot be enabled
```
mike-jumper, line 348: Do you mean user-level config?

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
```diff
@@ -341,6 +436,12 @@ public Map<String, Future<String>> getTokens(UserContext userContext, Connectabl
                         ksm.getRecordByDomain(filteredDomain));
             }
 
+            // Retrieve and define gateway server-specific tokens, if any
+            String gatewayHostname = parameters.get("gateway-hostname");
+            if (gatewayHostname != null && !gatewayHostname.isEmpty())
+                addRecordTokens(tokens, "KEEPER_GATEWAY_",
+                        ksm.getRecordByHost(filter.filter(gatewayHostname)));
+
```
mike-jumper, line 444: Why did this get moved?

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmConnection.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Override
+    public Map<String, String> getAttributes() {
+
+        // Make a copy of the existing map
+        Map<String, String> attributes = Maps.newHashMap(super.getAttributes());
+
+        // Add the configuration attribute
+        attributes.putIfAbsent(KsmAttributeService.KSM_CONFIGURATION_ATTRIBUTE, null);
+        return attributes;
```
mike-jumper, line 67: Unlike connection parameters, object attributes are not automatically filtered from visibility to the user. They won't be displayed in the UI, but they'll be present in the bodies of relevant REST API requests.

Given the nature of these attributes, we should hide them unless the user has appropriate permissions.

If the same issue is present with connection groups from that previous change, we should hide them there, as well.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmUser.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    public Map<String, String> getAttributes() {
+
+        // Make a copy of the existing map
+        Map<String, String> attributes = Maps.newHashMap(super.getAttributes());
+
+        // Add the configuration attribute
+        attributes.putIfAbsent(KsmAttributeService.KSM_CONFIGURATION_ATTRIBUTE, null);
+        return attributes;
```
mike-jumper, line 66: Here, as with connections and connection groups, we should not expose this attribute unless the user has appropriate permissions. I suggest hiding it from every user (even administrators) except the very user it applies to.

Another approach that might satisfy the requirements of all these objects could be to _never_ expose the value of the attribute. If it were strictly write-only, with only some sort of indicator for whether a value is set and a mechanism to clear it, then there would be no concern.

mike-jumper (changes requested):

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
```diff
@@ -52,218 +39,160 @@
...
-                // Set the newly updated attributes back to the original object
-                entity.setAttributes(attributes);
+                // Do not process the ConnectionGroup further if it does not exist
+                if (connectionGroup == null)
+                    return null;
 
-            }
+                // Unwrap the existing ConnectionGroup before rewrapping
+                if (connectionGroup instanceof KsmConnectionGroup)
+                connectionGroup = ((KsmConnectionGroup) connectionGroup).getUnderlyingConnectionGroup();
```
mike-jumper, line 131: Indentation of the `if` body is missing here.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
```diff
@@ -52,218 +39,160 @@
...
 
-                // Set the newly updated attributes back to the original object
-                entity.setAttributes(attributes);
+                // Do not process the ConnectionGroup further if it does not exist
+                if (connectionGroup == null)
+                    return null;
 
-            }
+                // Unwrap the existing ConnectionGroup before rewrapping
+                if (connectionGroup instanceof KsmConnectionGroup)
```
mike-jumper, line 130: I'm a bit confused by this. Under what circumstance would `super.get()` return a `KsmConnectionGroup`?

mike-jumper (changes requested):

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectory.java
```diff
@@ -0,0 +1,76 @@
...
+    @Override
+    public Collection<ObjectType> getAll(Collection<String> identifiers) throws GuacamoleException {
+
+        // Create a list of objects to return
+        List<ObjectType> objects = new ArrayList<ObjectType>(identifiers.size());
+
+        // Populate the list by looking up objects one by one
+        Iterator<String> identifierIterator = identifiers.iterator();
+        while(identifierIterator.hasNext()) {
+            objects.add(get(identifierIterator.next()));
```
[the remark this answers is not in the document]
jmuehlner: Ah, good call. I shall do so.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
jmuehlner: Oops, will fix.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectoryService.java
jmuehlner: Looks like this is just copypasta. I'll remove.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/user/KsmDirectory.java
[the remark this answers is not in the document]
jmuehlner: Updated - how does it look now?

mike-jumper (approved):

PR #753, opened 2022-08-08
GUACAMOLE-1661: Add domain search support for KSM vault extension.

jmuehlner:


extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
```diff
@@ -250,13 +291,38 @@ private void validateCache() throws GuacamoleException {
...
+                if (username != null && confService.getSplitWindowsUsernames()) {
+
+                    // Attempt to split the domain of the username
+                    WindowsUsername usernameAndDomain = (
+                            WindowsUsername.splitWindowsUsernameFromDomain(username));
+
+                    if (usernameAndDomain.hasDomain()) {
+
+                        // Update the username if a domain has been stripped off
+                        username = usernameAndDomain.getUsername();
```
jmuehlner, line 311: This probably should have been done before, even if we weren't using the domain here.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
```diff
@@ -179,6 +188,30 @@ public class KsmClient {
      */
     private final Set<String> cachedAmbiguousUsernames = new HashSet<>();
 
+    /**
+     * All records retrieved from Keeper Secrets Manager, where each key is the
+     * domain of the corresponding record. The domain of a record is
+     * determined by {@link Login} fields, thus a record may be associated with
+     * multiple users. If a record is associated with multiple users, there
```
mike-jumper, line 195: domains* x2

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
```diff
@@ -179,6 +188,30 @@ public class KsmClient {
...
+     * will be multiple references to that record within this Map. The contents
+     * of this Map are automatically updated if {@link #validateCache()}
+     * refreshes the cache. This Map must not be accessed without
+     * {@link #cacheLock} acquired appropriately. Before using a value from
+     * this Map, {@link #cachedAmbiguousDomains} must first be checked to
+     * verify that there is indeed only one record associated with that user.
```
mike-jumper, line 201: domain*

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
```diff
@@ -179,6 +188,30 @@ public class KsmClient {
...
+     */
+    private final Map<String, KeeperRecord> cachedRecordsByDomain = new HashMap<>();
+
+    /**
+     * The set of all domains that are associated with multiple records, and
+     * thus cannot uniquely identify a record. The contents of this Set are
+     * automatically updated if {@link #validateCache()} refreshes the cache.
+     * This Set must not be accessed without {@link #cacheLock} acquired
+     * appropriately.This Set must be checked before using a value retrieved
```
mike-jumper, line 210: Please add a space between the period and the beginning of the new sentence that follows.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
```diff
@@ -250,13 +291,38 @@ private void validateCache() throws GuacamoleException {
...
+
+                // Fetch the username
+                String username = recordService.getUsername(record);
+
+                // If domains should be split out from usernames
+                if (username != null && confService.getSplitWindowsUsernames()) {
+
+                    // Attempt to split the domain of the username
+                    WindowsUsername usernameAndDomain = (
+                            WindowsUsername.splitWindowsUsernameFromDomain(username));
```
mike-jumper, line 306: A couple points:

* Should this be abstracted away such that the caller need not concern themselves with whether they need to manually split domain from username?
* As currently written, `username` will be one of the following:
  * The raw username containing the domain, which will not match the split username and domain that would be part of the connection parameters.
  * The split username (with the domain stored in a custom field of the record), which will ambiguously match any user with that username regardless of domain.

I think we need to somehow take this into account and ensure that indexed usernames include the domain, and it would be good if the caller need not manually know whether they need to handle the splitting.

mike-jumper (changes requested):

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
jmuehlner: gah the copypasta

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
jmuehlner: What caller are you referring to, specifically?

mike-jumper: This extension.

jmuehlner: Ah, yes good call. The functionality was a bit goofed, but it should be better now. The `KEEPER_USER_` tokens will now ALSO match on domain (if configured to do so. false by default). 

The behavior seems a lot more sane now - let me know what you think.

mike-jumper (changes requested):

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
mike-jumper: This definitely looks much better.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/UserDomain.java
```diff
@@ -0,0 +1,129 @@
...
+            return false;
+
+        // .. and the same domain
+        if (domain == null) {
+            if (other.domain != null)
+                return false;
+        } else if (!domain.equals(other.domain))
+            return false;
+
+        return true;
```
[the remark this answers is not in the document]
jmuehlner: Hallelujah! I shall switch it.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/UserDomain.java
```diff
@@ -0,0 +1,129 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * A class intended for use as a key in KSM user record client cache. This
+ * class contains both a username and a password. When identifying a KSM
+ * record using token syntax like "KEEPER_USER_*", the user record will
+ * actually be identified by both the user and domain, if the appropriate
+ * settings are enabled.
+ */
+class UserDomain {
```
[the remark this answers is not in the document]
jmuehlner: Yeah, that would probably make more sense.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
```diff
@@ -236,12 +270,20 @@ private void validateCache() throws GuacamoleException {
...
+            cachedAmbiguousDomains.clear();
+            cachedRecordsByDomain.clear();
+
+            // Store all records, sorting each into host-based, login-based,
+            // and domain-based buckets
+            Iterator<KeeperRecord> recordIterator = records.iterator();
+            while(recordIterator.hasNext()) {
+
+                // Go through records one at a time
+                KeeperRecord record = recordIterator.next();
```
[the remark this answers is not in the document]
jmuehlner: Because it can now throw a `GuacamoleException`, and you can't have a java lambda that throws a checked exception. My level of frustration with the whole java lambda ecosystem is quite high right now.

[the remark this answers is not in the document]
mike-jumper: Ah, OK.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
```diff
@@ -399,32 +497,75 @@ public KeeperRecord getRecordByHost(String hostname) throws GuacamoleException {
...
-     * Returns the record associated with the given username. If no such record
-     * exists, or there are multiple such records, null is returned.
+     * Returns the record associated with the given username and domain. If no
+     * such record exists, or there are multiple such records, null is returned.
      *
      * @param username
      *     The username of the record to return.
      *
+     * @param domain
+     *     The domain of the record to return.
```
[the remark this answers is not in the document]
jmuehlner: Hmm, what specifically would you request that I add? All it means if this parameter is null is that it will fetch records for which the domain is null. It's just regular equality, nothing special.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/UserDomain.java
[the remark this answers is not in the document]
jmuehlner: While I'm at it, I also switched to the simplified `Objects.hashCode` method.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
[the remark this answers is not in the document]
mike-jumper: I think just adding `..., which may be null.` or `..., or null if there is no domain.` would be sufficient. Otherwise, it sounds like a domain is always required, and it's unclear that `null` is an expected value if there is no domain.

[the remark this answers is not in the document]
jmuehlner: Fair - added.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
```diff
@@ -250,13 +292,38 @@ private void validateCache() throws GuacamoleException {
...
+                }
+
+                // If domain matching is not enabled for user records,
+                // explicitly set all domains to null to allow matching
+                // on username only
+                if (!confService.getMatchUserRecordsByDomain())
+                    domain = null;
+
+                // Store the login by username and domain
+                addRecordForLogin(record, username, domain);
```
mike-jumper, line 324: Given a record with a login like `DOMAIN\Username`, no field named "domain", and with domain splitting and domain matching enabled, this looks like it will:

1. Index that record as the username `DOMAIN\Username`.
2. Index that record again with the domain+username pair.

Is that intended?

My thinking was that the intended behavior was:

* Domain (if present) and username are pulled from the record. If splitting is enabled, that domain _might_ come from splitting the username, in which case the username will be missing the domain.
* Records are indexed by domain (however that gets retrieved) and by either domain+username (if domain matching is enabled) or just username (if domain matching is not enabled).

If the above is true, then I'd think it would be better to abstract this away beneath the record service, such that `getUsername()` and `getDomain()` retrieve what we are considering to be the record's username and domain, transparently taking splitting into account.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
```diff
@@ -338,12 +332,59 @@ public Map<String, Future<String>> getTokens(UserContext userContext, Connectabl
                 addRecordTokens(tokens, "KEEPER_GATEWAY_",
                         ksm.getRecordByHost(filter.filter(gatewayHostname)));
 
+            // Retrieve and define domain tokens, if any
+            String domain = parameters.get("domain");
+            String filteredDomain = null;
+            if (domain != null && !domain.isEmpty()) {
+                filteredDomain = filter.filter(domain);
+                addRecordTokens(tokens, "{KEEPER_DOMAIN_",
```
mike-jumper, line 340: This should be `KEEPER_DOMAIN_`, not `{KEEPER_DOMAIN_`.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/UserLogin.java
```diff
@@ -0,0 +1,121 @@
...
+
+        final int prime = 31;
+
+        int result = 1;
+        result = prime * result + Objects.hashCode(domain);
+        result = prime * result + Objects.hashCode(username);
+
+        return result;
+
+    }
```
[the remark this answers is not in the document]
jmuehlner: Sure

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmSecretService.java
jmuehlner: Oops!

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
jmuehlner: Where do you see it indexing the record twice, with `DOMAIN\Username` and domain+username? As far as I can tell, the code is already doing what you're suggesting above.

jmuehlner: Where did you see it indexing by both the raw username, and domain+username? I only see a single call to `addRecordForLogin`, which should do either one of those depending on `getSplitWindowsUsernames()`.

There was a bug here where I didn't correctly remove the username from the domain if splitting is enabled, but that should be fixed now.

Everything should be working the way you suggest above.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
```diff
@@ -250,13 +292,40 @@ private void validateCache() throws GuacamoleException {
                 String hostname = recordService.getHostname(record);
                 addRecordForHost(record, hostname);
 
-                // Store based on username ONLY if no hostname (will otherwise
-                // result in ambiguous entries for servers tied to identical
-                // accounts)
-                if (hostname == null)
-                    addRecordForLogin(record, recordService.getUsername(record));
```
jmuehlner, line 257: Oops, I realize that I completely removed this logic for no good reason. I'll restore it.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
mike-jumper: You're right - I think I must have imagined an `addRecordForLogin()` call below `recordService.getUsername()`, but there isn't actually one there. It's just retrieving the username.

> Everything should be working the way you suggest above.

Do you think it's not worth moving the domain/username retrieval logic into the record service? I still find it a bit odd that record service abstracts away the details of retrieval for everything _but_ the splitting case, where the caller (in this case `KsmClient` and `KsmSecretService`) must always manually check and split.

jmuehlner: Sure, how does this look?

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
```diff
@@ -250,13 +290,26 @@ private void validateCache() throws GuacamoleException {
...
+                addRecordForDomain(record, domain);
+
+                // Get the username off of the record
+                String username = recordService.getUsername(record);
+
+                // If domain matching is not enabled for user records,
+                // explicitly set all domains to null to allow matching
+                // on username only
+                if (!confService.getMatchUserRecordsByDomain())
+                    domain = null;
```
mike-jumper, line 304: I recall from a previous review that the `forEach()` changed to an iterator specifically to support checked exceptions. Is this is the only case where that exception is thrown?

If so, then I think we should avoid both the repeated parsing of the config value and the limitations of Java's lambdas by just grabbing this config value once before we enter the `forEach()`. It doesn't need to be re-grabbed for each record. ;)

mike-jumper (commented):
Minor point about repeatedly parsing config in a loop vs. before entering that loop (and possible simplification that might result from that if this is the only checked exception in what used to be a lambda).

Otherwise LGTM.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
mike-jumper: Muuuuuuuuch better.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
jmuehlner: Well, that's not the only config parsing that's going on now - `getUsername` and `getDomain` now also parse the config every time they're called - a consequence of the refactor where that logic got shoved into the record service. I guess I could read the config once on service startup or something, or we could just undo that refactor and read all the configs once at the start of the loop here.

jmuehlner: I know that we read from a config file to parse the config so it's not going to change but it still feels weird to pull the config once at service creation time when there's no actual guarantee in the API that the config will never change.

mike-jumper: Very true, but I don't mean at creation time. I was thinking something like:

```java
final boolean matchUserRecordsByDomain = confService.getMatchUserRecordsByDomain();
records.forEach(record -> {
    ...
});
```

The config would then be refreshed each time the cache is refreshed, rather than once per record. But:

> Well, that's not the only config parsing that's going on now - `getUsername` and `getDomain` now also parse the config every time they're called ...

That's a really good point. It sounds like the minor performance hit and loss of nifty lambda isn't worth the added complexity.

jmuehlner: Sorry, I'm a little unclear exactly what you mean by "It sounds like the minor performance hit and loss of nifty lambda isn't worth the added complexity.". 

Are you suggesting I get rid of the refactor that moved the conf parsing into the record service?

I think that would be my preference - to just read all the configs once before this here `forEach` and get rid of all the exception-throwing stuff in the domain service.

We could go back to using a lambda that way too.

mike-jumper (approved):

PR #757, opened 2022-08-22
GUACAMOLE-1614: Include chromium-driver in Docker build for sake of JS unit tests.

mike-jumper:
This change corrects the Docker build failures noted on https://issues.apache.org/jira/browse/GUACAMOLE-1614, which are ultimately due to the Jasmine Maven plugin requiring chromium-driver to run its tests.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,6 +30,9 @@ ARG TOMCAT_JRE=jdk8
 # Use official maven image for the build
 FROM maven:3-jdk-8 AS builder
 
+# Install chromium-driver for sake of JavaScript unit tests
+RUN apt-get update && apt-get install -y chromium-driver
+
 # Use args to build radius auth extension such as
 # `--build-arg BUILD_PROFILE=lgpl-extensions`
 ARG BUILD_PROFILE


Dockerfile
```diff
@@ -30,6 +30,9 @@ ARG TOMCAT_JRE=jdk8
 # Use official maven image for the build
 FROM maven:3-jdk-8 AS builder
 
+# Install chromium-driver for sake of JavaScript unit tests
+RUN apt update && apt install -y chromium-driver
```
jmuehlner, line 34: We're already apt-updating on line 57 below - do we need to do it twice?

necouchman: Also, should this be `apt update` and then `apt install`, or `apt-get update` and `apt-get install`?

jmuehlner: Good point @necouchman - should probably be `apt-get`. `apt` is specifically discouraged from being used in scripts by the developers.

mike-jumper: > We're already apt-updating on line 57 below - do we need to do it twice?

Yes - this is for the build image of a multi-stage build, whereas line 57 is the runtime image.

mike-jumper: OK - amended the commit to use `apt-get`.

PR #760, opened 2022-09-12
GUACAMOLE-990: Guarantee auth banning occurs before all other auth processing.

mike-jumper:
This change removes guacamole-auth-ban's dependency on load order, relying instead on a new pre-authentication event to guarantee banning occurs before auth processing.


jmuehlner (approved):
Looks reasonable to me.

PR #762, opened 2022-09-26
GUACAMOLE-990: Ensure internal errors during auth reach global error handling/logging.

mike-jumper:
The recent changes adding support for automatically banning IP addresses that repeatedly fail to authenticate (#758) have resulted in the details of internal errors being omitted from the Tomcat logs, instead being logged as “User authentication was aborted”.

For example, if PostgreSQL is being used but the database server has not been started, the following is logged:

```
12:12:59.935 [http-nio-8080-exec-2] WARN  o.a.g.e.AuthenticationProviderFacade - The "postgresql" authentication provider has encountered an internal error which will halt the authentication process. If this is unexpected or you are the developer of this authentication provider, you may wish to enable debug-level logging. If this is expected and you wish to ignore such failures in the future, please set "skip-if-unavailable: postgresql" within your guacamole.properties.
12:12:59.949 [http-nio-8080-exec-2] ERROR o.a.g.rest.RESTExceptionMapper - Request could not be processed: User authentication was aborted.
```

These changes rethrow the underlying cause when authentication fails due to an unchecked internal error. For other errors (subclasses of `GuacamoleException`), the original underlying exception is unwrapped and rethrown as before.

### DIFF
--- a/guacamole/src/main/java/org/apache/guacamole/rest/auth/GuacamoleAuthenticationProcessException.java
+++ b/guacamole/src/main/java/org/apache/guacamole/rest/auth/GuacamoleAuthenticationProcessException.java
@@ -146,6 +146,38 @@ public class GuacamoleAuthenticationProcessException extends GuacamoleException 
         return guacCause;
     }
 
+    /**
+     * Rethrows the original GuacamoleException wrapped within this
+     * GuacamoleAuthenticationProcessException. If there is no such exception,
+     * and the cause of this failure is an unchecked RuntimeException or Error,
+     * that unchecked exception/error is rethrown as-is.
+     *
+     * @throws GuacamoleException
+     *     If the underlying cause of this exception is a checked
+     *     GuacamoleException subclass.
+     *
+     * @throws RuntimeException
+     *     If the underlying cause of this exception is an unchecked
+     *     RuntimeException.
+     *
+     * @throws Error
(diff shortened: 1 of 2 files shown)


guacamole/src/main/java/org/apache/guacamole/rest/auth/GuacamoleAuthenticationProcessException.java
```diff
@@ -146,6 +146,38 @@ public GuacamoleException getCauseAsGuacamoleException() {
...
+     *
+     * @throws Error
+     *     If the underlying cause of this exception is an unchecked Error.
+     */
+    public void rethrowCause() throws GuacamoleException, RuntimeException, Error {
+
+        // Rethrow any unchecked exceptions/errors as-is
+        Throwable cause = getCause();
+        if (cause instanceof RuntimeException)
+            throw (RuntimeException) cause;
```
jmuehlner, line 171: What's the benefit of casting the exception here?

jmuehlner: Ah gotcha. Wasn't sure if Java would automatically do that but I guess not

mike-jumper: Yep, it needs some handholding here.

PR #763, opened 2022-09-26
GUACAMOLE-1685: Optimize Dockerfile layering by minimizing run statem…

pbuehl-redhat:
…ents


mike-jumper (changes requested):
1. These changes appear to reduce the layer count by 1, at the expense of causing a full rebuild any time a build param is changed. Can you cite anything behind why you want to make these changes?
2. Please make sure your commits cover the high-level need behind the change being made (or are squashed into a commit that does). The commit "GUACAMOLE-1685: Add extra fix" does not add any information useful to the git history. See this helpful blog post: https://www.codelord.net/2015/03/16/bad-commit-messages-hall-of-shame/

adb014:
Given the review comment shouldn't this commit be "rejected" ?

PR #764, opened 2022-09-27
GUACAMOLE-1687: Leverage network activity to ensure keep-alive pings are sent.

mike-jumper:
From [GUACAMOLE-1687](https://issues.apache.org/jira/browse/GUACAMOLE-1687):

> Chrome, the various Chromium variants, and Safari will heavily throttle JavaScript timers within background tabs and windows. This prevents the tunnel from issuing its keep-alive ping and may result in a disconnect.

This change alters the way the keep-alive pings are sent, leveraging the pings received from the server to signal that another ping may need to be sent. Pings are still scheduled using JavaScript timers, but:

* Non-chained timeouts are used instead of chained timeouts or `setInterval()` (to avoid running into [Chrome's heavy throttling criteria](https://developer.chrome.com/blog/timer-throttling-in-chrome-88/)).
* Those pings will now be sent _immediately_ if enough time has elapsed since the last time a ping was sent.

**NOTE:** There are two types of pings involved here - the `nop` instruction sent every few seconds as a keep-alive, and the internal `ping` instruction used by the WebSocket tunnel to verify network stability.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Client.js
+++ b/guacamole-common-js/src/main/webapp/modules/Client.js
@@ -42,7 +42,36 @@ Guacamole.Client = function(tunnel) {
     var currentState = STATE_IDLE;
     
     var currentTimestamp = 0;
-    var pingInterval = null;
+
+    /**
+     * The rough number of milliseconds to wait between sending keep-alive
+     * pings. This may vary depending on how frequently the browser allows
+     * timers to run, as well as how frequently the client receives messages
+     * from the server.
+     *
+     * @private
+     * @constant
+     * @type {!number}
+     */
+    var KEEP_ALIVE_FREQUENCY = 5000;
+
+    /**
+     * The current keep-alive ping timeout ID, if any. This will only be set
+     * upon connecting.
+     *
+     * @private
+     * @type {number}
+     */
+    var keepAliveTimeout = null;
+
+    /**
+     * The timestamp of the point in time that the last keep-live ping was
+     * sent, in milliseconds elapsed since midnight of January 1, 1970 UTC.
+     *
+     * @private
+     * @type {!number}
+     */
+    var lastSentKeepAlive = 0;
 
     /**
      * Translation from Guacamole protocol line caps to Layer line caps.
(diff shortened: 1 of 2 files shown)


guacamole-common-js/src/main/webapp/modules/Client.js
```diff
@@ -42,7 +42,36 @@ Guacamole.Client = function(tunnel) {
...
+     * The rough number of milliseconds to wait between sending keep-alive
+     * pings. This may vary depending on how frequently the browser allows
+     * timers to run, as well as how frequently the client receives messages
+     * from the server.
+     *
+     * @private
+     * @constant
+     * @type {!number}
+     */
+    var KEEP_ALIVE_FREQUENCY = 5000;
```
jmuehlner, line 56: We use the `const` keyword pretty extensively in the webapp, though not yet in `guacamole-common-js`. Do you think it'd be appropriate to use that here?

guacamole-common-js/src/main/webapp/modules/Tunnel.js
```diff
@@ -907,6 +917,20 @@ Guacamole.WebSocketTunnel = function(tunnelURL) {
...
+     * @private
+     */
+    var sendPing = function sendPing() {
+        var currentTime = new Date().getTime();
+        tunnel.sendMessage(Guacamole.Tunnel.INTERNAL_DATA_OPCODE, 'ping', currentTime);
+        lastSentPing = currentTime;
+    };
+
     /**
      * Initiates a timeout which, if data is not received, causes the tunnel
```
jmuehlner, line 935: This comment really does not capture the totality of what's going on in this function. Would you mind adding a bit more detail here?

guacamole-common-js/src/main/webapp/modules/Client.js
mike-jumper: Yes, except that the guacamole-common-js build does not currently leverage the same sort of JavaScript compatibility compilation. We're safe to use new fanciness like `const` and `let` within the webapp, but guacamole-common-js would need to migrate from `minify-maven-plugin` to `frontend-maven-plugin` to get those build tools and associated benefits.

guacamole-common-js/src/main/webapp/modules/Tunnel.js
```diff
@@ -933,6 +958,16 @@ Guacamole.WebSocketTunnel = function(tunnelURL) {
             tunnel.setState(Guacamole.Tunnel.State.UNSTABLE);
         }, tunnel.unstableThreshold);
 
+        var currentTime = new Date().getTime();
+        var pingDelay = Math.max(lastSentPing + PING_FREQUENCY - currentTime, 0);
+
+        // Ping tunnel endpoint regularly to test connection stability, sending
```
jmuehlner, line 964: Where is this happening regularly? I see the keepalive ping above getting triggered by receiving stuff from the server, but don't see anything like that for this. As far as I can tell it's only ever called in `socket.onopen()`. Does that happen regularly?

guacamole-common-js/src/main/webapp/modules/Tunnel.js
mike-jumper: Sure.

guacamole-common-js/src/main/webapp/modules/Client.js
jmuehlner: Well, never mind for now.

guacamole-common-js/src/main/webapp/modules/Tunnel.js
jmuehlner: Oh nevermind I see it now.

PR #766, opened 2022-09-30
GUACAMOLE-1628: Docker image of guacamole client is missing unzip

jlancaster36:
Installs unzip and fixes docker start.sh to support setting LOGBACK_LEVEL to file path without delimiter problems

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,10 +53,10 @@ RUN /opt/guacamole/bin/build-guacamole.sh "$BUILD_DIR" /opt/guacamole "$BUILD_PR
 # For the runtime image, we start with the official Tomcat distribution
 FROM tomcat:${TOMCAT_VERSION}-${TOMCAT_JRE}
 
-# Install XMLStarlet for server.xml alterations
+# Install XMLStarlet for server.xml alterations and unzip for LOGBACK_LEVEL case
 RUN apt-get update -qq \
-    && apt-get install -y xmlstarlet \
-    && rm -rf /var/lib/apt/lists/*
+    && apt-get install -y xmlstarlet unzip\
+    && rm -rf /var/lib/apt/lists/* 
 
 # This is where the build artifacts go in the runtime image
 WORKDIR /opt/guacamole


Dockerfile
```diff
@@ -58,6 +58,9 @@ RUN apt-get update -qq \
     && apt-get install -y xmlstarlet \
     && rm -rf /var/lib/apt/lists/*
 
+# # Install unzip for LOGBACK_LEVEL case
```
mike-jumper, line 61: An extra `#` snuck in here.

Dockerfile
```diff
@@ -58,6 +58,9 @@ RUN apt-get update -qq \
...
+RUN apt-get update && apt-get install unzip
```
mike-jumper, line 62: It would be better to combine this with the existing `apt-get` line above, which also already invokes `apt-get update`.

mike-jumper (changes requested):
> Installs unzip ...

Makes sense.

> ... and fixes docker start.sh to support setting LOGBACK_LEVEL to file path without delimiter problems

Eh? Why? The only valid values for `LOGBACK_LEVEL` are the log levels. A file path has no meaning here.

jlancaster36:
My mistake. I assumed the LOGBACK_LEVEL values were a file path. Regardless, values containing forward slashes had them parsed out as delimiters. Changing the sed delimiter to semicolons should prevent future conflicts

mike-jumper:
> Changing the sed delimiter to semicolons should prevent future conflicts

Sure, except with semicolons and backslashes, as well as any character that will be interpreted by the XML parser like an ampersand or a quote.

mike-jumper (approved):

maglub:
It is quite weird that this change did not make it into neither the Dockerfile for the 1.5.0, nor 1.5.1-RC1 docker containers that are available on Dockerhub. When trying to start these containers, I get an error that `unzip` is not found.

In the Dockerfile:

```
RUN apt-get update -qq \
    && apt-get install -y xmlstarlet unzip\
    && rm -rf /var/lib/apt/lists/* 
```

* The change is correct in the Master branch, but not in the staging/1.5.1 branch
* But, this `apt-get` is not yet merged into the branch `stage/1.5.1`.
* It is also not available in the tag `1.5.1-RC1` (https://github.com/apache/guacamole-client/blob/1.5.1-RC1/Dockerfile)

That it did not make it into 1.5.0, I can understand. It was probably a mistake that could easily happen, but the change is correct in the master branch. I am quite surprised, though, that nobody tried to run the docker container for a month (the one here, which is broken https://hub.docker.com/layers/guacamole/guacamole/1.5.0/images/sha256-b5e3e65ce76c6f33d12fe652133c1b945a2431bafd8ae0d7eb3c5c5883e95d2d?context=explore)

I suggest that we cherry pick the Dockerfile from master, putting it into at least the `staging/1.5.1-RC1` branch, or to create a branch `staging-1.5.0.1`, fixing it there.

necouchman:
That it did not make it into 1.5.0, I can understand. It was probably a mistake that could easily happen, but the change is correct in the master branch. I am quite surprised, though, that nobody tried to run the docker container for a month (the one here, which is broken

@maglub It was not a mistake - the issue and fix were out of scope for 1.5.0 and went in already tagged for 1.6.0, and we did not pull it back into the 1.5.1 release. I'd like to wrap up the 1.6.0 release, soon, anyway, so hopefully it'll be out relatively soon - or we could consider pulling a couple of fixes into a 1.5.2 release. @mike-jumper ?

Also, it isn't that the container is broken, or that no one tried to run 1.5.0 or 1.5.1. It's that it is only required if you actually try to change the `LOGBACK_LEVEL` variable, which is only required if you're debugging a problem with the container.

maglub:
I also just came to that conclusion, looking through the startup script. Our home built helm chart is setting the LOGBACK_LEVEL. Thanks for a quick answer.

PR #768, opened 2022-10-05
GUACAMOLE-1224: Add global logging and event handling around object management.

mike-jumper:
This change expands the types of events provided by guacamole-ext, adding events for modifications to objects stored in a `Directory` off the `UserContext`. Combined with an event listener bundled with the webapp, the following events are now logged:

* Startup/shutdown of the webapp (though shutdown may not be visible in the logs if Tomcat stops its logger first)
* Creation, modification, and deletion of **all** objects (active connections, connections, connection groups, users, user groups, and sharing profiles)
* User logout / session expiration

Authentication events that used to be logged manually by the webapp have also now been migrated to that bundled, global listener.

Further changes will be made to add the following in a future PR:

* Permission changes
* Group membership changes
* Connection sharing and usage

### DIFF
--- a/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Credentials.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Credentials.java
@@ -236,4 +236,36 @@ public class Credentials implements Serializable {
         this.remoteHostname = remoteHostname;
     }
 
+    /**
+     * Returns whether this Credentials object does not contain any specific
+     * authentication parameters, including HTTP parameters and the HTTP header
+     * used for the authentication token. An authentication request that
+     * contains no parameters whatsoever will tend to be the first, anonymous,
+     * credential-less authentication attempt that results in the initial login
+     * screen rendering.
+     *
+     * @return
+     *     true if this Credentials object contains no authentication
+     *     parameters whatsoever, false otherwise.
+     */
+    public boolean isEmpty() {
+
+        // An authentication request that contains an explicit username or
+        // password (even if blank) is non-empty, regardless of how the values
+        // were passed
+        if (getUsername() != null || getPassword() != null)
+            return false;
+
+        // All further tests depend on HTTP request details
+        HttpServletRequest httpRequest = getRequest();
+        if (httpRequest == null)
+            return true;
+
+        // An authentication request is non-empty if it contains any HTTP
+        // parameters at all or contains an authentication token
+        return !httpRequest.getParameterNames().hasMoreElements()
+                && httpRequest.getHeader("Guacamole-Token") == null;
+
+    }
+
 }
--- a/guacamole-ext/src/main/java/org/apache/guacamole/net/event/IdentifiableObjectEvent.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/net/event/IdentifiableObjectEvent.java
@@ -26,29 +26,30 @@ import org.apache.guacamole.net.auth.Identifiable;
 
 /**
  * Abstract basis for events which affect or directly relate to objects that
- * may be stored within a {@link Directory}.
+ * are {@link Identifiable} and may be stored within a {@link Directory}.
  *
  * @param <ObjectType>
- *     The type of object stored within the {@link Directory}.
+ *     The type of object affected or related to the event.
  */
-public interface DirectoryObjectEvent<ObjectType extends Identifiable>
+public interface IdentifiableObjectEvent<ObjectType extends Identifiable>
         extends AuthenticationProviderEvent, UserEvent {
 
     /**
      * {@inheritDoc}
      * <p>
-     * NOTE: For subclasses of {@link DirectoryObjectEvent}, this will be the
-     * AuthenticationProvider associated with the affected object. This is not
-     * necessarily the same as the AuthenticationProvider that authenticated
-     * the user performing the operation, which can be retrieved via
-     * {@link #getAuthenticatedUser()} and {@link AuthenticatedUser#getAuthenticationProvider()}.
+     * NOTE: For subclasses of {@link IdentifiableObjectEvent}, this will be the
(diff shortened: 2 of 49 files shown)


guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Credentials.java
```diff
@@ -236,4 +236,35 @@ public void setRemoteHostname(String remoteHostname) {
         this.remoteHostname = remoteHostname;
     }
 
+    /**
+     * Returns whether this Credentials object does not contain any specific
+     * authentication parameters, including HTTP parameters and the HTTP header
+     * used for the authentication token. An authentication first, anonymous,
```
jmuehlner, line 242: Did some words get cut out of this when it was moved over? 

I'm not quite sure what
> An authentication first, anonymous, credential-less authentication attempt that results in the initial login screen rendering.

means here.

guacamole-ext/src/main/java/org/apache/guacamole/net/event/IdentifiableObjectEvent.java
```diff
@@ -0,0 +1,80 @@
...
+     * @return
+     *     The identifier of the object affected by the operation.
+     */
+    String getObjectIdentifier();
+
+    /**
+     * Returns the object affected by the operation, if available. Whether the
+     * affected object is available is context- and implementation-dependent.
+     * There is no general guarantee across all implementations of this event
+     * that the affected object will be available.If the object is not
```
jmuehlner, line 71: Missing space: `available.If`

guacamole/src/main/java/org/apache/guacamole/GuacamoleServletContextListener.java
```diff
@@ -125,7 +128,10 @@ public String getName() {
      */
     @Inject
     private List<File> temporaryFiles;
-    
+
+    @Inject
+    private ListenerService listenerService;
```
jmuehlner, line 133: Block comment?

guacamole/src/main/java/org/apache/guacamole/event/AffectedObject.java
```diff
@@ -0,0 +1,129 @@
...
+        if (object instanceof Nameable) {
+            try {
+                name = ((Nameable) object).getName();
+            }
+            catch (RuntimeException | Error e) {
+                logger.debug("Name of object \"{}\" could not be retrieved.", identifier, e);
+            }
+        }
+
+        /// Determine type of object
```
jmuehlner, line 73: Triple slash?

necouchman: That indicates a more important comment, right?

jmuehlner: Hmm, I guess it does have a specific meaning in javadoc, though it seems likely that it was accidental in this case.

guacamole/src/main/java/org/apache/guacamole/rest/auth/HashTokenSessionMap.java
```diff
@@ -226,6 +226,7 @@ public GuacamoleSession remove(String authToken) {
     @Override
     public void shutdown() {
         executor.shutdownNow();
+        sessionMap.values().stream().forEach(GuacamoleSession::invalidate);
```
jmuehlner, line 229: It is somewhat obvious what's happening here, but a comment explaining it could maybe still be nice, especially since it's new (and not entirely clear to me why it's needed now, but not before)

guacamole/src/main/java/org/apache/guacamole/rest/directory/DirectoryResource.java
```diff
@@ -213,7 +415,15 @@ public void patchObjects(List<APIPatch<String>> patches)
                 throw new GuacamoleClientException("Patch paths must start with \"/\".");
 
             // Remove specified object
-            directory.remove(path.substring(1));
+            String identifier = path.substring(1);
+            try {
+                directory.remove(path.substring(1));
```
jmuehlner, line 420: Can't you just use `identifier` here?

guacamole/src/main/java/org/apache/guacamole/event/AffectedObject.java
mike-jumper: Yep, this is definitely a typo on my part.

guacamole/src/main/java/org/apache/guacamole/rest/auth/HashTokenSessionMap.java
mike-jumper: Some of that should be captured by the commit message of the commit that introduced this line (affdcc8e425786ddd99a1375a93fce7c0a1307ef):

>  GUACAMOLE-1224: Ensure all user sessions are explicitly invalidated prior to application shutdown.

I can add a similar comment to clarify the code here.

jmuehlner:
Everything looks pretty straightforward and reasonable here, aside from the few minor quibbles I've noted above.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Credentials.java
mike-jumper: > Did some words get cut out of this when it was moved over?

Yep - not sure how. Fixed via rebase.

guacamole-ext/src/main/java/org/apache/guacamole/net/event/IdentifiableObjectEvent.java
mike-jumper: Oops. Fixed via rebase.

mike-jumper:
Alrighty - I've addressed the feedback thus far with a rebase.

jmuehlner:
Looks good to me!

PR #770, opened 2022-10-27
GUACAMOLE-1594: Update all Java and JavaScript dependencies to latest compatible.

mike-jumper:
This change updates all direct Java and JavaScript dependencies to their latest compatible versions.

NOTES:

* Several WebPack plugins could not be updated as all plugins within WebPack's "webpack-contrib" have dropped support for WebPack v4 entirely, but Closure Compiler only supports WebPack v4.
* A notable exception for Java dependencies is JRadius, which appears abandoned (hasn't been updated in over six years).


necouchman:
> A notable exception for Java dependencies is JRadius, which appears abandoned (hasn't been updated in over six years).

Yeah, and I haven't found a great replacement - RADIUS is dying a slow, painful death as a protocol, generally replaced by SSO protocols, but still widely-enough used to have some level of demand.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/pom.xml
```diff
@@ -57,14 +57,35 @@
...
         
         <!-- MyBatis Guice -->
         <dependency>
             <groupId>org.mybatis</groupId>
             <artifactId>mybatis-guice</artifactId>
-            <version>3.10</version>
+            <version>3.17</version>
+            <exclusions>
+
+                <!-- This dependency appears to be necessary only to provide an
```
jmuehlner, line 70: I understand that this dependency isn't necessary for us, but I'm still unclear why it's being explicitly excluded. Doesn't seem like it could be for license reasons since it's MIT-licensed, and furthermore there's an explicit inclusion in `extensions/pom.xml` for the same artifact.

jmuehlner:
Aside from my one question above, this looks good to me. Just to be extra sure, I checked it out and built and tested the webapp and everything works as expected.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/pom.xml
mike-jumper: Well, the inclusion in `extensions/pom.xml` was an initial experiment that was inadvertently committed - I'll get rid of that.

It's excluded because we have that build-time sanity check that fails the build if any extension bundles SLF4J binaries (to prevent compatibility issues with the SLF4J bundled with the webapp). Thus, we have to do one of the following:

* Exclude the dependency if it can be verified that it's unused (I went this route)
* Override that enforcer rule within this `pom.xml` to explicitly allow bundling of `jcl-over-slf4j`.

I went the exclusion route as it was easy to verify that the code in question doesn't use commons-logging.

jmuehlner: Cool, makes sense.

mike-jumper: Amended the commit to remove the change to `extensions/pom.xml`.

jmuehlner (approved):
LGTM

PR #772, opened 2022-11-07
GUACAMOLE-1594: (Re-)Update Java dependencies to latest compatible versions.

mike-jumper:
This change updates our Java dependencies to their latest versions (again). Several of the dependencies that we just updated via #770 received another set of updates, some fairly important like Commons Text.


extensions/guacamole-auth-ldap/pom.xml
```diff
@@ -72,6 +78,14 @@
             <version>3.12.0</version>
         </dependency>
 
+        <!-- Force latest version of commons-text (transitive dependency from
```
jmuehlner, line 81: What happens if a new version is released? This certainly won't be the latest version anymore, but will we then have multiple different versions of commons-text floating around at the same time?

Maybe the other one should be pinned as well?

mike-jumper: > What happens if a new version is released?

Whenever dependencies are updated, we have to revisit whether any of the exclusions and corresponding overrides are still necessary. All that is always a manual process. No dependency will be updating itself without someone going into the `pom.xml` and intentionally updating it.

> ... will we then have multiple different versions of commons-text floating around at the same time?

No - the override here ensures the version included is only that version. If we somehow did inadvertently include multiple versions of a dependency, we have build-time sanity checks that would fail.

jmuehlner: Okey doke

PR #776, opened 2022-11-13
GUACAMOLE-1716: Remove usage of AccessController (deprecated in Java 17 and later).

mike-jumper:


necouchman:
Seems okay to me. I haven't attempted to compile with 1.8, but doesn't look like it introduces anything that would break with 1.8.

necouchman (approved):

PR #778, opened 2022-11-29
GUACAMOLE-1721: Store only auth token from auth response in localStorage.

jmuehlner:
As discussed in [GUACAMOLE-1721](https://issues.apache.org/jira/browse/GUACAMOLE-1721), this change removes all auth response information except the auth token itself from local storage.

### DIFF
--- a/guacamole/src/main/frontend/src/app/auth/service/authenticationService.js
+++ b/guacamole/src/main/frontend/src/app/auth/service/authenticationService.js
@@ -61,17 +61,20 @@ angular.module('auth').factory('authenticationService', ['$injector',
     var cachedResult = null;
 
     /**
-     * The unique identifier of the local storage key which stores the result
-     * of the last authentication attempt.
+     * The unique identifier of the local storage key which stores the latest
+     * authentication token.
      *
      * @type String
      */
-    var AUTH_STORAGE_KEY = 'GUAC_AUTH';
+    var AUTH_TOKEN_STORAGE_KEY = 'GUAC_AUTH_TOKEN';
 
     /**
-     * Retrieves the last successful authentication result. If the user has not
+     * Retrieves the authentication result cached in memory. If the user has not
      * yet authenticated, the user has logged out, or the last authentication
      * attempt failed, null is returned.
+     *
+     * NOTE: setAuthenticationResult() will be called upon page load, so the
+     * cache should always be populated after the page has successfully loaded.
      *
      * @returns {AuthenticationResult}
      *     The last successful authentication result, or null if the user is not
@@ -84,12 +87,7 @@ angular.module('auth').factory('authenticationService', ['$injector',
             return cachedResult;
 
         // Return explicit null if no auth data is currently stored
-        var data = localStorageService.getItem(AUTH_STORAGE_KEY);
-        if (!data)
-            return null;
-
-        // Update cache and return retrieved auth result
-        return (cachedResult = new AuthenticationResult(data));
+        return null;
 
     };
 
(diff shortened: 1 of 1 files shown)


guacamole/src/main/frontend/src/app/auth/service/authenticationService.js
```diff
@@ -84,12 +84,7 @@ angular.module('auth').factory('authenticationService', ['$injector',
             return cachedResult;
 
         // Return explicit null if no auth data is currently stored
-        var data = localStorageService.getItem(AUTH_STORAGE_KEY);
-        if (!data)
-            return null;
-
-        // Update cache and return retrieved auth result
-        return (cachedResult = new AuthenticationResult(data));
+        return null;
```
mike-jumper, line 87: As this function will now only return the in-memory cached auth result provided via `setAuthenticationResult()`, we should note this in the function's docs, as well as the fact that `setAuthenticationResult()` will be called automatically upon page load.

guacamole/src/main/frontend/src/app/auth/service/authenticationService.js
```diff
@@ -115,9 +111,10 @@ angular.module('auth').factory('authenticationService', ['$injector',
             // Always store in cache
             cachedResult = data;
 
-            // Persist result past tab/window closure ONLY if not anonymous
+            // Persist auth token past tab/window closure ONLY if not anonymous
             if (data.username !== AuthenticationResult.ANONYMOUS_USERNAME)
-                localStorageService.setItem(AUTH_STORAGE_KEY, data);
+                localStorageService.setItem(
+                        AUTH_TOKEN_STORAGE_KEY, data.authToken);
```
mike-jumper, line 117: I think we should document that:

* Only the auth token is persisted locally, while the rest of the auth result is cached only in memory.
* The rest of the authentication result can be obtained by reauthenticating with the persisted token, which can be retrieved with `getCurrentToken()`.

mike-jumper (changes requested):

guacamole/src/main/frontend/src/app/storage/services/localStorageService.js
```diff
@@ -119,25 +140,56 @@ angular.module('storage').provider('localStorageService', [function localStorage
...
+     *
+     * @param {String} key
+     *     The arbitrary, unique key of the item to retrieve from localStorage.
+     *
+     * @returns {Object}
+     *     The value stored within localStorage under the given key,
+     *     automatically deserialized from JSON, or null if no such item is
+     *     present.
+     */
+    provider.getJsonItem = function getJsonItem(key) {
```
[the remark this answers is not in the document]
jmuehlner: Oh yeah, having the existing methods just handle both is a much better idea. I'll do that.

guacamole/src/main/frontend/src/app/auth/service/authenticationService.js
jmuehlner: Will do

guacamole/src/main/frontend/src/app/auth/service/authenticationService.js
jmuehlner: Done

guacamole/src/main/frontend/src/app/storage/services/localStorageService.js
```diff
@@ -119,10 +122,18 @@ angular.module('storage').provider('localStorageService', [function localStorage
...
+        // Pull from internal storage, if present
         var data = storedItems[key];
-        if (data)
-            return JSON.parse(data);
+        if (data) {
+
+            // Serialize to JSON if possible
+            try {
+                return JSON.stringify(value);
+            } catch (ignore) {}
```
mike-jumper, line 132: I think you mean `JSON.parse()` here, but I also don't think we should attempt to parse values that we expect may not be JSON, as doing so introduces ambiguity (a value intended to be stored as a string may happen to also be valid JSON).

We could universally store everything as JSON, in which case a string would just be stored as a JSON string, or we could provide functions that are specifically intended for storage/retrieval of only string values.

mike-jumper (changes requested):

guacamole/src/main/frontend/src/app/storage/services/localStorageService.js
mike-jumper: > A raw string isn't a valid JSON-parseable value, ...

Isn't it? I get:

Stringifying and parsing the JavaScript string 'foo'

jmuehlner: Err, yeah. I'm not sure what was going on when I tested that earlier.

jmuehlner: Well, that was a bunch of ado about nothing. Removed my changes to this service.

PR #779, opened 2022-11-29
GUACAMOLE-1772: Allow user configuration of KSM API call timeout.

jmuehlner:
As discussed in [GUACAMOLE-1772](https://issues.apache.org/jira/browse/GUACAMOLE-1722) this allows user configuration of the KSM API call timeout.

While I was in there, I also cleaned up some warnings about unchecked null references.

### DIFF
--- a/extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/conf/KsmConfigurationService.java
+++ b/extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/conf/KsmConfigurationService.java
@@ -28,6 +28,7 @@ import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.GuacamoleServerException;
 import org.apache.guacamole.environment.Environment;
 import org.apache.guacamole.properties.BooleanGuacamoleProperty;
+import org.apache.guacamole.properties.LongGuacamoleProperty;
 import org.apache.guacamole.properties.StringGuacamoleProperty;
 import org.apache.guacamole.vault.conf.VaultConfigurationService;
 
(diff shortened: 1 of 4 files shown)


extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/conf/KsmConfigurationService.java
```diff
@@ -176,6 +188,20 @@ public boolean getMatchUserRecordsByDomain() throws GuacamoleException {
...
+     *
+     * @return
+     *     The minimum number of milliseconds between KSM API calls.
+     *
+     * @throws GuacamoleException
+     *     If the value specified within guacamole.properties cannot be
+     *     parsed or does not exist.
+     */
+    public long getKsmApiTimeout() throws GuacamoleException {
+        return environment.getProperty(KSM_API_CALL_TIMEOUT, 60000L);
```
jmuehlner, line 203: The original default timeout was 5 seconds, which means that under heavy load, an entire monthly API call budget could be used up in a bit over 3 hours (the base subscription includes 2500 calls / month).

At 60 seconds, the budget would stretch to nearly 2 days.

mike-jumper: I'm not sure about this. Defaulting to 60 seconds would mean people will not see changes made to records until a full minute has elapsed, which could be surprising behavior for an admin troubleshooting a connection.

jmuehlner: What interval would you suggest?

jmuehlner: Eh, I'll go to 10. That will at least buy a little bit more time before a monthly allowance will run out without being too slow to refresh.

mike-jumper (approved):

PR #780, opened 2022-12-01
GUACAMOLE-1723: Enforce access time restrictions for logged in users.

jmuehlner:


extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/activeconnection/ActiveConnectionDirectory.java
```diff
@@ -44,34 +44,40 @@ public class ActiveConnectionDirectory extends RestrictedObject
 
     @Override
     public ActiveConnection get(String identifier) throws GuacamoleException {
+        validateUserAccess();
```
jmuehlner, line 47: I don't really love having to call this everywhere, but I'm not sure of a better / more elegant way to do this.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/pom.xml
```diff
@@ -94,6 +94,12 @@
             <artifactId>guice</artifactId>
         </dependency>
 
+        <!-- Guice Assisted Inject Extension -->
+        <dependency>
```
jmuehlner, line 98: I need this in order to create `AccessEnforcingDelegatingTunnel` instances. Since the `DelegatingGuacamoleTunnel` requires the wrapped tunnel at creation time, I can't use the same `Provider` + init method approach used elsewhere.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/activeconnection/ActiveConnectionDirectory.java
mike-jumper: What about a `RestrictedDirectory` base class?

jmuehlner: That sounds like a good idea - I'll give it a try!

jmuehlner: Ok, how does this look?

mike-jumper: Looks much more DRY, I'd say.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AccessEnforcingDelegatingTunnel.java
```diff
@@ -0,0 +1,196 @@
...
+            while (true) {
+
+                try {
+
+                    // Fetch an up-to-date user record from the DB to ensure
+                    // that any access restrictions modified while this tunnel
+                    // is open will be taken into account
+                    this.user.set(userService.retrieveUser(
+                        modeledAuthenticatedUser.getAuthenticationProvider(),
+                        modeledAuthenticatedUser));
```
mike-jumper, line 111: I think this might cause problems at scale if we're executing a SQL query every 10 seconds for every connection across every user. It should instead occur once across all logged-in users (ie: at the `AuthenticationProvider` level), presumably at most once per minute.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/tunnel/AccessEnforcingDelegatingTunnel.java
```diff
@@ -0,0 +1,196 @@
...
+
+        // Start periodically refreshing the user record
+        userRefreshThread.start();
+
+        // Filter received instructions, checking if the user's login
+        // is still valid for each one. If the login is invalid,
+        // log them out immediately and close the tunnel.
+        return new FilteredGuacamoleReader(
+            super.acquireReader(),
+            new GuacamoleFilter() {
```
mike-jumper, line 152: Rather than override `acquireReader()`, etc., it would be better to override `getSocket()` and use a `FilteredGuacamoleSocket`. The same filter implementation could be used for both reads and writes.

mike-jumper (changes requested):
It may be worth considering adding some means of requesting that a session be terminated outside implicit invalidation from throwing `GuacamoleUnauthorizedException`. For example, adding an `isValid()` to `UserContext` could allow extensions to lazily request session invalidation and rely on the webapp taking action within roughly 1 minute.

I definitely think we should switch to an `AuthenticationProvider`-level thread instead of per-connection threads (see comment), assuming the thread is still needed. If we do that, and we adopt an approach like `isValid()` suggested above, then things _could_ just be:

1. Spawn JDBC-specific thread for rechecking user validity. The thread could limit the scope of its search to a single query that covers the users currently logged in.
2. Ensure `isValid()` returns the correct value when invoked and rely on the webapp to automatically terminate sessions.

There'd then be no need to recheck things for each directory operation or for each tunnel read/write. Things would simply be automatically enforced within +/- 1 minute.

jmuehlner:
Alright, this is now significantly simplified. I realized that rechecking periodically whether the access window has been changed during an active session was making things a lot more complicated, and isn't really needed for this feature. So now the webapp checks once a minute if the user is outside of an access window, and if so, terminates the user session.

> It may be worth considering adding some means of requesting that a session be terminated outside implicit invalidation from throwing `GuacamoleUnauthorizedException`. For example, adding an `isValid()` to `UserContext` could allow extensions to lazily request session invalidation and rely on the webapp taking action within roughly 1 minute.
> 
> I definitely think we should switch to an `AuthenticationProvider`-level thread instead of per-connection threads (see comment), assuming the thread is still needed. If we do that, and we adopt an approach like `isValid()` suggested above, then things _could_ just be:
> 
> 1. Spawn JDBC-specific thread for rechecking user validity. The thread could limit the scope of its search to a single query that covers the users currently logged in.
> 2. Ensure `isValid()` returns the correct value when invoked and rely on the webapp to automatically terminate sessions.
> 
> There'd then be no need to recheck things for each directory operation or for each tunnel read/write. Things would simply be automatically enforced within +/- 1 minute.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledUserContext.java
```diff
@@ -27,8 +27,10 @@
 import com.google.inject.Provider;
 import java.util.Collection;
 import java.util.Date;
+
```
necouchman, line 30: Are we intentionally separating imports, now?

jmuehlner: No, this wasn't an intentional change. I'll remove it.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/UserContext.java
```diff
@@ -40,6 +40,23 @@ public interface UserContext {
      */
     User self();
 
+    /**
+     * Returns true if the session for the User associated with this user
+     * context is valid, or false otherwise. If the session is not valid, no
+     * user context actions may be peformed.
```
mike-jumper, line 46: I don't think we should commit to "no user context actions may be performed", as that would require more strict enforcement where we test `isValid()` before every action. I suggest instead only specifying that the user's session will be terminated by the web application.

It may also be useful to note here that the web application currently tests session validity every minute, though implementations should not rely on any particular schedule except that invalid sessions can be expected to be terminated in a reasonably timely manner.

mike-jumper (changes requested):

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/UserContext.java
jmuehlner: That makes sense - how does this look?

mike-jumper: Much better - LGTM!

mike-jumper (approved):

PR #782, opened 2022-12-13
Allow usage of custom keystore and custom certificats with guacd/ssl

ronansalmon:


necouchman:
@ronansalmon As with your other pull request, this will need a Jira issue and to be tagged correctly.

PR #783, opened 2022-12-28
GUACAMOLE-1728: Allow Null User Object attributes to be saved.

aleitner:
Attributes with a value of `null` get filtered out when saving objects. An empty `timezone` attribute is `null`, but we want to save null value attributes.

### DIFF
--- a/guacamole/src/main/java/org/apache/guacamole/rest/directory/DirectoryObjectTranslator.java
+++ b/guacamole/src/main/java/org/apache/guacamole/rest/directory/DirectoryObjectTranslator.java
@@ -145,8 +145,9 @@ public abstract class DirectoryObjectTranslator<InternalType extends Identifiabl
                 String attributeValue = attributes.get(attributeName);
 
                 // Include attribute value within filtered map only if
-                // (1) defined and (2) present within provided map
-                if (attributeValue != null)
+                // defined or present within provided map
+                if (attributeValue != null || 
+                        attributes.containsKey(attributeName))
                     filtered.put(attributeName, attributeValue);
 
             }


.gitignore
```diff
@@ -1,2 +1,3 @@
 *~
 target/
+.vscode/
```
necouchman, line 3: Adding .gitignore entries is fine, I'm just not sure it should be part of this change.

guacamole/src/main/java/org/apache/guacamole/rest/user/UserResource.java
```diff
@@ -150,6 +150,10 @@ public void updateObject(APIUser modifiedObject) throws GuacamoleException {
             throw e;
         }
 
+        /* Force updateObject to save the timezone when it is empty */
+        if (modifiedObject.getAttributes().get("timezone") == null)
+            modifiedObject.getAttributes().put("timezone", "");
+
```
necouchman, line 156: As mentioned on the Jira issue, I'm curious why setting an empty string is required or preferable to having a `null` or non-existent value? It doesn't look like this change provides any functionality based on that, just changes it to an empty string? Why?

necouchman (changes requested):

guacamole/src/main/java/org/apache/guacamole/rest/user/UserResource.java
aleitner: Upon saving the User preferences when the timezone is set to be empty, User Resource's `updateObject` get's called. This leads to the UserObjectTranslator performing a filter on the user object. The user object has attributes on it and any attributes that are `null` get filtered out. Meaning anything that has a `null` value does not get updated. Setting the attribute as an empty string will prevent the `timezone` attribute from being filtered out and thus allows a `timezone` to be set to empty.

.gitignore
aleitner: I can remove this from the pull request

guacamole/src/main/java/org/apache/guacamole/rest/user/UserResource.java
necouchman: Hmmm...it's possible that this is a bug, then, in the setting process, that needs to be corrected - IIRC, attributes that are null should be removed completely if they exist, not ignored. So, I suspect some amount of fixing is indeed needed, here; however, I'm not sure that changing from `null` to an empty string is the best way to accomplish this - without looking at things in detail, it seems like this just ignores the underlying problem and works around it rather than actually fixing it.

I very well could be mistaken, though...

aleitner: The alternative that I can think of would be to have the empty timezone be represented as an empty string rather than null on the javascript side

mike-jumper: Yeah, I agree this shouldn't be done this way:
1. Within Guacamole object attributes, there is a distinction between `null` (unset) and `""` (set to empty string). They are both distinct, legitimate values. Setting to `null` typically is used to delete an attribute.
2. The generic REST endpoint for attributes should definitely not be aware of the particular needs of any one attribute. It should just shuttle the data around.

I think we need to establish exactly why this attribute doesn't get cleared out and correct _that_. It may be that the JS for the timezone field isn't properly submitting its value, or it may be that the database isn't properly handling those received attribute values.

mike-jumper:
What does `setAttributes()` receive when the timezone field has been
cleared?

guacamole/src/main/java/org/apache/guacamole/rest/user/UserResource.java
aleitner: Okay I tackled the the root of the problem and allowed the ability to save nulls

guacamole/src/main/java/org/apache/guacamole/rest/directory/DirectoryObjectTranslator.java
```diff
@@ -145,8 +145,8 @@ public Map<String, String> filterAttributes(Collection<Form> schema,
                 String attributeValue = attributes.get(attributeName);
 
                 // Include attribute value within filtered map only if
-                // (1) defined and (2) present within provided map
-                if (attributeValue != null)
+                // present within provided map
+                if (attributes.containsKey(attributeName))
```
mike-jumper, line 149: Suggest `attributeValue != null || attributes.containsKey(attributeName)` to avoid the double lookup except where necessary.

mike-jumper (changes requested):
Please update your commit message to reflect the high-level nature of the change.

mike-jumper (approved):

PR #789, opened 2023-01-15
GUACAMOLE-1322: Add SAML_IDP_URL as required for Docker container absent a metadata file.

necouchman:
The `SAML_IDP_URL` environment variable should be required, in addition to the entity ID and callback URL, when no metadata file is provided. This fixes the Docker startup script to make sure that, absent a metadata file, all three environment variables are actually required.


necouchman:
Whoa, that was bad...not worth keeping this one around - let me fix it against the correct base...

PR #791, opened 2023-01-30
GUACAMOLE-1293: Add list/count of current users joined to a connection.

mike-jumper:
This change replaces the global connection message list with a per-connection share indicator that appears when at least one user has joined the connection:

User count showing 3 current users

Notifications appear noting whether a user has joined or left. These notifications fade away automatically after around 5 seconds. Up to 3 of these notifications will appear at any given time:

User join notification

The user can hover the mouse over the status indicator to see a list of all users currently on the connection, along with counts if that user has more than one copy of the connection open:

Tooltip showing the current users of the current connection

If multiple connections are tiled within the same browser tab, this indicator is part of the connection title:

Connection title containing number of current users

This change also adds `onjoin` and `onleave` handlers to `Guacamole.Client` so that developers need not implement their own message code handling within `onmsg`.


guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -128,10 +128,7 @@
 
         "INFO_CONNECTION_SHARED" : "This connection is now shared.",
         "INFO_NO_FILE_TRANSFERS" : "No file transfers.",
-
-        "MESSAGE_DEFAULT"      : "",
-        "MESSAGE_USER_JOINED"  : "User {ARGS_1} has joined the connection.",
```
jmuehlner, line 133: Huh, these were just unused?

mike-jumper: Nope, but they are now. They were used by the message display directive that these changes replace.

jmuehlner: Ah, nevermind. I see where they were used now.

guacamole-common-js/src/main/webapp/modules/Client.js
```diff
@@ -1416,8 +1450,24 @@ Guacamole.Client = function(tunnel) {
...
+
+            var msgid = parseInt(parameters[0]);
+            if (guac_client.onmsg)
+                guac_client.onmsg(msgid, parameters.slice(1));
+
+            switch (msgid) {
+
+                case Guacamole.Client.Message.USER_JOINED:
+                    if (guac_client.onjoin)
+                        guac_client.onjoin(parameters[1] /* User ID */, parameters[2] /* Name */);
```
jmuehlner, line 1462: These comments are helpful to know what these parameters are, but why not just put the values in appropriately named variables? That seems more straightforward.

guacamole/src/main/frontend/src/app/client/types/ManagedClient.js
```diff
@@ -174,14 +175,25 @@ angular.module('client').factory('ManagedClient', ['$rootScope', '$injector',
...
+         * user that originally started the connection. Duplicate connections
+         * from the same user are included in this total.
+         */
+        this.userCount = template.userCount || 0;
+
+        /**
+         * All users currently sharing this connection, excluding the user that
+         * originally started the connection. If the connection is not shared,
+         * this object will be empty. This map consists of key/value pairs
+         * where each key is the user's username and each value is an object
```
jmuehlner, line 190: Too bad we can't just use `Set`. I guess it's still a little too new for some browsers (looking at you IE and Edge < 12)

jmuehlner:
It looks like this is the case, but just to make sure, the referenced "global connection message list" was only used for connection join/leave notifications, correct?

guacamole-common-js/src/main/webapp/modules/Client.js
mike-jumper: Sure.

mike-jumper: Switched over to variables with a new commit.

mike-jumper:
> It looks like this is the case, but just to make sure, the referenced "global connection message list" was only used for connection join/leave notifications, correct?

Yes, that's correct.

necouchman:
I've always struggled with UX Design...looks much better... :smile:

guacamole-common-js/src/main/webapp/modules/Client.js
```diff
@@ -1416,8 +1450,31 @@ Guacamole.Client = function(tunnel) {
...
+                    break;
+
+                case Guacamole.Client.Message.USER_LEFT:
+                    userID = parameters[1];
+                    username = parameters[2];
+                    if (guac_client.onleave)
+                        guac_client.onleave(userID, username);
+                    break;
+
+            }
```
necouchman, line 1477: The way this is implemented, this will fire _both_ `guac_client.onmsg()`, if it's defined, as well as `guac_client.onjoin()` or `guac_client.onleave()`, if those are defined. I just want to make sure that's the intended behavior and how you see that implementation working? I'm not sure if there are other instructions that are similarly implemented, or if this is a first of its kind?

mike-jumper: What about firing `onmsg` first, and then `onjoin`/`onleave` if the default behavior isn't prevented by the `onmsg` handler?

necouchman: I'd be fine with that - or adding a `default:` to the `switch()` statement?

I don't know that I have a strong preference one way or the other - mainly I just wanted to make sure that firing both the `onmsg()` handler and then the `onjoin()` or `onleave()` handler was the intentional and not any sort of an "oops, forgot to handle `return` the `onmsg()` call, etc.

mike-jumper: Ah, gotcha. Nope - definitely on purpose. I wanted to provide an optional, more convenient way of handling these messages that doesn't require as much low-level familiarity with the Guacamole protocol.

mike-jumper: I'll push a commit with the revised behavior momentarily - `onmsg` first, followed by `onjoin`/`onleave` if default behavior is not prevented.

mike-jumper: Commit pushed: bd913274152fcc12a12e26955bb3acdaa71e52e8

necouchman (approved):
I'm good with it. @jmuehlner anything else?

PR #792, opened 2023-02-02
Docker start script OpenID options

ronansalmon:
docker/start: Add missing OpenID configuration options


necouchman:
@ronansalmon We definitely appreciate the contribution. All contributions need an associated Jira issue, and the PRs and commits will need to be tagged with that Jira issue. Unfortunately, due to abuse, Jira does not allow people to sign up at the moment, so you'll need to request an account, first.

PR #793, opened 2023-02-04
GUACAMOLE-1293: Display anonymous users within a shared session as "Anonymous".

mike-jumper:
This change corrects the rendering of anonymous users withing the new user count UI, displaying those users as "Anonymous" with distinct styling:

Anonymous users visible within user count UI

When these users join/leave a connection, the message displayed is also specific to anonymous users:

Anonymous user joining connection Anonymous user leaving connection


jmuehlner (approved):
Looks good to me.

PR #794, opened 2023-02-06
GUACAMOLE-1733: Add missing optional configuration options from docker start.sh script

ronansalmon:
Add missing optional configuration options from docker start.sh script

### DIFF
--- a/guacamole-docker/bin/start.sh
+++ b/guacamole-docker/bin/start.sh
@@ -664,19 +664,21 @@ END
     fi
 
     # Update config file
-    set_optional_property "radius-hostname"         "$RADIUS_HOSTNAME"
-    set_optional_property "radius-auth-port"        "$RADIUS_AUTH_PORT"
-    set_property          "radius-shared-secret"    "$RADIUS_SHARED_SECRET"
-    set_property          "radius-auth-protocol"    "$RADIUS_AUTH_PROTOCOL"
-    set_optional_property "radius-key-file"         "$RADIUS_KEY_FILE"
-    set_optional_property "radius-key-type"         "$RADIUS_KEY_TYPE"
-    set_optional_property "radius-key-password"     "$RADIUS_KEY_PASSWORD"
-    set_optional_property "radius-ca-file"          "$RADIUS_CA_FILE"
-    set_optional_property "radius-ca-type"          "$RADIUS_CA_TYPE"
-    set_optional_property "radius-ca-password"      "$RADIUS_CA_PASSWORD"
-    set_optional_property "radius-trust-all"        "$RADIUS_TRUST_ALL"
-    set_optional_property "radius-retries"          "$RADIUS_RETRIES"
-    set_optional_property "radius-timeout"          "$RADIUS_TIMEOUT"
+    set_optional_property "radius-hostname"                 "$RADIUS_HOSTNAME"
+    set_optional_property "radius-auth-port"                "$RADIUS_AUTH_PORT"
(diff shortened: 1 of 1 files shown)


guacamole-docker/bin/start.sh
```diff
@@ -1164,6 +1174,16 @@ if [ -n "$API_SESSION_TIMEOUT" ]; then
...
 fi
 
+# Maximum number of bytes to accept within the entity body of any particular HTTP request
+set_optional_property "api-max-request-size" "$API_MAX_REQUEST_SIZE"
+
+# A comma-separated whitelist of language keys to allow as display language 
+# choices within the Guacamole interface
+set_optional_property "allowed-languages" "$ALLOWED_LANGUAGES"
+
+set_optional_property "enable-environment-properties" "$ENABLE_ENVIRONMENT_PROPERTIES"
```
necouchman, line 1184: This should probably have a comment associated with it.

necouchman (changes requested):
Only one minor change from me.

guacamole-docker/bin/start.sh
ronansalmon: Missed that comment.
Fixed.

guacamole-docker/bin/start.sh
```diff
@@ -1064,6 +1069,11 @@ fi
 set_property "guacd-hostname" "$GUACD_HOSTNAME"
 set_property "guacd-port"     "$GUACD_PORT"
 
+# A comma-separated list of the identifiers of authentication providers that
+# should be allowed to fail internally without aborting the authentication process
+set_property "skip-if-unavailable"     "$SKIP_IF_UNAVAILABLE"
```
necouchman, line 1074: Shouldn't this be `set_optional_property`? Using `set_property` will put an empty string in there or possibly other unexpected results if it isn't present.

necouchman (changes requested):
One other change.

guacamole-docker/bin/start.sh
ronansalmon: makes sense!

guacamole-docker/bin/start.sh
```diff
@@ -1164,6 +1174,19 @@ if [ -n "$API_SESSION_TIMEOUT" ]; then
     associate_apisessiontimeout
 fi
 
+# Maximum number of bytes to accept within the entity body of any particular HTTP request
+set_optional_property "api-max-request-size" "$API_MAX_REQUEST_SIZE"
+
+# A comma-separated whitelist of language keys to allow as display language
```
necouchman, line 1180: I would suggest changing this to:
`A comma-separated list of language keys to allow...`

necouchman (changes requested):
And one more minor wording change in a comment :-).

guacamole-docker/bin/start.sh
ronansalmon: ok

necouchman:
Okay, the changes are good, I believe; however, the final remaining item is to clean up commits and commit messages. In particular:
* Each commit message should be descriptive of the change in a way that helps people in the future to know the purpose of the commit (see https://www.codelord.net/2015/03/16/bad-commit-messages-hall-of-shame/).
* Each commit message needs to be tagged with the same "GUACAMOLE-1733: " prefix at the front of it.

ronansalmon:
ok, squashed the commits !

necouchman (approved):
Looks good. Thanks!

PR #795, opened 2023-02-20
Allow overriding of guacd-[hostname/port/ssl] within the connection's json properties

marvin-enthus:
Hi,

this is just a simple extension to the guacamole-auth-json module allowing to specify `guacd-hostname`, `guacd-port` and `guacd-ssl` overrides within the [connection properties config](https://guacamole.apache.org/doc/gug/json-auth.html#json-format). Like it is also possible for connections configured via a database.

A config snipplet looks like be:
```json
{
    "username" : "arbitraryUsername",
    "expires" : TIMESTAMP,
    "connections" : {

        "Connection Name" : {
            "protocol" : "PROTOCOL",
            "parameters" : {
                "guacd-hostname" : "my-other-guacd-host",
                "guacd-port" : "4822",
                ...
            }
        },
        ...
    }
}
```

I'm opening this PR without a JIRA ticket as I don't have an ASF account and it's not stright-forward to register. Maybe someone creates the Issue on behalf of me ;)

### DIFF
--- a/extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/connection/ConnectionService.java
+++ b/extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/connection/ConnectionService.java
@@ -40,6 +40,7 @@ import org.apache.guacamole.net.InetGuacamoleSocket;
 import org.apache.guacamole.net.SSLGuacamoleSocket;
 import org.apache.guacamole.net.SimpleGuacamoleTunnel;
 import org.apache.guacamole.net.auth.GuacamoleProxyConfiguration;
+import org.apache.guacamole.net.auth.GuacamoleProxyConfiguration.EncryptionMethod;
 import org.apache.guacamole.protocol.ConfiguredGuacamoleSocket;
 import org.apache.guacamole.protocol.GuacamoleClientInformation;
 import org.apache.guacamole.protocol.GuacamoleConfiguration;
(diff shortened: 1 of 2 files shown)


mike-jumper:
@marvin-enthus if you shoot an email to private@guacamole.apache.org, I can create a JIRA account for you.

extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/connection/ConnectionService.java
```diff
@@ -176,6 +177,22 @@ public GuacamoleTunnel connect(UserData.Connection connection,
         String hostname = proxyConfig.getHostname();
         int port = proxyConfig.getPort();
 
+        // handle guacd-[hostname/port/ssl] overrides
+        Map<String, String> params = connection.getParameters();
```
mike-jumper, line 181: The guacd hostname, port, etc. are not connection parameters, which are strictly the protocol-specific key/value pairs for configuring the connection as defined by the protocol plugin that guacd ultimately loads. Additional properties configuring the details for guacd will need to be at a different level of the JSON, perhaps directly within the connection, not within the connection params.

mike-jumper (changes requested):

extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/connection/ConnectionService.java
marvin-enthus: Hi Mike. I Just pushed an other commit migrating this to connection level.

Currently i'm not sure about the best approach - I've allinged the java class with the current camelCase, so the settings are currently also in camelCase:

```
{
    "username" : "arbitraryUsername",
    "expires" : TIMESTAMP,
    "connections" : {

        "Connection Name" : {
            "protocol" : "PROTOCOL",
            "guacdHostname" : "my-other-guacd-host",
            "guacdPort" : 4822,
            "guacdSsl" : false,
            "parameters" : {
                "name1" : "value1",
                "name2" : "value2",
                ...
            }
        },
        ...
    }
}
```
But that's not matching the names in other places. I could set some annotations for the jackson data mapper, but haven't found any json annotation on the code so i'm not sure about this...

extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/connection/ConnectionService.java
```diff
@@ -176,6 +177,21 @@ public GuacamoleTunnel connect(UserData.Connection connection,
         String hostname = proxyConfig.getHostname();
         int port = proxyConfig.getPort();
 
+        // handle guacd-[hostname/port/ssl] overrides
+        String overrideGuacdHostname = connection.getGuacdHostname();
+        if (overrideGuacdHostname != null && !overrideGuacdHostname.isEmpty()) {
+            hostname = overrideGuacdHostname;
+        }
+        Integer overrideGuacdPort = connection.getGuacdPort();
```
necouchman, line 185: I think we prefer `int` over `Integer` unless you need the extra functionality of `Integer`.

extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/connection/ConnectionService.java
```diff
@@ -176,6 +177,21 @@ public GuacamoleTunnel connect(UserData.Connection connection,
...
+        if (overrideGuacdPort != null) {
+            port = overrideGuacdPort;
+        }
+        EncryptionMethod proxyMethod = proxyConfig.getEncryptionMethod();
+        Boolean overrideGuacdSSL = connection.getGuacdSsl();
```
necouchman, line 190: Similarly, I think we prefer `boolean` over `Boolean`, unless specifically requiring `Boolean`.

necouchman (changes requested):

giacomotontini:
Any update on this PR? Is there a reason for being stale ?

adb014:
This one seems pretty easy to rebase, test and get closed. It probably needs changes to the manual and so should stay against apache:main

PR #796, opened 2023-02-22
GUACAMOLE-1740: Don't display clipboard contents in the clipboard editor until it is focused on.

aleitner:
When bringing up the Guacamole menu (Ctrl+Alt+Shift) the contents of your clipboard are immediately visible within the clipboard editor. For privacy purposes, the contents of the clipboard editor should be hidden/collapsed until the user chooses to expand/show them.

This Pull request updates the Clipboard Directive to prevent displaying clipboard data in the clipboard editor until the clipboard editor has been focused on.

### DIFF
--- a/guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
+++ b/guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
@@ -49,11 +49,29 @@ angular.module('clipboard').directive('guacClipboard', ['$injector',
 
         /**
          * The DOM element which will contain the clipboard contents within the
-         * user interface provided by this directive.
+         * user interface provided by this directive. We populate the clipboard
+         * editor via this DOM element rather than updating a model so that we
+         * are prepared for future support of rich text contents.
          *
          * @type Element
          */
-        var element = $element[0];
+        var element = $element[0].querySelectorAll('.clipboard.active')[0];
+
+        /**
+         * When isActive is set to true then the Clipboard data will be
+         * displayed in the Clipboard Editor. When false, the Clipboard Editor
+         * will not be displayed with Clipboard data.
+         *
+         * @type Boolean
+         */
+        $scope.isActive = false;
+
+        /**
+         * Updates clipboard editor to be active.
+         */
+        $scope.setActive = function setActive() {
+            $scope.isActive = true;
+        };
 
         /**
          * Rereads the contents of the clipboard field, updating the
--- a/guacamole/src/main/frontend/src/app/clipboard/templates/guacClipboard.html
+++ b/guacamole/src/main/frontend/src/app/clipboard/templates/guacClipboard.html
@@ -1,1 +1,4 @@
-<textarea class="clipboard"></textarea>
+<div>
+    <textarea ng-show="isActive" class="clipboard active"></textarea>
+    <textarea ng-show="!isActive" class="clipboard inactive" ng-focus="setActive()">{{'CLIENT.TEXT_CLIPBOARD_AWAITING_FOCUS' | translate}}</textarea>
+</div>
--- a/guacamole/src/main/frontend/src/translations/en.json
+++ b/guacamole/src/main/frontend/src/translations/en.json
@@ -162,6 +162,7 @@
         "TEXT_USER_LEFT"                  : "{USERNAME} has left the connection.",
         "TEXT_RECONNECT_COUNTDOWN"        : "Reconnecting in {REMAINING} {REMAINING, plural, one{second} other{seconds}}...",
         "TEXT_FILE_TRANSFER_PROGRESS"     : "{PROGRESS} {UNIT, select, b{B} kb{KB} mb{MB} gb{GB} other{}}",
+        "TEXT_CLIPBOARD_AWAITING_FOCUS"   : "Click to view clipboard data...",
 
         "URL_OSK_LAYOUT" : "layouts/en-us-qwerty.json"
 
--- a/guacamole/src/main/frontend/src/app/clipboard/styles/clipboard.css
+++ b/guacamole/src/main/frontend/src/app/clipboard/styles/clipboard.css
@@ -31,8 +31,6 @@
     width: 100%;
     height: 2in;
     white-space: pre;
-    font-size: 1em;
-    overflow: auto;
     padding: 0.25em;
 }
 
@@ -58,4 +56,15 @@
     height: 1em;
     white-space: pre;
     overflow: hidden;
+}
+
+.active {
+    overflow: auto;
+    font-size: 1em;
+}
+
+.inactive {
+    overflow: hidden;
+    font-size: 0.9em;
+    opacity: 0.5;
 }
--- a/guacamole/src/main/frontend/src/translations/ja.json
+++ b/guacamole/src/main/frontend/src/translations/ja.json
@@ -132,7 +132,8 @@
         "TEXT_CLIENT_STATUS_DISCONNECTED" : "切断されました。",
         "TEXT_CLIENT_STATUS_UNSTABLE"     : "Guacamoleサーバへのネットワーク接続が不安定です。",
         "TEXT_CLIENT_STATUS_WAITING"      : "Guacamoleサーバに接続しました。応答を待っています",
-        "TEXT_RECONNECT_COUNTDOWN"        : "再接続しています... {REMAINING} {REMAINING, plural, one{second} other{seconds}}..."
+        "TEXT_RECONNECT_COUNTDOWN"        : "再接続しています... {REMAINING} {REMAINING, plural, one{second} other{seconds}}...",
+        "TEXT_CLIPBOARD_AWAITING_FOCUS"   : "クリックをしてコピー/カットされたテキストが表示されます..."
 
     },
 


jmuehlner:
Can you post a screenshot showing what this change would look like to a user?

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
```diff
@@ -82,7 +100,7 @@ angular.module('clipboard').directive('guacClipboard', ['$injector',
 
             // If the clipboard data is a string, render it as text
             if (typeof data.data === 'string')
-                element.value = data.data;
+                element.value = (shouldDisplay) ? data.data : '';
```
jmuehlner, line 103: It looks like this just shows an empty clipboard until the user clicks on the clipboard element, and then the data pops in, yeah? It'd probably be lot more obvious to users if there was some sort of styling to indicate an inactive clipboard, and some indication to the user about what to do if they want to see the clipboard.

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
```diff
@@ -92,6 +110,12 @@ angular.module('clipboard').directive('guacClipboard', ['$injector',
         // that indicate the clipboard field may have been changed
         element.addEventListener('input', updateClipboardData);
         element.addEventListener('change', updateClipboardData);
+        element.addEventListener('focus', displayAndUpdateClipboardEditor);
+
+        // Update remote clipboard if local clipboard changes
```
jmuehlner, line 115: Is this copy pasta? Looks like it's a verbatim copy of the block directly below.

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
```diff
@@ -47,6 +47,12 @@ angular.module('clipboard').directive('guacClipboard', ['$injector',
     config.controller = ['$scope', '$injector', '$element',
             function guacClipboardController($scope, $injector, $element) {
 
+        /**
+         * Set to true if the clipboard data should be displayed in the  
+         * Clipboard Editor. Otherwise, false.
+         */
+        var shouldDisplay = false;
```
jmuehlner, line 54: To match the existing code style of other variables declared here, this should have a `@type` annotation.

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
```diff
@@ -47,6 +47,12 @@ angular.module('clipboard').directive('guacClipboard', ['$injector',
     config.controller = ['$scope', '$injector', '$element',
             function guacClipboardController($scope, $injector, $element) {
 
+        /**
+         * Set to true if the clipboard data should be displayed in the
```
jmuehlner, line 51: You might consider rewording this comment to be about what this variable _does_ not what the code should set it to.

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
aleitner: Whoops. Overlooked this when checking the diff before I pulled

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
```diff
@@ -92,12 +119,20 @@ angular.module('clipboard').directive('guacClipboard', ['$injector',
...
+        element.addEventListener('focus', displayAndUpdateClipboardEditor);
+
 
         // Update remote clipboard if local clipboard changes
         $scope.$on('guacClipboard', function clipboardChanged(event, data) {
             updateClipboardEditor(data);
         });
 
+        // Update the prefocus clipboard editor message
+        guacTranslate('CLIENT.TEXT_CLIPBOARD_AWAITING_FOCUS', '').then(
```
jmuehlner, line 131: It might be a bit cleaner/easier to just do the translation in the template. That would also improve separation of concerns - the controller shouldn't have to care about what specific messages are being displayed to the user. All it should need to know is what the state of the clipboard is, exposing that state to the template which can then render different elements (with translated text in them) according to that state.

guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -162,6 +162,7 @@
         "TEXT_USER_LEFT"                  : "{USERNAME} has left the connection.",
         "TEXT_RECONNECT_COUNTDOWN"        : "Reconnecting in {REMAINING} {REMAINING, plural, one{second} other{seconds}}...",
         "TEXT_FILE_TRANSFER_PROGRESS"     : "{PROGRESS} {UNIT, select, b{B} kb{KB} mb{MB} gb{GB} other{}}",
+        "TEXT_CLIPBOARD_AWAITING_FOCUS"   : "Focus here to view clipboard data...",
```
jmuehlner, line 165: "Focus" seems like an excessively technical term here: one that people not fluent in web technologies might not understand. If the required action is to hover over the clipboard, it should probably just say that plainly.

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
jmuehlner: Also, what if a user literally has the text "Focus here to view clipboard data..." in their clipboard?

 Having some quick visual method to clearly identify that the clipboard is inactive would be nice. Rather than just sticking a message into the textarea - how about visually disabling (greying out, etc) the textarea and rendering a message in a different spot where it couldn't be confused with actual clipboard contents?

aleitner:
Screenshot_20230224_123359
Screenshot_20230224_123416

jmuehlner:
That looks pretty good! It does look like that screenshot has a  branding extension applied however - can you show what it looks like without the extension?

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
```diff
@@ -71,6 +89,32 @@ angular.module('clipboard').directive('guacClipboard', ['$injector',
...
 
+        /**
+         * Allows Clipboard Editor to begin displaying clipboard data, reads the
+         * clipboard data and updates the content of the clipboard editor to the
+         * given data.
+         */
+        var displayAndUpdateClipboardEditor = function displayAndUpdateClipboardEditor() { 
+            isActive = true;
+            element.value = ''
+            element.classList.remove("inactive");
```
jmuehlner, line 100: You can do this same thing with `ng-class` in the template - no need to put this kind of functionality directly in the controller.

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
```diff
@@ -40,13 +40,31 @@ angular.module('clipboard').directive('guacClipboard', ['$injector',
...
         restrict    : 'E',
         replace     : true,
+        scope: {
+            /**
+             * The message that will be displayed within the Clipboard Editor
+             * when it does not have focus.
+             *
+             * @type String
+             */
+            inactiveContent : '=',
```
jmuehlner, line 50: It's a little weird for the clipboard element to have to have the parent tell it what its own inactive text should be. This functionality should really be internal to the clipboard directive.

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
```diff
@@ -40,13 +40,31 @@ angular.module('clipboard').directive('guacClipboard', ['$injector',
...
             function guacClipboardController($scope, $injector, $element) {
 
+        /**
+         * When isActive is set to true then the Clipboard data will be
+         * displayed in the Clipboard Editor. When false, the Clipboard Editor
+         * will not be updated with Clipboard data.
+         * 
+         * @type Boolean
+         */
+        var isActive = false;
```
jmuehlner, line 66: If you expose this to the template (on `$scope`), you could have the template itself render differently based on the state. That would move all the display logic (css styles, translations, etc) out of the controller and into the view where it belongs.

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
```diff
@@ -92,17 +141,23 @@ angular.module('clipboard').directive('guacClipboard', ['$injector',
...
-        });
+            // Don't update with clipboard data unless focused.
+            if (!isActive) {
+                return
+            }
 
-        // Init clipboard editor with current clipboard contents
-        clipboardService.getClipboard().then((data) => {
             updateClipboardEditor(data);
-        }, angular.noop);
```
jmuehlner, line 104: Why d

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
```diff
@@ -71,6 +89,32 @@ angular.module('clipboard').directive('guacClipboard', ['$injector',
...
+
+            clipboardService.getClipboard().then((data) => {
+                updateClipboardEditor(data);
+            }, angular.noop);
+        };
+
+        /**
+         * Prevent the Clipboard Editor from displaying clipboard data.
+         */
+        var hideAndUpdateClipboardEditor = function hideAndUpdateClipboardEditor() {
```
jmuehlner, line 110: This entire function probably doesn't need to exist - all this can be done directly in the template with conditional rendering depending on the value of `isActive`.

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
```diff
@@ -92,17 +141,23 @@ angular.module('clipboard').directive('guacClipboard', ['$injector',
         // that indicate the clipboard field may have been changed
         element.addEventListener('input', updateClipboardData);
         element.addEventListener('change', updateClipboardData);
+        element.addEventListener('focus', displayAndUpdateClipboardEditor);
+        element.addEventListener('blur', hideAndUpdateClipboardEditor);
```
jmuehlner, line 145: Do we really want to hide the contents of the clipboard every time it's blurred? That would get pretty annoying. I'd think that just initially hiding the clipboard every time the sidebar is opened and re-hiding when the sidebar is closed should be good enough. 

I'd welcome other opinions on this, however.

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
jmuehlner: Or alternatively, you could consider rendering completely different elements based on `isActive` using e.g. `ng-view="isActive"` or `ng-view="!isActive"`. 

That might end up being easier and probably wouldn't require dynamically fiddling CSS classes.

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
aleitner: Previously I had the translator injected into this and would retrieve the text from there, but I was asked to instead just pass the text from the client. This sounds like you are asking me to go back to the way the code originally was. Am I understanding that correctly?

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
aleitner: This is how I originally had planned to do it, but I felt like it's more secure to just hide whenever it isn't focused in case of password copying

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
jmuehlner: What I meant was to have the translation done in the template, as opposed to the controller. It _is_ technically done in a template now, but it should be the clipboard template, not the client template, and doesn't need to be passed through any controllers or javascript code at all.

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
jmuehlner: I don't think there's any one single correct answer for this one - I could see some users (like me) getting frustrated when the clipboard contents keep getting hidden while working in a connection, making it more difficult to keep track of the clipboard content.

On the other hand, I could see why some users might want to automatically rehide every time, in case they copy a password locally and have it get populated into their clipboard in the background without realizing it.

I almost wonder if this should be a configuration setting...

@mike-jumper thoughts?

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
```diff
@@ -40,33 +40,38 @@ angular.module('clipboard').directive('guacClipboard', ['$injector',
...
         replace     : true,
+        scope: {
+            /**
+             * When isActive is set to true then the Clipboard data will be
+             * displayed in the Clipboard Editor. When false, the Clipboard Editor
+             * will not be updated with Clipboard data.
+             * 
+             * @type Boolean
+             */
+            isActive : '=?',
```
jmuehlner, line 51: I don't think you need this? This `scope` in the directive config is used for passing arguments to the directive - which you're not doing that I can see. You can just set stuff directly on `$scope` in the controller.

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
```diff
@@ -40,33 +40,38 @@ angular.module('clipboard').directive('guacClipboard', ['$injector',
...
-        /**
-         * The DOM element which will contain the clipboard contents within the
-         * user interface provided by this directive.
-         *
-         * @type Element
-         */
-        var element = $element[0];
-
+        // Check if isActive was defined. If not default to false
+        $scope.isActive = angular.isDefined($scope.isActive) ? $scope.isActive : false;
```
jmuehlner, line 61: What is this logic for? When would `$scope.isActive` be defined at controller initialization time?

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
aleitner: I was originally intending for it to be optional so that it could be set to true for any reason. This doesn't really make much sense so I removed it

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
aleitner: This was for when I made isActive an optional argument

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
```diff
@@ -47,26 +47,21 @@ angular.module('clipboard').directive('guacClipboard', ['$injector',
...
-        /**
-         * The DOM element which will contain the clipboard contents within the
-         * user interface provided by this directive.
-         *
-         * @type Element
-         */
-        var element = $element[0];
-
+        // Default to inactive Clipboard Editor so any contents would be hidden.
+        $scope.isActive = false;
```
jmuehlner, line 51: To match the styling of controllers here, this should have a block comment with a `@type` annotation. The one you had for the `scope` config in the controller definition would work here.

guacamole/src/main/frontend/src/app/clipboard/templates/guacClipboard.html
```diff
@@ -1 +1,4 @@
-<textarea class="clipboard"></textarea>
+<div>
+    <textarea ng-model="clipboardData" ng-change="updateClipboardData(this.value)" ng-show="isActive" class="clipboard"></textarea>
+    <textarea ng-focus="isActive=true" ng-show="!isActive" class="clipboard inactive">{{'CLIENT.TEXT_CLIPBOARD_AWAITING_FOCUS' | translate}}</textarea>
```
jmuehlner, line 3: This mostly looks good, though I'm not a big fan of directly setting scope variables like this inside the the template like this: `ng-focus="isActive=true"`.

This looks like it's moving controller logic a bit too far into the template, sort of the opposite problem as before. If you created a `setActive()` function or similar on the controller scope, that would probably restore the balance, so to speak.

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
```diff
@@ -49,19 +49,37 @@ angular.module('clipboard').directive('guacClipboard', ['$injector',
...
+         * When isActive is set to true then the Clipboard data will be
+         * displayed in the Clipboard Editor. When false, the Clipboard Editor
+         * will not be displayed with Clipboard data.
+         *
+         * @type Boolean
+         */
+        $scope.isActive = false;
+
+        /**
+         * Updates clipboard editor be active.
```
jmuehlner, line 70: I think you're missing a preposition here.

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
```diff
@@ -49,19 +49,37 @@ angular.module('clipboard').directive('guacClipboard', ['$injector',
 
         /**
          * The DOM element which will contain the clipboard contents within the
-         * user interface provided by this directive.
+         * user interface provided by this directive. We populate the clipboard
+         * editor via this DOM element rather than updating a model so that we
+         * have support for rich text contents.
```
jmuehlner, line 54: I think we're getting a bit ahead of ourselves claiming that we "have support for rich text contents". Might want to clarify that this is to enable future support for this functionality, since we certainly don't have it now.

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
```diff
@@ -49,19 +49,37 @@ angular.module('clipboard').directive('guacClipboard', ['$injector',
...
+        };
 
         /**
          * Rereads the contents of the clipboard field, updating the
          * ClipboardData object on the scope as necessary. The type of data
          * stored within the ClipboardData object will be heuristically
          * determined from the HTML contents of the clipboard field.
          */
-        var updateClipboardData = function updateClipboardData() {
+        const updateClipboardData = function updateClipboardData() {
```
jmuehlner, line 82: I'm a fan of using `const` over `var` for this sort of thing for sure - but since no actual changes were made to this function, we should just leave it unchanged as much as possible - to ensure clean git history. Same with the removal of line 105 below. Let's just leave things the way they were.

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
```diff
@@ -49,11 +49,29 @@ angular.module('clipboard').directive('guacClipboard', ['$injector',
...
          * The DOM element which will contain the clipboard contents within the
-         * user interface provided by this directive.
+         * user interface provided by this directive. We populate the clipboard
+         * editor via this DOM element rather than updating a model so that we
+         * are prepared for future support of rich text contents.
          *
          * @type Element
          */
-        var element = $element[0];
+        var element = $element[0].querySelector('.clipboard');
```
jmuehlner, line 58: Hmm, now that I look at this, it makes me a little uncomfortable. This is just selecting the first element with the `clipboard` class, but now that there's 2 of those, it's only the ordering in the template that makes this work - if somebody came in later and changed the order around, clipboard sync would just stop working properly and it might not be obvious why.

How about updating this to specifically only match the active clipboard? Either by modifying this selector to match only elements that have the `clipboard` class, but _not_ the `inactive` class, or (probably better) by adding an `active` class to the active clipboard and querying based on that?

guacamole/src/main/frontend/src/app/clipboard/directives/guacClipboard.js
```diff
@@ -49,11 +49,29 @@ angular.module('clipboard').directive('guacClipboard', ['$injector',
...
+         *
+         * @type Boolean
+         */
+        $scope.isActive = false;
+
+        /**
+         * Updates clipboard editor to be active.
+         */
+        $scope.setActive = function setActive() {
+            $scope.isActive = true
```
jmuehlner, line 73: Semicolon missing here

jmuehlner (approved):

PR #797, opened 2023-02-22
GUACAMOLE-839: Add webapp SSO support for certificates / smart cards.

mike-jumper:
This change adds a new SSO extension that leverages SSL/TLS authentication to determine user identity. In practice, this means that users can sign in with a certificate issued to their browser by an 
employer or using a smart card and card reader (including PIV/CAC cards).

**NOTE:** This _does not_ mean that you can pass through that authentication result to an RDP server that accepts smart cards. That's an entirely different can of worms.

# Overall Authentication Flow

If configured, users are presented with an option to log in with a smart card or certificate, similar to existing support for SAML, etc.:

Login screen with smartcard option

Clicking on "Certificate / Smart Card" requests that the browser present a certificate to authentication, typically resulting in a prompt from the OS to select a card:

OS-specific card selection prompt

Authentication flow at that point is entirely between the OS, browser, and SSL termination. If the card requires a PIN, the OS should prompt you for that:

OS-specific prompt for card PIN

If everything checks out, the user is authenticated with a username extracted from the subject DN in their certificate:

Successful smart card authentication

If not, the user sees a typical authentication failure:

Unsuccessful smart card authentication

# How This Works

SSL/TLS client authentication is both simple and complex. Requiring it is easy in principle - you just set up SSL termination with something like Nginx and configure your SSL termination to perform SSL client authentication. In practice, things are more complex:

* The identity asserted via SSL/TLS client authentication is bound to that SSL session and may be cached by the browser or OS.
* If the OS or browser does manage to cache the certificate, then another user that happens to use that same system after the previous user logs out could inherit their identity.
* It is not possible for the application sitting behind SSL termination to force revalidation of the certificate.

To protect against this, these changes _dynamically generate subdomains_. When you click the link to sign in with a smart card, Guacamole actually:

1. Generates a random subdomain with a lengthy nonce value.
2. Redirects you to that subdomain.
3. Verifies that (a) the certificate presented at that subdomain is valid and (b) the subdomain is a subdomain that we generated and has never been used before.
4. Regardless of the result of validation, the user is issued a temporary auth session token representing the authentication result.
5. The temporary session token is presented to the main domain as the user's credentials. If it represents a valid identity, the user is authenticated and the token is discarded. If not, authentication fails.

Thus, any potentially cached credentials are tied to a unique subdomain that can only be used once, guaranteeing auth attempts always require fresh validation of the user's certificate.

Prerequisites for configuring this are:

* Wildcard DNS configured to point at the Guacamole server
* A wildcard certificate that is valid for the Guacamole server
* SSL termination pre-configured to request SSL/TLS client authentication for any subdomain of the Guacamole server, but _not_ for the Guacamole server itself.

# Configuration Options

## Required

* `ssl-primary-uri` - The URI of this instance without SSL/TLS client authentication required. This must be a URI that points to THIS instance of Guacamole, but behind SSL termination that DOES NOT require or request SSL/TLS client authentication.
* `ssl-client-auth-uri` - The URI that should be used to authenticate users with SSL/TLS client authentication. This must be a URI that points to THIS instance of Guacamole, but behind SSL termination that requires SSL/TLS client authentication, and the domain must start with a wildcard (which Guacamole will replace with uniquely-generated subdomains).

## Optional

* `ssl-subject-username-attribute` - A comma-separated list of all attributes that may contain the user's username in the subject DN of their certificate. The least-significant (leftmost) attribute of the DN is always used, so a subject DN without one of these attributes in the leftmost position will not be accepted at all. By default, all attributes are allowed.
* `ssl-subject-base-dn` - The base DN to require of any subject DN. Only subject DNs beneath this base DN will be accepted. By default, no restriction is placed on the subject DN.
* `ssl-client-certificate-header` - The name of the header to use to retrieve the URL-encoded client certificate from an HTTP request received from the SSL termination service. This defaults to `X-Client-Certificate`. See:
  * [`$ssl_client_escaped_cert`](https://nginx.org/en/docs/http/ngx_http_ssl_module.html#var_ssl_client_escaped_cert) (Nginx)
  * [`SSL_CLIENT_CERT`](https://httpd.apache.org/docs/2.4/mod/mod_ssl.html) (Apache HTTP Server)
* `ssl-client-verified-header` - The name of the header containing the validation result from the SSL termination service. The value of this header must be either `SUCCESS` (successful), `NONE` (no certificate present), or `FAILED:reason` where "reason" is a description of the error. Guacamole will assume that any value that doesn't match these represents generic failure. By default, the header `X-Client-Verified` is used. See:
  * [`$ssl_client_verify`](https://nginx.org/en/docs/http/ngx_http_ssl_module.html#var_ssl_client_escaped_cert) (Nginx)
  * [`SSL_CLIENT_VERIFY`](https://httpd.apache.org/docs/2.4/mod/mod_ssl.html) (Apache HTTP Server)
* `ssl-max-domain-validity` - The maximum amount of time that a generated subdomain should remain valid if left unused, in minutes. By default, domains are valid for 5 minutes. Each subdomain immediately ceases to be valid upon use.
* `ssl-max-token-validity` - The maximum amount of time that the temporary token representing the authentication result should be valid if left unused, in minutes. By default, tokens are valid for 5 minutes. Each temporary token immediately ceases to be valid upon use.

## Example Configuration

```
ssl-primary-uri: https://example.net/guacamole
ssl-client-auth-uri: https://*.example.net/guacamole

# My test cards contain subject DNs like:
# "CN=Test Cardholder IV,OU=Test Agency,OU=Test Department,O=U.S. Government,C=US"

ssl-subject-username-attribute: cn
ssl-subject-base-dn: ou=test department,o=u.s. government,c=us
```


extensions/guacamole-auth-sso/modules/guacamole-auth-sso-base/src/main/java/org/apache/guacamole/auth/sso/IdentifierGenerator.java
```diff
@@ -0,0 +1,106 @@
...
+        // Generate a base64 identifier if we're allowed to vary by case
+        if (caseSensitive) {
+            int minBytes = (minBits + 23) / 24 * 3; // Round up to nearest multiple of 3 bytes, as base64 encodes blocks of 3 bytes at a time
+            byte[] bytes = new byte[minBytes];
+            secureRandom.nextBytes(bytes);
+            return BaseEncoding.base64().encode(bytes);
+        }
+
+        // Generate base32 identifiers if we cannot vary by case
+        minBits = (minBits + 4) / 5 * 5; // Round up to nearest multiple of 5 bites, as base32 encodes 5 bits at a time
```
jmuehlner, line 101: `bites`?

mike-jumper: Oops ... bits*

I'll fix that.

guacamole/src/main/frontend/src/app/auth/service/authenticationService.js
```diff
@@ -18,25 +18,46 @@
  */
 
 /**
- * A service for authenticating a user against the REST API.
+ * A service for authenticating a user against the REST API. Invoking the
+ * authenticate() or login() functions of this service
```
jmuehlner, line 22: It looks like the second half of this sentence is missing?

jmuehlner:
Ok, after my first read-through I don't notice any major issues, just a couple of minor typos. 

@necouchman when you have a free hour or two, I'd really appreciate a second set of eyes on this one as well since it's so involved and touches so many things.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-ssl/src/main/resources/guac-manifest.json
```diff
@@ -0,0 +1,35 @@
...
+    ],
+
+    "css" : [
+        "styles/sso-providers.css"
+    ],
+
+    "js" : [ "ssl.min.js" ],
+
+    "html" : [
+        "html/sso-providers.html",
```
jmuehlner, line 19: What is `sso-providers.html`? I don't see this file anywhere.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-ssl/src/main/resources/guac-manifest.json
```diff
@@ -0,0 +1,35 @@
...
+        "html/sso-provider-ssl.html"
+    ],
+
+    "translations" : [
+        "translations/ca.json",
```
jmuehlner, line 24: Where are all these translations? I don't see any of these files (even `en.json`) in the `guacamole-auth-sso-ssl` package.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-ssl/src/main/resources/guac-manifest.json
```diff
@@ -0,0 +1,35 @@
...
+
+    "name"      : "SSL Authentication Extension",
+    "namespace" : "ssl",
+
+    "authProviders" : [
+        "org.apache.guacamole.auth.ssl.SSLAuthenticationProvider"
+    ],
+
+    "css" : [
+        "styles/sso-providers.css"
```
jmuehlner, line 13: Where is this file? I don't see it anywhere.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-ssl/src/main/resources/guac-manifest.json
mike-jumper: Ah - that's inherited from the "guacamole-auth-sso-base" project: https://github.com/apache/guacamole-client/blob/78c084be6568ae2ae8097daddb1cbf94903df060/extensions/guacamole-auth-sso/modules/guacamole-auth-sso-base/src/main/resources/html/sso-providers.html

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-ssl/src/main/resources/guac-manifest.json
mike-jumper: These also are inherited from "guacamole-auth-sso-base": https://github.com/apache/guacamole-client/tree/master/extensions/guacamole-auth-sso/modules/guacamole-auth-sso-base/src/main/resources/translations

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-ssl/src/main/resources/guac-manifest.json
mike-jumper: This, too, is inherited from "guacamole-auth-sso-base": https://github.com/apache/guacamole-client/blob/78c084be6568ae2ae8097daddb1cbf94903df060/extensions/guacamole-auth-sso/modules/guacamole-auth-sso-base/src/main/resources/styles/sso-providers.css

guacamole/src/main/frontend/src/app/auth/service/authenticationService.js
mike-jumper: Whoops - looks like I

necouchman:
Sure, @jmuehlner, I'll try to have a look over it today.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-ssl/src/main/resources/guac-manifest.json
jmuehlner: Ahhhh right

necouchman (changes requested):
Overall looks okay to me - just some minor questions...

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-ssl/src/main/java/org/apache/guacamole/auth/ssl/SSLAuthenticationProvider.java
```diff
@@ -0,0 +1,48 @@
...
+import org.apache.guacamole.auth.sso.SSOAuthenticationProvider;
+
+/**
+ * Guacamole authentication backend which authenticates users using SSL/TLS
+ * client authentication provided by some external SSL termination system. This
+ * SSL termination system must be configured to provide access to this same
+ * instance of Guacamole and must have both a wildcard certificate and wildcard
+ * DNS. No storage for connections is provided - only authentication. Storage
+ * must be provided by some other extension.
+ */
```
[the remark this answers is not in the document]
mike-jumper: My research hasn't revealed any, but I would be very happy to hear of a better approach.

The current approach with wildcard certificates and wildcard DNS is rooted in what I encountered while researching how PIV/CAC auth could be implemented. I had access to:

* A subset of the set of NIST PIV test cards.
* A USB card reader.
* https://secure.login.gov/login/piv_cac as a real-world example of what a good PIV/CAC auth flow can be.

From the above, I observed:

* Browsers cache credentials from X.509 certificates. Initially, I had hoped to work around this by checking SSL session IDs, but that was not sufficient - these were _new_ sessions established with cached credentials.
* There is no way for anything not directly handling the SSL/TLS communication to force any sort of behavior with respect to smart cards.
* Behavior of an SSL/TLS connection is tied to the domain. It's not possible to require SSL client auth for only a path beneath a domain, because this negotiation occurs when the SSL connection is established and before anything like the HTTP request path would be available.
* The `login.gov` site appears to leverage randomly-generated hostnames and wildcard DNS to ensure that clicking the "Insert your PIV/CAC" button always produces a fresh prompt for your card. This occurs through a series of redirects that are not visible unless you take a look at browser dev tools:

  1. User visits https://secure.login.gov/login/piv_cac
  2. User clicks "Insert your PIV/CAC"
  3. Internally, a request is issued to `https://secure.login.gov/login/present_piv_cac`
  4. That request is redirected to `https://XXXX.pivcac.prod.login.gov/?nonce=NONCE&...` where `XXXX` is a 4-character random subdomain and `NONCE` is a lengthy, unique nonce value.
  5. _Only_ the request to `XXXX.pivcac.prod.login.gov` requests SSL/TLS auth. It then redirects back to the original domain and includes a lengthy token in the query parameters of the redirect URL.
  6. The result of auth is obtained through using that token and reported to the user.

  While I _believe_ the `login.gov` site does things this way for the same reason I did (to ensure the user is guaranteed a fresh auth attempt), I can find no supporting documentation for this sort of requirement, including for established smart card auth solutions that purport to support PIV/CAC.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-ssl/src/main/java/org/apache/guacamole/auth/ssl/SSLClientAuthenticationResource.java
```diff
@@ -0,0 +1,414 @@
...
+        }
+        catch (UnsupportedEncodingException e) {
+            // This should never happen, as UTF-8 is a standard charset that
+            // the JVM is required to support
+            throw new UnsupportedOperationException("Unexpected lack of UTF-8 support.", e);
+        }
+    }
+
+    /**
+     * Extracts a user's username from the provided LDAP DN. If specific
```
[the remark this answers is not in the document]
mike-jumper: It's a fair point. I'll rephrase to note that this extracts the username from the subject name of an X.509 certificate, which is expected to be in the format of an LDAP DN.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-ssl/src/main/java/org/apache/guacamole/auth/ssl/SSLAuthenticationProvider.java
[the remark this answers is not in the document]
necouchman: Okay, if that's the accepted way, that's fine - I've not ever implemented a SSL certificate authentication system, so I'm not familiar with the details, but sounds like that is best practice.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-ssl/src/main/java/org/apache/guacamole/auth/ssl/SSLClientAuthenticationResource.java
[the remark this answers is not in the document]
mike-jumper: Fixed via rebase.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-ssl/src/main/java/org/apache/guacamole/auth/ssl/SSLAuthenticationSessionManager.java
```diff
@@ -0,0 +1,60 @@
...
+     * is returned.
+     *
+     * @param identifier
+     *     The unique string returned by the call to defer(). For convenience,
+     *     this value may safely be null.
+     *
+     * @return
+     *     The identity asserted by the external SSL termination service at the
+     *     end of the authentication process represented by the authentication
+     *     session the the given identifier, or null if there is no such
```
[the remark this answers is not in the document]
mike-jumper: Fixed via rebase.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-ssl/src/main/java/org/apache/guacamole/auth/ssl/SSLAuthenticationProvider.java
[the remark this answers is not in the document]
mike-jumper: > Okay, if that's the accepted way, that's fine ...

Well, it's the way that I determined `login.gov` was using, and the only way I can think of that would ensure there is no caching of credentials, but I have not found any documentation explicitly supporting that approach.

The fact that you've not dealt with similar in your own use of X.509 auth, and that no other SSO system leveraging X.509 via the browser appears to be documented as doing the same, concerns me that I may have overengineered this ... but then again, I'm reasonably confident this is what `login.gov` is doing.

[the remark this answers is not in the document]
necouchman: Sounds good. My main concern is for those who may not have quite so broad control of DNS systems, or have DNS and/or SSL Certificate Admins who frown upon wildcard subdomains (not that I have any experience with those types of DNS admins...), and how they might implement such a system. But, that's probably a small contingent of folks who would be implementing this type of authentication.

mike-jumper:
Forgot to include the new SSO extension within the dist archive: b9958fa3318423f4b98cec66f734bd10f31ac2c3

mike-jumper:
@necouchman Any other changes needed?

necouchman:
@mike-jumper I'll try to take another sweep through this afternoon, but it looked pretty solid to me.

jmuehlner:
@necouchman looks good to me!

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-ssl/src/main/java/org/apache/guacamole/auth/ssl/SSLClientAuthenticationResource.java
```diff
@@ -0,0 +1,415 @@
...
+     * Decodes the provided URL-encoded string as UTF-8, returning the result.
+     *
+     * @param value
+     *     The URL-encoded string to decode.
+     *
+     * @return
+     *     The decoded string.
+     *
+     * @throws GuacamoleException
+     *     If the provided value is not a value URL-encoded string.
```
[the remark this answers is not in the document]
mike-jumper: Ah, yep - oops.

mike-jumper:
OK - I've pushed some commits to address the typos.

There was also an issue encountered during testing related to Java returning a `byte[]` for the value of the `serialNumber` attribute for any DN that uses `serialNumber` as the username, which I've addressed here by switching to BouncyCastle for parsing the certificate.

necouchman (approved):
Okay, good with me. :shipit:

lordneeko:
@mike-jumper Apologies for ressurecting this, but in your OP, you added
"NOTE: This does not mean that you can pass through that authentication result to an RDP server that accepts smart cards. That's an entirely different can of worms."

Is that something that is at all technically possible to be added within Guacamole? Are there technical limitations that prevents this from being possible? I know client side installed RDP software can typically perform USB redirects (Microsoft AVD, AWS Workspaces, etc), but is there any possibility for browser based RDP to do this?
Thanks.

necouchman:
@lordneeko I believe it is technically possible - RDP (and I believe FreeRDP) supports the pass-through of Smart Cards, so it's really just a matter of being able to (safely) pull that data/stream/connection from the client-side browser and sent it to the FreeRDP library so that it can provide it to the RDP connection. That may be an over-simplified explanation of it, but I think it's within the realm of possibility.

mike-jumper:
@lordneeko Yes, please don't resurrect old PRs to ask questions about Guacamole. The place to ask such things would be the user@guacamole.apache.org mailing list:

https://guacamole.apache.org/support/#mailing-lists

PR #798, opened 2023-02-27
GUACAMOLE-1744: Automatically clean up UI if session has expired in the background.

mike-jumper:
This change adds a periodic check that automatically clears out the local UI state if the server has invalidated the session. The main purpose of this is to ensure the UI always reflects server-side state within a reasonable timeframe, especially for the case that an admin may inadvertently leave a page open containing sensitive data.

The check is implemented with a `HEAD` request to `.../api/session` - a new REST API mechanism which returns HTTP 204 if the session is valid but does not affect that session's last access timestamp (all other REST API requests for resources within a session will update that timestamp).

### DIFF
--- a/guacamole/src/main/java/org/apache/guacamole/rest/auth/HashTokenSessionMap.java
+++ b/guacamole/src/main/java/org/apache/guacamole/rest/auth/HashTokenSessionMap.java
@@ -208,12 +208,12 @@ public class HashTokenSessionMap implements TokenSessionMap {
         if (authToken == null)
             return null;
 
-        // Update the last access time and return the GuacamoleSession
-        GuacamoleSession session = sessionMap.get(authToken);
-        if (session != null)
-            session.access();
-
-        return session;
+        // Return the GuacamoleSession having the given auth token (NOTE: We
+        // do not update the access time here, as it is necessary to be able
+        // to retrieve and check the session without causing that session to
+        // be marked as active. Instead, those updates occur as needed when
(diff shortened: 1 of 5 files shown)


guacamole/src/main/java/org/apache/guacamole/rest/auth/HashTokenSessionMap.java
```diff
@@ -209,11 +209,7 @@ public GuacamoleSession get(String authToken) {
             return null;
 
         // Update the last access time and return the GuacamoleSession
-        GuacamoleSession session = sessionMap.get(authToken);
-        if (session != null)
-            session.access();
-
-        return session;
+        return sessionMap.get(authToken);
```
necouchman, line 212: I'm guessing the comment, here, needs to be updated, as well - while it isn't specifically called out, I'm guessing the point is that we no longer want to actually update the session access time when checking the token, because the session has actually been idle?

necouchman (changes requested):
Just one comment that needs to be tweaked and I think this is good.

guacamole/src/main/java/org/apache/guacamole/rest/auth/HashTokenSessionMap.java
mike-jumper: Yep - good catch.

mike-jumper: Fixed via rebase.

PR #805, opened 2023-03-03
GUACAMOLE-1746: Docker Allow usage of custom keystore and custom certificat

ronansalmon:
The start script handles custom keystore/certificat.

### DIFF
--- a/guacamole-docker/README.md
+++ b/guacamole-docker/README.md
@@ -228,32 +228,15 @@ The process for doing this via the `sqlcmd` utilities included
 with SQLServer is documented in
 [the Guacamole manual](http://guacamole.apache.org/doc/gug/jdbc-auth.html#jdbc-auth-sqlserver).
 
-Enabling guacd ssl
+Enabling ssl
 ================
-This explains how to enable ssl between guacamole and guacd using a self signed certificat.
+This explains how to enable ssl between guacamole and guacd using a self signed certificate.
 
-1. Generate a new certificat
-You need to create the new certificat on the guacd host.
+You need to create the new certificate on the guacd host, see https://github.com/apache/guacamole-server/blob/master/README 
+or https://github.com/apache/guacamole-server/blob/master/src/guacd-docker/README.md depending
+on the version you will use (standalone vs docker).
 
-```shell
-openssl genrsa -out /etc/guacd/server.key 2048
-openssl req -new -key /etc/guacd/server.key -out /etc/guacd/cert.csr
-openssl x509 -in /etc/guacd/cert.csr -out /etc/guacd/server.crt -req -signkey /etc/guacd/server.key -days 3650
-openssl pkcs12 -export -in /etc/guacd/server.crt -inkey /etc/guacd/server.key  -out /etc/guacd/server.p12 -CAfile ca.crt -caname root
-```
-2. Configure guacd
-
-On debian, edit /etc/default/guacd and modify the following variables.
-```
-# listen on all interface
-LISTEN_ADDRESS=0.0.0.0
-
-# certificats
-DAEMON_ARGS=-C /etc/guacd/server.crt -K /etc/guacd/server.key
-```
-restart guacd! 
-
-3. Deploy Guacamole
+Copy the SSL certificate server.p12 to /path/guacamole/certs
 
 ```shell
 docker run --name some-guacamole                                  \
@@ -262,13 +245,12 @@ docker run --name some-guacamole                                  \
   -e GUACD_SSL=true                                               \
   -e GUACD_PORT=4822                                              \
   -e GUACD_HOSTNAME=hostname                                      \
-  -v <path to certificat>:/home/guacamole/certs                   \
+  -v /path/guacamole/certs:/home/guacamole/certs                   \
   ...
   -d -p 8080:8080 guacamole/guacamole
 ```
 
-4. From the guacamole web interface, add a new connexion and enable SSL/TLS whenever using a guacd proxy.
-
+From the guacamole web interface, add a new connection and enable SSL/TLS whenever using a guacd proxy.
 
 
 Reporting issues
--- a/guacamole-docker/README.md
+++ b/guacamole-docker/README.md
@@ -228,6 +228,49 @@ The process for doing this via the `sqlcmd` utilities included
 with SQLServer is documented in
 [the Guacamole manual](http://guacamole.apache.org/doc/gug/jdbc-auth.html#jdbc-auth-sqlserver).
 
+Enabling guacd ssl
+================
+This explains how to enable ssl between guacamole and guacd using a self signed certificat.
+
+1. Generate a new certificat
+You need to create the new certificat on the guacd host.
+
+```shell
+openssl genrsa -out /etc/guacd/server.key 2048
+openssl req -new -key /etc/guacd/server.key -out /etc/guacd/cert.csr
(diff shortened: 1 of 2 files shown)


guacamole-docker/README.md
```diff
@@ -228,6 +228,49 @@ The process for doing this via the `sqlcmd` utilities included
 with SQLServer is documented in
 [the Guacamole manual](http://guacamole.apache.org/doc/gug/jdbc-auth.html#jdbc-auth-sqlserver).
 
+Enabling guacd ssl
+================
+This explains how to enable ssl between guacamole and guacd using a self signed certificat.
```
necouchman, line 233: `certificate`

Here, and several other places below in this file.

guacamole-docker/README.md
```diff
@@ -228,6 +228,49 @@ The process for doing this via the `sqlcmd` utilities included
...
+
+On debian, edit /etc/default/guacd and modify the following variables.
+```
+# listen on all interface
+LISTEN_ADDRESS=0.0.0.0
+
+# certificats
+DAEMON_ARGS=-C /etc/guacd/server.crt -K /etc/guacd/server.key
+```
+restart guacd!
```
necouchman, line 254: Why are these steps required for a Docker install? Wouldn't this be steps that would work if you're running guacd natively?

guacamole-docker/README.md
```diff
@@ -228,6 +228,49 @@ The process for doing this via the `sqlcmd` utilities included
...
+  -e GUACOMOLE_SSL_KEYSTORE_PASS=changeme                         \
+  -e GUACD_SSL=true                                               \
+  -e GUACD_PORT=4822                                              \
+  -e GUACD_HOSTNAME=hostname                                      \
+  -v <path to certificat>:/home/guacamole/certs                   \
+  ...
+  -d -p 8080:8080 guacamole/guacamole
+```
+
+4. From the guacamole web interface, add a new connexion and enable SSL/TLS whenever using a guacd proxy.
```
necouchman, line 270: `connexion` -> `connection`

necouchman (changes requested):
A few spelling corrections, but I'm also concerned about the nature of the instructions added to the README.md file.

The instructions seem to cover a scenario where you're running guacd natively on a system but you're running guacamole-client in a Docker container. This is a perfectly valid scenario; however, I'm not sure that the README for guacamole-client within Docker should include instructions that are specific to guacd - either Docker-based or natively installed. It might be better to document that in either the gaucamole-server repo or the Docker chapter in the User Guide (guacamole-manual), and just put a note, here, that refers people to one of those two places?

guacamole-docker/README.md
ronansalmon: @necouchman ,
Yes, I added documentation on how to setup the guacd standalone only with ssl mode.

The guacd doc should probably go somewhere else indeed. I don't think it can go in the Docker chapter in the User Guide since one can use either the guacd standalone or guacd docker or both.

I don't know where it should go ? What would you recommend ?

necouchman: This may be more suited for the Configuring guacd section of the user guide.

ronansalmon: @necouchman 
can I use the same Jira to update guacd section of the user guide ?

necouchman: Sure, I think that would be fine.

ronansalmon: All is now over here
[guacamole-server/pull/419](https://github.com/apache/guacamole-server/pull/419)

guacamole-docker/README.md
```diff
@@ -228,6 +228,31 @@ The process for doing this via the `sqlcmd` utilities included
 with SQLServer is documented in
 [the Guacamole manual](http://guacamole.apache.org/doc/gug/jdbc-auth.html#jdbc-auth-sqlserver).
 
+Enabling ssl
+================
+This explains how to enable ssl between guacamole and guacd using a self signed certificate.
+
+You need to create the new certificate on the guacd host, see https://github.com/apache/guacamole-server/blob/master/README 
+or https://github.com/apache/guacamole-server/blob/master/src/guacd-docker/README.md depending
+on the version you will use (standalone vs docker).
```
necouchman, line 237: I suspect we should document this somewhere in the Guacamole User Guide (Manual), and link to that instead of linking directly to the Github repository.

necouchman (changes requested):

necouchman:
@ronansalmon Any chance we can wrap this one up?

ronansalmon:
Hey !
This PR is linked to https://github.com/apache/guacamole-server/pull/419.
I was waiting for the PR 419 to be merged (if accepted) to go any further.

So, how should we proceed ?

sirux88:
I would add some thoughts:
1) What is the environment variabale `GUACD_SSL` for? As far as I can see it's not used on `guacamole-client` nor on `guacamole-server` anywhere (except within the changed files). Seems obsolete to me
2) Maybe the environment variables `GUACD_SSL_KEYSTORE_FILE` and `GUACD_SSL_KEYSTORE_PASS` should be renamed to something fitting better to their true purpose like `JAVA_KEYSTORE_FILE` and `JAVA_KEYSTORE_PASS` 

For reference: I had a similar approach here https://github.com/sirux88/guacamole-client/commit/e02d2bef55ee428c1da7eb1ecc7f77c7514588ad. My use case was using a self signed cert for LDAP

mike-jumper:
The `GUACD_SSL` environment variable controls the `guacd-ssl` property, which is used to enable SSL/TLS encryption between the webapp and guacd. It's uncommon to use this, as that communication is typically purely over loopback or at least on a trusted network route, but it's not obsolete.

mike-jumper:
> > The `GUACD_SSL` environment variable controls the `guacd-ssl` property, which is used to enable SSL/TLS encryption between the webapp and guacd. It's uncommon to use this, as that communication is typically purely over loopback or at least on a trusted network route, but it's not obsolete.
> 
> I still can't find it. But if you're saying it's used I'm ok with this

For reference, here's the code path is:

### Property Declaration (standardized as part of guacamole-ext)

https://github.com/apache/guacamole-client/blob/ac6e501eff8b11f0c4b9a5bad653e237b979b14d/guacamole-ext/src/main/java/org/apache/guacamole/environment/Environment.java#L61-L69

### Property Retrieval (via a `GuacamoleProxyConfiguration` from `LocalEnvironment`)

https://github.com/apache/guacamole-client/blob/ac6e501eff8b11f0c4b9a5bad653e237b979b14d/guacamole-ext/src/main/java/org/apache/guacamole/environment/LocalEnvironment.java#L379-L390

### (Voluntary) Usage of Retrieved Values

Implementations aren't strictly _required_ to use `getDefaultGuacamoleProxyConfiguration()`, but most do, either explicitly or implicitly via `SimpleConnection`.

https://github.com/apache/guacamole-client/blob/ac6e501eff8b11f0c4b9a5bad653e237b979b14d/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleConnection.java#L206

https://github.com/apache/guacamole-client/blob/ac6e501eff8b11f0c4b9a5bad653e237b979b14d/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/connection/ModeledConnection.java#L462

etc.

ronansalmon:
Hi, @tiborauer I gave up on this PR and #419. Feel free to close or take over on both of them.

PR #807, opened 2023-03-08
GUACAMOLE-839: Correct handling of possible null values within SSL/TLS client auth support.

mike-jumper:
This change:

* Updates `WildcardURIGuacamoleProperty` such that it explicitly handles possible `null` inputs (omitted properties).
* Updates handling of received PEM such that anything but a single X.509 certificate is rejected, including PEM that contains nothing at all.


jmuehlner (approved):
Looks reasonable to me.

PR #808, opened 2023-03-14
GUACAMOLE-1219: Add support for disabling TOTP for specific users and groups.

necouchman:
Continuing the work started in #577, this adds an attribute to the TOTP module at both a user and group level that allows for disabling the TOTP requirement for a certain user or users who are members of one or more groups that have the requirement disabled.

This closes #577.

### DIFF
--- a/extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/user/UserVerificationService.java
+++ b/extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/user/UserVerificationService.java
@@ -26,19 +26,23 @@ import java.security.InvalidKeyException;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Set;
 import javax.servlet.http.HttpServletRequest;
 import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.GuacamoleSecurityException;
 import org.apache.guacamole.GuacamoleUnsupportedException;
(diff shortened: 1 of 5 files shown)


extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/user/UserVerificationService.java
```diff
@@ -203,6 +207,65 @@ private boolean setKey(UserContext context, UserTOTPKey key)
...
+        Set<String> userGroups = authenticatedUser.getEffectiveUserGroups();
+        Directory<UserGroup> directoryGroups = context.getPrivileged().getUserGroupDirectory();
+        for (String userGroup : userGroups) {
+            UserGroup thisGroup = directoryGroups.get(userGroup);
+            if (thisGroup == null)
+                continue;
+            
+            Map<String, String> attributes = thisGroup.getAttributes();
+            if (attributes != null 
+                    && !attributes.isEmpty()
```
jmuehlner, line 254: This check doesn't seem necessary - if a map is empty it will never contain the key. For that matter, line 255 shouldn't be needed either - if the key doesn't exist in the map, it'll return `null`, and that will fail the check on line 256.

necouchman: Yeah, that makes sense - I was having some issues with `NullPointerException`s in/around this area, which is why I put all of the extra checks in place.

Fixed via rebase.

jmuehlner (approved):

PR #809, opened 2023-03-14
GUACAMOLE-926: Batch Connection Import

jmuehlner:
~There's still a bit of cleanup to be done here, but the functionality is essentially done, so if people want to have a look, I'd appreciate it.~

This PR implements a batch import UI for connections, allowing an admin to provide a CSV, YAML, or JSON file with a list of connections, which will be submitted to a modified PATCH implementation on the backend that now allows creation of bulk directory objects.

Screenshot 2023-03-14 112023
Screenshot 2023-03-14 111613
Screenshot 2023-03-14 111632
Screenshot 2023-03-14 111658
Screenshot 2023-03-14 111743


mike-jumper:
How is the "Row #" heading intended to be interpreted when importing from JSON or YAML?

jmuehlner:
> How is the "Row #" heading intended to be interpreted when importing from JSON or YAML?

It's the index (+1) within the array of connection objects in the file. If you have a different suggestion for the header I'm down to change it.

jmuehlner:
Ok, I think this is now cleaned up enough to be ready for review. Looking forward to hearing people's thoughts, nitpicks, etc.

necouchman:
I've always got some good nits to pick.

doc/licenses/process-nextick-args-2.0.1/README
```diff
@@ -0,0 +1,8 @@
+process-nextick-args (https://github.com/calvinmetcalf/process-nextick-args)
+---------------------------------------------
+
+    Version: 2.0.1
+    From: 'Calvin Metcalf' (https://github.com/calvinmetcalf)
+    License(s):
+        MIT (bundled/process-nextick-args-2.0.1/LICENSE)
```
mike-jumper, line 7: The bundled file is actually `license.md`.

doc/licenses/readable-stream-2.3.7/README
```diff
@@ -0,0 +1,8 @@
+readable-stream (https://github.com/nodejs/readable-stream)
+---------------------------------------------
+
+    Version: 2.3.7
+    From: 'Node.js' (https://github.com/nodejs)
```
mike-jumper, line 5: According to the license, this should be "Node.js contributors, Joyent, Inc., and other Node contributors".

doc/licenses/setimmediate-1.0.5/README
```diff
@@ -0,0 +1,8 @@
+setImmediate.js (https://github.com/YuzuJS/setImmediate)
+---------------------------------------------
+
+    Version: 1.0.5
+    From: 'Yuzu (by Barnes & Noble Education)' (https://github.com/YuzuJS)
+    License(s):
+        MIT (bundled/setimmediate-1.0.5/LICENSE)
```
mike-jumper, line 7: The bundled file is actually `LICENSE.txt`.

doc/licenses/stream-browserify-2.0.2/README
```diff
@@ -0,0 +1,8 @@
+stream-browserify (https://github.com/browserify/stream-browserify)
+---------------------------------------------
+
+    Version: 2.0.2
+    From: 'browserify' (https://github.com/browserify)
```
mike-jumper, line 5: From the license, this should be "James Halliday".

doc/licenses/string_decoder-1.1.1/README
```diff
@@ -0,0 +1,8 @@
+string_decoder (https://github.com/nodejs/string_decoder)
+---------------------------------------------
+
+    Version: 1.1.1
+    From: 'Node.js' (https://github.com/nodejs)
```
mike-jumper, line 5: According to the license, this should be "Node.js contributors, Joyent, Inc., and other Node contributors".

doc/licenses/timers-browserify-2.0.12/README
```diff
@@ -0,0 +1,8 @@
+timers-browserify (https://github.com/browserify/timers-browserify)
+---------------------------------------------
+
+    Version: 2.0.12
+    From: 'browserify' (https://github.com/browserify)
```
mike-jumper, line 5: According to the license, this should be "J. Ryan Stinnett".

doc/licenses/timers-browserify-2.0.12/README
```diff
@@ -0,0 +1,8 @@
...
+    License(s):
+        MIT (bundled/timers-browserify-2.0.12/LICENSE)
```
mike-jumper, line 7: The bundled file is actually `LICENSE.md`.

mike-jumper (changes requested):
Initial review of just the additional bundled dependency licenses:

* Some of the names of the copyright holders (`From: ...`) need to be corrected to align with the copyright holders declared by the license.
* Some of the references to `.../LICENSE` need to be corrected to the name of the bundled license file (`.../LICENSE.md`, `.../license.md`, etc.).

doc/licenses/csv-6.2.5/dep-coordinates.txt
```diff
@@ -0,0 +1,2 @@
+csv:6.2.5
+csv-parse:5.3.3
```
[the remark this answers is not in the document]
jmuehlner: Yep

doc/licenses/csv-6.2.5/README
```diff
@@ -0,0 +1,7 @@
+node-csv (https://github.com/adaltas/node-csv)
+---------------------------------------------
+
+    Version: 6.2.5
+    From: 'adaltas' (https://github.com/adaltas)
```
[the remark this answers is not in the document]
jmuehlner: > Initial review of just the additional bundled dependency licenses:
> 
> * Some of the names of the copyright holders (`From: ...`) need to be corrected to align with the copyright holders declared by the license.
> * Some of the references to `.../LICENSE` need to be corrected to the name of the bundled license file (`.../LICENSE.md`, `.../license.md`, etc.).

Oof, sorry about all the mistakes. These should all be fixed now.

guacamole/src/main/java/org/apache/guacamole/rest/jsonpatch/APIPatchOutcome.java
```diff
@@ -0,0 +1,110 @@
...
+     */
+    private final String path;
+
+    /**
+     * Create an outcome associated with a submitted patch, as part of a JSON
+     * patch API request.
+     *
+     * @param op
+     * @param identifier
+     * @param path
```
necouchman, line 60: Looks like these parameters are missing documentation?

guacamole/src/main/java/org/apache/guacamole/rest/directory/DirectoryResource.java
```diff
@@ -386,46 +417,210 @@ public Map<String, ExternalType> getObjects(
...
     @PATCH
-    public void patchObjects(List<APIPatch<String>> patches)
+    public APIPatchResponse patchObjects(List<APIPatch<ExternalType>> patches)
             throws GuacamoleException {
 
-        // Apply each operation specified within the patch
-        for (APIPatch<String> patch : patches) {
+        // An outcome for each patch included in the request. This list
+        // may include both success and failure responses, though the
+        // presense of any failure would indicated that the entire
```
necouchman, line 444: presense -> presence

?

guacamole/src/main/java/org/apache/guacamole/rest/directory/DirectoryResource.java
```diff
@@ -386,46 +417,210 @@ public Map<String, ExternalType> getObjects(
...
+
+                // Apply each operation specified within the patch
+                for (APIPatch<ExternalType> patch : patches) {
+
+                    // Retrieve and validate path
+                    String path = patch.getPath();
+                    if (!path.startsWith("/"))
+                        throw new GuacamoleClientException("Patch paths must start with \"/\".");
+
+                    if(patch.getOp() == APIPatch.Operation.add) {
```
necouchman, line 485: I hesitate to question this at all, because I know you know what you're doing, but i"ll throw a couple of things out there - feel free to ignore them if you've already considered and have reasons for doing this the way you have it written, here...
* Throughout this section, there are multiple calls to `patch.getOp()`. I don't know what the cost is in terms of doing that versus assigning the value to a variable and checking throughout this section? Maybe it's negligible, and doesn't matter, and it makes sense to keep it this way? Or maybe it occupies more memory than it's worth and the cost of doing it this way is less than going the variable route?
* I know there are only two operations here (three if you count the catch-all at the end), but any reason to go `if/else` rather than a `switch()` statement?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/JDBCDirectory.java
```diff
@@ -0,0 +1,44 @@
...
+import org.apache.guacamole.GuacamoleException;
+import org.apache.guacamole.net.auth.AtomicDirectoryOperation;
+import org.apache.guacamole.net.auth.Directory;
+import org.apache.guacamole.net.auth.Identifiable;
+import org.mybatis.guice.transactional.Transactional;
+
+/**
+ * An implementation of Directory that uses database transactions to guarantee
+ * atomicity for any operations supplied to tryAtomically().
+ */
```
necouchman, line 31: Does this implementation, coupled with the fact that it is used elsewhere, mean that all JDBC operations will be attempted atomically? And does that have any potential adverse impacts from a performance perspective, particularly when you have multiple people accessing the Guacamole UI, starting and stopping connections, which will involve DB reads to retrieve connection data and DB writes for history entries, etc.

Sorry if that's a stupid or obvious question...

guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -5,7 +5,7 @@
     "APP" : {
 
         "NAME"    : "Apache Guacamole",
-        "VERSION" : "${project.version}",
+        "VERSION" : "1.5.0",
```
necouchman, line 8: Any reason this is being set statically, here?

guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -183,6 +183,79 @@
...
+        "TABLE_HEADER_PROTOCOL" : "Protocol",
+        "TABLE_HEADER_ERRORS" : "Errors",
+        "TABLE_HEADER_ROW_NUMBER": "Row #",
+
+        "UPLOAD_FILE_TYPES": "CSV, JSON, or YAML",
+        "UPLOAD_HELP_LINK": "View Format Tips",
+        "UPLOAD_DROP_TITLE": "Drop a File Here",
+        "UPLOAD_BROWSE_LINK": "Browse for File"
+        
+    },
```
necouchman, line 257: Seems like the style, here, doesn't match the other sections, in a couple of ways:
* Generally things are organized alphabetically, and so `ERROR_` should be between `DIALOG_` and `FIELD_`.
* There are several line breaks here between the key and value pairs.

guacamole/src/main/java/org/apache/guacamole/rest/APIError.java
```diff
@@ -21,6 +21,8 @@
 
 import java.util.Collection;
 import java.util.Collections;
+import java.util.List;
+
```
necouchman, line 25: Extra line.

necouchman (changes requested):
I've got an initial review done, with a few issues - mostly style nitpicks, as usual :-).

guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -183,6 +183,79 @@
 
     },
 
+    "IMPORT": {
+        
+        "ACTION_ACKNOWLEDGE"          : "@:APP.ACTION_ACKNOWLEDGE",
+
+        "BUTTON_CANCEL": "Cancel",
+        "BUTTON_CLEAR": "Clear",
+        "BUTTON_IMPORT": "Import Connections",
```
mike-jumper, line 192: The prefix for actions that may be taken, button or not, is `ACTION_`. Here, for generic actions like "Cancel" and "Clear", they should be defined at the `APP` level and then referenced.

guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -183,6 +183,79 @@
...
+
+        "CONNECTIONS_IMPORTED_SUCCESS": "{NUMBER} connections imported successfully.",
```
mike-jumper, line 194: You'll need to use pluralization here so that "connection" vs. "connections" is chosen appropriately based on the value of `NUMBER`.

guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -183,6 +183,79 @@
...
+
+        "DIALOG_HEADER_ERROR" : "@:APP.DIALOG_HEADER_ERROR",
+        "DIALOG_HEADER_SUCCESS": "Success",
+
+        "FIELD_PLACEHOLDER_FILTER" : "@:APP.FIELD_PLACEHOLDER_FILTER",
+
+        "HEADER": "Connection Import",
```
mike-jumper, line 201: The name of this key needs to be more specific to the meaning of the string, not just `HEADER`. Assuming this is for a section of a page, the prefix should also be `SECTION_HEADER_`.

guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -183,6 +183,79 @@
...
+
+        "HELP_HEADER": "Connection Import File Format",
```
mike-jumper, line 203: Same here - this needs to be more specific to the meaning of the string and use the established translation key prefixes.

guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -183,6 +183,79 @@
...
+
+        "HELP_FILE_TYPE_HEADER": "File Types",
```
mike-jumper, line 205: `SECTION_HEADER_...`?

guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -183,6 +183,79 @@
...
+
+        "TABLE_HEADER_NAME" : "Name",
+        "TABLE_HEADER_PROTOCOL" : "Protocol",
+        "TABLE_HEADER_ERRORS" : "Errors",
+        "TABLE_HEADER_ROW_NUMBER": "Row #",
+
+        "UPLOAD_FILE_TYPES": "CSV, JSON, or YAML",
+        "UPLOAD_HELP_LINK": "View Format Tips",
+        "UPLOAD_DROP_TITLE": "Drop a File Here",
+        "UPLOAD_BROWSE_LINK": "Browse for File"
```
mike-jumper, line 255: This is probably an `ACTION_`.

guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -183,6 +183,79 @@
...
+        "ERROR_FILE_SINGLE_ONLY": "Please upload only a single file at a time",
+
+        "TABLE_HEADER_NAME" : "Name",
+        "TABLE_HEADER_PROTOCOL" : "Protocol",
+        "TABLE_HEADER_ERRORS" : "Errors",
+        "TABLE_HEADER_ROW_NUMBER": "Row #",
+
+        "UPLOAD_FILE_TYPES": "CSV, JSON, or YAML",
+        "UPLOAD_HELP_LINK": "View Format Tips",
+        "UPLOAD_DROP_TITLE": "Drop a File Here",
```
mike-jumper, line 254: This is probably a `HELP_`.

guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -183,6 +183,79 @@
...
+
+        "ERROR_FILE_SINGLE_ONLY": "Please upload only a single file at a time",
+
+        "TABLE_HEADER_NAME" : "Name",
+        "TABLE_HEADER_PROTOCOL" : "Protocol",
+        "TABLE_HEADER_ERRORS" : "Errors",
+        "TABLE_HEADER_ROW_NUMBER": "Row #",
+
+        "UPLOAD_FILE_TYPES": "CSV, JSON, or YAML",
+        "UPLOAD_HELP_LINK": "View Format Tips",
```
mike-jumper, line 253: `ACTION_...`?

guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -183,6 +183,79 @@
...
+            "No connection name found in the provided file",
+
+        "ERROR_FILE_SINGLE_ONLY": "Please upload only a single file at a time",
+
+        "TABLE_HEADER_NAME" : "Name",
+        "TABLE_HEADER_PROTOCOL" : "Protocol",
+        "TABLE_HEADER_ERRORS" : "Errors",
+        "TABLE_HEADER_ROW_NUMBER": "Row #",
+
+        "UPLOAD_FILE_TYPES": "CSV, JSON, or YAML",
```
mike-jumper, line 252: `INFO_...`? Or `HELP_...`?

guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -183,6 +183,79 @@
...
+
+        "HELP_CSV_HEADER": "CSV Format",
+        "HELP_CSV_DESCRIPTION": "A connection import CSV file has one connection record per row. Each column will specify a connection field. At minimum the connection name and protocol must be specified.",
+        "HELP_CSV_MORE_DETAILS": "The CSV header for each row specifies the connection field. The connection group ID that the connection should be imported into may be directly specified with \"parentIdentifier\", or the path to the parent group may be specified using \"group\" as shown below. In most cases, there should be no conflict between fields, but if needed, an \" (attribute)\" or \" (parameter)\" suffix may be added to disambiguate. Lists of user or user group identifiers must be semicolon-seperated.¹",
```
mike-jumper, line 210: Is that `¹` at the end intentional?

guacamole/src/main/frontend/src/images/action-icons/guac-monitor-add-many.svg
mike-jumper, line 1: I think it would make more sense for this icon to show multiple monitors than multiple plus signs.

guacamole/src/main/frontend/src/translations/en.json
mike-jumper, line 194: This needs a prefix - presumably `INFO_`?

guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -905,6 +978,7 @@
     "SETTINGS_CONNECTIONS" : {
 
         "ACTION_ACKNOWLEDGE"          : "@:APP.ACTION_ACKNOWLEDGE",
+        "ACTION_IMPORT_CONNECTIONS"   : "Import",
```
mike-jumper, line 981: This should probably be declared generically as `ACTION_IMPORT` at the `APP` level, and then referenced here.

mike-jumper (changes requested):
This review covers mainly the translation strings. Overall:

* The strings need prefixes that align with the established naming conventions.
* The strings need names that describe the nature of the string.
* The primary definition of some of these strings should probably be moved up to the `APP` level.
* Need to use pluralization.

guacamole/src/main/frontend/src/app/import/directives/connectionImportErrors.js
```diff
@@ -0,0 +1,234 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/* global _ */
```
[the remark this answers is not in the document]
jmuehlner: This is a [JSLint ](https://www.jslint.com/help.html) directive indicating that the `_` variable is a global defined outside this file. This makes Netbeans happy, among other possible benefits.

There's a prior example here: https://github.com/apache/guacamole-client/blob/master/guacamole/src/main/frontend/src/app/client/types/ManagedClient.js#L20

guacamole/src/main/java/org/apache/guacamole/rest/directory/DirectoryResource.java
jmuehlner: Oops yes. Will fix.

guacamole/src/main/frontend/src/app/import/directives/connectionImportErrors.js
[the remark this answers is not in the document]
jmuehlner: Definitely not needed per se - it builds and runs fine, but helpful yes. 😃

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/JDBCDirectory.java
jmuehlner: The only place this is used is in the [Directory PATCH endpoint](https://github.com/jmuehlner/guacamole-client/blob/GUACAMOLE-926-batch-connection-import/guacamole/src/main/java/org/apache/guacamole/rest/directory/DirectoryResource.java#L449), so it should have no effects on performance or anything else for any other JDBC call. 

So no, not all JDBC operations will be attempted atomically.

guacamole/src/main/java/org/apache/guacamole/rest/directory/DirectoryResource.java
jmuehlner: I would expect the cost of calling `getOp()` to be negligible considering that it doesn't do any calculation and just returns a value. There wouldn't be any harm in creating a copy of the variable in the local scope either though. 

I suspect that the JDK or JVM might optimize away the cost of the extra stack frame for the `getOp()` call, but I'm not 100% sure about that. I'll change it over to a local variable and then we don't have to worry about it.

As for the switch statement - I mostly just wanted to avoid another layer of indentation, since an `if` adds one level, and 
```
switch(op) {
    case add:
         ... stuff ...
```
adds two.

I'll admit that a switch statement is probably the "right way to do it" though. If people have strong preferences here, I'm happy to change it over.

guacamole/src/main/frontend/src/translations/en.json
jmuehlner: Yes.

jmuehlner:
> 

Ok, I think these should all be resolved now...

mike-jumper (changes requested):

guacamole/src/main/frontend/src/images/action-icons/guac-monitor-add-many.svg
mike-jumper: What do you think of:

Multiple monitors, overlapping

or:

Multiple smaller monitors, overlapping

?

(GitHub doesn't seem to provide a way to set the background color of a comment, paragraph, section, etc., so the above may be invisible...)

guacamole/src/main/frontend/src/app/import/templates/connectionImportFileHelp.html
```diff
@@ -0,0 +1,93 @@
...
+    <p>{{'IMPORT.HELP_FILE_TYPE_DESCRIPTION' | translate}}</p>
+
+    <h2>{{'IMPORT.HELP_CSV_HEADER' | translate}}</h2>
+    <p>{{'IMPORT.HELP_CSV_DESCRIPTION' | translate}}</p>
+    <p>{{'IMPORT.HELP_CSV_MORE_DETAILS' | translate}}</p>
+    <pre>name,protocol,hostname,group,users,groups,guacd-encryption (attribute)
+conn1,vnc,conn1.web.com,ROOT,guac user 1;guac user 2,Connection 1 Users,none
+conn2,rdp,conn2.web.com,ROOT/Parent Group,guac user 1,,ssl
+conn3,ssh,conn3.web.com,ROOT/Parent Group/Child Group,guac user 2;guac user 3,,
+conn4,kubernetes,,,,,</pre>
```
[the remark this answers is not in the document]
jmuehlner: Should it? I had assumed that this sort of example file content usually isn't. Plus doing all this formatting within the translation system seems like it'd be a pain - probably having to split this apart line by line.

guacamole/src/main/frontend/src/images/action-icons/guac-monitor-add-many.svg
jmuehlner: I do like the look of those better than the current icon, but unfortunately they don't render well in actual use on the button. The first one looks pretty similar to the current button, with a bit more hair on top.
image

And the bottom one just looks like a blur to me.
image

guacamole/src/main/java/org/apache/guacamole/rest/directory/DirectoryResource.java
necouchman: No strong preference, just curious.

guacamole/src/main/frontend/src/app/import/templates/connectionImportFileHelp.html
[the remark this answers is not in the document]
mike-jumper: Since the example CSV itself contains English that's meant to be human-readable, I think it should be (if possible). Does angular-translate strip newlines from the translated text? If not, I'd think we could have something like:

    "IMPORT" : {
        "INFO_EXAMPLE_CSV" : "a,b,c\n1,2,3\n4,5,6"
    },

and:

    <pre>{{ 'IMPORT.INFO_EXAMPLE_CSV' | translate }}</pre>

guacamole/src/main/frontend/src/app/rest/services/connectionService.js
```diff
@@ -24,9 +24,11 @@ angular.module('rest').factory('connectionService', ['$injector',
         function connectionService($injector) {
 
     // Required services
-    var requestService        = $injector.get('requestService');
     var authenticationService = $injector.get('authenticationService');
     var cacheService          = $injector.get('cacheService');
+
+    // Required types
+    const Error = $injector.get('Error');
```
mike-jumper, line 31: Is this used? I don't see any usages in the relevant part of the diff.

guacamole/src/main/frontend/src/app/index/config/indexRouteConfig.js
```diff
@@ -126,6 +126,23 @@ angular.module('index').config(['$routeProvider', '$locationProvider',
...
+        // Connection import page
+        .when('/import/:dataSource/connection', {
+            title         : 'APP.NAME',
+            bodyClassName : 'settings',
+            templateUrl   : 'app/import/templates/connectionImport.html',
+            controller    : 'importConnectionsController',
+            resolve       : { updateCurrentToken: updateCurrentToken }
+        })
+
+        // Connection import page
```
mike-jumper, line 138: Copypasta?

guacamole/src/main/frontend/src/app/import/types/ParseError.js
```diff
@@ -0,0 +1,68 @@
...
+
+        /**
+         * The object which should be passed through to the translation service
+         * for the sake of variable substitution. Each property of the provided
+         * object will be substituted for the variable of the same name within
+         * the translation string.
+         *
+         * @type Object
+         */
+        this.variables = template.variables;
```
mike-jumper, line 62: Should this extend or otherwise leverage `TranslatableMessage`?

guacamole/src/main/frontend/src/app/import/types/DisplayErrorList.js
```diff
@@ -0,0 +1,83 @@
...
+     */
+    const DisplayErrorList = function DisplayErrorList(messages) {
+
+        // Use empty message list by default
+        this.messages = messages || [];
+
+        // The single String message composed of all messages concatenated
+        // together. This will be used for filtering / sorting, and should only
+        // be calculated once.
+        this.cachedMessage = null;
```
mike-jumper, line 42: Please document object properties using JSDoc.

mike-jumper, line 42: This doesn't appear to be used, though there is a reference to `concatenatedMessage`.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/AtomicDirectoryOperation.java
```diff
@@ -0,0 +1,57 @@
...
+ * the Directory.
+ */
+public interface AtomicDirectoryOperation<ObjectType extends Identifiable>  {
+
+    /**
+     * Attempt the operation atomically. If the Directory does not support
+     * atomic operations, the atomic flag will be set to false. If the atomic
+     * flag is set to true, the provided directory is guaranteed to perform
+     * the operations within this function atomically. Atomicity of the
+     * provided directory outside this function, or of the directory invoking
```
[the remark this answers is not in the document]
jmuehlner: if?

guacamole/src/main/java/org/apache/guacamole/rest/directory/DirectoryResource.java
```diff
@@ -386,46 +417,211 @@ public Map<String, ExternalType> getObjects(
...
+            public void executeOperation(boolean atomic, Directory<InternalType> directory)
+                    throws GuacamoleException {
+
+                // If the underlying directory implentation does not support
+                // atomic operations, abort the patch operation. This REST
+                // endpoint requires that operations be performed atomically.
+                if (!atomic)
+                    throw new GuacamoleUnsupportedException(
+                            "Atomic operations are not supported. " +
+                            "The patch cannot be executed.");
```
mike-jumper, line 461: Recommend rephrasing this such that the message will indicate to the administrator the nature of the failure. "Atomic operations are not supported" sounds like Guacamole overall does not support the operation, whereas the issue is that an extension does not support atomic operations and therefore can't reliably perform patching.

guacamole/src/main/java/org/apache/guacamole/rest/jsonpatch/APIPatchError.java
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * Return the error associated with the patch failure.
+     *
+     * @return
+     *     The error associated with the patch failure.
+     */
+    public String getError() {
+        return error;
+    }
```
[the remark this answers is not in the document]
jmuehlner: Yeah, probably actually. 

I can't guarantee that the caught errors will actually be translateable, but I can at least check if it implements `Translatable` and handle appropriately if so.

guacamole/src/main/frontend/src/app/rest/types/DirectoryPatch.js
```diff
@@ -0,0 +1,89 @@
...
+
+        /**
+         * Adds the specified object to the relation.
+         */
+        ADD : "add",
+
+        /**
+         * Removes the specified object from the relation.
+         */
+        REMOVE : "remove"
```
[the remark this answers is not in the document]
jmuehlner: Oops!

guacamole/src/main/frontend/src/app/import/controllers/importConnectionsController.js
```diff
@@ -0,0 +1,760 @@
...
+
+        // Choose the appropriate parse function based on the mimetype
+        if (mimeType.endsWith("json"))
+            processDataCallback = connectionParseService.parseJSON;
+
+        else if (mimeType.endsWith("csv"))
+            processDataCallback = connectionParseService.parseCSV;
+
+        else if (mimeType.endsWith("yaml"))
+            processDataCallback = connectionParseService.parseYAML;
```
mike-jumper, line 473: I think we should map the specific, known mimetypes (`application/json`, `text/csv`, etc.) to their corresponding handlers rather than forgivingly accepting anything that ends with specific substrings.

guacamole/src/main/frontend/src/app/import/controllers/importConnectionsController.js
```diff
@@ -0,0 +1,760 @@
...
+     * complete.
+     */
+    $scope.import = () => processData($scope.mimeType, $scope.fileData);
+
+    /**
+     * @return {Boolean}
+     *     True if import should be disabled, or false if import should be
+     *     allowed.
+     */
+    $scope.importDisabled = () =>
```
mike-jumper, line 505: Overall description missing from this JSDoc.

guacamole/src/main/frontend/src/app/import/controllers/importConnectionsController.js
```diff
@@ -0,0 +1,760 @@
...
+     */
+    const uploadContainer = angular.element(
+            $document.find('.file-upload-container'));
+
+    /**
+     * The location where files can be dragged-and-dropped to.
+     *
+     * @type Element
+     */
+    const dropTarget = uploadContainer.find('.drop-target');
```
mike-jumper, line 646: Should this be moved to a directive such that the controller for this page does not directly manipulate the DOM?

guacamole/src/main/frontend/src/app/import/directives/connectionImportErrors.js
```diff
@@ -0,0 +1,234 @@
...
+
+/**
+ * A directive that displays errors that occured during parsing of a connection
```
mike-jumper, line 23: occurred*

guacamole/src/main/frontend/src/app/import/types/ParseResult.js
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * Service which defines the ParseResult class.
+ */
+angular.module('import').factory('ParseResult', [function defineParseResult() {
+
+    /**
+     * The result of parsing a connection import file - containing a list of
+     * API patches ready to be submitted to the PATCH REST API for batch
+     * connection creation, a set of users and user groups to grant access to
+     * each connection, and any errors that may have occured while parsing
```
mike-jumper, line 29: occurred*

guacamole/src/main/frontend/src/app/import/types/ParseResult.js
```diff
@@ -0,0 +1,89 @@
...
+         * identifiers.
+         *
+         * @type {Object.<String, Boolean>}
+         */
+        this.groups = template.users || {};
+
+        /**
+         * An array of errors encountered while parsing the corresponding
+         * connection (at the same array index). Each connection should have a
+         * an array of errors. If empty, no errors occured for this connection.
```
mike-jumper, line 71: occurred*

guacamole/src/main/frontend/src/app/import/controllers/importConnectionsController.js
```diff
@@ -0,0 +1,760 @@
...
+            // If an error occurs while trying to users or groups, or while trying
+            // to assign permissions to users or groups, clean up the already-created
+            // connections, displaying an error to the user along with a blank slate
+            // so they can fix their problems and try again.
+            .catch(error => {
+                cleanUpConnections(connectionResponse);
+                handleError(error);
+            }))
+
+        // If an error occured when the call to create the connections was made,
```
mike-jumper, line 391: occurred*

guacamole/src/main/frontend/src/app/import/services/connectionCSVService.js
```diff
@@ -0,0 +1,430 @@
...
+ * under the License.
+ */
+
+/* global _ */
+
+// A suffix that indicates that a particular header refers to a parameter
+const PARAMETER_SUFFIX = ' (parameter)';
+
+// A suffix that indicates that a particular header refers to an attribute
+const ATTRIBUTE_SUFFIX = ' (attribute)';
```
mike-jumper, line 26: Please document with JSDoc.

mike-jumper (changes requested):

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/AtomicDirectoryOperation.java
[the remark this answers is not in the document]
mike-jumper: Never mind - I misread this. There's no typo here.

guacamole/src/main/frontend/src/app/import/templates/connectionImportFileHelp.html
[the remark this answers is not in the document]
jmuehlner: Adding `\n`s into the translation value in the JSON _does_ work! I'll switch all these over.

guacamole/src/main/frontend/src/app/rest/services/connectionService.js
jmuehlner: Nope. I'll get rid of it.

guacamole/src/main/frontend/src/app/import/controllers/importConnectionsController.js
jmuehlner: That's the way I originally had it, but the directive and controller got so intertangled that it ended up being cleaner to merge them.

guacamole/src/main/frontend/src/app/import/types/DisplayErrorList.js
jmuehlner: Oops, that's supposed to be `concatenatedMessage`.

guacamole/src/main/java/org/apache/guacamole/rest/directory/DirectoryResource.java
jmuehlner: How does this look?

guacamole/src/main/frontend/src/app/import/controllers/importConnectionsController.js
jmuehlner: Sure, seems fair.

guacamole/src/main/java/org/apache/guacamole/rest/jsonpatch/APIPatchFailureException.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.util.List;
+
+import org.apache.guacamole.GuacamoleClientException;
+
+/**
+ * An exception describing a failure to apply the patches from a JSON Patch
+ * request. A list of outcomes is included, one for each patch in the request.
+ */
+public class APIPatchFailureException extends GuacamoleClientException {
```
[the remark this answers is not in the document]
jmuehlner: The `message` here is never displayed to the user, so I'm not sure how much this matters? 

The only things that seem like they really need translation are the the errors in the `patches` list.

jmuehlner:
I'm seeing this error on CI (though not locally)
```
Error:  Failed to execute goal com.github.searls:jasmine-maven-plugin:3.0-beta-02:test (default) on project guacamole-common-js: The jasmine-maven-plugin encountered an exception:: session not created: This version of ChromeDriver only supports Chrome version 112
Error:  Current browser version is 111.0.5563.110 with binary path /usr/bin/chromium
Error:  Build info: version: '3.141.59', revision: 'e82be7d358', time: '2018-11-14T08:17:03'
Error:  System info: host: 'a24ad6ba7e99', ip: '172.17.0.2', os.name: 'Linux', os.arch: 'amd64', os.version: '5.15.0-1034-azure', java.version: '1.8.0_342'
Error:  Driver info: driver.version: ChromeDriver
```

Hopefully it goes away by itself soon...

guacamole/src/main/java/org/apache/guacamole/rest/jsonpatch/APIPatchError.java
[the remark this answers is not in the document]
mike-jumper: Can also leverage `APP.TEXT_UNTRANSLATED` to turn any untranslated string into a translated string:

https://github.com/apache/guacamole-client/blob/165bd413c0ab9c55b0c0e3614fcb829ef5f8ca95/guacamole/src/main/java/org/apache/guacamole/rest/APIError.java#L41-L52

https://github.com/apache/guacamole-client/blob/165bd413c0ab9c55b0c0e3614fcb829ef5f8ca95/guacamole/src/main/java/org/apache/guacamole/rest/APIError.java#L205-L208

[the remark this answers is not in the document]
jmuehlner: Yes good call I will do that.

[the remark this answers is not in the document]
jmuehlner: Or wait, I already did this.

jmuehlner:
Updated the connection icon to something a bit more distinct. (thanks to @mike-jumper for this one)
image

jmuehlner:
Converting to draft while it give it one more once-over.

mike-jumper:
Looks good to me! Any further comments @necouchman?

PR #802, opened 2023-03-02
GUACAMOLE-1738: Update version of MySQL connector

kelderek:
GUACAMOLE-1738 Update version of MySQL connector used to resolve issue 1738 where it can no longer connect to the MySQL server using the new extension and old connector: https://issues.apache.org/jira/projects/GUACAMOLE/issues/GUACAMOLE-1738

PR #810, opened 2023-03-15
GUACAMOLE-1738: Update version of MySQL connector

mike-jumper:
This is an exact copy of #802, cherry-picked onto `staging/1.5.1`.

Testing a merge of this to `staging/1.5.1`, and then `staging/1.5.1` back to `master`, git seems happy and is able to handle the merges without conflicts.


necouchman:
@mike-jumper So, do I need to go through the same process of doing the "Merging 1.5.1 changes back to master"? It looks like it works fine, we just end up with duplicate commits...

necouchman:
@mike-jumper Nevermind, re-read your original comment and went ahead and pushed to master. We do end up with two sets of GUACAMOLE-1738 commits in the master branch, but the code is fine.

In the future, especially right after a release, we'll just have to make sure to get the bugfix branch created very early so we can push to that.

PR #811, opened 2023-03-21
GUACAMOLE-1754: Run docker build for PR CI.

jmuehlner:


jmuehlner:
Not sure why this job isn't running at all. I'd expect that it would not run automatically and would require [manual approval](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks), but it looks like it's not detecting the job at all, which is strange.

jmuehlner:
> I'm also not sure why the job isn't running, but then again it makes some sense that GitHub wouldn't run workflows that haven't been approved and merged (otherwise any contributor could run anything).
> 
> Looks like there are additional security concerns for GitHub Actions that we would need to mitigate if we intend to use GitHub Actions to perform PR verification builds:
> 
> https://cwiki.apache.org/confluence/display/BUILDS/GitHub+Actions+status

Do you see any documentation anywhere specifying that the actions have to be merged before being runnable? That would be kind of a pain for testing... I can see that PRs on forks would require manual approval [here](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks), which makes a lot of sense to me.

jmuehlner:
I do see [workflow failures](https://github.com/jmuehlner/guacamole-client/actions) when I removed required parts of the YAML, so it is getting processed...

lorengordon:
This has happened to be before also. If you push the branch to apache/guacamole-client, and open the PR from there, it should trigger the actions. I've only had to do this the very first workflow added to a project...

jmuehlner:
> This has happened to me before also. If you push the branch to apache/guacamole-client, and open the PR from there, it should trigger the actions. I've only had to do this the very first workflow added to a project...

Thanks for the info - that's very helpful!

jmuehlner:
@mike-jumper @necouchman Any objections to me moving this PR to the `apache/guacamole-client` repo as detailed above to get this workflow running?

necouchman:
Good with me, particularly if that's what Infra is recommending.

jmuehlner:
Looks like it (getting a CI job by running on the main repo) worked: https://github.com/apache/guacamole-client/pull/812.

jmuehlner:
Ok, looks like the job successfully passed. However, I didn't see any manual approval required for the job, which I think we should have - my idea is that a maintainer can manually approve the workflow for any PR that's fully reviewed and ready to be merged.

jmuehlner:
Ok, looks like it is indeed only the first time that the workflow needs to be approved.

I had to approve [this workflow run](https://github.com/apache/guacamole-client/pull/813), but not [this one](https://github.com/apache/guacamole-client/pull/814).

I think this is probably ok, though I'm glad to hear other's opinions on this. If this looks good, I can make a similar change for [guacamole-server](https://github.com/apache/guacamole-server).

jmuehlner:
> > Ok, looks like it is indeed only the first time that the workflow needs to be approved.
> 
> So long as the build itself can somehow be guaranteed to not be able to impact the repository, the runner it runs on, etc., then this is probably OK as long as we can always manually stop the build.
> 
> If there is any possibility that the build could be used maliciously, then implicit approval for follow-up builds would be deal-breaking IMHO. It would be far too easy for someone to open a PR, await approval of that PR's build, and then push new and malicious changes.

So it looks like github itself [requires workflow approvals for new users (and does some sort of github-initiated banning of malicious contributors)](https://github.blog/2021-04-22-github-actions-update-helping-maintainers-combat-bad-actors/).

The rest would be up to the configuration of the runner. It won't be able to make changes to the repo unless it has the appropriate credentials as [exemplified here](https://stackoverflow.com/questions/57921401/push-to-origin-from-github-action/58393457#58393457).

The same goes for doing bad things to the runner - a malicious user can run whatever sort of nasty things that the Apache runner is configured to let them do. 

**TLDR** There's probably no way to guarantee that a user can't mine cryptocurrency or such, but they shouldn't be able to destroy the runner or push bad changes to the repo itself.

lorengordon:
You can also modify the repository settings to [always require approval for all outside contributors](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#controlling-changes-from-forks-to-workflows-in-public-repositories) to run the actions. That gives you a chance to review the changes first.

jmuehlner:
@lorengordon thanks, that's exactly what I'm looking for! I don't seem to have permission to see that settings page, however. So I guess it's no surprise I couldn't find it. @mike-jumper do you have permission to edit these workflow approval settings by any chance?

jmuehlner:
Well, according to Daniel Gruno in https://issues.apache.org/jira/browse/INFRA-24388, the settings are already configured as we'd want, though it doesn't seem like it works that way in practice.

jmuehlner:
Looks like approvals _are_ working as expected now. See
https://issues.apache.org/jira/browse/INFRA-24388
https://github.com/apache/guacamole-client/pull/820

mike-jumper (approved):

mike-jumper:
:shipit:

PR #814, opened 2023-03-22
GUACAMOLE-1754: Run docker build for PR CI. part 2 the return of khan

aleitner:


jmuehlner:
Ah ok, looks like it only required approval the very first time...

PR #815, opened 2023-03-22
GUACAMOLE-1756: Display login failures regardless of whether the interactive login form was used.

mike-jumper:
This change adds an additional overall application state for the case that an error occurred during authentication without the use of a login form, and provides an alternative location for that error to be rendered when no login form is present.

For example, if authentication is attempted via some sort of SSO, but is repeatedly failing and is now being throttled by "guacamole-auth-ban", the following will appear (instead of a mysterious blank page):

A modal error dialog presenting a login failure message above a "Re-login" button. The error message reads: "Too many failed authentication attempts. Please try again later."

### DIFF
--- a/guacamole/src/main/frontend/src/app/index/controllers/indexController.js
+++ b/guacamole/src/main/frontend/src/app/index/controllers/indexController.js
@@ -35,6 +35,7 @@ angular.module('index').controller('indexController', ['$scope', '$injector',
     const SESSION_VALIDITY_RECHECK_INTERVAL = 15000;
 
     // Required types
+    const Error              = $injector.get('Error');
     const ManagedClientState = $injector.get('ManagedClientState');
 
     // Required services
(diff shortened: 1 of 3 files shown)


guacamole/src/main/frontend/src/app/index/controllers/indexController.js
```diff
@@ -323,6 +329,25 @@ angular.module('index').controller('indexController', ['$scope', '$injector',
...
+
+        // All errors related to an interactive login form are handled elsewhere
+        if ($scope.applicationState === ApplicationState.AWAITING_CREDENTIALS
+                || error.type === Error.Type.INSUFFICIENT_CREDENTIALS
+                || error.type === Error.Type.INVALID_CREDENTIALS)
+            return;
+
+        $scope.applicationState = ApplicationState.AUTOMATIC_LOGIN_REJECTED;
+        $scope.page.title = 'APP.NAME';
+        $scope.page.bodyClassName = '';
```
jmuehlner, line 344: I see these 2 lines duplicated in 4 different places in this controller. Maybe time to create a common function or something? It would be more DRY and make it a bit easier to know what these do / are for.

mike-jumper: Sure - I can make this better.

jmuehlner (approved):

PR #821, opened 2023-04-03
GUACAMOLE-1253: Update select all queries to process in batches.

aleitner:
SQL Server limits all queries to a maximum of 2100 parameters. This can cause queries to fail within Guacamole if the number of objects being queried would exceed that limit.

This pull request updates the SQL-Server, MySQL, and PostgreSQL jdbc code to perform batch operations when attempting to query large collections of objects

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/JDBCEnvironment.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/JDBCEnvironment.java
@@ -68,6 +68,19 @@ public abstract class JDBCEnvironment extends DelegatingEnvironment {
     public abstract int getAbsoluteMaxConnections() throws GuacamoleException;
 
     /**
+     * Returns the maximum number of identifiers/parameters to be 
+     * included in a single batch when executing SQL statements.
+     *
+     * @return
+     *     The maximum number of identifiers/parameters to be included 
+     *     in a single batch.
+     *
+     * @throws GuacamoleException
+     *     If an error occurs while retrieving the property.
+     */
+    public abstract int getBatchSize() throws GuacamoleException;
+
+    /**
      * Returns the default maximum number of concurrent connections to allow to
      * any one connection, unless specified differently on an individual
      * connection. Zero denotes unlimited.
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
@@ -19,13 +19,18 @@
 
 package org.apache.guacamole.auth.jdbc.base;
 
+import com.google.common.collect.Lists;
+import com.google.inject.Inject;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 import org.apache.guacamole.auth.jdbc.user.ModeledAuthenticatedUser;
 import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.GuacamoleSecurityException;
+import org.apache.guacamole.auth.jdbc.JDBCEnvironment;
 import org.apache.guacamole.auth.jdbc.permission.ObjectPermissionMapper;
 import org.apache.guacamole.auth.jdbc.permission.ObjectPermissionModel;
 import org.apache.guacamole.auth.jdbc.user.UserModel;
@@ -65,6 +70,12 @@ public abstract class ModeledDirectoryObjectService<InternalType extends Modeled
         ObjectPermission.Type.DELETE,
         ObjectPermission.Type.ADMINISTER
     };
+    
+    /**
+     * The environment of the Guacamole server.
+     */
+    @Inject
+    private JDBCEnvironment environment;
     
     /**
      * Returns an instance of a mapper for the type of object used by this
@@ -347,10 +358,10 @@ public abstract class ModeledDirectoryObjectService<InternalType extends Modeled
      *     A new collection containing only the strings within the provided
      *     collection which are valid identifiers.
      */
-    protected Collection<String> filterIdentifiers(Collection<String> identifiers) {
+    protected List<String> filterIdentifiers(Collection<String> identifiers) {
 
         // Obtain enough space for a full copy of the given identifiers
-        Collection<String> validIdentifiers = new ArrayList<String>(identifiers.size());
+        List<String> validIdentifiers = new ArrayList<>(identifiers.size());
 
         // Add only valid identifiers to the copy
         for (String identifier : identifiers) {
@@ -387,26 +398,36 @@ public abstract class ModeledDirectoryObjectService<InternalType extends Modeled
             Collection<String> identifiers) throws GuacamoleException {
 
         // Ignore invalid identifiers
-        identifiers = filterIdentifiers(identifiers);
+        List<String> filteredIdentifiers = filterIdentifiers(identifiers);
 
         // Do not query if no identifiers given
-        if (identifiers.isEmpty())
+        if (filteredIdentifiers.isEmpty())
             return Collections.<InternalType>emptyList();
 
-        Collection<ModelType> objects;
+        int batchSize = environment.getBatchSize();
 
-        // Bypass permission checks if the user is privileged
-        if (user.isPrivileged())
-            objects = getObjectMapper().select(identifiers);
+        boolean userIsPrivileged = user.isPrivileged();
 
-        // Otherwise only return explicitly readable identifiers
-        else
-            objects = getObjectMapper().selectReadable(user.getUser().getModel(),
-                    identifiers, user.getEffectiveUserGroups());
-        
+        // Process the filteredIdentifiers in batches using Lists.partition() and flatMap
+        Collection<ModelType> allObjects = Lists.partition(filteredIdentifiers, batchSize).stream()
+                .flatMap(chunk -> {
+                    Collection<ModelType> objects;
+
+                    // Bypass permission checks if the user is privileged
+                    if (userIsPrivileged)
+                        objects = getObjectMapper().select(chunk);
+
+                    // Otherwise only return explicitly readable identifiers
+                    else
+                        objects = getObjectMapper().selectReadable(user.getUser().getModel(),
+                                chunk, user.getEffectiveUserGroups());
+
+                    return objects.stream();
+                })
+                .collect(Collectors.toList());
+
         // Return collection of requested objects
-        return getObjectInstances(user, objects);
-        
+        return getObjectInstances(user, allObjects);
     }
 
     /**
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/conf/MySQLEnvironment.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/conf/MySQLEnvironment.java
@@ -104,6 +104,20 @@ public class MySQLEnvironment extends JDBCEnvironment {
      * The default SSL mode for connecting to MySQL servers.
      */
     private final MySQLSSLMode DEFAULT_SSL_MODE = MySQLSSLMode.PREFERRED;
+    
+    /**
+     * The default maximum number of identifiers/parameters to be included in a 
+     * single batch when executing SQL statements for MySQL and MariaDB.
+     * 
+     * MySQL and MariaDB impose a limit on the maximum size of a query, 
+     * determined by the max_allowed_packet configuration variable. A value of 
+     * 1000 is chosen to accommodate the max_allowed_packet limit without 
+     * exceeding it.
+     *
+     * @see https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_max_allowed_packet
+     * @see https://mariadb.com/kb/en/server-system-variables/#max_allowed_packet
+     */
+    private static final int DEFAULT_BATCH_SIZE = 1000;
 
     /**
      * Constructs a new MySQLEnvironment, providing access to MySQL-specific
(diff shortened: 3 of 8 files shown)


mike-jumper:
>  This can cause queries to fail within **KCM** ...

Guacamole*

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
```diff
@@ -387,25 +396,36 @@ public Collection<InternalType> retrieveObjects(ModeledAuthenticatedUser user,
...
 
-        Collection<ModelType> objects;
+        int batchSize = environment.getDefaultBatchSize();
 
-        // Bypass permission checks if the user is privileged
-        if (user.isPrivileged())
-            objects = getObjectMapper().select(identifiers);
+        Collection<ModelType> allObjects = new ArrayList<>(filteredIdentifiers.size());
+
+        for (int startIndex = 0; startIndex < filteredIdentifiers.size(); startIndex += batchSize) {
```
jmuehlner, line 409: A comment before this loop would be very handy for reader's to understand what's going on here. There's a lot of dense code that's understandable with some thinking, but at a glance it's not super obvious (at least not to me).

jmuehlner, line 409: A comment before this loop would be very handy for readers to understand what's going on here. There's a lot of dense code that's understandable with some thinking, but at a glance it's not super obvious (at least not to me).

jmuehlner: That said, there's also probably easier ways to do this (splitting a list into chunks). Have a look at https://guava.dev/releases/19.0/api/docs/com/google/common/collect/Lists.html#partition(java.util.List,%20int) for example (the JDBC extensions already include guava as a dependency).

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/UserDirectory.java
```diff
@@ -21,9 +21,11 @@
 
 
 import com.google.inject.Inject;
+
```
jmuehlner, line 24: ?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
aleitner: I switched to a more functional map reduce style of code

jmuehlner: I do dig functional-style programming, but this still looks a little overly complex. Did you look into the function I mentioned above? All this IntStream + filter + map stuff looks like it could be replaced by

```
        Collection<ModelType> allObjects = Lists.partition(filteredIdentifiers, batchSize).stream()
                .flatMap(chunk -> {
                     ...
```

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/JDBCEnvironment.java
```diff
@@ -67,6 +67,19 @@ public JDBCEnvironment() {
...
+     * included in a single batch when executing SQL statements.
+     *
+     * @return
+     *     The default maximum number of identifiers/parameters to be included 
+     *     in a single batch.
+     *
+     * @throws GuacamoleException
+     *     If an error occurs while retrieving the property.
+     */
+    public abstract int getDefaultBatchSize() throws GuacamoleException;
```
mike-jumper, line 81: Is it the default batch size? Or is it just the batch size?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
```diff
@@ -387,26 +398,38 @@ public Collection<InternalType> retrieveObjects(ModeledAuthenticatedUser user,
...
+        int batchSize = environment.getDefaultBatchSize();
 
-        // Bypass permission checks if the user is privileged
-        if (user.isPrivileged())
-            objects = getObjectMapper().select(identifiers);
+        boolean userIsPrivileged = user.isPrivileged();
+
+        // Process the filteredIdentifiers in batches using IntStream.range and flatMap
+        Collection<ModelType> allObjects = IntStream.range(0, filteredIdentifiers.size())
+                .filter(index -> index % batchSize == 0)
```
mike-jumper, line 413: While I do like Java's fancy streams, this seems an inefficient way to generate the starting indices of batches. I think we should directly calculate the starting indices rather than doing a brute-force search through all indices.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/conf/MySQLEnvironment.java
```diff
@@ -104,6 +104,12 @@ public class MySQLEnvironment extends JDBCEnvironment {
      * The default SSL mode for connecting to MySQL servers.
      */
     private final MySQLSSLMode DEFAULT_SSL_MODE = MySQLSSLMode.PREFERRED;
+    
+     /**
+     * The default maximum number of identifiers/parameters to be included in a 
+     * single batch when executing SQL statements.
+     */
+    private static final int DEFAULT_BATCH_SIZE = 2147483647;
```
mike-jumper, line 112: Is there documentation that might be referenced here that shows `2147483647` is the relevant limit for both MySQL and MariaDB?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/conf/PostgreSQLEnvironment.java
```diff
@@ -115,6 +115,12 @@ public class PostgreSQLEnvironment extends JDBCEnvironment {
      * The default value to use for SSL mode if none is explicitly configured.
      */
     private final PostgreSQLSSLMode DEFAULT_SSL_MODE = PostgreSQLSSLMode.PREFER;
+    
+    /*
+     * The default maximum number of identifiers/parameters to be included in a 
+     * single batch when executing SQL statements.
+     */
+    private static final int DEFAULT_BATCH_SIZE = 2147483647;
```
mike-jumper, line 123: Is there documentation that might be referenced here that shows `2147483647` is the relevant limit for PostgreSQL?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/conf/PostgreSQLGuacamoleProperties.java
```diff
@@ -302,5 +302,17 @@ private PostgreSQLGuacamoleProperties() {}
...
+     * The maximum number of identifiers/parameters to be included in a single batch when
+     * executing SQL statements.
+     */
+    public static final IntegerGuacamoleProperty POSTGRES_BATCH_SIZE =
+            new IntegerGuacamoleProperty() {
+
+        @Override
+        public String getName() { return "postgres-batch-size"; }
+
+    };
```
mike-jumper, line 316: Please maintain established naming convention (`postgresql-*` for the property, `POSTGRESQL_*` for the constant).

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/conf/SQLServerEnvironment.java
```diff
@@ -88,6 +88,12 @@ public class SQLServerEnvironment extends JDBCEnvironment {
      */
     public static final SQLServerDriver SQLSERVER_DEFAULT_DRIVER = SQLServerDriver.MICROSOFT_2005;
 
+    /**
+     * The default maximum number of identifiers/parameters to be included in a 
+     * single batch when executing SQL statements.
+     */
+    private static final int DEFAULT_BATCH_SIZE = 1000;
```
mike-jumper, line 95: Is there documentation that might be referenced here that shows `1000` is the relevant limit for SQL Server?

mike-jumper (changes requested):

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
jmuehlner: @mike-jumper see also the conversation regarding the use of https://guava.dev/releases/19.0/api/docs/com/google/common/collect/Lists.html#partition(java.util.List,%20int).

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/JDBCEnvironment.java
aleitner: Yeah it's just the batch size

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
aleitner: Updated to use `Lists.partition`

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/conf/PostgreSQLGuacamoleProperties.java
aleitner: Updated to be more consistent with naming conventions

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/conf/MySQLEnvironment.java
aleitner: It looks like MySQL and MariaDB have a limitation on the maximum size of the SQL query, which is determined by the 1max_allowed_packet1 system variable. This variable defines the maximum packet size that the server can send or receive. I could change the code to using this value with some calculation to determine a batch size

aleitner: The value 2147483647 represents the maximum signed integer value in MySQL and MariaDB, which can be used as the maximum allowed number of parameters in a single SQL query. However, the actual limit might depend on the specific version and configuration of the database system. Instead of referring to the maximum signed integer value, it could be more appropriate to reference the practical limits of MySQL and MariaDB in this context.

aleitner:
@mike-jumper 

PostgreSQL:
Suggested batch size: 5000

Rationale: While PostgreSQL does not have a specific limit on the number of parameters per query, the maximum number of parameters per prepared statement is 65535. Choosing a smaller value than the maximum limit can help avoid potential performance issues or query execution errors.

Relevant documentation: https://www.postgresql.org/docs/current/runtime-config-resource.html#GUC-MAX-PREPARED-STATEMENT-ARGS

MySQL and MariaDB:
Suggested batch size: 1000

Rationale: MySQL and MariaDB impose a limit on the maximum size of a query, determined by the max_allowed_packet configuration variable. This value is typically much larger than the actual number of parameters used in a single query. A smaller batch size like 1000 is often used in practice to accommodate the max_allowed_packet limit without exceeding it.

Relevant documentation:

MySQL: https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_max_allowed_packet
MariaDB: https://mariadb.com/kb/en/server-system-variables/#max_allowed_packet
SQL Server:
Suggested batch size: 2000

Rationale: SQL Server supports a maximum of 2100 parameters per query, as mentioned in the error message you provided. To stay within this limit, you can set the batch size to a smaller value like 2000.

Relevant documentation: https://docs.microsoft.com/en-us/sql/sql-server/maximum-capacity-specifications-for-sql-server

Keep in mind these are all just suggestions as there isn't an actual limit for mysql and postgres

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/conf/MySQLEnvironment.java
```diff
@@ -104,6 +104,20 @@ public class MySQLEnvironment extends JDBCEnvironment {
...
+     * 
+     * MySQL and MariaDB impose a limit on the maximum size of a query, 
+     * determined by the max_allowed_packet configuration variable. A value of 
+     * 1000 is chosen to accommodate the max_allowed_packet limit without 
+     * exceeding it.
+     *
+     * @see https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_max_allowed_packet
+     * @see https://mariadb.com/kb/en/server-system-variables/#max_allowed_packet
+     */
+    private static final int BATCH_SIZE = 1000;
```
mike-jumper, line 120: As the default value for `*-batch-size`, this should be `DEFAULT_BATCH_SIZE`.

mike-jumper (changes requested):
Except that the constants representing the default batch size should be named `DEFAULT_BATCH_SIZE` rather than `BATCH_SIZE`, this looks good to me.

mike-jumper (approved):
LGTM!

mike-jumper:
Build failure :(

Will rerun after #827 is ready and merged.

mike-jumper:
The verification build is for some reason verifying a merge against commit 6ba8971fa610ba73f42ed78dfaa8b0df2a1926f7, rather than against actual current `master` (b8770589f3e0ac14c620fc7a287c33ee16e4658b)...

mike-jumper:
@aleitner Can you rebase against current `master`? The PR verification build is otherwise failing from the issue that #827 corrected.

PR #825, opened 2023-04-06
TEST: Use maven-frontend-plugin instead of jasmine-frontend-plugin to fix build?

jmuehlner:
Just testing to see if this approach will work in CI.


jmuehlner:
closed in favor of https://github.com/apache/guacamole-client/pull/826

PR #830, opened 2023-04-11
GUACAMOLE-1020: Implement extension to enable additional restrictions

necouchman:
I've taken a run at implementing a decorating extension that allows users, groups, connections, and connection groups, to be restricted beyond the defaults provided by the base Guacamole implementation:
* Restrict the times at which users can log in based on a "Day of the Week" schedule. This is implemented using both an "Allow at certain times" field, which, if present, will restrict the user to logins only during those times, and a "Deny at certain times" field, which, if present, will block the user from logging in during the specified times. This is implemented at both the individual user level, as well as a setting that can be applied to a group and will impact all the members of that group.
Screenshot_2023-04-11_09-33-34
Screenshot_2023-04-11_10-46-48

* Restrict the hosts from which users can log in, based on hostname, IP address, or CIDR notation. I've attempted to implement both IPv4 and IPv6 restrictions. Hostnames will be reverse-queried to resolve to IPs, and then they are checked against the user's login IP, if it's available.
Screenshot_2023-04-11_10-50-45

* Restrict the times at which connections and/or connection groups (of the Balancing variety) can be accessed, in the same "Day of the Week" schedule.
* Restrict the hosts from which connections and/or connection groups (of the Balancing variety) can be accessed, using hostname, IP address, and/or CIDR range.
Screenshot_2023-04-11_11-06-47


extensions/guacamole-auth-restrict/src/main/resources/types/timeRestrictionEntry.js
```diff
@@ -0,0 +1,69 @@
...
+        /**
+         * The numerical representation of the day of the week this restriction
+         * applies to.
+         *
+         * @type Number
+         */
+        this.weekDay = template.weekDay;
+        
+        /**
+         * The hour aand minute that this restriction starts, in 24-hour time,
```
jmuehlner, line 50: Typo: `aand`

extensions/guacamole-auth-restrict/src/main/resources/controllers/timeRestrictionFieldController.js
```diff
@@ -0,0 +1,222 @@
...
+    /**
+     * Parse the provided string into an array containing the objects that
+     * represent each of entries that can then be displayed as a more
+     * user-friendly field.
+     * 
+     * @param {String} restrString
+     *     The string that contains the restrictions, un-parsed and as stored
+     *     in the underlying field.
+     *     
+     * @returns {Array|Object}
```
jmuehlner, line 110: This doesn't look correct - this is saying that the return type could be an array _or_ an object. An array of objects would be `Object[]`.

extensions/guacamole-auth-restrict/src/main/resources/controllers/timeRestrictionFieldController.js
```diff
@@ -0,0 +1,222 @@
...
+
+        // Loop through split string and process each item
+        for (let i = 0; i < restrArray.length; i++) {
+            if (restrictionRegex.test(restrArray[i])) {
+                var currArray = restrArray[i].match(restrictionRegex);
+                var restrDays = currArray[1].split("-");
+                
+                if (restrDays.length > 1) {
+                    for (let j = restrDays[0]; j <= restrDays[1]; j++) {
+                        var entry = {};
```
jmuehlner, line 135: I'd consider defining a type for these restriction objects to make it clearer what exactly the fields are, and what the format of each field should be.

extensions/guacamole-auth-restrict/src/main/resources/controllers/hostRestrictionFieldController.js
```diff
@@ -0,0 +1,165 @@
...
+    /**
+     * The restrictions, as objects, that are used by the HTML template to
+     * present the restrictions to the user via the web interface.
+     */
+    $scope.restrictions = [];
+    
+    /**
+     * Remove the current entry from the list.
+     * 
+     * @param {type} entry
```
jmuehlner, line 62: `{type}`?

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/calendar/TimeRestrictionParser.java
```diff
@@ -0,0 +1,135 @@
...
+/**
+ * A class for parsing time-based restrictions stored in a String into other
+ * formats that can be used by Guacamole.
+ */
+public class TimeRestrictionParser {
+    
+    /**
+     * The compiled regular expression that matches one or more instances of
+     * a restriction string, which specifies at least one day and time range
+     * that the restriction applies to.
```
jmuehlner, line 39: An example of what a valid string looks like would be really handy here, since this is hard to read at a glance.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/calendar/TimeRestrictionParser.java
```diff
@@ -0,0 +1,135 @@
...
+            }
+            
+        }
+        
+        // Return the list of restrictions
+        return restrictions;
+        
+    }
+    
+    public static String toString(List<DailyRestriction> restrictions) {
```
jmuehlner, line 125: Is this used? If so it needs to be documented.

extensions/guacamole-auth-restrict/src/main/resources/controllers/timeRestrictionFieldController.js
```diff
@@ -0,0 +1,222 @@
...
+         * model.
+         *
+         * @type String
+         */
+        timezone : 'UTC'
+
+    };
+    
+    // Required types
+    // const TimeRestrictionEntry = $injector.get('TimeRestrictionEntry');
```
jmuehlner, line 54: ?

extensions/guacamole-auth-restrict/src/main/resources/controllers/timeRestrictionFieldController.js
```diff
@@ -0,0 +1,222 @@
...
+                    || !Object.hasOwn(restrictions[i], 'startTime')
+                    || restrictions[i].startTime === null 
+                    || !Object.hasOwn(restrictions[i], 'endTime') 
+                    || restrictions[i].endTime === null)
+                continue;
+            
+            // If this is not the first item, then add a semi-colon separator
+            if (restrString.length > 0)
+                restrString += ';';
+            var currString = '' + restrictions[i].weekDay;
```
jmuehlner, line 186: This is a pretty dense code block - some newlines between these `if` statements (and some comments about what they're doing) would make this a lot more readable.

extensions/guacamole-auth-restrict/src/main/resources/controllers/timeRestrictionFieldController.js
necouchman: Yeah - I actually had it that way, in fact, the code is still there. I was just running into issues with it and temporarily removed it, then forgot to go back and put it back...

extensions/guacamole-auth-restrict/src/main/resources/controllers/hostRestrictionFieldController.js
```diff
@@ -0,0 +1,165 @@
...
+     * in an underlying module.
+     * 
+     * @param {Object[]} restrictions
+     *     The array of restrictions that will be converted to a string.
+     * 
+     * @returns {String}
+     *     The string containing the restriction data that can be stored in e.g.
+     *     a database.
+     */
+    var storeRestrictions = function storeRestrictions(restrictions) {
```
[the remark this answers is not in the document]
necouchman: Makes sense.

extensions/guacamole-auth-restrict/src/main/resources/restrictModule.js
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * Module which provides handling for TOTP multi-factor authentication.
```
jmuehlner, line 21: Looks like some copypasta in this file - here and on line `27`.

extensions/guacamole-auth-restrict/src/main/resources/controllers/timeRestrictionFieldController.js
necouchman: Okay, maybe you can help me spot what I'm doing wrong, here. I have the timeRestrictionEntry.js file, which defines this TimeRestrictionEntry - however, when I try to include it via the injector and then use the `new TimeRestrictionEntry()` constructor, it causes the fields to not be processed correctly and the following errors in the console:
```
angular.min.js:129 Failed to retrieve field with type "GUAC_TIME_RESTRICTION"
angular.min.js:129 Failed to retrieve field with type "GUAC_HOST_RESTRICTION"
angular.min.js:129 Failed to retrieve field with type "GUAC_TIME_RESTRICTION"
angular.min.js:129 Failed to retrieve field with type "GUAC_HOST_RESTRICTION"
```
I'm sure I have a typo somewhere or am doing something wrong that's really simple, but I cannot figure out what it is.

extensions/guacamole-auth-restrict/src/main/resources/types/timeRestrictionEntry.js
necouchman: Fixed via rebase.

extensions/guacamole-auth-restrict/src/main/resources/controllers/timeRestrictionFieldController.js
necouchman: Fixed via rebase.

extensions/guacamole-auth-restrict/src/main/resources/controllers/hostRestrictionFieldController.js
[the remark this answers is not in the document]
necouchman: Changed to `const` via rebase.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/calendar/TimeRestrictionParser.java
necouchman: Added a list of examples.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/calendar/TimeRestrictionParser.java
necouchman: Nope, not used - removed it.

extensions/guacamole-auth-restrict/src/main/resources/controllers/timeRestrictionFieldController.js
necouchman: Spread it out and added some comments.

extensions/guacamole-auth-restrict/src/main/resources/controllers/timeRestrictionFieldController.js
necouchman: Yeah, I'm still stumped by this. If I try to add this line into the `timeRestrictionFieldController.js` file, the field does not work at all:
```
const TimeRestrictionEntry = $injector.get('TimeRestrictionEntry');
```

At present, I've just defined the `TimeRestrictionEntry` and `HostRestrictionEntry` "types" in the controller files, which works fine since they aren't actually used outside of those files. Any insight on why the `factory()` method doesn't seem to be working as expected to create these data types would be appreciated.

extensions/guacamole-auth-restrict/src/main/resources/controllers/timeRestrictionFieldController.js
necouchman: Fixed via rebase.

extensions/guacamole-auth-restrict/src/main/resources/controllers/timeRestrictionFieldController.js
jmuehlner: Well it looks like your type is called `timeRestrictionEntry` in [the factory](https://github.com/apache/guacamole-client/pull/830/files#diff-e1c9d1ad21ea5614d05826fa67243d11c4808c1c28d7b3ca166b583c2b4bac19R23) (note the case difference).

necouchman: Yep, that was it. Should be good, now.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/TranslatableInvalidHostConnectionException.java
```diff
@@ -0,0 +1,73 @@
...
+     * Create a new host-based connection exception with the given message and
+     * translation string that can be processed by Guacamole's translation
+     * service.
+     * 
+     * @param message
+     *     The non-translatable, human-readable message containing details
+     *     of the exception.
+     * 
+     * @param translatableMessage 
+     *     The translation key for this exception that can be processed through
```
jmuehlner, line 47: This isn't necessarily just a key - consider rewording this param description.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/RestrictionVerificationService.java
```diff
@@ -0,0 +1,337 @@
...
+            // Pull host-based restrictions for this group and verify
+            String grpAllowedHostString = grpAttributes.get(RestrictUserGroup.RESTRICT_HOSTS_ALLOWED_ATTRIBUTE_NAME);
+            String grpDeniedHostString = grpAttributes.get(RestrictUserGroup.RESTRICT_HOSTS_DENIED_ATTRIBUTE_NAME);
+            if ((grpAllowedHostString != null && !grpAllowedHostString.isEmpty()) || (grpDeniedHostString != null && !grpDeniedHostString.isEmpty()))
+                if (!allowedByHostRestrictions(grpAllowedHostString, grpDeniedHostString, remoteAddress))
+                    throw new TranslatableInvalidHostLoginException("User \"" 
+                            + authenticatedUser.getIdentifier() 
+                            + "\" is not allowed to log in from this host due to restrictions on group \"" 
+                            + userGroup + "\".",
+                        "RESTRICT.ERROR_USER_LOGIN_NOT_ALLOWED_FROM_HOST");
```
jmuehlner, line 289: This looks to be mis-indented.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/RestrictionVerificationService.java
```diff
@@ -0,0 +1,337 @@
...
+        
+        if (context.self().getEffectivePermissions().getSystemPermissions().hasPermission(SystemPermission.Type.ADMINISTER)) {
+            LOGGER.warn("User \"{}\" has System Administration permissions; additional restrictions will be bypassed.", authenticatedUser.getIdentifier());
+            return;
+        }
+            
+        // Verify time-based restrictions specific to the user
+        String allowedTimeString = userAttributes.get(RestrictUser.RESTRICT_TIME_ALLOWED_ATTRIBUTE_NAME);
+        String deniedTimeString = userAttributes.get(RestrictUser.RESTRICT_TIME_DENIED_ATTRIBUTE_NAME);
+        if ((allowedTimeString != null && !allowedTimeString.isEmpty() || (deniedTimeString != null && !deniedTimeString.isEmpty())))
```
jmuehlner, line 237: It doesn't look like all these null and empty checks are needed here - `allowedByTimeRestrictions()` does all the same checks. The same goes for the host checks (also in `verifyConnectionRestrictions()`).

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/host/HostRestrictionParser.java
```diff
@@ -0,0 +1,77 @@
...
+        // First split the string by semicolons and process each entry
+        for (String host : hostString.split(";")) {
+            
+            HostName hostName = new HostName(host);
+            try {
+                hostName.validate();
+                addressList.add(hostName);
+            }
+            catch (HostNameException e) {
+                LOGGER.error("Invalid host name or IP: {}", host);
```
jmuehlner, line 67: Not sure this should be `error` level - this could just be bad user-supplied data, no? `warning` seems more appropriate.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/RestrictionVerificationService.java
```diff
@@ -0,0 +1,337 @@
...
+        
+        // If attributes do not exist or are empty then the action is allowed.
+        if ((allowedHostsString == null || allowedHostsString.isEmpty()) 
+                && (deniedHostsString == null || deniedHostsString.isEmpty()))
+            return true;
+        
+        // If the remote address cannot be determined, and restrictions are
+        // in effect, log an error and deny the action.
+        if (remoteAddress == null || remoteAddress.isEmpty()) {
+            LOGGER.error("Host-based restrictions are present, but the remote address is invalid or could not be resolved.");
```
jmuehlner, line 148: `error` seems like a bit overkill for a case of something that we expected might be not present not being present. Seems like this should be `warning` at worst, or maybe even `info` if not knowing the remote address is really expected.

jmuehlner:
Had you considered allowing the weekday or time parts of the restrictions to be left out, rather than requiring both? It looks like most of the code would already be pretty close to supporting that.

I think it could be pretty handy - for example:

An admin might prefer to add rule a that a user can access a connection from 9:00 to 17:00 every day, and also add a couple of rules that the user cannot access the connection on Saturday or Sunday.

As opposed to right now it looks like they'd have to add 5 rules, one for each day of the week, and if they wanted to change the hours, they'd have to change all 5 of the rules.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/connectiongroup/RestrictConnectionGroup.java
```diff
@@ -0,0 +1,183 @@
...
+    @Override
+    public Map<String, String> getAttributes() {
+
+        // Create independent, mutable copy of attributes
+        Map<String, String> attributes = new HashMap<>(super.getAttributes());
+        
+        for (String attribute : RESTRICT_CONNECTIONGROUP_ATTRIBUTES) {
+            String value = attributes.get(attribute);
+            if (value == null || value.isEmpty())
+                attributes.put(attribute,  null);
```
jmuehlner, line 148: double space here after the comma

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/connectiongroup/RestrictConnectionGroup.java
```diff
@@ -0,0 +1,183 @@
...
+        return getDelegateConnectionGroup();
+    }
+    
+    @Override
+    public Map<String, String> getAttributes() {
+
+        // Create independent, mutable copy of attributes
+        Map<String, String> attributes = new HashMap<>(super.getAttributes());
+        
+        for (String attribute : RESTRICT_CONNECTIONGROUP_ATTRIBUTES) {
```
jmuehlner, line 145: A quick comment explaining what this loop is for / is doing would be handy - same for the very similar loop in `setAttributes()`.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/TranslatableInvalidHostLoginException.java
```diff
@@ -0,0 +1,72 @@
...
+ * An exception that represents an invalid login or connection due to
+ * restrictions based on the host from which the action should be allowed.
+ */
+public class TranslatableInvalidHostLoginException 
+        extends TranslatableGuacamoleClientException {
+
+    /**
+     * The serial version ID of this class.
+     */
+    private static final long serialVersionUID = 1L;
```
jmuehlner, line 35: Good call on adding these - I had forgotten that `Exception` is `Serializable`, so this seems like the right thing to do. That said, do we actually serialize any `GuacamoleException`s? We have a lot of subclasses floating around without this field being defined...

necouchman: I'm not sure - NetBeans recommended it, so I added it. I'm happy to go either way, although it's unclear to me what the value is if the superclasses don't also have it. I'm also happy to go back and add it to others, if we should do that.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/RestrictionVerificationService.java
necouchman: I'm happy to make this `warn` or `info` - my rationale for going with `error` is that, under this condition, the login will actually be denied, which seems to make it an error condition.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/host/HostRestrictionParser.java
necouchman: Yeah, this makes sense that it should be a warning.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/RestrictionVerificationService.java
necouchman: Yeah, I guess I should probably pick one place or the other to do the checks. I was trying to avoid calling the `verify` methods at all if they weren't present; however, I need to do the checks within those methods due to the logic of either rejecting or accepting the connection, so I should probably just skip the checks, here.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/RestrictionVerificationService.java
necouchman: Ugh, yes. The new version of NetBeans has a "feature" enabled that puts extra stuff at the beginning of the parameters, and it's messing up my ability to indent correctly...

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/TranslatableInvalidHostConnectionException.java
necouchman: Fixed via rebase.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/connectiongroup/RestrictConnectionGroup.java
necouchman: Fixed via rebase.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/connectiongroup/RestrictConnectionGroup.java
necouchman: Added via rebase.

necouchman:
> Had you considered allowing either the weekday or time parts of the restrictions to be left out, rather than requiring both? It looks like most of the code would already be pretty close to supporting that.

I had thought about it, but not quite so thorougly.

> An admin might prefer to add rule a that a user can access a connection from 9:00 to 17:00 every day, and also add a couple of rules that the user cannot access the connection on Saturday or Sunday.

This should be pretty easy to do - I could add a RegEx/parsing rule that looks for an `*` to create a rule for every day. I could also look for values like "WD" for Week Day and "WE" for Week End.

> As opposed to right now it looks like they'd have to add 5 rules, one for each day of the week, and if they wanted to change the hours, they'd have to change all 5 of the rules.

Yep, that could be quite cumbersome.

> If this is hard to implement, I'm fine with leaving it as a future enhancement,

Nah, I'll take a run at it, I think it should be pretty easy. Thanks for the suggestion!

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/host/HostRestrictionParser.java
necouchman: Fixed via rebase.

necouchman:
@jmuehlner I've taken a run at implementing what I think you were getting at with the multi-day options.
Screenshot_2023-04-13_13-20-14

I also tweaked it so that 1) date is always stored in UTC in the database, and 2) the front-end form sticks with the user's timezone for the field itself, then translates to UTC when storing in the backend.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/RestrictionVerificationService.java
necouchman: Removed extra checks.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/RestrictionVerificationService.java
necouchman: Okay, I changed this to `warn`, after reviewing some of the other Guacamole Client code and seeing what was used in other cases.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/user/RestrictUserContext.java
```diff
@@ -0,0 +1,165 @@
...
+/**
+ * A UserContext implementation for additional login and connection restrictions
+ * which wraps the UserContext of some other extension.
+ */
+public class RestrictUserContext extends DelegatingUserContext {
+
+    /**
+     * The remote address from which this user logged in.
+     */
+    String remoteAddress;
```
jmuehlner, line 48: Hmm, why is this not `private`? Or `final`? I don't see it updated or accessed anywhere outside this class.

necouchman: Probably just couldn't make up my mind how it was going to be updated (constructor or `set` method) and forgot to come back and fix it. Fixed via rebase.

extensions/guacamole-auth-restrict/src/main/resources/controllers/hostRestrictionFieldController.js
```diff
@@ -0,0 +1,168 @@
...
+         *
+         * @type String
+         */
+        timezone : 'UTC'
+
+    };
+    
+    /**
+     * The restrictions, as objects, that are used by the HTML template to
+     * present the restrictions to the user via the web interface.
```
jmuehlner, line 58: This array should be annotated with the `@type` of item that it holds.

jmuehlner:
This is all looking pretty reasonable to me - @mike-jumper thoughts?

extensions/guacamole-auth-restrict/src/main/resources/controllers/hostRestrictionFieldController.js
necouchman: Fixed via rebase.

necouchman:
@mike-jumper Ping...

mike-jumper:
:wave: Sorry about that. I'll take a look over this later today.

extensions/guacamole-auth-restrict/src/main/resources/guac-manifest.json
```diff
@@ -0,0 +1,29 @@
+{
+
+    "guacamoleVersion" : "1.5.0",
```
mike-jumper, line 3: This should match the version of guacamole-ext that the extension was built against.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/calendar/TimeRestrictionParser.java
```diff
@@ -0,0 +1,180 @@
...
+            DayOfWeek.FRIDAY
+    );
+    
+    /**
+     * A list of DayOfWeek items that make up weekends.
+     */
+    private static final List<DayOfWeek> RESTRICTION_WEEKEND = Arrays.asList(
+            DayOfWeek.SATURDAY,
+            DayOfWeek.SUNDAY
+    );
```
mike-jumper, line 87: Though this particular definition is very common, not all locales use a M-F workweek and Sat/Sun weekend:

https://en.wikipedia.org/wiki/Workweek_and_weekend#Around_the_world

Any thoughts on how this might work where the definition of a weekday or weekend is locale-dependent?

extensions/guacamole-auth-restrict/pom.xml
```diff
@@ -0,0 +1,191 @@
...
+        </plugins>
+    </build>
+
+    <dependencies>
+
+       <!-- Guacamole Extension API -->
+        <dependency>
+            <groupId>org.apache.guacamole</groupId>
+            <artifactId>guacamole-ext</artifactId>
+            <version>1.5.1</version>
```
mike-jumper, line 130: We should build against the current version (1.5.2). Same for the 1.5.1 references above.

extensions/guacamole-auth-restrict/src/main/resources/templates/hostRestrictionField.html
```diff
@@ -0,0 +1,23 @@
+<div class="hostField">
+    <table class="restrictionList" ng-show="restrictions !== null && restrictions.length > 0">
+        <tr>
+            <th>Host</th>
+            <th><img src="images/x-red.svg" class="restrictionListHeader" alt="X"></th>
```
mike-jumper, line 5: What is the purpose of the red X's in the table headers?

mike-jumper (changes requested):
Initial partial review - I'm about 50% done reading through the full set of changes.

extensions/guacamole-auth-restrict/src/main/resources/guac-manifest.json
necouchman: Ah, yeah, need to go through and update everything to 1.5.2...

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/calendar/TimeRestrictionParser.java
necouchman: Ah, good point, I had not considered that. I guess I see the following options:
* See if there's an existing Java class that contains the various locales and Weekday/Weekend mappings.
* Create a custom class and/or enum that contains this data, and try to somehow map that to the current system locale and/or user locale.
* Create a guacamole.properties item to configure weekend and weekday days.
* Rework the UX a little bit, removing these specific Weekday/Weekend entries, but then creating a multi-select interface of some sort where multiple days could be selected.

Or, something else, if anyone has any other ideas?

extensions/guacamole-auth-restrict/pom.xml
necouchman: Yep, will rework for 1.5.2.

extensions/guacamole-auth-restrict/src/main/resources/templates/hostRestrictionField.html
necouchman: Just to denote that it is a column that contains the X to remove the entry from the table. I suppose that should be obvious based on the X, there, and I can probably remove the header.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/calendar/TimeRestrictionParser.java
jmuehlner: On possibility would be to just get rid of the weekday / weekend distinction - you can always still create a "employees can only log in between 9 and 5 on M-F" sort of restriction like what I suggested in https://github.com/apache/guacamole-client/pull/830#issuecomment-1506083282.

So I guess option 3 would be what I lean towards. Allowing configurable weekdays / weekends sounds a bit overkill to me. I'm happy to be convinced otherwise though.

necouchman: > One possibility would be to just get rid of the weekday / weekend distinction - you can always still create a "employees can only log in between 9 and 5 on M-F" sort of restriction like what I suggested in [#830 (comment)](https://github.com/apache/guacamole-client/pull/830#issuecomment-1506083282).

Maybe I misunderstood what you were getting at with this in your original comment. Are you thinking of some way that people would be able to enter a custom restriction set, maybe even using the way it is stored in the database, that allows for easier entry than selecting a day and enter start and end times? Or just enter the days (M-F) for the day field instead of selecting from the drop-down?

> 
> So I guess option 3 would be what I lean towards. Allowing configurable weekdays / weekends sounds a bit overkill to me. I'm happy to be convinced otherwise though. You could even possibly just skip the multi drop down, and just add e.g. individual rules for Saturday and for Sunday.

So, you're saying, allow "weekend" and "weekday" to be defined in guacamole.properties, and not worry about a separate class where you track different weekend and weekday values for each locale (option 2)?

extensions/guacamole-auth-restrict/src/main/resources/guac-manifest.json
necouchman: Fixed via rebase and force-push.

extensions/guacamole-auth-restrict/pom.xml
necouchman: Fixed via rebase and force-push.

extensions/guacamole-auth-restrict/src/main/resources/templates/hostRestrictionField.html
necouchman: Removed via rebase.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/calendar/TimeRestrictionParser.java
jmuehlner: Well, what I was imagining was that you could both blacklist and whitelist rules, each of which can contain a day and/or time.

So for the workday-only case, you could have 3 rules:
1. Whitelist of 9AM to 5PM
2. Blacklist Saturday
3. Blacklist Sunday

That wouldn't require the code knowing about the concept of a weekend or require the field to be able to handle different days in a single rule, and it'd be pretty easy to either change the restricted days (i.e. weekends) or change the working hours.

necouchman: Ah, okay - so that should be doable given the way it's currently implemented. The "Allow" rules are evaluated first, followed by the "Deny" rules, which means the "Deny" rules would take precedence. So, you should be able to:
1. Add an Allow rule for "All Days" and select 09:00 - 17:00.
2. Add a Deny rule for Saturday, and one for Sunday, and select 00:00 to 23:59.

necouchman: Okay, pulled out the Weekend and Weekday definitions and parsing, but left the "*" for all days in there, which hopefully allows the right level of flexibility without causing confusion over weekend and weekdays across different locales.

necouchman:
@mike-jumper or @jmuehlner, anything else that needs tweaking on this one?

extensions/guacamole-auth-restrict/src/main/resources/controllers/timeRestrictionFieldController.js
```diff
@@ -0,0 +1,221 @@
...
+        // Loop through split string and process each item
+        for (let i = 0; i < restrArray.length; i++) {
+            
+            // Test if our regex matches
+            if (restrictionRegex.test(restrArray[i])) {
+                var currArray = restrArray[i].match(restrictionRegex);
+                var entry = new TimeRestrictionEntry();
+                entry.weekDay = '' + currArray[1];
+                entry.startTime = new Date(Date.UTC(1970, 1, 1, parseInt(currArray[2].slice(0,2)), parseInt(currArray[2].slice(2)), 0, 0));
+                entry.endTime = new Date(Date.UTC(1970, 1, 1, parseInt(currArray[3].slice(0,2)), parseInt(currArray[3].slice(2)), 0, 0));
```
necouchman, line 131: Apparently `Date.UTC()` does not deal with DST ("summer time") correctly:

```
var testDate1 = new Date();
undefined
testDate1.getHours();
16
testDate1.getTimezoneOffset();
240
testDate1.getUTCHours();
20
var testDate2 = new Date(Date.UTC(2023, 11, 3, 20, 00, 00, 00));
undefined
testDate2.getHours()
15
testDate2.getTimezoneOffset();
300
```

:face_with_spiral_eyes:

jmuehlner: I'm seeing consistent behavior on chrome regardless of whether I use `Date.UTC()` or not
```
new Date(Date.UTC(2023, 10, 1, 20, 00, 00, 00)).getTimezoneOffset();
420

new Date(2023, 10, 1, 20, 00, 00, 00).getTimezoneOffset();
420

new Date(Date.UTC(2023, 11, 1, 20, 00, 00, 00)).getTimezoneOffset();
480

new Date(2023, 11, 1, 20, 00, 00, 00).getTimezoneOffset();
480```

Comment (jmuehlner) [reply extensions/guacamole-auth-restrict/src/main/resources/controllers/timeRestrictionFieldController.js:131]: It's also worth keeping in mind that the `Date()` and `Date.UTC()` constructors are zero-indexed for the month argument -- so `0` means January and `11` means December.

Comment (necouchman) [reply extensions/guacamole-auth-restrict/src/main/resources/controllers/timeRestrictionFieldController.js:131]: > It's also worth keeping in mind that the Date() and Date.UTC() constructors are zero-indexed for the month argument -- so 0 means January and 11 means December.

Ah, yes, okay, that was at least part of the issue I was seeing. I've still got something going on in my code that isn't translating things quite correctly - the translation of the hour from local/browser time (the field) to the database is always off by an hour, and I've no idea why.

Comment (jmuehlner) [reply extensions/guacamole-auth-restrict/src/main/resources/controllers/timeRestrictionFieldController.js:131]: Those `1970, 1, 1` arguments will throw things off by a month, and we're right at the cusp of a DST change so it makes sense that that could do it.

Comment (necouchman) [reply extensions/guacamole-auth-restrict/src/main/resources/controllers/timeRestrictionFieldController.js:131]: Yeah, I realized that was causing issues and got rid of that, already, but I was still seeing some other oddities. Or maybe I just hadn't recompiled. Weird - seems to be working right now with changes I'll commit shortly.

Comment (necouchman) [reply extensions/guacamole-auth-restrict/src/main/resources/controllers/timeRestrictionFieldController.js]: Okay, I think I've fixed that issue...

Comment (jmuehlner) [reply extensions/guacamole-auth-restrict/src/main/resources/controllers/timeRestrictionFieldController.js]: While we're at it - why not use the current date instead of January, 1970?

There may have been changes in switchover dates for some locations for DST onset date since 1970 which may also screw up these comparisons.

Comment (necouchman) [reply extensions/guacamole-auth-restrict/src/main/resources/controllers/timeRestrictionFieldController.js]: Yep, that's pretty much what I did - just pull the current date and use the year, month, and date from that, and then set the hour and minute from the configuration.

File: extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/calendar/DailyRestriction.java [hunk 26]
```diff
@@ -0,0 +1,132 @@
...
+     *     true if the current time of day falls within this restriction,
+     *     otherwise false.
+     */
+    public boolean appliesNow() {
+        DayOfWeek currentDay = LocalDate.now().getDayOfWeek();
+        LocalTime currentTime = LocalTime.now(ZoneId.of("UTC"));
+        
+        // Check that we are in the specified time restriction
+        return (weekDays.contains(currentDay) && currentTime.isAfter(startTime) && currentTime.isBefore(endTime));
+    }
```
necouchman, line 130: Okay, my next issue is that this, or some other part of the restriction verification, doesn't actually correctly handle situations where UTC rolls into the next day (or, if you're ahead of UTC, the previous day). So, for example, if I have a restriction that I create in US Eastern Daylight Time, allowing logins from 0800 - 2200, the representation of that in the DB will be 1200 - 0200, and it won't validate correctly because it's the next day.

necouchman: So, I think there are two solutions to this:
* When storing the restriction, we should adjust the day such that the start time is always on the day of the restriction. So, for example, if you're in a UTC+8 hour timezone, and you put a restriction in that says Monday (1) from 0600-0800, the UTC is going to be 2200 on Sunday, so the day should shift to 0 (or 7). If you're in UTC-9 and you put a restriction that starts at 1700 on Monday (1), UTC is Tuesday at 0100, so the day would shift to 2. The reverse will need to be done when unpacking the restrictions to properly display them in local time.
* When checking the restriction in the `appliesNow()` method above, we'll need to check at least the case where `endTime` is before `startTime` so that we can "wrap" the day around to the next day.

necouchman: Alright, I believe I've resolved this issue...

Toumassa:
Hi, is there any chance of having this merged with the release of the version 1.6.0, as it is not in the milestone? Thank you for your work.

necouchman:
@Toumassa Yes, I would think we would target it for 1.6.0. I've updated the Jira issue accordingly.

stcbus:
I can try and come up with something if there's no time, but would it be possible to display somewhere ther active time restrictions for a connection? I imagine maybe an icon  next to the connection group and hovering over it could display the time restrictions.

jmuehlner (approved):
Looks good to me! I'll wait to see if @mike-jumper has anything further to say since he was also requested as a reviewer.

necouchman:
@mike-jumper Ping.

necouchman:
I've made a few updates to it, while waiting for reviews - I implemented a `Restrictable` interface to make some of the methods common to anything that can be restricted, and reworked some of the code around that.

extensions/guacamole-auth-restrict/pom.xml
```diff
@@ -0,0 +1,178 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0
+                        http://maven.apache.org/maven-v4_0_0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>org.apache.guacamole</groupId>
+    <artifactId>guacamole-auth-restrict</artifactId>
+    <packaging>jar</packaging>
+    <version>1.5.5</version>
```
jmuehlner, line 29: This should be 1.6.0 now, yeah?

extensions/guacamole-auth-restrict/src/main/resources/templates/timeRestrictionField.html
```diff
@@ -0,0 +1,38 @@
+<div class="timeRestrictionField">
+    <table class="restrictionList" ng-show="restrictions !== null && restrictions.length > 0">
+        <tr>
+            <th>Day</th>
```
jmuehlner, line 4: These should probably be translated.

extensions/guacamole-auth-restrict/src/main/resources/templates/hostRestrictionField.html
```diff
@@ -0,0 +1,23 @@
+<div class="hostField">
+    <table class="restrictionList" ng-show="restrictions !== null && restrictions.length > 0">
+        <tr>
+            <th>Host</th>
```
jmuehlner, line 4: These should also probably be translated.

extensions/guacamole-auth-restrict/src/main/resources/guac-manifest.json
```diff
@@ -0,0 +1,29 @@
+{
+
+    "guacamoleVersion" : "1.5.5",
```
jmuehlner, line 3: This should be 1.6.0.

extensions/guacamole-auth-restrict/src/main/resources/controllers/timeRestrictionFieldController.js
```diff
@@ -0,0 +1,302 @@
...
+            
+            // If this is not the first item, then add a semi-colon separator
+            if (restrString.length > 0)
+                restrString += ';';
+            
+            // When these fields first gets a value, the default year is 1970
+            // In order to avoid issues with Daylight Savings Time, we have to
+            // work around this.
+            if (restrictions[i].startTime instanceof Date && restrictions[i].startTime.getFullYear() === 1970) {
+                let startHour = restrictions[i].startTime.getHours();
```
jmuehlner, line 218: I'd consider moving this date fixing stuff into a shared function (for the start and end times).

necouchman:
@jmuehlner : I think I've addressed all the issues...without breaking things...but I need to test. Anything else jumps out, let me know.

necouchman:
I've put this back into draft mode for the moment - tracking down a couple of final things that aren't working as expected. Hopefully resolved tonight/tomorrow.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/user/RestrictedUserContext.java
```diff
@@ -0,0 +1,197 @@
...
+    
+    @Override
+    public boolean isValid() {
+        try {
+            // Verify whether or not time restrictions still apply.
+            RestrictionVerificationService.verifyTimeRestrictions(this, effectiveUserGroups);
+            return true;
+        }
+        catch (GuacamoleException e) {
+            LOGGER.debug("User account is now restricted and is no longer valid.");
```
jmuehlner, line 192: Maybe should include the exception here in the debug log?

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/calendar/RestrictionType.java
```diff
@@ -0,0 +1,109 @@
...
+     * 
+     * @param restriction2
+     *     The second restriction to compare.
+     * 
+     * @return 
+     *     Return which of the two restrictions is the higher-priority.
+     */
+    public static RestrictionType getHigherPriority(RestrictionType restriction1, RestrictionType restriction2) {
+        
+        // If the second is higher than the first, resturn the second.
```
jmuehlner, line 90: Typo: "resturn"

extensions/guacamole-auth-restrict/src/main/resources/controllers/timeRestrictionFieldController.js
```diff
@@ -0,0 +1,320 @@
...
+ * time restrictions that apply to user logins and connections.
+ */
+angular.module('guacRestrict').controller('timeRestrictionFieldController', ['$scope', '$injector',
+        function timeRestrictionFieldController($scope, $injector) {
+    
+    // Required types
+    const TimeRestrictionEntry = $injector.get('TimeRestrictionEntry');
+    
+    // Required services
+    const $log = $injector.get('$log');
```
jmuehlner, line 32: Unused?

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/calendar/RestrictionType.java
necouchman: Oops...corrected.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/calendar/DailyRestriction.java
```diff
@@ -0,0 +1,140 @@
...
+
+/**
+ * A class that stores a daily time restriction that can be used to determine
+ * whether or not a user can log in on a certain day of the week and during
+ * a certain time window.
+ */
+public class DailyRestriction {
+    
+    /**
+     * The day of the week that this restriction applies to.
```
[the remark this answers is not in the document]
necouchman: Yes, indeed.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/user/RestrictedUserContext.java
necouchman: Yep, good idea. Done.

necouchman:
> LGTM, but it looks like there's an outstanding change request by @mike-jumper

I'm not seeing any unresolved issues, but I'll scroll through it again and make sure, and wait for Mike's approval.

jmuehlner:
> > LGTM, but it looks like there's an outstanding change request by @mike-jumper
> 
> I'm not seeing any unresolved issues, but I'll scroll through it again and make sure, and wait for Mike's approval.

Ah, looks like it was just the comment:
> Initial partial review - I'm about 50% done reading through the full set of changes.

mike-jumper:
> > > LGTM, but it looks like there's an outstanding change request by @mike-jumper
> > 
> > 
> > I'm not seeing any unresolved issues, but I'll scroll through it again and make sure, and wait for Mike's approval.
> 
> Ah, looks like it was just the comment:
> 
> > Initial partial review - I'm about 50% done reading through the full set of changes.

Heh - nice. :)

I'll read through things now.

jmuehlner:
> > > > LGTM, but it looks like there's an outstanding change request by @mike-jumper
> > > 
> > > 
> > > I'm not seeing any unresolved issues, but I'll scroll through it again and make sure, and wait for Mike's approval.
> > 
> > 
> > Ah, looks like it was just the comment:
> > > Initial partial review - I'm about 50% done reading through the full set of changes.
> 
> Heh - nice. :)
> 
> I'll read through things now.

@mike-jumper any objections if I merge this? Did you stil have more you wanted to read through?

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/user/RestrictedUser.java
```diff
@@ -0,0 +1,184 @@
...
+
+        // Loop through extension-specific attributes, only sending ones
+        // that are non-null and non-empty to the underlying storage mechanism.
+        for (String attribute : RESTRICT_USER_ATTRIBUTES) {
+            String value = attributes.get(attribute);
+            if (value != null && value.isEmpty())
+                attributes.put(attribute, null);
+        }
+
+        super.setAttributes(attributes);
```
mike-jumper, line 166: Depending on context, the attributes controlling user restrictions will need to be removed from the `Map` before passing things along to `super.setAttributes()`, to ensure users cannot manipulate their own restrictions.

mike-jumper (changes requested):
Going to double-check the restriction logic, but everything else looks good.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/user/RestrictedUser.java
necouchman: Hmmm, I guess I thought the underlying storage mechanism (that this extension decorates) would enforce the correct permissions, but I've added the admin check in to make sure that only admins can modify user-level restrictions.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/user/RestrictedUserContext.java
```diff
@@ -0,0 +1,209 @@
...
+        return Collections.unmodifiableCollection(connectionGroupAttrs);
+    }
+    
+    @Override
+    public Directory<User> getUserDirectory() throws GuacamoleException {
+        return new DecoratingDirectory<User>(super.getUserDirectory()) {
+
+            @Override
+            protected User decorate(User object) throws GuacamoleException {
+                Permissions currentPermissions = self().getEffectivePermissions();
```
mike-jumper, line 143: To avoid potentially requerying the effective permissions of the current user for every other user in the directory, the permission sets involved here should instead be retrieved at entry to `getUserDirectory()`.

mike-jumper (changes requested):
One minor change needed to ensure we don't issue too many queries. Otherwise, LGTM!

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/user/RestrictedUserContext.java
necouchman: Yep, makes sense. I've moved it up a couple of lines...

mike-jumper: I think the calls to retrieve the permission sets (`currentPermissions.getSystemPermissions()` and `currentPermissions.getUserPermissions()`) need to be moved, as well, in case implementations defer retrievals of permissions until the permission type is known.

If possible, it would also be good to avoid calls to `hasPermission()` until specifically needed, such as when invoking `setAttributes()`, as that may also result in a query per user. I think our current JDBC implementation provides a `hasPermission()` that behaves this way, performing one `selectOne()` query per call:

https://github.com/apache/guacamole-client/blob/40620e445cca88c0687e14769b2d7375b3eac703/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/permission/ObjectPermissionSet.java#L100-L104

https://github.com/apache/guacamole-client/blob/40620e445cca88c0687e14769b2d7375b3eac703/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/permission/ModeledObjectPermissionService.java#L173-L187

Alternatively, the entire set of identifiers that are accessible by the user for operations requiring a specific set of permissions can be retrieved in bulk using `getAccessibleObjects()`.

necouchman: Ah, yeah, again, makes sense. I was trying to avoid either having to pass around and store copies of stuff in seemingly unrelated places, or having to implement a bunch of dependency injection just to make that happen, but I'll figure something out.

mike-jumper: What about pulling the `Set` of accessible identifiers when `getUserDirectory()` is invoked and passing that in instead of a `hasAdmin` flag?

necouchman: Okay, I've done a simpler implementation in the interim that just uses `getAccessibleObjects()` to pull the objects for which the current user has `ADMINISTER` privileges, and moved those calls out of the individual `decorate()` calls so that it only gets called once.

I'll dig into the possibility of using the `Supplier` class - I was having a little bit of trouble with that because it seems as though the `memoize()` method expects to get a function that takes zero arguments, or maybe at most one argument, which would mean needing to do some additional wrapping to pull the `getAccessibleObjects()` method in (would have to write some sort of inline method or something that calls `getAccessibleObjects()` with the `ADMINISTER` type and the identifier set from the user directory. I'll work on that if I have time, but, if the route I went works I'm trying to avoid this holding up the release...

mike-jumper: Sure - makes sense. I think what you describe sounds good.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/user/RestrictedUserContext.java
```diff
@@ -0,0 +1,214 @@
...
+                        Collections.singletonList(ObjectPermission.Type.ADMINISTER), super.getUserDirectory().getIdentifiers());
+        
+        return new DecoratingDirectory<User>(super.getUserDirectory()) {
+
+            @Override
+            protected User decorate(User object) throws GuacamoleException {
+                
+                // Check and see if the logged in user has admin privileges -
+                // either system-level or for that particular object.
+                boolean hasAdmin = isAdmin || adminIdentifiers.contains(object.getIdentifier());
```
mike-jumper, line 156: Nice. I like this.

mike-jumper (approved):

PR #831, opened 2023-04-12
Use  header instead of the http parameter for session/tunnels/<tunnel ID>/protocol

aresliharris:
Hello,

This change hides the authentication token from the http parameter by using the predefined header "Guacamole-Token" instead for the following API call.

`GET /api/session/tunnels/<tunnel ID>/protocol?token=<token>`

image

Related Jira: https://issues.apache.org/jira/browse/GUACAMOLE-956


aresliharris:
Closing this one as a [new PR](https://github.com/apache/guacamole-client/pull/832) was created based on the format that @jmuehlner suggested.

PR #833, opened 2023-04-13
GUACAMOLE-926: Fix loading bug and fuzzier group matching

jmuehlner:
This change does two things:
* Fix a bug where the loading spinner would be applied and never go away when file parsing succeeded, with connection errors
* Allow a bit more leeway for users in how they declare their group declarations:
   - They can (as before) specify the full path like "ROOT/parent/child"
   - They can use just a leading slash instead of the root group like "/parent/child"
   - They can leave off the leading slash, like "parent/child"
   - They can optionally include a trailing slash like "parent/child/"

### DIFF
--- a/guacamole/src/main/frontend/src/app/import/services/connectionParseService.js
+++ b/guacamole/src/main/frontend/src/app/import/services/connectionParseService.js
@@ -46,6 +46,14 @@ angular.module('import').factory('connectionParseService',
     const service = {};
 
     /**
+     * The identifier of the root connection group, under which all other groups
+     * and connections exist.
+     * 
+     * @type String
+     */
+    const ROOT_GROUP_IDENTIFIER = 'ROOT';
+
+    /**
      * Perform basic checks, common to all file types - namely that the parsed
      * data is an array, and contains at least one connection entry. Returns an
      * error if any of these basic checks fails.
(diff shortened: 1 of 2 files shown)


guacamole/src/main/frontend/src/app/import/services/connectionParseService.js
```diff
@@ -167,7 +196,7 @@ angular.module('import').factory('connectionParseService',
 
             // Set the parent identifier now that it's known
             return {
-                ...connection,
+                ..._.omit(connection, 'group'),
```
jmuehlner, line 199: This isn't strictly necessary - when passed to the `Connection` constructor later, this will automatically get removed, and will actually incur a small performance penalty from copying the connection object.

If people don't think the added cleanliness of removing the `group` field at this point is worth it, I'd be ok with removing this.

necouchman: If the performance penalty is negligible and doesn't change depending on the number of connections you import, then I'm fine leaving it. If the performance penalty scales as you increase the number of connections, then it seems like removing this is probably a good idea, particularly if it is going to get taken off elsewhere.

necouchman (approved):

PR #834, opened 2023-04-13
GUACAMOLE-1769: Handle case that KSM's getCustom() returns null.

mike-jumper:


extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmRecordService.java
```diff
@@ -229,6 +231,10 @@ private <T> String getSingleStringValue(List<T> values, Function<T, String> mapp
     private <T extends KeeperRecordField> T getField(List<KeeperRecordField> fields,
             Class<T> fieldClass, Pattern labelPattern) {
 
+        // There are no fields if no List was provided at all
+        if (fields == null)
```
jmuehlner, line 235: I'm not suggesting that you do this right now, since it's a bit of a refactor-cat-in-bathtub thing, but having `@Nullable` and `@Nonnull` annotations is pretty handy for this sort of thing. The compiler would complain if you try to pass anything that might be null in here.

mike-jumper: I'm actually surprised the compiler didn't catch this, as `getCustom()` is annotated with `@Nullable`.

jmuehlner: It would be nice if it worked like that, but as far as I can tell, it needs the explicit `@NotNull` for the compiler to do anything about it.

PR #835, opened 2023-04-13
GUACAMOLE-926: Validate parentIdentifiers before attempting connection import.

jmuehlner:
I found yet another bug 😓. I wasn't properly validating user-supplied `parentIdentifier` group IDs on imported connections, and it turns out that these are not validated on the backend either, so bad user input could result in a 500 error with no indication to the user what went wrong.

This change adds a check to the Javascript to make sure that all group IDs are validating before attempting any import.


guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -208,8 +208,7 @@
         "ERROR_INVALID_CSV_HEADER": "Invalid CSV Header \"{HEADER}\" is neither an attribute or parameter",
         "ERROR_INVALID_MIME_TYPE": "Unsupported file type: \"{TYPE}\"",
         "ERROR_INVALID_GROUP": "No group matching \"{GROUP}\" found",
-        "ERROR_INVALID_USER_GROUP_IDENTIFIERS": "User Groups not found: {IDENTIFIER_LIST}",
```
jmuehlner, line 211: These are vestigal translations from an earlier version of the code would error out rather than creating unknown users and groups. They are not used anywhere in the app.

mike-jumper (approved):

PR #836, opened 2023-04-14
Add Portuguese and Romanian keyboard layout options

vanntile:
These are the tickets GUACAMOLE-1770 and GUACAMOLE-1771


necouchman:
@vanntile : Unfortunately you'll need to do two separate pull requests on the client side, as well, tagged for each of the Jira issues. The pull request and the commits needs to be a 1-to-1 relationship with the Jira issue.

PR #838, opened 2023-04-14
GUACAMOLE-926: Fix attribute handling bug and improve error handling for invalid files.

jmuehlner:
This change does 2 things:
* Fix a bug where CSV parsing fails when the ` (attribute)` disambiguation suffix is used
* Improved error messaging if invalid files masquerading as valid files are provided (e.g. excel files with a `.csv` extension)


guacamole/src/main/frontend/src/app/import/controllers/importConnectionsController.js
```diff
@@ -640,8 +655,23 @@ angular.module('import').controller('importConnectionsController', ['$scope', '$
...
+                // Check if the file has a header of a known-bad type
+                if (_.some(ZIP_SIGNATURES,
+                        signature => fileData.startsWith(signature))) {
+
+                    // Throw an error and abort processing
+                    handleError(new ParseError({
+                        message: "Invalid file type detected",
+                        key: 'IMPORT.ERROR_DETECTED_INVALID_TYPE'
+                    }));
+                    return;
```
necouchman, line 669: I guess I'm a little conflicted about this change. On the one hand, I understand that the likelihood of someone trying to upload, for example, an XLSX or ODS file in place of a CSV may be higher than someone trying to upload, say, a MP4 video file.

On the other hand, what is the behavior if someone tries to drop a MP4 or tar file, and why is it important to handle errors with common spreadsheets differently from errors with completely unrelated files?

jmuehlner: If somebody just directly uploads a `.mp4` or `.tar` file, the MIME type check will reject it before it ever gets here, and they'll get an error specifically telling them that they've provided an invalid file type, e.g. `Unsupported file type: "video/mp4"`.

This is specifically to handle the case that somebody uploads a CSV, YAML, or JSON file that's actually an excel document, which is probably much more likely than them uploading a `.csv` file that's actually a video.

If binary garbage _is_ provided to the parsers, this is what seems to generally happen for each case:

- CSV parser throws a an error with message `Argument must be a Buffer` (so ugly we specifically catch and handle it)
- JSON parser throws an error with a message like `Unexpected token 'ë', "ëÑ§%øtÅ9¡<"... is not valid JSON`
- The YAML parser seems to actually parse the binary garbage, generally producing an object that's caught by one of our checks further on, producing this error `The provided file must contain a list of connections`

Obviously there's no way we can guarantee that a file won't have binary garbage in it without preprocessing the entire file, which has a huge performance cost. 

This check is intended to catch the likely most common case and throw a friendlier error before an attempt is made to parse the file at all, since the errors generated by the parsers themselves are less friendly.

necouchman: Ah, okay - I think my confusion was that I expected that the MIME type check would be "smarter" than just reading the file extension - I thought it would actually be looking at file magic, etc. With that assumption gone, and seeing that the browser is just looking at file extension, I see why this is required.

guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -207,17 +207,21 @@
         "ERROR_EMPTY_FILE": "The provided file is empty",
         "ERROR_INVALID_CSV_HEADER": "Invalid CSV Header \"{HEADER}\" is neither an attribute or parameter",
         "ERROR_INVALID_MIME_TYPE": "Unsupported file type: \"{TYPE}\"",
+        "ERROR_DETECTED_INVALID_TYPE": "Unsupported file type. Please make sure the file is valid CSV, JSON, or YAML.",
```
necouchman, line 210: Since you're fixing issues - maybe the style in this section could be cleaned up to match the style used throughout the rest of the translation files - mainly that colons are generally aligned with the items above and below? Unless we're working to change that style and go with something more like what's in this section across the board? I'm good either way, just going for consistency... :smile:

jmuehlner: Oh, good call! It's done.

PR #840, opened 2023-04-19
GUACAMOLE-926: Add options to replace connections / permissions.

jmuehlner:
This change allows an admin to configure the connection batch importer to replace connections, and if replacing connections, additionally to also replace the permissions associated with that connection.

I've added the `REPLACE` verb to the directory PATCH endpoint to enable the former, and the latter is accomplished by completely removing and recreating the connection - to ensure that any existing permissions are wiped out.

I've added some little clickable help bubbles to each of these options to explain them, though I'm not feeling super good about the text therein. If anyone has any suggestions for concising these explanations, I'd appreciate them.
image
image

In order to make it easier to understand what's a duplicate, and why, I've added a "group" column to the error display, showing the user-provided group (normalized if possible, or translated from the provided parentIdentifier, if valid), as well as modified the error messages to be more explicit about the exact connection update conflicts, if there are any.
image


guacamole/src/main/frontend/src/app/import/controllers/importConnectionsController.js
```diff
@@ -322,52 +318,6 @@ angular.module('import').controller('importConnectionsController', ['$scope', '$
         return $q.all({ ...userRequests, ...groupRequests });
     }
 
-    // Given a PATCH API response, create an array of patches to delete every
-    // entity created in the original request that generated this response
-    const createDeletionPatches = creationResponse =>
```
jmuehlner, line 327: I removed all this rollback functionality because:
* It would be very hard / impossible to effectively implement in some cases, like in the case of a connection being removed and then re-added to reset the permissions.
* It would be unlikely to work anyway - if something like a create user or grant permission call fails, it's most likely because of the browser lost connection to the backend, or the backend experienced some sort of error. In that case, the subsequent calls to put things back the way they were wouldn't work anyway.

guacamole/src/main/frontend/src/app/settings/templates/settingsConnections.html
```diff
@@ -10,7 +10,7 @@
         <div class="action-buttons">
 
             <a class="import-connections button"
-                ng-show="canCreateConnections()"
+                ng-show="canAdminister()"
```
jmuehlner, line 13: Changed this to only show for administrators, due to the wide variety of permissions actually required to do the import.

guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -200,38 +200,52 @@
...
+        "ERROR_EMPTY_FILE"                   : "The provided file is empty",
+        "ERROR_INVALID_CSV_HEADER"           : "Invalid CSV Header \"{HEADER}\" is neither an attribute or parameter",
+        "ERROR_INVALID_MIME_TYPE"            : "Unsupported file type: \"{TYPE}\"",
+        "ERROR_INVALID_GROUP"                : "No group matching \"{GROUP}\" found",
+        "ERROR_INVALID_GROUP_IDENTIFIER"     : "No connection group with identifier \"{IDENTIFIER}\" found",
+        "ERROR_NO_FILE_SUPPLIED"             : "Please select a file to import",
+        "ERROR_PARSE_FAILURE_CSV"            : "Please make sure your file is valid CSV. Parsing failed with error \"{ERROR}\". ",
+        "ERROR_PARSE_FAILURE_JSON"           : "Please make sure your file is valid JSON. Parsing failed with error \"{ERROR}\". ",
+        "ERROR_PARSE_FAILURE_YAML"           : "Please make sure your file is valid YAML. Parsing failed with error \"{ERROR}\". ",
+        "ERROR_REJECT_UPDATE_CONNECTION"     : "Disallowed update to existing connection at \"{PATH}\"",
```
mike-jumper, line 218: As written, this makes sense to me only in the context of an update, but:

* A connection import isn't an update unless the "Replace/Update ..." box is checked.
* This error doesn't occur if the "Replace/Update" box is checked.

Assuming the name can be included, recommend rephrasing to: `"Connection \"{NAME}\" already exists at \"{PATH}\"."`

guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -200,38 +200,52 @@
...
-        "HELP_JSON_MORE_DETAILS"            :  "The connection group ID that the connection should be imported into may be directly specified with a \"parentIdentifier\" field, or the path to the parent group may be specified using a \"group\" field as shown below. An array of user and user group identifiers to grant access to may be specified per connection.",
+        "HELP_JSON_MORE_DETAILS"            : "The connection group ID that the connection should be imported into may be directly specified with a \"parentIdentifier\" field, or the path to the parent group may be specified using a \"group\" field as shown below. An array of user and user group identifiers to grant access to may be specified per connection.",
+        "HELP_REPLACE_CONNECTION_TITLE"     : "Replacing existing connections",
+        "HELP_REPLACE_CONNECTION_CONTENT"   : "Checking this box will allow existing connections to be updated, if an imported connection has the same name and parent group as an existing connection. If unchecked, attempts to update existing connections will be treated as an error.",
```
mike-jumper, line 240: I'd recommend:

> Entirely replace/update existing connections if their names and parent connection groups match the values in the provided file. If unchecked, attempting to import a connection with the same name and parent connection group of an existing connection will be considered an error.

guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -200,38 +200,52 @@
...
+        "HELP_REPLACE_PERMISSION_TITLE"     : "Replacing connection permissions",
+        "HELP_REPLACE_PERMISSION_CONTENT"   : "If replacement of existing connections is enabled, checking this box will allow full replacement of connection permissions. If checked, access permission will only be granted to users and groups specified in the import file for this connection. If unchecked, specified users and groups will be granted access in addition to any existing permissions.",
```
mike-jumper, line 242: I'd recommend:

> Fully reset the permissions granted for all connections in the provided file to the permissions specified in that file. If no permissions are specified, all relevant connection permissions will be revoked. If unchecked, existing permissions are preserved, and any permissions specified in the file will be added.

guacamole/src/main/frontend/src/app/import/controllers/importConnectionsController.js
```diff
@@ -688,9 +637,42 @@ angular.module('import').controller('importConnectionsController', ['$scope', '$
...
+            'IMPORT.HELP_REPLACE_CONNECTION_TITLE',
+            'IMPORT.HELP_REPLACE_CONNECTION_CONTENT');
+
+    /**
+     * Display a modal with information about the existing connection permission
+     * replacement option.
+     */
+    $scope.showPermissionReplaceHelp = () => showModal(
+            'IMPORT.HELP_REPLACE_PERMISSION_TITLE',
+            'IMPORT.HELP_REPLACE_PERMISSION_CONTENT');
```
mike-jumper, line 676: What about showing this information as a tooltip visible on hover (such as with the `title` attribute), rather than a modal dialog?

guacamole/src/main/java/org/apache/guacamole/rest/directory/DirectoryResource.java
```diff
@@ -590,7 +591,56 @@ public void executeOperation(boolean atomic, Directory<InternalType> directory)
...
+
+                        // The identifier of the object to be replaced
+                        String identifier = path.substring(1);
+
+                        InternalType original = null;
+
+                        try {
+
+                            // Fetch the object to be updated
+                            original = directory.get(identifier);
```
mike-jumper, line 604: `get()` has its own success/failure event type that should be dispatched separately from `update()`. See:

https://github.com/apache/guacamole-client/blob/05c675bb215111643fc1213f661b07867a98aeb0/guacamole/src/main/java/org/apache/guacamole/rest/directory/DirectoryResource.java#L747-L757

guacamole/src/main/java/org/apache/guacamole/rest/directory/DirectoryResource.java
```diff
@@ -590,7 +591,56 @@ public void executeOperation(boolean atomic, Directory<InternalType> directory)
...
+                            
+                            // Apply the changes to the original object
+                            translator.applyExternalChanges(
+                                    original, patch.getValue());
+
+                            // Update the directory
+                            directory.update(original);
+
+                            replacedObjects.add(original);
```
mike-jumper, line 613: If the requested object does not exist, `original` will be `null` here.

guacamole/src/main/frontend/src/app/import/types/ParseResult.js
```diff
@@ -65,10 +69,19 @@ angular.module('import').factory('ParseResult', [function defineParseResult() {
...
+         * @type {Object.<String, String>}
+         */
+        this.groupPaths = template.groupPaths || {};
+
         /**
          * An array of errors encountered while parsing the corresponding
-         * connection (at the same array index). Each connection should have a
-         * an array of errors. If empty, no errors occurred for this connection.
+         * connection (at the same array index in the patches array). Each 
+         * connection should have a an array of errors. If empty, no errors
```
mike-jumper, line 83: "a an array" -> "an array"

guacamole/src/main/frontend/src/app/import/types/ConnectionImportConfig.js
```diff
@@ -0,0 +1,103 @@
...
+         * permissions.
+         */
+        ADD : "ADD",
+
+        /**
+         * Any existing permissions will be removed, ensuring that only the
+         * users or groups specified in the import file will be granted to the
+         * replaced connection after import.
+         */
+        REPLACE : "REPLACE"
```
mike-jumper, line 97: Recommend renaming `ADD` to `PRESERVE` so that the mode is consistently worded from the same perspective (an action performed upon existing permissions).

To me, `REPLACE` makes sense as an "existing permission mode" (we're replacing existing permissions), but not `ADD` (we're not adding existing permissions).

guacamole/src/main/frontend/src/app/import/types/ConnectionImportConfig.js
```diff
@@ -0,0 +1,103 @@
...
+        this.existingPermissionMode = template.existingPermissionMode
+                || ConnectionImportConfig.ExistingPermissionMode.ADD;
+
+    };
+
+    /**
+     * Valid modes for the behavior of the importer when attempts are made to
+     * update existing connections.
+     */
+    ConnectionImportConfig.ReplaceConnectionMode = {
```
mike-jumper, line 64: Similar to `ExistingPermissionMode`, I suggest renaming this to `ExistingConnectionMode`.

mike-jumper (changes requested):

guacamole/src/main/frontend/src/images/question.svg
```diff
@@ -0,0 +1,64 @@
...
+       xml:space="preserve"
+       style="font-style:normal;font-weight:normal;font-size:40px;line-height:1.25;font-family:sans-serif;fill:#000000;fill-opacity:1;stroke:none"
+       x="17.083984"
+       y="52.78125"
+       id="text4900"><tspan
+         id="tspan4898"
+         x="17.083984"
+         y="52.78125"
+         sodipodi:role="line"
+         style="font-style:normal;font-variant:normal;font-weight:normal;font-stretch:normal;font-size:56px;font-family:sans-serif;-inkscape-font-specification:'sans-serif, Normal';font-variant-ligatures:normal;font-variant-caps:normal;font-variant-numeric:normal;font-variant-east-asian:normal">?</tspan></text>
```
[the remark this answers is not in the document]
jmuehlner: Good call - will do.

guacamole/src/main/java/org/apache/guacamole/rest/directory/DirectoryResource.java
jmuehlner: Ok, I've updated it. I think this is the right logic now - if an attempt is made to replace a non-existant object, it will now log both a GET and UPDATE failure event.

jmuehlner:
Ok @mike-jumper I think I've addressed all your concerns / suggestions!

jmuehlner:
> Ok @mike-jumper I think I've addressed all your concerns / suggestions!

Oops, sorry I realized I forgot to update an error message. Let me fix that.

guacamole/src/main/java/org/apache/guacamole/rest/directory/DirectoryResource.java
mike-jumper: If a user attempts to update a single connection via the non-batch REST API, but that connection does not exist, they'll currently get a `GET` failure event but no `UPDATE` failure event (the attempt to update doesn't occur because the retrieval failed). Do you think these changes should match that behavior, or that the REST API changes that are part of GUACAMOLE-1224 should be modified to ensure both are fired?

jmuehlner: Hmm, it seems to me that a failed attempt to update a record (even if it's because it doesn't exist), should trigger an UPDATE failure event. What do you think?

mike-jumper: Yeah, I agree. There'd otherwise be no context for failures in audit logs.

mike-jumper (changes requested):

guacamole/src/main/frontend/src/app/import/templates/connectionImport.html
```diff
@@ -38,6 +38,28 @@ <h2>{{'IMPORT.SECTION_HEADER_CONNECTION_IMPORT' | translate}}</h2>
...
 
+        <ul class="import-config">
+            <li>
+                <input type="checkbox"
+                    id="existing-connection-mode" ng-model="importConfig.existingConnectionMode"
+                    ng-true-value="'REPLACE'" ng-false-value="'REJECT'" />
+                <label for="existing-connection-mode">
+                    {{'IMPORT.FIELD_HEADER_EXISTING_CONNECTION_MODE' | translate}}
+                </label>
+                <span title="{{'IMPORT.HELP_EXISTING_CONNECTION_MODE' | translate}}" class="help"></span>
```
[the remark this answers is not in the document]
jmuehlner: Good call! Will do.

guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -200,38 +200,50 @@
...
-        "HELP_FILE_TYPE_DESCRIPTION"        : "Three file types are supported for connection import: CSV, JSON, and YAML. The same data may be specified by each file type. This must include the connection name and protocol. Optionally, a connection group location, a list of users and/or user groups to grant access, connection parameters, or connection protocols may also be specified. Any users or user groups that do not exist in the current data source will be automatically created.",
+        "HELP_FILE_TYPE_DESCRIPTION"        : "Three file types are supported for connection import: CSV, JSON, and YAML. The same data may be specified by each file type. This must include the connection name and protocol. Optionally, a connection group location, a list of users and/or user groups to grant access, connection parameters, or connection protocols may also be specified. Any users or user groups that do not exist in the current data source will be automatically created. Note that any existing connection permissions will not be removed for updated connections.",
```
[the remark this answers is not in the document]
jmuehlner: Also good call - also will do.

guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -200,38 +200,50 @@
...
+        "ERROR_REQUIRED_PROTOCOL"            : "No connection protocol found in the provided file",
 
         "FIELD_PLACEHOLDER_FILTER" : "@:APP.FIELD_PLACEHOLDER_FILTER",
 
+        "FIELD_HEADER_EXISTING_CONNECTION_MODE" : "Replace/Update existing connections",
+        "FIELD_HEADER_EXISTING_PERMISSION_MODE" : "Reset permissions",
+
+        "FIELD_OPTION_DUPLICATE_REPLACE" : "Replace duplicates",
+        "FIELD_OPTION_DUPLICATE_IGNORE"  : "Ignore duplicates",
+        "FIELD_OPTION_DUPLICATE_ERROR"   : "Disallow duplicates",
```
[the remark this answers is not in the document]
jmuehlner: Err... nope

mike-jumper (approved):

PR #842, opened 2023-04-22
GUACAMOLE-926: Batch import - fix UI issue with permission checkbox, fix replace error when using decorating extension.

jmuehlner:
Yet again I failed to find these when I was fixing the earlier issues that I also failed to find earlier.

Anyway, this PR fixes 2 more problems:
* The permission replacement checkbox wasn't getting disabled properly
* An exception is thrown when trying to use the REPLACE verb in the patch directory endpoint, if a decorating extension e.g. `guacamole-history-recording-storage` is used.

### DIFF
--- a/guacamole/src/main/java/org/apache/guacamole/rest/directory/DirectoryResource.java
+++ b/guacamole/src/main/java/org/apache/guacamole/rest/directory/DirectoryResource.java
@@ -430,6 +430,10 @@ public abstract class DirectoryResource<InternalType extends Identifiable, Exter
      * @param identifier
      *     The identifier of the object to retrieve from the directory.
      *
+     * @param directory
+     *     The directory to fetch the object from. If null, the directory
+     *     associated with this DirectoryResource instance will be used.
+     *
      * @return
      *     The object from the directory with the provided identifier.
      *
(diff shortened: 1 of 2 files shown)


guacamole/src/main/java/org/apache/guacamole/rest/directory/DirectoryResource.java
```diff
@@ -639,10 +649,11 @@ else if (op == APIPatch.Operation.replace) {
...
 
-                            // Fetch the object to be updated. If no object is
-                            // found, a directory GET failure event will be
-                            // logged, and the update attempt will be aborted.
-                            original = getObjectByIdentifier(identifier);
+                            // Fetch the object to be updated from the atomic
+                            // directory instance. If no object is found, a 
+                            // directory GET failure event will be logged, and
+                            // the update attempt will be aborted.
+                            original = getObjectByIdentifier(identifier, directory);
```
jmuehlner, line 656: This is what happened when I enabled the history recording extension:

```
java.lang.ClassCastException: org.apache.guacamole.history.connection.HistoryConnection cannot be cast to org.apache.guacamole.auth.jdbc.connection.ModeledConnection
	at org.apache.guacamole.auth.jdbc.connection.ConnectionDirectory.update(ConnectionDirectory.java:71)
	at org.mybatis.guice.transactional.TransactionalMethodInterceptor.invoke(TransactionalMethodInterceptor.java:100)
	at org.apache.guacamole.rest.directory.DirectoryResource$3.executeOperation(DirectoryResource.java:652)
	at org.apache.guacamole.auth.jdbc.base.JDBCDirectory.tryAtomically(JDBCDirectory.java:42)```
	
In other words, it was pulling a connection from the history recording extension directory, which had been decorated the extension into a different `Connection` subclass, and trying to write it to JDBC directory, which did not like that one bit.

We have to use the directory provided by `tryAtomically()` since it's the only directory we have that's guaranteed to be atomic.

necouchman (approved):

PR #846, opened 2023-04-27
GUACAMOLE-1780: Fix TOTP interaction with SAML and SSL

jmuehlner:
This change allows the SAML auth extension to put a session token into a "pending" state when the SAML authentication attempt succeeds, but the overall auth attempt fails. 

This will allow a user to enter their TOTP code and resubmit the same auth token to resume their previous SAML auth session. After the login _does_ succeed, the SAML auth session is invalidated, ensuring that it can't be replayed.

I think there's probably a lot of different ways to solve this issue - this one seemed like a pretty reasonable way to handle it.
Let me know what you think!


guacamole-ext/src/main/java/org/apache/guacamole/net/auth/AuthenticationProvider.java
```diff
@@ -247,5 +247,42 @@ UserContext redecorate(UserContext decorated, UserContext context,
...
+     *      provider.
+     *
+     * @param credentials
+     *     The credentials which were most recently submitted for the given
+     *     AuthenticatedUser. These are not guaranteed to be the same as the
+     *     credentials associated with the AuthenticatedUser object, which are
+     *     the credentials provided when the user originally authenticated.
+     */
+    default void notifyAuthenticationSuccess(
+                AuthenticatedUser authenticatedUser, Credentials credentials) {};
```
mike-jumper, line 286: Why this vs. the existing `AuthenticationSuccessEvent` and `AuthenticationFailureEvent`?

jmuehlner: Sure, that works. How does this look?

jmuehlner:
Putting this in draft while I investigate an issue.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/AuthenticationProvider.java
jmuehlner: I had some ideas when I was first working on this that it was important that failures should only be reported to auth providers that had actually succeeded, but that turned out to not really matter for this feature - and it's pretty easy for auth providers to keep track of that manually if required. 

I've switched over to the standard `Listener` approach now. Much simpler.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/src/main/java/org/apache/guacamole/auth/saml/AuthenticationEventListener.java
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * Manager of active SAML authentication attempts.
+     *
+     * Requires static injection due to the fact that the webapp just calls the
+     * constructor directly when creating new Listeners. The instances will not
+     * be constructed by guice.
+     */
+    @Inject
+    private static SAMLAuthenticationSessionManager sessionManager;
```
jmuehlner, line 43: Doing static injection is not my preference, but I think it's better than trying to rig up some sort of system by which the webapp has to figure out the correct way to construct each `Listener` class depending on whether the particular extension would like to construct it via guice or the normal constructor method.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-base/src/main/java/org/apache/guacamole/auth/sso/AuthenticationSessionManager.java
```diff
@@ -97,9 +149,21 @@ public String generateInvalid() {
      *     value was returned by defer().
      */
     public T resume(String identifier) {
-
         if (identifier != null) {
-            T session = sessions.remove(identifier);
+             T session = sessions.get(identifier);
```
mike-jumper, line 153: Indentation seems to have gotten wonky here (misaligned by one space).

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-base/src/main/java/org/apache/guacamole/auth/sso/AuthenticationSessionManager.java
```diff
@@ -97,9 +149,21 @@ public String generateInvalid() {
...
+            if (pendingSessions.contains(identifier)) {
+                invalidateSession(identifier);
+                return null;
+            }
+
+            // Otherwise, mark the session as pending. NOTE: Unless explicitly
+            // removed from pending status via a call to reactivateSession(),
+            // the next attempt to resume this session will fail
+            else
+                pendingSessions.add(identifier);
```
mike-jumper, line 165: Though the underlying collections are threadsafe, this access pattern is not. The intent appears to be that, given two separate calls to `resume("someValue")`, the first call succeeds and returns a non-null session while the second call returns `null`, but this may not happen. Consider:

| Thread 1                                              | Thread 2                      |
| ----------------------------------------------------- | ------------------------------------------------------ |
|`resume("someValue")`||
|                                                       | `resume("someValue")`                                  |
| `if (identifier != null) {` (true)                    |                              |
|                                                       | `if (identifier != null) {` (true)  |
| `T session = sessions.get(identifier);`               |                                                        |
|                                                       | `T session = sessions.get(identifier);`                |
| `if (pendingSessions.contains(identifier)) {` (false) |                                                        |
|                                                       |  `if (pendingSessions.contains(identifier)) {` (false) |
| `pendingSessions.add(identifier);`                    |  |
|                                                       | `pendingSessions.add(identifier);`                   |
| `if (session != null && session.isValid())` (true)    |  |
|                                                       | `if (session != null && session.isValid())` (true)  |
| `return session;`                                     |  |
|                                                       | `return session;`   |

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/src/main/java/org/apache/guacamole/auth/saml/AuthenticationEventListener.java
```diff
@@ -0,0 +1,70 @@
...
+
+            // If the SSL auth succeeded, but other auth providers failed to
+            // authenticate the user associated with the credentials in this
+            // failure event, they may wish to make another login attempt. To 
+            // avoid an infinite login attempt loop, re-enable the session
+            // associated with these credentials, allowing the auth attempt to be
+            // resumed without requiring another round trip to the SAML provider.
+            sessionManager.reactivateSession(
+                AuthenticationProviderService.getSessionIdentifier(
+                    ((AuthenticationFailureEvent) event).getCredentials()));
```
mike-jumper, line 66: This should probably only happen for specific exceptions, namely:

* `GuacamoleInsufficientCredentialsException`
* A `GuacamoleClientException` that isn't a `GuacamoleSecurityException`

Other failures like `GuacamoleInvalidCredentialsException`, `GuacamoleUnauthorizedException`, or some unknown internal failure, should instead fully invalidate the current flow as they did previously.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/src/main/java/org/apache/guacamole/auth/saml/AuthenticationProviderService.java
```diff
@@ -69,6 +69,26 @@ public class AuthenticationProviderService implements SSOAuthenticationProviderS
...
+    /**
+     * Return the value of the session identifier associated with the given
+     * credentials, or null if no session identifier is found in the
+     * credentials.
+     *
+     * @param credentials The credentials from which to extract the session
+     * identifier.
+     *
+     * @return The session identifier associated with the given credentials, or
+     * null if no identifier is found.
```
mike-jumper, line 81: Please reformat to match style of other JavaDoc.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-ssl/src/main/java/org/apache/guacamole/auth/ssl/AuthenticationEventListener.java
```diff
@@ -0,0 +1,74 @@
...
+            // avoid an infinite login attempt loop, re-enable the session
+            // associated with these credentials, allowing the auth attempt to be 
+            // resumed without requiring another round trip to the SSL service.
+            sessionManager.reactivateSession(
+                AuthenticationProviderService.getSessionIdentifier(
+                    ((AuthenticationFailureEvent) event).getCredentials()));
+
+    }
+
+}
```
mike-jumper, line 74: Could this and the SAML implementation perhaps be brought up into `guacamole-auth-sso-base` as a base class that both implementations might leverage? It looks like they apply the same logic.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-ssl/src/main/java/org/apache/guacamole/auth/ssl/AuthenticationEventListener.java
```diff
@@ -0,0 +1,74 @@
...
+
+            // If the SSL auth succeeded, but other auth providers failed to
+            // authenticate the user associated with the credentials in this
+            // failure event, they may wish to make another login attempt. To
+            // avoid an infinite login attempt loop, re-enable the session
+            // associated with these credentials, allowing the auth attempt to be 
+            // resumed without requiring another round trip to the SSL service.
+            sessionManager.reactivateSession(
+                AuthenticationProviderService.getSessionIdentifier(
+                    ((AuthenticationFailureEvent) event).getCredentials()));
```
mike-jumper, line 70: Same here - this should probably only happen for specific exceptions, with other exceptions invalidating the session.

mike-jumper (changes requested):

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-ssl/src/main/java/org/apache/guacamole/auth/ssl/AuthenticationEventListener.java
jmuehlner: Sure, done!

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-base/src/main/java/org/apache/guacamole/auth/sso/AuthenticationSessionManager.java
jmuehlner: Oops, my bad. Yeah that's not going to work. I've updated it and this issue should be fixed now.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-base/src/main/java/org/apache/guacamole/auth/sso/SSOAuthenticationEventListener.java
```diff
@@ -0,0 +1,115 @@
...
+        // nothing to do
+        if (sessionIdentifier == null)
+            return;
+
+        // If the SSL auth succeeded, but other auth providers failed to
+        // authenticate the user associated with the credentials in this
+        // failure event, they may wish to make another login attempt. To
+        // avoid an infinite login attempt loop, re-enable the session
+        // associated with these credentials, allowing the auth attempt to be
+        // resumed without requiring another round trip to the SSL service.
```
mike-jumper, line 61: SSO*?

mike-jumper (changes requested):
Minor typo - otherwise, LGTM.

mike-jumper (approved):

PR #848, opened 2023-04-28
GUACAMOLE-926: Further batch import fixes and improvements

jmuehlner:
This change fixes a few incorrect error messages, and adds extra error checking around invalid/missing values in import files.

In addition, I added a `username` and `password` parameters to all the example files, since these might be likely values that users would want to populate during import.


jmuehlner:
Actually realized I should add similar checks to the "users" and "groups" fields.

necouchman (approved):

PR #849, opened 2023-05-02
GUACAMOLE-926: Batch Import: Attempt to correct incorrectly-cased parameter values if possible

jmuehlner:
Yet another batch import bugfix / improvement!? Nobody could have seen _this_ coming!

This change does two things:
* Fixes an issue where file-level CSV errors didn't get surfaced to the user
* Attempts to correct any invalidly-cased user-provided connection parameter values, when the parameter has a limited number of possible options.

The latter is intended to help out with the case where a user may, while editing a CSV in many popular spreadsheets, have their "true" values for boolean parameters converted to "TRUE", which the backend will not understand.

### DIFF
--- a/guacamole/src/main/frontend/src/app/import/services/connectionParseService.js
+++ b/guacamole/src/main/frontend/src/app/import/services/connectionParseService.js
@@ -200,7 +200,8 @@ angular.module('import').factory('connectionParseService',
     /**
      * Returns a promise that will resolve to a transformer function that will
      * perform various checks and transforms relating to the connection group
-     * tree heirarchy. It will:
+     * tree hierarchy, pushing any errors into the resolved connection object.
+     * It will:
      * - Ensure that a connection specifies either a valid group path (no path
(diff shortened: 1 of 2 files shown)


guacamole/src/main/frontend/src/app/import/services/connectionParseService.js
```diff
@@ -200,7 +200,8 @@ angular.module('import').factory('connectionParseService',
     /**
      * Returns a promise that will resolve to a transformer function that will
      * perform various checks and transforms relating to the connection group
-     * tree heirarchy. It will:
+     * tree heirarchy, pushing any errors into the resolved connection object.
```
mike-jumper, line 203: hierarchy*

mike-jumper (changes requested):
Looks good - only issues I see are a couple typos.

jmuehlner:
> Looks good - only issues I see are a couple typos.

Spelling is hard

mike-jumper (approved):

PR #853, opened 2023-05-05
GUACAMOLE-615: Correct potential parse failures in webapp parser implementations.

mike-jumper:
The Java and JavaScript implementations of the Guacamole protocol parser may cause unexpected parsing failures due to those languages' use of UTF-16 internally for string representations and string length. In particular, characters involving high/low surrogates may end up with miscalculated length prefixes, ultimately causing the resulting instruction to fail to parse.

These changes refactor both implementations to calculate lengths in a way that takes high/low surrogates into account, and refactor the tunnel implementations to use a common parser.

The internal implementations of the length calculation (`codePointCount()` in Java and regex in JavaScript) were chosen after benchmarking the parser to verify that performance is not impacted. The "naive" solution (walking the string and counting codepoints manually) is _significantly_ slower, whereas the solution here is as fast as the old implementation (if not faster).

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Parser.js
+++ b/guacamole-common-js/src/main/webapp/modules/Parser.js
@@ -86,6 +86,29 @@ Guacamole.Parser = function Parser() {
     var elementCodepoints = 0;
 
     /**
+     * The number of parsed characters that must accumulate in the begining of
+     * the parse buffer before processing time is expended to truncate that
+     * buffer and conserve memory.
+     *
+     * @private
+     * @constant
+     * @type {!number}
+     */
+    var BUFFER_TRUNCATION_THRESHOLD = 4096;
+
+    /**
+     * The lowest Unicode codepoint to require a surrogate pair when encoded
+     * with UTF-16. In UTF-16, characters with codepoints at or above this
+     * value are represented with a surrogate pair, while characters with
+     * codepoints below this value are represented with a single character.
+     *
+     * @private
+     * @constant
+     * @type {!number}
+     */
+    var MIN_CODEPOINT_REQUIRES_SURROGATE = 0x10000;
+
+    /**
      * Appends the given instruction data packet to the internal buffer of
      * this Guacamole.Parser, executing all completed instructions at
      * the beginning of this buffer, if any.
@@ -109,7 +132,7 @@ Guacamole.Parser = function Parser() {
         else {
 
             // Truncate buffer as necessary
-            if (startIndex > 4096 && elementEnd >= startIndex) {
+            if (startIndex > BUFFER_TRUNCATION_THRESHOLD && elementEnd >= startIndex) {
 
                 buffer = buffer.substring(startIndex);
 
@@ -157,7 +180,7 @@ Guacamole.Parser = function Parser() {
                 // a high and low surrogate, elementEnd points to the low
                 // surrogate and NOT the element terminator. We must shift the
                 // end and reevaluate.
-                else if (elementCodepoints && buffer.codePointAt(elementEnd - 1) >= 0x10000) {
+                else if (elementCodepoints && buffer.codePointAt(elementEnd - 1) >= MIN_CODEPOINT_REQUIRES_SURROGATE) {
                     elementEnd++;
                     continue;
                 }
--- a/guacamole-common-js/src/main/webapp/modules/Parser.js
+++ b/guacamole-common-js/src/main/webapp/modules/Parser.js
@@ -22,14 +22,16 @@ var Guacamole = Guacamole || {};
 /**
  * Simple Guacamole protocol parser that invokes an oninstruction event when
  * full instructions are available from data received via receive().
- * 
+ *
  * @constructor
  */
-Guacamole.Parser = function() {
+Guacamole.Parser = function Parser() {
 
     /**
      * Reference to this parser.
+     *
      * @private
+     * @type {!Guacamole.Parser}
      */
     var parser = this;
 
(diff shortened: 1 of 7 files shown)


guacamole-common-js/src/main/webapp/modules/Parser.js
```diff
@@ -63,80 +92,135 @@ Guacamole.Parser = function() {
...
+            buffer = packet;
 
-            buffer = buffer.substring(start_index);
+        else {
 
-            // Reset parse relative to truncation
-            element_end -= start_index;
-            start_index = 0;
+            // Truncate buffer as necessary
+            if (startIndex > 4096 && elementEnd >= startIndex) {
```
jmuehlner, line 112: Any chance we could get a constant for this magic number?

guacamole-common-js/src/main/webapp/modules/Parser.js
```diff
@@ -63,80 +92,135 @@ Guacamole.Parser = function() {
...
+                if (codepoints < elementCodepoints) {
+                    elementEnd += elementCodepoints - codepoints;
+                    continue;
+                }
+
+                // If the current element ends with a character involving both
+                // a high and low surrogate, elementEnd points to the low
+                // surrogate and NOT the element terminator. We must shift the
+                // end and reevaluate.
+                else if (elementCodepoints && buffer.codePointAt(elementEnd - 1) >= 0x10000) {
```
jmuehlner, line 160: A constant would be nice for this number `0x10000` as well. I'm sure you're well familiar with what it means, but I'm less familiar with UTF8, and don't.

guacamole-common-js/src/main/webapp/modules/Parser.js
```diff
@@ -63,80 +92,135 @@ Guacamole.Parser = function() {
...
-            var length_end = buffer.indexOf(".", start_index);
-            if (length_end != -1) {
+            var lengthEnd = buffer.indexOf('.', startIndex);
+            if (lengthEnd !== -1) {
 
                 // Parse length
-                var length = parseInt(buffer.substring(element_end+1, length_end));
-                if (isNaN(length))
-                    throw new Error("Non-numeric character in element length.");
+                elementCodepoints = parseInt(buffer.substring(elementEnd+1, lengthEnd));
```
jmuehlner, line 208: To match existing code style, there should probably be a space between the `+` operator and its operands here.

guacamole-common-js/src/main/webapp/modules/Parser.js
```diff
@@ -157,3 +241,74 @@ Guacamole.Parser = function() {
...
+ *     should stop (the character after the last character being counted). If
+ *     omitted, all characters after the start location will be counted.
+ *
+ * @returns {!number}
+ *     The number of Unicode codepoints within the requested portion of the
+ *     given string.
+ */
+Guacamole.Parser.codePointCount = function codePointCount(str, start, end) {
+    str = str.substring(start || 0, end);
+    var surrogatePairs = str.match(/[\uD800-\uDBFF][\uDC00-\uDFFF]/g);
```
jmuehlner, line 272: A quick comment (or commented constant) on what this regex is doing would be cool.

guacamole-common-js/src/main/webapp/modules/Tunnel.js
```diff
@@ -546,14 +520,36 @@ Guacamole.HTTPTunnel = function(tunnelURL, crossDomain, extraTunnelHeaders) {
...
         var dataUpdateEvents = 0;
 
-        // The location of the last element's terminator
-        var elementEnd = -1;
+        var parser = new Guacamole.Parser();
+        parser.oninstruction = function instructionReceived(opcode, args) {
 
-        // Where to start the next length search or the next element
-        var startIndex = 0;
+            // Switch to next request end-of-stream is signalled
```
jmuehlner, line 526: I think there's a word missing in this comment.

guacamole-common-js/src/main/webapp/modules/Parser.js
mike-jumper: Sure, I'll add that.

guacamole-common-js/src/main/webapp/modules/Parser.js
mike-jumper: Sure, I'll add that, too. This value is actually not relevant to UTF-8, but UTF-16 - it's the cutoff for characters requiring surrogate pairs.

jmuehlner:
Everything looks pretty reasonable to me. Any concerns @necouchman (or anyone else)? This touches a lot of pretty important things so it'd be nice to get a second pair of eyes before I merge this.

guacamole-common-js/src/main/webapp/modules/Parser.js
```diff
@@ -22,39 +22,91 @@ var Guacamole = Guacamole || {};
...
+    /**
+     * The number of parsed characters that must accumulate in the begining of
+     * the parse buffer before processing time is expended to truncate that
+     * buffer and conserve memory.
+     *
+     * @private
+     * @constant
+     * @type {!number}
+     */
+    var BUFFER_TRUNCATION_THRESHOLD = 4096;
```
jmuehlner, line 97: Consider using `const` for constants.

mike-jumper: Unfortunately, so long as we continue to support IE, we can't do that within guacamole-common-js unless the build system is migrated to something that will compile to older JS.

We could do this if we migrate guacamole-common-js to the same build system and plugins used by the webapp.

necouchman:
Sure @jmuehlner, I'll try to have a look this afternoon/evening.

guacamole-common-js/src/main/webapp/modules/Parser.js
jmuehlner: Ah right right IE. Definitely not worth migrating the build system for just this.

necouchman:
Working through the review, hopefully be done with it today...

guacamole-common-js/src/main/webapp/modules/Tunnel.js
```diff
@@ -823,6 +749,8 @@ Guacamole.WebSocketTunnel = function(tunnelURL) {
      */
     var tunnel = this;
 
+    var parser = null;
```
necouchman, line 752: Guessing this should get some documentation?

necouchman (changes requested):
Looks pretty good - just one variable missing some documentation that the others around it seem to have. After that's fixed: :shipit: .

guacamole-common-js/src/main/webapp/modules/Tunnel.js
mike-jumper: Oops - yes, definitely.

mike-jumper: Fixed via rebase.

necouchman (approved):

jmuehlner (approved):

PR #862, opened 2023-05-11
GUACAMOLE-926: Fix SQL Server JDBC auth extension by disabling batch executor.

jmuehlner:
Turns out SQL Server didn't work at all after my changes in https://github.com/apache/guacamole-client/pull/809. 

Login attempts fail with the following error:
```
Error committing transaction.  
Cause: org.apache.ibatis.executor.ExecutorException: Error getting generated key or setting result to parameter object. 
Cause: com.microsoft.sqlserver.jdbc.SQLServerException: The statement must be executed before any results can be obtained.
```

This change disables the batch executor entirely for SQL Server and adds a warning to that effect.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/conf/SQLServerEnvironment.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/conf/SQLServerEnvironment.java
@@ -295,6 +295,21 @@ public class SQLServerEnvironment extends JDBCEnvironment {
                 true);
     }
 
+    @Override
+    public boolean shouldUseBatchExecutor() {
+
+        // The SQL Server driver does not work when batch execution is enabled.
+        // Specifically, inserts fail with com.microsoft.sqlserver.jdbc.SQLServerException:
+        // The statement must be executed before any results can be obtained.
+        // See https://github.com/microsoft/mssql-jdbc/issues/358 for more.
+        logger.warn(
+                "JDBC batch executor is disabled for SQL Server Connections. "
+                + "Large batched updates may run slower."
+        );
+        return false;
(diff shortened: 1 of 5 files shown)


extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/conf/SQLServerEnvironment.java
```diff
@@ -295,4 +295,36 @@ public boolean enforceAccessWindowsForActiveSessions() throws GuacamoleException
                 true);
     }
 
+    @Override
+    public boolean shouldUseBatchExecutor() {
```
jmuehlner, line 299: This change isn't required to fix the problem with the batch executor, but it is awfully handy for testing using self-hosted SQL Server instances with self-signed certificates.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-sqlserver/src/main/java/org/apache/guacamole/auth/sqlserver/conf/SQLServerEnvironment.java
```diff
@@ -295,4 +295,36 @@ public boolean enforceAccessWindowsForActiveSessions() throws GuacamoleException
...
+    /**
+     * Returns true if all server certificates should be trusted, including
+     * those signed by an unknown certificate authority, such as self-signed
+     * certificates, or false otherwise.
+     *
+     * @throws GuacamoleException
+     *     If an error occurs while retrieving the property value, or if the
+     *     value was not set, as this property is required.
+     */
+    public boolean trustAllServerCertificates() throws GuacamoleException {
```
jmuehlner, line 322: This isn't required to fix the problem with the batch executor, but it is awfully handy for testing using self-hosted SQL Server instances with self-signed certificates.

PR #875, opened 2023-05-22
GUACAMOLE-1792: Updated czech translation

Vict0rC:
Correction of some inaccurately translated words in the previous translation. Completion of the translation of the "import" section according to the English original.

### DIFF
--- a/guacamole-common-js/jsdoc-conf.json
+++ b/guacamole-common-js/jsdoc-conf.json
@@ -4,7 +4,7 @@
     },
     "opts" : {
         "recurse" : true,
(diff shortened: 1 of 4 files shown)


mike-jumper (changes requested):
Please update your commit message to include the ":" after "GUACAMOLE-1792". If possible, it would be nice if the commit message noted anything about the nature of these changes to differentiate them from any other update to the Czech translations.

Other than the commit message, these changes look fine.

Vict0rC:
Is it ok like this?
https://github.com/apache/guacamole-client/pull/876

PR #877, opened 2023-05-23
GUACAMOLE-1793: Fix creation of SessionRecording from blob.

leonard2901:
When creating a new SessionRecording from a blob an error is thrown in the `readNextBlock` function when reading `blob.size` since `blob` is undefined. My understanding is that this will always be the case when creating a new SessionRecording from a blob.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/SessionRecording.js
+++ b/guacamole-common-js/src/main/webapp/modules/SessionRecording.js
@@ -398,8 +398,10 @@ Guacamole.SessionRecording = function SessionRecording(source) {
     };
 
     // Read instructions from provided blob, extracting each frame
-    if (source instanceof Blob)
+    if (source instanceof Blob) {
+        recordingBlob = source;
         parseBlob(recordingBlob, loadInstruction, notifyLoaded);
+    }
 
     // If tunnel provided instead of Blob, extract frames, etc. as instructions
     // are received, buffering things into a Blob for future seeks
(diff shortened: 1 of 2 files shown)


mike-jumper (commented):
Looks good, however from your comment on JIRA:

>
> ...
>
> To reproduce this, run the following test:
>
> ```js
> describe("Guacamole.SessionRecording", function EventSpec() {
>     it("should create new SessionRecording instance from Blob", async > function () {
>         const url = 'https://raw.githubusercontent.com/apache/' +
>             'guacamole-client/master/doc/guacamole-playback-example/src/main/webapp/recording.guac';
>         const response = await fetch(url);
>         const blob = await response.blob();
>         const recording = new Guacamole.SessionRecording(blob);
>         expect(recording).not.toBeNull();
>     });
> });
> ```
>
> ...
>

Perhaps that test should be added to the unit tests within guacamole-common-js?

I'm otherwise good with these changes, though I think we should wait to merge until after 1.5.2-RC1 is promoted to 1.5.2, so that we can create a bugfix release branch and this can be part of the next bugfix release.

mike-jumper (changes requested):
I went ahead and created the `staging/1.5.3` branch for release-after-next. Please rebase and retarget this PR against that when you can.

I also highly recommend adding your unit test.

leonard2901:
Instead of fetching a remote recording I created a blob from a simple command string. I hope it's okay like this.

mike-jumper:
Sounds good to me.

mike-jumper (approved):

PR #880, opened 2023-05-24
GUACAMOLE-1795: Support new private key field name for KSM Pam User.

jmuehlner:
As discussed in https://issues.apache.org/jira/browse/GUACAMOLE-1795, the "Pam User" record type in KSM has private keys stored in a standard (non-custom) hidden field called "private pem key" or similar. 

This change modifies the private key function to look in this location as well to support this new record type.


necouchman (approved):
Looks okay to me.

PR #881, opened 2023-06-01
GUACAMOLE-1796: Reduce SQL Server batch size to ensure fetching large lists of connections still works. 

jmuehlner:
As discussed in https://issues.apache.org/jira/browse/GUACAMOLE-1796, a batch size of `1000` is too big when fetching large lists of  connections.

I tested and verified that the suggested batch size of `500` works correctly when fetching thousands of connections, where `1000` does not.


necouchman (approved):

jmuehlner:
@necouchman good call! Rebased.

PR #882, opened 2023-06-01
GUACAMOLE-1799: Add Name and Version properties to all translation files

leonard2901:
Only the English and Polish translation files contain the APP.NAME and APP.VERSION properties. The values of these properties are used by the login page for example.

Currently this does not seem to be a problem because the English translation file is *somehow* always loaded before the translation file of the language that the user requested.

To not have to rely on this behavior and to prepare things for the Angular frontend I would like to include these properties in all translation files.

### DIFF
--- a/guacamole/src/main/frontend/src/translations/ca.json
+++ b/guacamole/src/main/frontend/src/translations/ca.json
@@ -4,6 +4,9 @@
     
     "APP" : {
 
+        "NAME"    : "Apache Guacamole",
+        "VERSION" : "${project.version}",
+
         "ACTION_ACKNOWLEDGE"        : "OK",
         "ACTION_CANCEL"             : "Cancel·lar",
         "ACTION_CLONE"              : "Clon",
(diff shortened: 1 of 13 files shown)


necouchman:
> Currently this does not seem to be a problem because the English translation file is somehow always loaded before the translation file of the language that the user requested.

This is intentional - it is done this way so that, if things are not translated into a particular language, they will still at the very least be displayed in English, rather than showing up as a very odd translation string.

> To not have to rely on this behavior and to prepare things for the Angular frontend I would like to include these properties in all translation files.

Why is relying on this behavior problematic?

mike-jumper:
Closing as the described behavior is intentional. @leonard2901, if there is a technical reason that this is impractical with Angular vs. AngularJS, please let us know. I wouldn't expect Angular to prevent us from providing the same functionality, though.

leonard2901:
Using the English translations as a fallback makes sense. I did not consider that translations might be missing. Thanks for the clarification!

PR #884, opened 2023-06-09
GUACAMOLE-1806: Update Java dependencies to patched versions

Inperpetuammemoriam:
These changes should address the following (potentially relevant) vulnerabilities:
- CVE-2022-21724
- CVE-2022-26520
- CVE-2022-31197
- CVE-2022-40151
- CVE-2022-40152
- CVE-2022-41946
- CVE-2023-20861
- CVE-2023-20862
- CVE-2023-20863
- GHSA-673j-qm5f-xpv8


mike-jumper (changes requested):
Please:

1. Tag the relevant JIRA issue in your commit messages (see contribution guidelines and established git history).
2. Rebase and retarget this against the `staging/1.5.3` branch.

We routinely update to the latest stable+compatible versions of dependencies each release, and I just created a JIRA issue for that:

https://issues.apache.org/jira/browse/GUACAMOLE-1807

Are there any other dependencies that can and should be updated here?

Inperpetuammemoriam:
> Are there any other dependencies that can and should be updated here?

There are two more Java dependencies that should be taken care of in one way or another. (see [GUACAMOLE-1806](https://issues.apache.org/jira/browse/GUACAMOLE-1806))

mike-jumper (changes requested):
Please make sure all commits tag the relevant JIRA issue. There's still one commit that needs updating:

image

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/pom.xml
```diff
@@ -74,6 +74,23 @@
...
+                    <artifactId>woodstox-core</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <!-- Woodstox -->
+        <dependency>
+            <groupId>com.fasterxml.woodstox</groupId>
+            <artifactId>woodstox-core</artifactId>
+            <version>5.4.0</version>
```
necouchman, line 93: Was this change thoroughly tested to make sure that the SAML library functions as expected with this dependency being swapped out from under it?

Inperpetuammemoriam: I built a simple container environment including a SAML Identity Provider (Keycloak) and was still able to perform a login.

Just to be clear: I am far away from being an expert. So any advice on getting this done the right way is highly appreciated.

mike-jumper: I believe the upstream SAML library corrected this by updating their version of xmlsec, and the change there is minimal:

https://github.com/SAML-Toolkits/java-saml/pull/377

We could do that instead of overriding Woodstox, and trust that the change is equivalent to what java-saml has done but not released.

Inperpetuammemoriam: Correct me, if I'm wrong: Would updating xmlsec to 2.3.0 not also mean to update woodstox-core to 6.2.6? That seems to be an even bigger change.

The version 6.2.6 of woodstox-core furthermore contains the following vulnerabilities that version 5.4.0 does not:
- CVE-2022-40156
- CVE-2022-40155
- CVE-2022-40154
- CVE-2022-40153
- CVE-2022-40152

mike-jumper: Oof, yeah - good catch and never mind. Bigger change is not necessarily bad for a dependency we don't directly use ourselves (presumably the library using it has already tested their own compatibility), but we definitely don't want to pull in new issues.

mike-jumper (approved):

PR #886, opened 2023-06-12
GUACAMOLE-1810: Fix meta handling on Mac Firefox.

jmuehlner:
Tested on firefox, chrome, and safari. Handling appears to work correctly on all with both left and right option.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Keyboard.js
+++ b/guacamole-common-js/src/main/webapp/modules/Keyboard.js
@@ -1314,9 +1314,10 @@ Guacamole.Keyboard = function Keyboard(element) {
 
             var keydownEvent = new KeydownEvent(e);
 
-            // Ignore (but do not prevent) the "composition" keycode sent by some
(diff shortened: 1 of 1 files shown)


guacamole-common-js/src/main/webapp/modules/Keyboard.js
```diff
@@ -1381,24 +1380,37 @@ Guacamole.Keyboard = function Keyboard(element) {
...
+            if (e.data && !e.isComposing)
                 guac_keyboard.type(e.data);
-            }
+
+        };
+
+        /**
+         * Handles the given "compositionstart" event, automatically removing
+         * the "input" event handler, as "input" events should only be handled
+         * IFF composition events are not provided by the browser.
```
necouchman, line 1391: IFF = ?

Maybe I'm just getting old and don't understand all the new-fangled language ;-).

jmuehlner: Oh, If and Only If. Though, saying "only If and Only If" is awfully redundant... let me fix that.

necouchman: I suspected it might be that, but not terribly familiar with it...

necouchman (approved):

PR #890, opened 2023-06-18
GUACAMOLE-1813: Add support for history recording storage extension to Docker

necouchman:
This adds support for the guacamole-history-storage-extension to the Docker container. I went ahead and based it against 1.5.3 - if anyone feels it should get pushed out to 1.6.0 that's fine - it's a bit on the fence as to whether it's actually a bugfix or not.


mike-jumper:
I agree with this being a bugfix and worth basing on 1.5.3.

It may even be worth expanding the concept of what can be included in these patch releases to anything that doesn't require changes to the docs (ie: anything that we can merge and then call it "done").

PR #891, opened 2023-06-20
GUACAMOLE-1803: Fix frame scheduling logic in recording playback.

jmuehlner:
This is fixing an issue created in https://github.com/apache/guacamole-client/commit/7176ccce682202fcc0f804d52e9fba5efeee5a83 where frames are (in effect) never scheduled during recording playback, so whatever frame you seek to is just stuck forever.

I guess when I tested this I must not have properly rebuilt things, but I confirmed it was properly built and tested now.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/SessionRecording.js
+++ b/guacamole-common-js/src/main/webapp/modules/SessionRecording.js
@@ -149,13 +149,13 @@ Guacamole.SessionRecording = function SessionRecording(source, refreshInterval) 
     var currentFrame = -1;
 
     /**
-     * The timestamp of the frame when playback began, in milliseconds. If
+     * The position of the recording when playback began, in milliseconds. If
      * playback is not in progress, this will be null.
      *
      * @private
      * @type {number}
      */
-    var startVideoTimestamp = null;
+    var startVideoPosition = null;
 
     /**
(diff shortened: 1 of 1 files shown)


guacamole-common-js/src/main/webapp/modules/SessionRecording.js
```diff
@@ -750,29 +750,38 @@ Guacamole.SessionRecording = function SessionRecording(source, refreshInterval)
             // Pull the upcoming frame
             var next = frames[currentFrame + 1];
 
-            // The position at which the next frame should be rendered, taking
-            // into account any accumulated delays from rendering frames so far
-            var nextFramePosition = next.timestamp - startVideoTimestamp + startRealTimestamp;
```
jmuehlner, line 755: The big problem here is that this was supposed to be a _position_ within the recording, but it's actually a browser _timestamp_, so it's going to be way larger than the delay until the next `refreshInterval` update, meaning that the timer will continuously tick up but no frames were ever rendered.

jmuehlner: Also, the whole idea of scheduling a frame at some different position from the actual frame position makes no sense.

mike-jumper (approved):

PR #893, opened 2023-06-21
GUACAMOLE-1796: Batch updates to permissions so large permission changes can succeed.

jmuehlner:
The issue here is that if you do a connection batch import with a file with 1000 connections, granting permission to each of these connections to a single user:

1. The connection creation succeeds since the connection creation is properly batched
2. The attempt to grant permissions causes an error since the permission updates were _not_ batched

This PR fixes point 2.


mike-jumper:
Is the issue specific to batch import? I'd think the same issue would arise if someone attempts to grant permissions to a bunch of connections at once, either via the REST API or through quite a lot of box-checking.

mike-jumper:
If specific to batch import, then this should be [GUACAMOLE-926](https://issues.apache.org/jira/browse/GUACAMOLE-926). If not, then GUACAMOLE-1796 is correct, but:

* This should target `staging/1.5.3` as a bug fix.
* The fix shouldn't be specific to batch import.

jmuehlner:
Good call, this isn't batch specific. I've rebased against 1.5.3 and gotten rid of connection-import-specific language.

PR #894, opened 2023-06-24
GUACAMOLE-1820: Create UI for viewing, searching, and navigating to key events in session recording player.

jmuehlner:
image


guacamole-common-js/src/main/webapp/modules/SessionRecording.js
```diff
@@ -872,6 +911,19 @@ Guacamole.SessionRecording = function SessionRecording(source, refreshInterval)
...
+     * is available.
+     *
+     * @event
+     * @param {!String} text
+     *     The typed text associated with the batch of text.
+     *
+     * @param {!number} timestamp
+     *     The relative timestamp associated with the batch of text.
+     */
+    this.onText = null;
```
mike-jumper, line 925: In keeping with the style of other events, this should be `ontext`.

guacamole/src/main/frontend/src/app/player/directives/textView.js
```diff
@@ -0,0 +1,172 @@
...
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
```
mike-jumper, line 45: Does this file come from https://github.com/glyptodon/glyptodon-enterprise-player? The original player implementation did, but I don't recall this being part of that.

guacamole/src/main/frontend/src/app/player/directives/player.js
```diff
@@ -161,56 +182,36 @@ angular.module('player').directive('guacPlayer', ['$injector', function guacPlay
...
+         * format, deferring to the time service for the implementation.
          *
          * @param {!number} value
          *     The time to format, in milliseconds.
          *
          * @returns {!string}
-         *     The given quantity of milliseconds formatted as "DD:HH:MM:SS".
+         *     A human readable string representing the provided number of
+         *     milliseconds.
          */
```
mike-jumper, line 213: Can just annotate:

```js
/**
 * @borrows playerTimeService.formatTime
 */
```

rather than copy the docs verbatim.

guacamole/src/main/frontend/src/app/player/directives/textView.js
```diff
@@ -0,0 +1,172 @@
...
+         * Convert the given quantity of milliseconds into a human readable
+         * format, deferring to the time service for the implementation.
+         *
+         * @param {!number} value
+         *     The time to format, in milliseconds.
+         *
+         * @returns {!string}
+         *     A human readable string representing the provided number of
+         *     milliseconds.
+         */
```
mike-jumper, line 166: Same here - you can just annotate:

```js
/**
 * @borrows playerTimeService.formatTime
 */
```

and avoid multiple copies of the same docs.

guacamole-common-js/src/main/webapp/modules/SessionRecording.js
```diff
@@ -872,6 +911,19 @@ Guacamole.SessionRecording = function SessionRecording(source, refreshInterval)
...
+    /**
+     * Fired whenever a new batch of typed text extracted from key events
+     * is available.
+     *
+     * @event
+     * @param {!String} text
+     *     The typed text associated with the batch of text.
+     *
+     * @param {!number} timestamp
+     *     The relative timestamp associated with the batch of text.
```
mike-jumper, line 923: Could this event should be richer? There definitely needs to be the rough text equivalent available for easy consumption, so this much is pretty good, but:

* This doesn't appear to provide a way for consuming code to distinguish between non-printable keypresses (like "<Backspace">) and the text equivalent to how we represent them (typing "<", "B", "a", ... etc.), which would prevent implementations from styling such keys differently or taking reliable automated action based on the presumed text typed.
* Timing information for keypresses vs. frames is lost (ie: there's no way for consuming code to choose to represent the keys pressed/released one key at a time)

Perhaps there's a way we can represent that in an object received by this event?

guacamole-common-js/src/main/webapp/modules/KeyEventBatcher.js
```diff
@@ -0,0 +1,481 @@
...
+ * NOTE: The event processing logic and output format is based on the `guaclog`
+ * tool, with the addition of batching support.
+ *
+ * @constructor
+ * @param {number} [batchSeperation=5000]
+ *     The minimum number of milliseconds that must elapse between subsequent
+ *     batches of key-event-generated text. If 0 or negative, no splitting will
+ *     occur, resulting in a single batch for all provided key events.
+ */
+Guacamole.KeyEventBatcher = function KeyEventBatcher(batchSeperation) {
```
mike-jumper, line 36: I'd recommend `KeyEventInterpreter` over `KeyEventBatcher`.

guacamole-common-js/src/main/webapp/modules/KeyEventBatcher.js
```diff
@@ -0,0 +1,481 @@
...
+        {keysym: 0xFFFF, name: 'Delete' }
+    ];
+
+    /**
+     * All known keys, as a map of X11 keysym to KeyDefinition.
+     *
+     * @constant
+     * @type {Object.<String, KeyDefinition>}
+     */
+    var KNOWN_KEYS = {};
```
mike-jumper, line 191: This should also be annotated `@private`.

guacamole-common-js/src/main/webapp/modules/KeyEventBatcher.js
```diff
@@ -0,0 +1,481 @@
...
+    };
+
+    /**
+     * A precursor array to the KNOWN_KEYS map. The objects contained within
+     * will be constructed into full KeyDefinition objects.
+     *
+     * @constant
+     * @type {Object[]}
+     */
+    var _KNOWN_KEYS = [
```
mike-jumper, line 97: This should also be annotated `@private`.

guacamole-common-js/src/main/webapp/modules/KeyEventBatcher.js
```diff
@@ -0,0 +1,481 @@
...
+
+    /**
+     * A definition for a known key.
+     *
+     * @constructor
+     * @param {KEY_DEFINITION|object} [template={}]
+     *     The object whose properties should be copied within the new
+     *     KEY_DEFINITION.
+     */
+    var KeyDefinition = function KeyDefinition(template) {
```
mike-jumper, line 58: This should also be annotated `@private`.

guacamole-common-js/src/main/webapp/modules/KeyEventBatcher.js
```diff
@@ -0,0 +1,481 @@
...
+
+    /**
+     * Returns true if the currently-pressed keys are part of a shortcut, or
+     * false otherwise.
+     *
+     * @returns {!boolean}
+     *     True if the currently-pressed keys are part of a shortcut, or false
+     *     otherwise.
+     */
+    function isShortcut() {
```
mike-jumper, line 244: This should also be annotated `@private`.

mike-jumper (changes requested):
Initial review - pretty spiffy.

There are a good few functions in `KeyEventBatcher.js` that are private but not annotated with `@private`. I've commented on a couple of these, but please look through for others.

With respect to the UI-design, I think we should also:

* Add some degree of padding between the filter and the edge of the key event viewer (the border of the filter is currently directly against the edge of the dialog).
* If possible, visibly distinguish between typed characters and non-printable keys.
* If possible, visibly distinguish between characters that have been typed, are being typed, and will be typed.

guacamole/src/main/frontend/src/app/player/directives/textView.js
jmuehlner: Oops, no.

jmuehlner:
> * If possible, visibly distinguish between typed characters and non-printable keys.
> * If possible, visibly distinguish between characters that have been typed, are being typed, and will be typed.

Ok, I've visually distinguished these by bolding the non-printable keys and control characters, and changing the color of future keypresses to a slightly lighter gray.
image

Currently, every single key event has its own DOM element, so you can watch the characters flipping from gray to black (which is pretty fun TBH).

With my test recordings, the performance seemed just the same as before. I also tried doing an SSH recording where I held down keys on the keyboard for 20 minutes straight, resulting in 23499 DOM elements being rendered for the individual characters.

I tested this recording on Firefox and Chrome, and it performed about the same in both. Seeking around was pretty slow, but that seemed to be mostly rendering. The initial video processing was pretty smooth, taking ~10 seconds or so. When playing back the recording, the browser did use quite a bit of CPU, but never crashed or had major usability issues.

I'd be curious to know people's thoughts on this approach. If this is too performance-heavy, I could also somewhat streamline things by combining sequential similarly-styled key events into a single DOM element, though we'd lose granularity as far as which events occurred before or after the current recording playback position.

necouchman:
> Currently, every single key event has its own DOM element, so you can watch the characters flipping from gray to black (which is pretty fun TBH).

Like a Mickey Mouse sing-a-long! :laughing:

mike-jumper:
> > Currently, every single key event has its own DOM element, so you can watch the characters flipping from gray to black (which is pretty fun TBH).
> 
> Like a Mickey Mouse sing-a-long! 😆

Excellent ... now all we need to do is write a key-raoke extension ...

jmuehlner:
> If this is too performance-heavy, I could also somewhat streamline things by combining sequential similarly-styled key events into a single DOM element, though we'd lose granularity as far as which events occurred before or after the current recording playback position.

Well, I went ahead and implemented this in a seperate commit that we can always pull off if people don't like it. Let me know what you think.

guacamole-common-js/src/main/webapp/modules/SessionRecording.js
jmuehlner: This is addressed now.

jmuehlner: This is addressed now.

jmuehlner:
It's unclear why the build failure is happening only for my branch and not `master`, since I didn't modify any of the affected projects, but the problem seems to be that `guacamole-auth-ban` and `guacamole-auth-sso` both have the following in their `pom.xml`:
```
<parent>
    <artifactId>extensions</artifactId>
    <version>1.5.2</version>
    <relativePath>../</relativePath>
</parent>
```

This is an issue because there's no `extensions` project at version 1.5.2 located at `../`. It's now 1.5.3 instead.

mike-jumper (changes requested):

guacamole-common-js/src/main/webapp/modules/KeyEventInterpreter.js
```diff
@@ -0,0 +1,578 @@
...
+    /**
+     * If the provided keysym corresponds to a valid UTF-8 character, return
+     * a KeyDefinition for that keysym. Otherwise, return null.
+     *
+     * @private
+     * @param {Number} keysym
+     *     The keysym to produce a UTF-8 KeyDefinition for, if valid.
+     *
+     * @returns
+     *     Return a KeyDefinition for the provided keysym, if it it's a valid
```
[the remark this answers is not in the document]
jmuehlner: image

guacamole-common-js/src/main/webapp/modules/KeyEventInterpreter.js
```diff
@@ -0,0 +1,578 @@
...
+        }
+
+        // Set initial byte
+        byteView[0] = mask | codepoint;
+
+        // Convert to UTF8 string
+        var name = new TextDecoder("utf-8").decode(byteArray);
+
+        // Create and return the definition
+        return new KeyDefinition({keysym: keysym.toString(), name: name, value: name, modifier: false});
```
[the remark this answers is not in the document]
jmuehlner: I don't see where that's happening. Can you be a bit more specific?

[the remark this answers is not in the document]
jmuehlner: Ah, I do see that calling `toString()` here is pointless. I'll get rid of that.

guacamole-common-js/src/main/webapp/modules/KeyEventInterpreter.js
```diff
@@ -0,0 +1,579 @@
...
+
+    /**
+     * Handles a raw key event, potentially appending typed text to the
+     * current batch, and calling onBatch with the current batch, if the
+     * callback is set and a new batch is about to be started.
+     *
+     * @param {!string[]} args
+     *     The arguments of the key event.
+     */
+    interpreter.handleKeyEvent = function handleKeyEvent(args) {
```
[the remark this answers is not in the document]
jmuehlner: Uh, not sure, but it's easiest to just match convention and do `this.nameOfTheThing`.

guacamole/src/main/frontend/src/app/player/styles/textView.css
```diff
@@ -0,0 +1,137 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*
+ * NOTE: This session recording player implementation is based on the Session
+ * Recording Player for Glyptodon Enterprise which is available at
+ * https://github.com/glyptodon/glyptodon-enterprise-player under the
+ * following license:
```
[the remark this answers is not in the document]
jmuehlner: I guess not, no. Removed.

guacamole-common-js/src/main/webapp/modules/SessionRecording.js
```diff
@@ -104,6 +104,16 @@ Guacamole.SessionRecording = function SessionRecording(source, refreshInterval)
...
+    /**
+     * The minimum number of milliseconds which must elapse between key events
+     * before text can be split across multiple frames.
+     *
+     * @private
+     * @constant
+     * @type {Number}
+     */
+    var TYPED_TEXT_INTERVAL = 5000;
+
```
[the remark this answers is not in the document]
jmuehlner: Yeah, it was vestigal. Now gone.

guacamole-common-js/src/main/webapp/modules/KeyEventInterpreter.js
```diff
@@ -0,0 +1,579 @@
...
+    // Default to 0 seconds to keep the raw timestamps
+    if (startTimestamp === undefined || startTimestamp === null)
+        startTimestamp = 0;
+
+    /**
+     * A definition for a known key.
+     *
+     * @constructor
+     * @private
+     * @param {KEY_DEFINITION|object} [template={}]
```
mike-jumper, line 65: `KEY_DEFINITION` is not a type. Should this be `KeyDefinition`?

guacamole-common-js/src/main/webapp/modules/KeyEventInterpreter.js
```diff
@@ -0,0 +1,579 @@
...
+        for (var i = 1; i < bytes; i++) {
+            byteView[bytes - i] = 0x80 | (codepoint & 0x3F);
+            codepoint >>= 6;
+        }
+
+        // Set initial byte
+        byteView[0] = mask | codepoint;
+
+        // Convert to UTF8 string
+        var name = new TextDecoder("utf-8").decode(byteArray);
```
mike-jumper, line 312: Can you clarify why this is necessary? I read:

1. X11 keysym is converted to a Unicode codepoint.
2. Codepoint is manually encoded as UTF-8 bytes.
3. UTF-8 bytes are decoded to text.

Seems like a verrrry roundabout way to do:

```js
var name = String.fromCodePoint(codepoint);
```

mike-jumper (changes requested):

guacamole-common-js/src/main/webapp/modules/KeyEventInterpreter.js
mike-jumper: Or even:

```js
var name = String.fromCharCode(codepoint);
```

so long as we know `codepoint` will always be no greater than `0xFFFF`.

jmuehlner: Uh yeah, we definitely know that it will be no greater than `0xFFFF` considering `var codepoint = keysym & 0xFFFF;` on line 281. I'll use `fromCharCode`. All this stuff was carried over from the `guaclog` codebase and is indeed not needed.

Good catch.

guacamole-common-js/src/main/webapp/modules/KeyEventInterpreter.js
```diff
@@ -0,0 +1,550 @@
...
+     * incomplete, as more key events might be processed before the next
+     * batch starts.
+     *
+     * @returns {Guacamole.KeyEventInterpreter.KeyEventBatch}
+     *     The current batch of text.
+     */
+    this.getCurrentBatch = function getCurrentBatch() {
+        return currentBatch;
+    }
+}
```
mike-jumper, line 469: Missing semicolon.

guacamole-common-js/src/main/webapp/modules/KeyEventInterpreter.js
```diff
@@ -0,0 +1,550 @@
...
+    /**
+     * The simplified, human-readable value representing the key events for
+     * this batch, equivalent to concatenating the `text` field of all key
+     * events in the batch.
+     *
+     * @type {!String}
+     */
+    this.simpleValue = simpleValue || '';
+
+}
```
mike-jumper, line 550: Missing semicolon.

guacamole-common-js/src/main/webapp/modules/KeyEventInterpreter.js
```diff
@@ -0,0 +1,550 @@
...
+     * Return the current batch of typed text. Note that the batch may be
+     * incomplete, as more key events might be processed before the next
+     * batch starts.
+     *
+     * @returns {Guacamole.KeyEventInterpreter.KeyEventBatch}
+     *     The current batch of text.
+     */
+    this.getCurrentBatch = function getCurrentBatch() {
+        return currentBatch;
+    }
```
mike-jumper, line 468: Missing semicolon.

guacamole-common-js/src/main/webapp/modules/KeyEventInterpreter.js
```diff
@@ -0,0 +1,550 @@
...
+                // Add the key to the current batch
+                currentBatch.simpleValue += keyText;
+                currentBatch.events.push(new Guacamole.KeyEventInterpreter.KeyEvent(
+                        keyText, typed, relativeTimestap));
+
+            }
+
+        }
+
+    }
```
mike-jumper, line 456: Missing semicolon.

mike-jumper (changes requested):
Alrighty - just a few missing semicolons, but otherwise LGTM.

jmuehlner:
> Alrighty - just a few missing semicolons, but otherwise LGTM.

Semicolons fixed.

mike-jumper (approved):

PR #898, opened 2023-07-05
GUACAMOLE-1829: Correct versions of dependencies for remaining 1.5.2 versioned extensions.

jmuehlner:
[This PR build](https://github.com/apache/guacamole-client/actions/runs/5468351262/jobs/9955867163?pr=894) is failing because a few extensions didn't have their versions bumped when 1.5.3 changes were merged back to master. 

Note that none of the affected extensions were actually modified in that PR. This error likely hasn't happened on the master branch due to stale docker layers.

This change corrects the versions so that everything builds correctly.

### DIFF
--- a/extensions/guacamole-auth-ban/pom.xml
+++ b/extensions/guacamole-auth-ban/pom.xml
@@ -26,14 +26,14 @@
     <groupId>org.apache.guacamole</groupId>
     <artifactId>guacamole-auth-ban</artifactId>
     <packaging>jar</packaging>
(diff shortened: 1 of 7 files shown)


jmuehlner:
Note that these versions weren't bumped in the 1.5.3 branch because these extensions aren't _in_ the 1.5.3 release. That's also why this fix is against master, not 1.5.3 (which is already fine).

necouchman (approved):

PR #899, opened 2023-07-10
GUACAMOLE-1807: Update to latest compatible versions of Java dependencies and JDBC drivers.

mike-jumper:
This change:

* Adds a convenient `checkDependencies` property that can be set to `true` during the build to automatically check for dependencies with updates or CVEs, producing `dependency-update-report.txt` and `dependency-check-report.html` respectively.
* Updates to the latest compatible versions of the various Java dependencies (there are no applicable updates to the runtime JavaScript dependencies)
* Adds Docker build arguments for controlling the versions of the bundled JDBC drivers, and updates those to the latest.

**BUT**, beware that:
* While the update check catches _almost_ everything, not all available updates will be caught. For example, major updates to Guava are usually compatible, but won't show up in the update report as they involve a major version bump.
* The CVE check always produces false positives, though the false positives are usually pretty easy to recognize.


mike-jumper:
Converting to draft while I double-check that this covers everything...

jmuehlner:
Why is that?
> beware that major updates to Guava are usually compatible but not included in these results

mike-jumper:
> > beware that major updates to Guava are usually compatible but not included in these results
>
> Why is that?

Google frequently releases updates to Guava as major version bumps, but there has yet to be a change affecting compatibility that affects the parts of Guava used by Guacamole.

mike-jumper:
Should be good for review now.

jmuehlner (approved):
LGTM

PR #902, opened 2023-07-21
GUACAMOLE-1239: Make identifier comparison case-insensitive.

necouchman:
This is an attempt to make comparison of usernames case-insensitive, as most authentication systems do not distinguish between case in usernames. This goes about it by 1) converting all identifiers to lower-case, and 2) making the comparison of identifiers case-insensitive.

I'm really more doing this to get the conversation started on the best way to go about this. I realize that this method is a sort of atomic bomb approach - it's simple, but has a lot of (potentially negative) implications. It's also not in any way configurable. I'm happy to discuss other ways that we can accomplish it, but both personal experience with case-sensitive usernames in Guacamole along with the other folks who are asking for it makes me believe it's something we need to address.


mike-jumper:
This looks reasonable to me, though I think it should be something that's determined by the implementation.

What about adding a function like `isIdentifierCaseSensitive()` to `Identifiable`, with a default implementation that just returns `true`? The `equals()` comparison could then take that into account (`hashCode()` would need to do the same).

The other issue will be with database queries. IIRC, MySQL/MariaDB is case-insensitive for username queries, but PostgreSQL is case-sensitive. We would need to somehow allow for MySQL/MariaDB to optionally consider case and for PostgreSQL to optionally ignore case ... ideally without making any schema changes.

necouchman:
> The other issue will be with database queries. IIRC, MySQL/MariaDB is case-insensitive for username queries, but PostgreSQL is case-sensitive. We would need to somehow allow for MySQL/MariaDB to optionally consider case and for PostgreSQL to optionally ignore case ... ideally without making any schema changes.

Are you talking about the actual DB usernames (as in what would be configured in guacamole.properties), or just the normal string-based queries in the tables? These changes, here, shouldn't actually impact the Postgres (or MySQL or MSSQL) usernames and/or passwords that Guacamole uses to access the database, and I would think that all databases consider case when doing comparisons for `varchar` fields, like the `name` field in the `guacamole_entity` table?

mike-jumper:
I'm referring to comparisons against the `guacamole_entity` table. There does seem to be a difference in how basic string comparisons are performed between the databases, probably due to differing defaults in collation.

When I try things against PostgreSQL, `guacadmin` and `GUACADMIN` are different:

```
guacamole_db=# SELECT * FROM guacamole_entity WHERE type = 'USER' AND name = 'guacadmin';
 entity_id |   name    | type 
-----------+-----------+------
         1 | guacadmin | USER
(1 row)

guacamole_db=# SELECT * FROM guacamole_entity WHERE type = 'USER' AND name = 'GUACADMIN';
 entity_id | name | type 
-----------+------+------
(0 rows)
```

When I try against MySQL, they're the same:

```
MariaDB [guacamole_db]> SELECT * FROM guacamole_entity WHERE type = 'USER' AND name = 'guacadmin';
+-----------+-----------+------+
| entity_id | name      | type |
+-----------+-----------+------+
|         1 | guacadmin | USER |
+-----------+-----------+------+
1 row in set (0.000 sec)

MariaDB [guacamole_db]> SELECT * FROM guacamole_entity WHERE type = 'USER' AND name = 'GUACADMIN';
+-----------+-----------+------+
| entity_id | name      | type |
+-----------+-----------+------+
|         1 | guacadmin | USER |
+-----------+-----------+------+
1 row in set (0.001 sec)
```

necouchman:
Got it - I'll take a run at it.

necouchman:
The generally accepted method for PostgreSQL to do case-insensitive searches/comparisons seems to be to use the `lower()` function to convert both strings to lower-case before the comparison.

MySQL appears to have a couple of options:
* Create or edit collation of the column in question to a case-sensitive option. The downside of this is that the schema has to either be changed or modified on-the-fly.
* Use a binary option when doing the column comparison, which might have a performance hit - views on that are mixed - because it may bypass indices.

necouchman:
> The other issue will be with database queries. IIRC, MySQL/MariaDB is case-insensitive for username queries, but PostgreSQL is case-sensitive. We would need to somehow allow for MySQL/MariaDB to optionally consider case and for PostgreSQL to optionally ignore case ... ideally without making any schema changes.

@mike-jumper : So I take it you're thinking that this should be configurable - default to case-sensitive, perhaps, but with a `guacamole.properties` option to make it case-insensitive?

mike-jumper:
I think the ideal would be a bit of a combination of all worlds:

* Some mechanism for an implementation to declare whether its identifiers are case-sensitive. If an implementation is absolutely inherently case-insensitive (like LDAP) or case-sensitive (like local Linux accounts), it would make sense for that fact to be made available to other implementations consuming those identifiers.
* Configuration options for any implementations where case sensitivity is within implementation control and it makes sense to provide such an option (I'd think the database backends would fall into this category).

necouchman:
Converted to a draft for now - I think I've accomplished just about everything except the ability to force the JDBC modules to deal with usernames in one form or the other Testing so far shows that an LDAP username entered in a different case will 1) get permissions assigned in JDBC from the username regardless of case, and 2) will not create duplicate user entries auto-created in a Postgres database.

necouchman:
@mike-jumper 

I'm working through the JDBC changes required to make case sensitivity configurable, and I'm getting a bit confused in some a couple of the mapper files. For example:

https://github.com/apache/guacamole-client/blob/3611be5c41b5c08d2dc224ff197274112eea0302/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/resources/org/apache/guacamole/auth/jdbc/permission/UserPermissionMapper.xml#L26-L32

The `affected_name` column is labeled as a `_name`, not `_identifier`; however, it maps to the `object_identifier` column and has a type of `INTEGER`. If I look at the `ObjectPermissionModel` class, though:

https://github.com/apache/guacamole-client/blob/3611be5c41b5c08d2dc224ff197274112eea0302/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/permission/ObjectPermissionModel.java#L30-L33

`objectIdentifier` has a type of `String`. So, shouldn't the jdbcType in the mapper file be `VARCHAR`, and not `INTEGER`?

Furthermore, going a little further in that file, if I look at the `selectOne` query definition:

https://github.com/apache/guacamole-client/blob/3611be5c41b5c08d2dc224ff197274112eea0302/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/resources/org/apache/guacamole/auth/jdbc/permission/UserPermissionMapper.xml#L54-L74

You can see, here, that we're selecting the columns `entity_id`, `permission`, and `name` into the aforementioned `UserPermissionResultMap`, but the `name` column coming out of the `guacamole_entity` table is definitely a `varchar` field, not an integer, but it's being returned as the user identifier (`objectIdentifier`) in that `ObjectPermissionModel` class.

What am I missing, here?

mike-jumper:
Here, since the relevant column from the database result is `affected_name` and that is indeed a string, the `jdbcType` here _should_ also be `VARCHAR`. This probably only works as-is because:

* The databases we support and their drivers are forgiving in this area.
* According to [the MyBatis documentation](https://mybatis.org/mybatis-3/sqlmap-xml.html#id-result), the `jdbcType` attribute is only required for nullable columns during `INSERT`/`UPDATE`/`DELETE`. BUT, since result maps are only relevant to `SELECT` queries, this seems a rather long-winded way of stating that the declared `jdbcType` has no impact on result maps.

I can't find any examples in the MyBatis documentation that show `jdbcType` ever being used for any `<result>`, so this may well be unnecessary.

You're correct that absolutely all of the `Identifiable` objects within the JDBC auth use identifiers that happen to be integers, with the exceptions of users and groups which instead use the username / group name. The integer IDs of users and groups are used only internally.

necouchman:
@mike-jumper @jmuehlner 

Appreciate any feedback you have on this most recent round of changes. I've spent some time trying to re-work the various JDBC modules to handle configurable case sensitivity settings. For each of the databases:
* PostgreSQL is case-sensitive by default - the widely-accepted way to make it case-insensitive is to convert everything to either lower or upper-case when comparing. I opted for `LOWER()`.
* MySQL is case-insensitive by default, and the community seems quite split over the best way to accomplish case-sensitive searches. The _easiest_ way, and the way I opted to go, is to use the `BINARY` keyword in the `WHERE` and `JOIN` clauses that need to be converted to case-sensitive comparisons. The _better_ way is to change the collation used in the queries; however, the challenge, here, is that it requires some knowledge of what collation is being used to begin with, which (I don't think) can be programmatically detected in a way that we can use. Maybe I'm wrong, and there is a way to do this?? The `BINARY` method has two pitfalls that I read about: 1) it ends up bypassing indexing, so there are some other creative things you have to do to insure you make use of that, and 2) it has some potential pitfalls when double-byte character sets are in use.
* SQLServer - haven't even dealt with this one, yet - figured I'd get consensus on how to handle Postgres and MySQL before tackling Microsoft. From what I've read, the only way to do this is to change the collation during query time, which is going to have the same challenge as that route with MySQL - you have to know something of what collation is in use to begin with.

Aside from database-specific issues, the other big thing to look at is how I've actually changed the interfaces that MyBatis uses with its mapper files to generate the required code...
* I was hoping to come up with some way to use something within the XML mapper files that could query the injected `JDBCEnvironment` object and call the `getCaseSensitiveUsernames()` method without having to modify the arguments of any of the queries. I could not figure out a way - if I've overlooked it, PLEASE tell me so that I can clean this up.
* Having not been able to figure out a way to do injection into the XML mapper files, I went with just modifying the arguments of the classes that need to know about the case-sensitivity, which leans to the side of breaking interface inheritance over passing useless parameters to methods that don't need it.
* The other thing to consider would be if we want to (continue to) limit the case sensitivity setting to _only_ usernames (as I've done, here), or if we want to make it where we're configuring all string comparison within a given DB environment (usernames, group names, connection names...whatever strings get compared). Or somewhere in between.

jmuehlner:
> The better way is to change the collation used in the queries; however, the challenge, here, is that it requires some knowledge of what collation is being used to begin with, which (I don't think) can be programmatically detected in a way that we can use. Maybe I'm wrong, and there is a way to do this??

You should be able to query which collations are available on a database using `SHOW COLLATION`, but yeah it sounds like there's differing collations available in various version of MySQL/MariaDB. Trying to build something that queries which is available and tries to find the best one sounds like a mess.

So short answer no I don't know of a better way, sadly.

necouchman:
> > The better way is to change the collation used in the queries; however, the challenge, here, is that it requires some knowledge of what collation is being used to begin with, which (I don't think) can be programmatically detected in a way that we can use. Maybe I'm wrong, and there is a way to do this??
> 
> You should be able to query which collations are available on a database using `SHOW COLLATION`, but yeah it sounds like there's differing collations available in various version of MySQL/MariaDB. Trying to build something that queries which is available and tries to find the best one sounds like a mess.
> 
> So short answer no I don't know of a better way, sadly.
> 
> This is all more of a pain than I would have ever guessed, but generally your approach looks pretty reasonable to me so far.

Thanks for the feedback. Yeah, it seems like, even using the `SHOW COLLATION` option, we'd have to have some sort of look-up table that said "if collation is x, switch to y", but that means maintaining a lookup table...

necouchman:
I suppose the other route to consider, here, is to make the JDBC changes specific to forcing the PostgreSQL module into case-insensitive mode, and not worry about the settings/options/changes that force MySQL and/or SQL Server into case-sensitive mode? It seems like the most frequent request is to make the look-ups and comparisons case-_in_sensitive, and since MySQL and SQL Server already do that by default, we could just go with that.

@mike-jumper @jmuehlner Further thoughts?

necouchman:
@jmuehlner or @mike-jumper Ping on this one, so I can either clean it up for 1.6.0, or de-scope it if we need to. I'd rather it make it into this release if at all possible...

jmuehlner:
> @jmuehlner or @mike-jumper Ping on this one, so I can either clean it up for 1.6.0, or de-scope it if we need to. I'd rather it make it into this release if at all possible...

I'm fine with the idea of offering an option to enable case-insensitive comparison, and just let it use the default case sensitiy behavior of the database otherwise. I think this would cover 95% of the use case of this feature.

I'm fine with including this in 1.6.0 if it's ready.

necouchman:
> > @jmuehlner or @mike-jumper Ping on this one, so I can either clean it up for 1.6.0, or de-scope it if we need to. I'd rather it make it into this release if at all possible...
> 
> I'm fine with the idea of offering an option to enable case-insensitive comparison, and just let it use the default case sensitiy behavior of the database otherwise. I think this would cover 95% of the use case of this feature.
> 
> I'm fine with including this in 1.6.0 if it's ready.

Okay, I think it's ready for review. The gist of the changes is:
* I've added case-insensitive comparisons across the board for the various extensions and guacamole-ext items.
* For PostgreSQL, specifically, I've added the capability to do case-insensitive queries for usernames.
* I've left the queries in MySQL and SQL Server alone - they are already case-insensitive, and I've avoided any work to try to make them behave in a case sensitive fashion due to the challenges mentioned above. There's also a warning that will be logged for the MySQL and SQL Server extensions when the `getCaseSensitiveUsernames()` method is called. I don't know if this is necessary or helpful, but I didn't want someone to try to throw "sql-server-case-sensitive-usernames: true" into their properties file and wonder why it doesn't work.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/JDBCEnvironment.java
```diff
@@ -271,5 +278,24 @@ public boolean shouldUseBatchExecutor() {
...
+     * @return
+     *     true if usernames should be treated as case-sensitive, or false if
+     *     usernames should be treated as case-insensitive.
+     * 
+     * @throws GuacamoleException 
+     *     If guacamole.properties cannot be parsed.
+     */
+    public boolean getCaseSensitiveUsernames() throws GuacamoleException {
+        
+        LOGGER.warn("This database extension does not support case-sensitive username comparisons.");
```
jmuehlner, line 295: I wonder if we could only warn when the connection is actually configured to use case-sensitive username comparisons. 

It's not clear that it's particularly useful to warn admins about something that they're not even trying to do.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/permission/UserPermissionMapper.java
```diff
@@ -19,7 +19,141 @@
...
+    
+    /**
+     * Deletes the given permissions from the database. If any permissions do
+     * not exist, they will be ignored.
+     *
+     * @param permissions
+     *     The permissions to delete.
+     *
+     * @param caseSensitive
+     *     Whether or not string comparisons will be done in a case-sensitive
```
jmuehlner, line 40: Specifically user name string comparisons, yeah? If so, this should be more clear.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/JDBCEnvironment.java
necouchman: Yeah, that's fair. I was trying to be too clever with a default implementation...

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/permission/UserPermissionMapper.java
necouchman: Yep.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/UserMapper.java
```diff
@@ -33,10 +34,65 @@ public interface UserMapper extends ModeledDirectoryObjectMapper<UserModel> {
      *
      * @param username
      *     The username of the user to return.
+     * 
+     * @param caseSensitive
```
jmuehlner, line 38: All these other mappers should also be updated to be more clear about what specifically is case sensitive.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/usergroup/UserGroupMemberUserMapper.java
```diff
@@ -19,10 +19,65 @@
...
 
 /**
  * Mapper for the one-to-many relationship between a user group and its user
  * members.
  */
-public interface UserGroupMemberUserMapper extends ObjectRelationMapper<UserGroupModel> {}
+public interface UserGroupMemberUserMapper extends ObjectRelationMapper<UserGroupModel> {
+
+    /**
+     * Inserts rows as necessary to establish the one-to-many relationship
```
jmuehlner, line 33: This stuff about `RelatedObjectSet` is pretty abstract. It's not even clear just from reading this file what it's even about. Can you rephrase this (and any others) in terms of the specific relationships actually being modified?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/conf/MySQLEnvironment.java
```diff
@@ -439,7 +439,26 @@ public boolean enforceAccessWindowsForActiveSessions() throws GuacamoleException
...
                 MySQLGuacamoleProperties.MYSQL_ENFORCE_ACCESS_WINDOWS_FOR_ACTIVE_SESSIONS,
-                true);
+                true
+        );
+    }
+    
+    @Override
+    public boolean getCaseSensitiveUsernames() throws GuacamoleException {
+        
+        // Get the configured or value for the property.
```
jmuehlner, line 449: configured or value?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/conf/PostgreSQLEnvironment.java
```diff
@@ -398,5 +398,17 @@ public boolean enforceAccessWindowsForActiveSessions() throws GuacamoleException
...
     }
+    
+    @Override
+    public boolean getCaseSensitiveUsernames() throws GuacamoleException {
+        
+        // By default, PostgreSQL does use case-sensitive string searches, so
+        // we will honor case-sensitive usernames.
+        return getProperty(
+                PostgreSQLGuacamoleProperties.POSTGRESQL_CASE_SENSITIVE_USERNAMES,
+                true
```
jmuehlner, line 409: Should this not default to false like it does elsewhere?

necouchman: Here's my thought process:
* In most, if not all, of the other cases (LDAP, OpenID, SAML, etc.), the back-end authentication system is already dealing with usernames in a case-insensitive fashion, so defaulting to "false" will not meaningfully change the behavior of the system.
* In the case of PostgreSQL, the current and default behavior, without this option, is a case-sensitive username comparison. I think that most people are okay with - and even want - case-insensitive username comparisons, storage, etc., but, in the interest of not changing things without the admin specifically making the decision, I left this as `true`, which is the way it behaves today.

I'm happy to go either way with this - I really do think that most people want case-insensitivity with usernames - but then we'll just need to make sure this is documented prominently in the release notes so that admins are as unsurprised by the change as possible.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/conf/MySQLEnvironment.java
necouchman: LoL. Yeah, that needs to be fixed.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/src/main/java/org/apache/guacamole/auth/postgresql/conf/PostgreSQLEnvironment.java
jmuehlner: That's a fair point yeah. I'm definitely a fan of not changing behavior unexpectedly when people upgrade. Backwards compatibility is a good thing to have. It might just be worth putting a comment here about why this is defaulting to true instead of false.

necouchman:
> > > @jmuehlner or @mike-jumper Ping on this one, so I can either clean it up for 1.6.0, or de-scope it if we need to. I'd rather it make it into this release if at all possible...
> > 
> > 
> > I'm fine with the idea of offering an option to enable case-insensitive comparison, and just let it use the default case sensitiy behavior of the database otherwise. I think this would cover 95% of the use case of this feature.
> > I'm fine with including this in 1.6.0 if it's ready.
> 
> Okay, I think it's ready for review. The gist of the changes is:
> 
> * I've added case-insensitive comparisons across the board for the various extensions and guacamole-ext items.
> * For PostgreSQL, specifically, I've added the capability to do case-insensitive queries for usernames.
> * I've left the queries in MySQL and SQL Server alone - they are already case-insensitive, and I've avoided any work to try to make them behave in a case sensitive fashion due to the challenges mentioned above. There's also a warning that will be logged for the MySQL and SQL Server extensions when the `getCaseSensitiveUsernames()` method is called. I don't know if this is necessary or helpful, but I didn't want someone to try to throw "sql-server-case-sensitive-usernames: true" into their properties file and wonder why it doesn't work.

Well, I changed my mind a bit and went a slightly different route that I believe is more consistent across the modules:
* I've implemented the case-insensitive searching across all the JDBC modules using the `LOWER()` method.
* MySQL and SQL Server modules still default to `false` for the case-sensitivity; PostgreSQL defaults to `true` with the additional comments regarding the reason for that.

My thought process (now) is that the modules will behave as they have been (principle of least surprise). However, should an admin, for some reason or another, want to change the database collation for either SQL Server or MySQL to one that does case-sensitive string comparisons, they need only enable it and it will work as expected (as PostgreSQL does). This also keeps the queries fairly consistent across each of the DB modules.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/conf/MySQLGuacamoleProperties.java
```diff
@@ -301,6 +301,14 @@ private MySQLGuacamoleProperties() {}
         @Override
         public String getName() { return "mysql-batch-size"; }
 
-    };    
+    };
+    
+    public static final BooleanGuacamoleProperty MYSQL_CASE_SENSITIVE_USERNAMES =
+            new BooleanGuacamoleProperty() {
```
mike-jumper, line 307: Please document.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledUser.java
```diff
@@ -780,5 +788,15 @@ public Permissions getEffectivePermissions() throws GuacamoleException {
...
     }
+    
+    @Override
+    public boolean isCaseSensitive() {
+        try {
+            return environment.getCaseSensitiveUsernames();
+        }
+        catch (GuacamoleException e) {
+            return true;
+        }
```
mike-jumper, line 799: Should the failure be logged here?

extensions/guacamole-auth-header/src/main/java/org/apache/guacamole/auth/header/user/AuthenticatedUser.java
```diff
@@ -58,6 +66,16 @@ public void init(String username, Credentials credentials) {
...
     }
 
+    @Override
+    public boolean isCaseSensitive() {
+        try {
+            return confService.getCaseSensitiveUsernames();
+        }
+        catch (GuacamoleException e) {
+            return false;
+        }
```
mike-jumper, line 76: Should the failure and its implications be logged here?

mike-jumper (changes requested):

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/AbstractUser.java
```diff
@@ -49,6 +49,11 @@ public String getPassword() {
     public void setPassword(String password) {
         this.password = password;
     }
+    
+    @Override
+    public boolean isCaseSensitive() {
+        return false;
+    }
```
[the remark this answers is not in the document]
necouchman: Yeah, that's a fair point.

necouchman:
Thanks, @mike-jumper - one question for both you and @jmuehlner: Following the principle of least surprise, should I just default all comparisons across usernames for all of the extensions to `true`, requiring admins to specifically take action in order to change the behavior? Or are some of the assumptions made about systems that are likely to be case-insensitive (like LDAP) safe? I'm leaning toward going back and defaulting to `true` (case-sensitive) across the board...

mike-jumper:
I agree that keeping things case-sensitive by default for now is probably the best option. It's the least likely to cause trouble for those that are expecting the established behavior, and it's easy to change for those than want case-insensitive.

Switching to case-insensitive by default for backends that are generally case-insensitive in practice (LDAP and some SSO providers) probably does make sense but the risk of surprise feels like it's worth being part of 2.0.0.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/JDBCEnvironment.java
```diff
@@ -30,7 +30,7 @@
  * intended for use within JDBC based authentication providers.
  */
 public abstract class JDBCEnvironment extends DelegatingEnvironment {
-
+
```
jmuehlner, line 33: Whitespace error?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/JDBCEnvironment.java
```diff
@@ -271,5 +271,29 @@ public boolean shouldUseBatchExecutor() {
         return true;
 
     }
+    
+    /**
+     * Returns a boolean value that indicates whether or not usernames should
+     * be treated as case-sensitive.
+     * 
+     * @return
+     *     true if usernames should be treated as case-sensitive, or false if
```
jmuehlner, line 280: I noticed that `@return` annotation in `ConfigurationService.java` capitalizes the initial "True", whereas this doesn't. I'm ok with either, but it we should probably try to be consistent.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/UserMapper.java
```diff
@@ -33,10 +34,82 @@ public interface UserMapper extends ModeledDirectoryObjectMapper<UserModel> {
...
+     *
+     * @return 
+     *     A Collection of all objects having the given identifiers.
+     */
+    Collection<UserModel> select(@Param("identifiers") Collection<String> identifiers,
+                                 @Param("caseSensitive") boolean caseSensitive);
+    
+    /**
+     * Selects all objects which have the given identifiers and are explicitly
+     * readably by the given user. If an identifier has no corresponding
```
jmuehlner, line 70: Typo "readably"

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/UserMapper.java
```diff
@@ -33,10 +34,82 @@ public interface UserMapper extends ModeledDirectoryObjectMapper<UserModel> {
...
      *
      * @return
      *     The user having the given username, or null if no such user exists.
      */
-    UserModel selectOne(@Param("username") String username);
+    UserModel selectOne(@Param("username") String username,
+                        @Param("caseSensitive") boolean caseSensitive);
+    
+    /**
+     * Selects all objects which have the given identifiers. If an identifier
```
jmuehlner, line 49: Aren't these just users? Why be vague and call them objects? Unless I'm misunderstanding something here.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-ssl/src/main/java/org/apache/guacamole/auth/ssl/conf/SSLEnvironment.java
```diff
@@ -0,0 +1,42 @@
...
+ * under the License.
+ */
+
+package org.apache.guacamole.auth.ssl.conf;
+
+import org.apache.guacamole.GuacamoleException;
+import org.apache.guacamole.auth.sso.conf.SSOEnvironment;
+
+/**
+ * An environment for retrieving SAML-related properties from the Guacamole
```
jmuehlner, line 26: Copypasta: "SAML"

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/JDBCEnvironment.java
necouchman: I think I've made this standard/consistent across all of the changes :-).

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-ssl/src/main/java/org/apache/guacamole/auth/ssl/conf/SSLEnvironment.java
necouchman: Should be fixed, now.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/UserMapper.java
necouchman: Okay, I think I've fixed this - it was copied from the class it used to inherit it from, so I just failed to change the wording.

necouchman:
Okay, I've gone through and tried to make sure comments are consistent. I think it's fixed up. I also believe I got the injection of the `SSOEnvironment` stuff correct in the SSO modules, but I still need to test, and it might be worth someone having a closer look at that.

I also had another thought - I've done all of this work on a per-module basis, creating configuration options, etc. Does it make sense, instead, to do something like:
* Global option for case-sensitivity across the entire Guacamole installation, instead of per-extension (`case-sensitive-usernames:`)?
* Make the check for case sensitivity part of the `Environment` class, so that anything/everything that has access to that class can pull the status.
* If username case-sensitivity is enabled, it is enabled across the board, for all the extensions; if it's disabled, it's disabled across the board for all extensions.

jmuehlner:
> Global option for case-sensitivity across the entire Guacamole installation, instead of per-extension (`case-sensitive-usernames:`)?

Eh, I'm not sure. Would there be situations potentially where somebody might want to configure the application to do case-insensitive comparisons for _some_ extensions, but not others? If so we wouldn't want to prevent them from configuring it that way. 

I could see how having a global option would be really handy, but maybe it could be overriden on a per-extension basis?

necouchman:
> > Global option for case-sensitivity across the entire Guacamole installation, instead of per-extension (`case-sensitive-usernames:`)?
> 
> Eh, I'm not sure. Would there be situations potentially where somebody might want to configure the application to do case-insensitive comparisons for _some_ extensions, but not others? If so we wouldn't want to prevent them from configuring it that way.
> 
> I could see how having a global option would be really handy, but maybe it could be overriden on a per-extension basis?

Yeah, that's why I was asking. I wasn't sure if we could come up with scenarios one way or the other - it seems like, most of the time, the request is going to be to make it behave one way or the other across the board, particularly between various extensions (JDBC + LDAP, for example). But always good to have the flexibility there to let folks do what they like with it. I'll give it a go - I have an idea, and I think it'll be reasonably easy to do...

necouchman:
Okay, I think I've got things fixed up. I have some testing to do, yet, but both global and per-extension options exist, now, with the per-extension options overriding the global option.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/conf/ConfigurationService.java
```diff
@@ -21,7 +21,6 @@
 
 import com.google.inject.Inject;
 import java.util.Collection;
-import java.util.List;
```
jmuehlner, line 24: We should probably avoid doing whitespace/import fixes for files not otherwise modified for this change, to keep the git history cleaner.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-cas/src/main/java/org/apache/guacamole/auth/cas/conf/ConfigurationService.java
```diff
@@ -26,10 +26,10 @@
 import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.GuacamoleServerException;
 import org.apache.guacamole.auth.cas.group.GroupFormat;
-import org.apache.guacamole.environment.Environment;
```
jmuehlner, line 29: This change should probably also be left out if there's no more substantive changes to this file.

extensions/guacamole-auth-quickconnect/src/main/java/org/apache/guacamole/auth/quickconnect/conf/ConfigurationService.java
necouchman: Sounds good. I'll just collect all the little fix-ups I'm accumulating and open a Jira issue for that.

jmuehlner:
> Okay, I think I've got things fixed up. I have some testing to do, yet, but both global and per-extension options exist, now, with the per-extension options overriding the global option.

Looks good to me! Any further concerns, @mike-jumper?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/JacksonLDAPConfiguration.java
```diff
@@ -203,6 +203,13 @@ public class JacksonLDAPConfiguration implements LDAPConfiguration {
      */
     @JsonProperty("member-attribute-type")
     private String memberAttributeType;
+    
+    /**
+     * The raw YAML value of {@link LDAPGuacamoleProperties#LDAP_USERNAMES_CASE_SENSITIVE}.
+     * If not set within the YAML, this will currently default to true.
+     */
+    @JsonProperty("ldap-usernames-case-sensitive")
+    private String caseSensitiveUsernames;
```
mike-jumper, line 212: These values do follow the names of the properties, but:

* Since the values within `ldap-servers.yaml` are always LDAP-related, the `ldap-` prefix used by the Guacamole properties is omitted.
* The Guacamole property that `LDAP_USERNAMES_CASE_SENSITIVE` points to is actually `ldap-case-sensitive-usernames`, not `ldap-usernames-case-sensitive`.

Since all the other properties use the `*-case-sensitive-usernames` convention, as well, I think:

* This should be renamed to `@JsonProperty("case-sensitive-usernames")`.
* The constant in `LDAPGuacamoleProperties` defining the property should be renamed to `LDAP_CASE_SENSITIVE_USERNAMES`.

mike-jumper (changes requested):
Looks pretty good to me. I only found one issue with property naming consistency that I think is worth addressing.

mike-jumper (approved):

PR #906, opened 2023-08-23
GUACAMOLE-1849: Fix a bug in Parser.js where receive would reparse instructions multiple times.

darren-foley-bravurasecurity:
This fixes instruction buffers from xhr-based tunnels being processed multiple times in some cases due to the `startIndex` getting reset in the parser.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Parser.js
+++ b/guacamole-common-js/src/main/webapp/modules/Parser.js
@@ -208,7 +208,7 @@ Guacamole.Parser = function Parser() {
                     // Immediately truncate buffer if its contents have been
                     // completely parsed, so that the next call to receive()
                     // need not append to the buffer unnecessarily
-                    if (elementEnd + 1 === buffer.length) {
+                    if (!isBuffer && elementEnd + 1 === buffer.length) {
                         elementEnd = -1;
                         buffer = '';
                     }


guacamole-common-js/src/main/webapp/modules/Parser.js
```diff
@@ -209,8 +209,14 @@ Guacamole.Parser = function Parser() {
...
-                        buffer = '';
+                        if (isBuffer) {
+                            startIndex = buffer.length;
+                            break;
+                        }
+                        else {
+                            elementEnd = -1;
+                            buffer = '';
+                        }
                     }
```
mike-jumper, line 220: Would it make more sense to just add `!isBuffer &&` to the overall condition here? I would think that the existing logic will already ensure `startIndex` is set correctly and that the loop breaks once the buffer is exhausted.

mike-jumper: Or perhaps I'm just not seeing under what conditions `startIndex` would be incorrectly set...

darren-foley-bravurasecurity: That seems reasonable. I'll test it!

darren-foley-bravurasecurity: I tested your suggestion: it was equally effective and much more concise :) The PR has been updated

mike-jumper (approved):

PR #907, opened 2023-08-29
GUACAMOLE-1851: Add Inject notation to RequestValidationService constructor to fix Guice loading.

necouchman:
Adds the `@Inject` notation to the constructor in `RequestValidationService` class so that it loads correctly with Guice, correcting an issue introduced by the addition of the parameterized construction from GUACAMOLE-1809.

### DIFF
--- a/extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/RequestValidationService.java
+++ b/extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/RequestValidationService.java
@@ -42,16 +42,18 @@ public class RequestValidationService {
      * Service for retrieving configuration information regarding the
      * JSONAuthenticationProvider.
      */
-    @Inject
     private ConfigurationService confService;
 
     /**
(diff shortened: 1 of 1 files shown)


extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/RequestValidationService.java
```diff
@@ -52,6 +52,7 @@ public class RequestValidationService {
      * @param confService
      *     The (mock) instance of ConfigurationService
      */
+    @Inject
```
mike-jumper, line 55: With this annotation present, does the `@Inject` on the `confService` member variable have any effect? Or are things now injected purely through invoking this constructor?

If only the constructor is used, then the docs for the constructor should probably be updated to capture its true usage, which is no longer specific to unit testing.

necouchman: Good point - I'll take a look at both those items and refactor.

PR #908, opened 2023-08-29
GUACAMOLE-1820: Allow newlines within key events.

jmuehlner:
This change ensures that newlines are rendered within individual key events.

This ensures that, for example, if you type several commands into a terminal, pressing enter after each command, these commands will render on subsequent lines in the key event viewer, rather than being compressed all into a single line.


guacamole-common-js/src/main/webapp/modules/KeyEventInterpreter.js
```diff
@@ -96,6 +96,21 @@ Guacamole.KeyEventInterpreter = function KeyEventInterpreter(batchSeperation, st
...
+         * directly, followed by the name being printed as if it was a
+         * non-printable or non-typed character.
+         *
+         * If the key was not directly typed, this flag will have no effect on
+         * behavior.
+         *
+         * @type {!boolean}
+         * @default false
+         */
+        this.printAndRender = template.printAndRender || false;
```
mike-jumper, line 112: I'm on the fence regarding this being something `Guacamole.KeyEventInterpreter` should represent and decide. My initial thinking would be that this should be an implementation decision.

Thoughts?

mike-jumper: Never mind - I see this is `@private` _and_ only ever feeds into the `simpleText` convenience value. Implementations are still free to do whatever they wish.

guacamole-common-js/src/main/webapp/modules/KeyEventInterpreter.js
```diff
@@ -317,6 +332,19 @@ Guacamole.KeyEventInterpreter = function KeyEventInterpreter(batchSeperation, st
...
+    /**
+     * Return a formatted, display-ready name for the provided key definition.
+     *
+     * @param {KeyDefinition} keyDefinition
+     *     A key definition to return the formatted name for.
+     *
+     * @returns
+     *     The formatted, display-ready name of the provided key definition.
+     */
+    function getPrintedKeyName(keyDefinition) {
```
mike-jumper, line 344: This should be annotated `@private`.

guacamole-common-js/src/main/webapp/modules/KeyEventInterpreter.js
```diff
@@ -424,6 +452,29 @@ Guacamole.KeyEventInterpreter = function KeyEventInterpreter(batchSeperation, st
...
+                    // Display the name first, followed by the value
+                    currentBatch.simpleValue += printedValue;
+                    currentBatch.simpleValue += renderedValue;
+
+                    // Add pair of key events in the same order, with the name
+                    // displayed as non-typed to ensure proper styling
+                    currentBatch.events.push(new Guacamole.KeyEventInterpreter.KeyEvent(
+                        printedValue, false, relativeTimestap));
+                    currentBatch.events.push(new Guacamole.KeyEventInterpreter.KeyEvent(
+                        renderedValue, true, relativeTimestap));
```
mike-jumper, line 473: I don't think individual events should possibly be split into multiple events purely for the sake of rendering, as it's not guacamole-common-js that should dictate that rendering.

I think we should instead settle on a structure that represents both the expected typed result (if any) of a logical event and the keys pressed to produce that result, leaving it up to the implementation to decide how that is ultimately rendered.

mike-jumper (changes requested):

guacamole-common-js/src/main/webapp/modules/KeyEventInterpreter.js
jmuehlner: So currently the `KeyEventIntepreter` is documented to "produce human readable text batches". 

It sounds like you're advocating removing all of the rendering logic out of this service (i.e. everything that produces the `text` or `simpleValue` fields, modifier/shortcut tracking, probably even batch separation). This would be probably all move to a service in the webapp. 

I think that separation of concerns makes a lot of sense, though I was unsure how much heavy handed I should be in changing the way the interpreter works.

jmuehlner: It doesn't look like we've released this code yet, so it seems like it should be ok.

jmuehlner: @mike-jumper How does this look?

guacamole-common-js/src/main/webapp/modules/KeyEventInterpreter.js
```diff
@@ -460,91 +253,85 @@ Guacamole.KeyEventInterpreter = function KeyEventInterpreter(batchSeperation, st
...
 
     /**
-     * All key events for this batch.
+     * The key definition for the pressed key.
      *
-     * @type {!Guacamole.KeyEventInterpreter.KeyEvent[]}
+     * @type {!Guacamole.KeyEventInterpreter.KeyDefinition}
      */
-    this.events = events || [];
+    this.definition = template.definition;
```
mike-jumper, line 320: `Guacamole.KeyEventInterpreter.KeyDefinition` is marked as `@private`, but this property is public.

guacamole-common-js/src/main/webapp/modules/KeyEventInterpreter.js
```diff
@@ -460,91 +253,85 @@ Guacamole.KeyEventInterpreter = function KeyEventInterpreter(batchSeperation, st
...
+ * and the timestamp when the event occured.
  *
  * @constructor
- * @param {!Guacamole.KeyEventInterpreter.KeyEvent[]} events
- *     The interpreted key events for this batch.
- *
- * @param {!String} simpleValue
- *     The simplified, human-readable value representing the key events for
- *     this batch.
+ * @param {Guacamole.KeyEventInterpreter.KeyDefinition|object} [template={}]
```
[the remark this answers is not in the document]
jmuehlner: Yes!

guacamole-common-js/src/main/webapp/modules/KeyEventInterpreter.js
jmuehlner: Ah, right you are. It shouldn't be private.

guacamole-common-js/src/main/webapp/modules/KeyEventInterpreter.js
```diff
@@ -460,91 +253,84 @@ Guacamole.KeyEventInterpreter = function KeyEventInterpreter(batchSeperation, st
...
  * @constructor
- * @param {!Guacamole.KeyEventInterpreter.KeyEvent[]} events
- *     The interpreted key events for this batch.
- *
- * @param {!String} simpleValue
- *     The simplified, human-readable value representing the key events for
- *     this batch.
+ * @param {Guacamole.KeyEventInterpreter.KeyEvent|object} [template={}]
+ *     The object whose properties should be copied within the new
+ *     KeyDefinition.
```
mike-jumper, line 307: KeyEvent*?

mike-jumper (changes requested):

guacamole-common-js/src/main/webapp/modules/KeyEventInterpreter.js
jmuehlner: Yes

guacamole-common-js/src/main/webapp/modules/KeyEventInterpreter.js
```diff
@@ -283,7 +180,8 @@ Guacamole.KeyEventInterpreter = function KeyEventInterpreter(batchSeperation, st
         var name = String.fromCharCode(codepoint);
 
         // Create and return the definition
-        return new KeyDefinition({keysym: keysym, name: name, value: name, modifier: false});
+        return new Guacamole.KeyEventInterpreter.KeyDefinition({
+                keysym: keysym, name: name, value: name, modifier: false});
```
mike-jumper, line 184: `Guacamole.KeyEventInterpreter.KeyDefinition` does not appear to have a `modifier` property that would be copied from this template.

jmuehlner: Oops, indeed it doesn't. Removed.

mike-jumper (approved):

guacamole-common-js/src/main/webapp/modules/KeyEventInterpreter.js
```diff
@@ -283,7 +179,8 @@ Guacamole.KeyEventInterpreter = function KeyEventInterpreter(batchSeperation, st
         var name = String.fromCharCode(codepoint);
 
         // Create and return the definition
-        return new KeyDefinition({keysym: keysym, name: name, value: name, modifier: false});
+        return new Guacamole.KeyEventInterpreter.KeyDefinition({
+                keysym: keysym, name: name, value: name);
```
mike-jumper, line 183:
```
#17 31.59 [INFO] BUILD FAILURE
#17 31.59 [INFO] ------------------------------------------------------------------------
#17 31.59 [INFO] Total time:  30.193 s
#17 31.59 [INFO] Finished at: 2023-09-27T17:49:53Z
#17 31.59 [INFO] ------------------------------------------------------------------------
#17 31.59 [ERROR] Failed to execute goal com.github.buckelieg:minify-maven-plugin:2.0.1:minify (default-minify) on project guacamole-common-js: java.lang.RuntimeException: JSCompiler errors
#17 31.59 [ERROR] all.js:8606: ERROR - [JSC_PARSE_ERROR] Parse error. '}' expected
#17 31.59 [ERROR]                 keysym: keysym, name: name, value: name);
#17 31.59 [ERROR]                                                        ^
#17 31.59 [ERROR] 
#17 31.59 [ERROR] -> [Help 1]
```

mike-jumper (changes requested):
Except for build failure, LGTM.

guacamole-common-js/src/main/webapp/modules/KeyEventInterpreter.js
jmuehlner: Oops, sorry. I think my IDE ate that closing brace. Should be fixed now.

mike-jumper: 😋

mike-jumper (approved):

PR #911, opened 2023-09-26
GUACAMOLE-1855: Allow MFA modules to be configured to bypass or enforce for specific hosts.

necouchman:
This pull request implements the feature proposed in Jira for GUACAMOLE-1855 - the Duo and TOTP modules can be configured to either be bypassed for users logging in from specific hosts, or can be configured to be explicitly required for users logging in from specific hosts, while other hosts are bypassed.

This allows situations where users connecting from the Internet ought to be required to complete MFA, but users logging in from behind a VPN or on a corporate network may not need to perform the extra authentication step.

### DIFF
--- a/extensions/guacamole-auth-duo/src/main/java/org/apache/guacamole/auth/duo/UserVerificationService.java
+++ b/extensions/guacamole-auth-duo/src/main/java/org/apache/guacamole/auth/duo/UserVerificationService.java
@@ -40,6 +40,7 @@ import org.apache.guacamole.language.TranslatableMessage;
 import org.apache.guacamole.net.auth.AuthenticatedUser;
 import org.apache.guacamole.net.auth.Credentials;
 import org.apache.guacamole.net.auth.credentials.CredentialsInfo;
+import org.apache.guacamole.properties.IPAddressListProperty;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -120,52 +121,30 @@ public class UserVerificationService {
         if (username == null || username.equals(AuthenticatedUser.ANONYMOUS_IDENTIFIER))
             return;
         
-        // We enforce by default
-        boolean enforceHost = true;
-        
-        // Check for a list of addresses that should be bypassed and iterate
+        // Pull address lists to check from configuration. Note that the enforce
+        // list will override the bypass list, which means that, if the client
+        // address happens to be in both lists, Duo MFA will be enforced.
         List<IPAddress> bypassAddresses = confService.getBypassHosts();
-        for (IPAddress bypassAddr : bypassAddresses) {
-
-            // If the address contains current client address, flip enforce flag
-            // and break out
-            if (clientAddr != null && clientAddr.isIPAddress()
-                    && bypassAddr.getIPVersion().equals(clientAddr.getIPVersion())
-                    && bypassAddr.contains(clientAddr)) {
-                enforceHost = false;
-                break;
-            }
-        }
-        
-        // Check for a list of addresses that should be enforced and iterate
         List<IPAddress> enforceAddresses = confService.getEnforceHosts();
+        
+        // Check if the bypass list contains the client address, and set the
+        // enforce flag to the opposite.
+        boolean enforceHost = !(IPAddressListProperty.addressListContains(bypassAddresses, clientAddr));
         
         // Only continue processing if the list is not empty
         if (!enforceAddresses.isEmpty()) {
             
             // If client address is not available or invalid, MFA will
             // be enforced.
-            if (clientAddr == null || !clientAddr.isIPAddress()) {
+            if (clientAddr == null || !clientAddr.isIPAddress())
                 enforceHost = true;
-            }
             
-            else {
-                // With addresses set, this default changes to false.
-                enforceHost = false;
-
-                for (IPAddress enforceAddr : enforceAddresses) {
-                    
-                    // If there's a match, flip the enforce flag and break out of the loop
-                    if (enforceAddr.getIPVersion().equals(clientAddr.getIPVersion())
-                            && enforceAddr.contains(clientAddr)) {
-                        enforceHost = true;
-                        break;
-                    }
-                }
-            }
+            // Check the enforce list for the client address and set enforcement flag.
+            else
+                enforceHost = IPAddressListProperty.addressListContains(enforceAddresses, clientAddr);
         }
             
-        // If the enforce flag has been changed, exit, bypassing Duo MFA.
+        // If the enforce flag is not true, bypass Duo MFA.
         if (!enforceHost)
             return;
 
--- a/extensions/guacamole-auth-duo/src/main/java/org/apache/guacamole/auth/duo/UserVerificationService.java
+++ b/extensions/guacamole-auth-duo/src/main/java/org/apache/guacamole/auth/duo/UserVerificationService.java
@@ -23,10 +23,13 @@ import com.duosecurity.Client;
 import com.duosecurity.exception.DuoException;
 import com.duosecurity.model.Token;
 import com.google.inject.Inject;
+import inet.ipaddr.IPAddress;
+import inet.ipaddr.IPAddressString;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Collections;
 import java.util.concurrent.TimeUnit;
+import java.util.List;
 import javax.servlet.http.HttpServletRequest;
 import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.GuacamoleServerException;
@@ -107,9 +110,63 @@ public class UserVerificationService {
     public void verifyAuthenticatedUser(AuthenticatedUser authenticatedUser)
             throws GuacamoleException {
 
-        // Ignore anonymous users (unverifiable)
+        // Pull the original HTTP request used to authenticate
+        Credentials credentials = authenticatedUser.getCredentials();
+        HttpServletRequest request = credentials.getRequest();
+        IPAddress clientAddr = new IPAddressString(request.getRemoteAddr()).getAddress();
+
+        // Ignore anonymous users
         String username = authenticatedUser.getIdentifier();
-        if (username.equals(AuthenticatedUser.ANONYMOUS_IDENTIFIER))
+        if (username == null || username.equals(AuthenticatedUser.ANONYMOUS_IDENTIFIER))
+            return;
+        
+        // We enforce by default
+        boolean enforceHost = true;
+        
+        // Check for a list of addresses that should be bypassed and iterate
+        List<IPAddress> bypassAddresses = confService.getBypassHosts();
+        for (IPAddress bypassAddr : bypassAddresses) {
+
+            // If the address contains current client address, flip enforce flag
+            // and break out
+            if (clientAddr != null && clientAddr.isIPAddress()
+                    && bypassAddr.getIPVersion().equals(clientAddr.getIPVersion())
+                    && bypassAddr.contains(clientAddr)) {
+                enforceHost = false;
+                break;
+            }
+        }
+        
+        // Check for a list of addresses that should be enforced and iterate
+        List<IPAddress> enforceAddresses = confService.getEnforceHosts();
+        
+        // Only continue processing if the list is not empty
+        if (!enforceAddresses.isEmpty()) {
+            
+            // If client address is not available or invalid, MFA will
+            // be enforced.
+            if (clientAddr == null || !clientAddr.isIPAddress()) {
+                enforceHost = true;
+            }
+            
+            else {
+                // With addresses set, this default changes to false.
+                enforceHost = false;
+
+                for (IPAddress enforceAddr : enforceAddresses) {
+                    
+                    // If there's a match, flip the enforce flag and break out of the loop
+                    if (enforceAddr.getIPVersion().equals(clientAddr.getIPVersion())
+                            && enforceAddr.contains(clientAddr)) {
+                        enforceHost = true;
+                        break;
+                    }
+                }
+            }
+        }
+            
+        // If the enforce flag has been changed, exit, bypassing Duo MFA.
+        if (!enforceHost)
             return;
 
         // Obtain a Duo client for redirecting the user to the Duo service and
@@ -136,11 +193,6 @@ public class UserVerificationService {
             throw new GuacamoleServerException("Duo authentication service is "
                     + "not currently available (failed health check).", e);
         }
-
-        // Pull the original HTTP request used to authenticate, as well as any
-        // associated credentials
-        Credentials credentials = authenticatedUser.getCredentials();
-        HttpServletRequest request = credentials.getRequest();
 
         // Retrieve signed Duo authentication code and session state from the
         // request (these will be absent if this is an initial authentication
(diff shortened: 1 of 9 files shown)


mike-jumper:
Looks pretty neat. What's the intended behavior for the case that an address matches subnets/addresses within both the enforce list and the bypass list?

necouchman:
> What's the intended behavior for the case that an address matches subnets/addresses within both the enforce list and the bypass list?

That's a good point - probably should give some more intentional thought to that. As it is currently written, as soon as an address is found in the bypass list, the user verification function will return, bypassing MFA authentication, and the enforcement list will never be processed.

I'm tempted to re-arrange things a bit where enforcement would actually "take precedence", and that a host in both lists would always be enforced. That seems like the more security-conscious way to go.

I'm open to suggestions, opinions, or discussion, though!

necouchman:
Also figured out that I missed the actual code for host checking on the TOTP side. Oops.

necouchman:
Okay, I've restructured things a bit such that having hosts listed in the enforce list _should_ override anything listed in the bypass list. I need to do some testing before it's ready, so I'll mark this as a draft for the moment - in the meantime if you see anything else amiss, please let me know.

mike-jumper:
Sounds good! And I agree regarding giving enforcement priority over bypass. Better for a mistake in configuration to result in excessive enforcement of MFA than unexpected bypassing of MFA.

necouchman:
Okay, I think I've got the checks working. I'm currently hitting three issues:
* Build is impacted by the odd Bouncycastle version update. Once that's fixed the build will work properly.
* If I try to use Tomcat with its default IPv4 + IPv6 dual-stack enabled, the `request.getRemoteAddr()` call periodically flips between the actual IPv4 client IP and an IPv6 loopback IP (0:0:0:0:0:0:0:0:1 - or some number of 0s). I was able to pretty well remove this by adding `-Djava.net.preferIPv4Stack=true` to `CATALINA_OPTS` in the Tomcat startup, but ideally this would be able to reliably handle both IPv4 and IPv6. If anyone has any hints for that, I'm open.
* I had to make some unusual changes to the IPAddress detection code because I couldn't loop through a `List` of `IPAddress` types - I'd get an error about how `IPv4Address` couldn't be cast to `IPAddress`. What's even more weird is that I converted everything to `IPAddressString` data types, and I can't even do a `for (IPAddressString ipAddr : ipAddrList)` - I get the same error about being unable to cast it. Again, if anyone has hints, I'm open to them.

necouchman:
Alrighty then - this should be ready for review, now. I've tested with the TOTP module and it appears to work as expected, at least with IPv4 addresses. I still have a bit of an issue with the IPv4 + IPv6 config and the client address flipping between a real IPv4 address and IPv6 loopback, but that seems to be more of a Nginx + Tomcat issue than it is anything to do with this.

mike-jumper:
> I still have a bit of an issue with the IPv4 + IPv6 config and the client address flipping between a real IPv4 address and IPv6 loopback, but that seems to be more of a Nginx + Tomcat issue than it is anything to do with this.

I've seen this before, as well - I think it's due to the local OS' networking stack occasionally choosing the IPv6 address for `localhost` instead of IPv4, but the default `RemoteIpValve` regex only matches IPv4 addresses.

necouchman:
> I've seen this before, as well - I think it's due to the local OS' networking stack occasionally choosing the IPv6 address for localhost instead of IPv4, but the default RemoteIpValve regex only matches IPv4 addresses.

Ah, okay - and, now that you mention it, it seems like something we've discussed before. Maybe I'll put in a Jira issue and update the manual page with that information.

extensions/guacamole-auth-duo/src/main/java/org/apache/guacamole/auth/duo/UserVerificationService.java
```diff
@@ -71,10 +74,60 @@ public void verifyAuthenticatedUser(AuthenticatedUser authenticatedUser)
...
         Credentials credentials = authenticatedUser.getCredentials();
         HttpServletRequest request = credentials.getRequest();
+        IPAddress clientAddr = new IPAddressString(request.getRemoteAddr()).getAddress();
 
         // Ignore anonymous users
         if (authenticatedUser.getIdentifier().equals(AuthenticatedUser.ANONYMOUS_IDENTIFIER))
             return;
+        
+        // We enforce by default
+        boolean enforceHost = true;
```
jmuehlner, line 84: Isn't this almost exactly the same code as in `UserVerificationService.java` below? Can you create a shared function?

jmuehlner: This would also be a nice place to explicitly document the behavior when an address is on both the bypass _and_ enforce lists.

necouchman: Sure, will try to address both concerns.

necouchman: > Isn't this almost exactly the same code as in `UserVerificationService.java` below? Can you create a shared function?

When you say "below", do you mean the code common to checking both enforce and bypass lists within this class, or were you talking about the code across the classes (shared between Duo and TOTP)? I can certainly create some classes within `guacamole-ext` that could be shared among these - and, honestly, probably other - use-cases.

jmuehlner: I was mostly talking about [this code block](https://github.com/apache/guacamole-client/pull/911/files#diff-5ce38f439ffddba700c3d8ca58d67bd7d1a2e377f6ac852783b1109df0107d8cR84-R130) from lines 84 to 130, and the similar code block in the TOTP `UserVerificationService.java` (sorry, that was unclear...)

If there's other similar code that you think should be shared, I'm definitely open to that as well.

necouchman: Okay - I took some of the common code and moved it into a utility class in `guacamole-ext`. I was trying to balance collecting common code against putting too much extension-specific code into the `guacamole-ext` code, but if you think there's more that should get consolidated into it, I'm happy to take another go at it.

guacamole-ext/src/main/java/org/apache/guacamole/net/util/IPAddressUtil.java
```diff
@@ -0,0 +1,62 @@
...
+                    
+            // If version matches and entry contains it, return true
+            if (ipEntry.getIPVersion().equals(ipAddr.getIPVersion())
+                    && ipEntry.contains(ipAddr))
+                return true;
+        
+        // No match, so return false
+        return false;
+        
+    }
```
mike-jumper, line 60: I think this would have a better life as either a static convenience function of `IPAddressListProperty`, or as a `contains()` instance method of `IPAddressListProperty` that pulls the list from itself automatically.

guacamole-ext/src/main/java/org/apache/guacamole/properties/IPAddressListProperty.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.regex.Pattern;
+import org.apache.guacamole.GuacamoleException;
+import org.apache.guacamole.GuacamoleServerException;
+
+/**
+ * A GuacamoleProperty implementation that parses a String for a comma-separated
+ * list of IP addresses and/or IP subnets, both IPv4 and IPv6, and returns the
+ * list of those valid IP addresses/subnets.
+ */
+public abstract class IPAddressListProperty implements GuacamoleProperty<List <IPAddress>> {
```
mike-jumper, line 37: Stray space here between `List` and `<IPAddress>`.

guacamole-ext/src/main/java/org/apache/guacamole/properties/IPAddressListProperty.java
```diff
@@ -0,0 +1,81 @@
...
+    /**
+     * A pattern which matches against the delimiters between values. This is
+     * currently simply a comma and any following whitespace. Parts of the
+     * input string which match this pattern will not be included in the parsed
+     * result.
+     */
+    private static final Pattern DELIMITER_PATTERN = Pattern.compile(",\\s*");
+    
+    @Override
+    public List<IPAddress> parseValue(String values) throws GuacamoleException {
```
mike-jumper, line 48: I'm on the fence about directly exposing an object from a third-party library as part of the extension API. Should we ever decide to switch to a different IP address library, this would encumber that switch.

Would Java's standard `InetAddress` work? I see `IPAddress` provides a `toInetAddress()`.

extensions/guacamole-auth-duo/src/main/java/org/apache/guacamole/auth/duo/UserVerificationService.java
```diff
@@ -71,10 +75,38 @@ public void verifyAuthenticatedUser(AuthenticatedUser authenticatedUser)
...
+        // address happens to be in both lists, Duo MFA will be enforced.
+        List<IPAddress> bypassAddresses = confService.getBypassHosts();
+        List<IPAddress> enforceAddresses = confService.getEnforceHosts();
+        
+        // Check if the bypass list contains the client address, and set the
+        // enforce flag to the opposite.
+        boolean enforceHost = !(IPAddressUtil.addressListContains(bypassAddresses, clientAddr));
+        
+        // Only continue processing if the list is not empty
+        if (enforceAddresses != null && !enforceAddresses.isEmpty()) {
```
mike-jumper, line 95: `getEnforceHosts()` cannot return `null`. If the property is omitted, `getEnforceHosts()` returns an empty list.

extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/user/UserVerificationService.java
```diff
@@ -288,6 +292,45 @@ private boolean totpDisabled(UserContext context,
...
+        // address happens to be in both lists, Duo MFA will be enforced.
+        List<IPAddress> bypassAddresses = confService.getBypassHosts();
+        List<IPAddress> enforceAddresses = confService.getEnforceHosts();
+        
+        // Check the bypass list for the client address, and set the enforce
+        // flag to the opposite.
+        boolean enforceHost = !(IPAddressUtil.addressListContains(bypassAddresses, clientAddr));
+        
+        // Only continue processing if the list is not empty
+        if (enforceAddresses != null && !enforceAddresses.isEmpty()) {
```
mike-jumper, line 317: Same here: `getEnforceHosts()` cannot return `null`.

mike-jumper (changes requested):

guacamole-ext/src/main/java/org/apache/guacamole/properties/IPAddressListProperty.java
necouchman: I think the main challenge with that approach is that the list, as implemented with the `IPAddress` class, is designed to take a combination of IPv4 addresses (1.2.3.4), IPv4 CIDR notation (1.2.3.0/24), IPv6 addresses, and IPv6 CIDR notation, and handle all of those possibilities in a single list. I haven't found anything conclusive, nor have I tested it, but I'm pretty sure the built-in `InetAddress` class can only handle IP addresses - possibly either IPv4 or IPv6 - but not CIDR notation.

guacamole-ext/src/main/java/org/apache/guacamole/net/util/IPAddressUtil.java
necouchman: Makes sense - went with the static method for now, since the `IPAddressListProperty` doesn't currently actually store the value - just processes the argument and returns the list.

guacamole-ext/src/main/java/org/apache/guacamole/properties/IPAddressListProperty.java
necouchman: Fixed via rebase.

extensions/guacamole-auth-duo/src/main/java/org/apache/guacamole/auth/duo/UserVerificationService.java
necouchman: Removed `null` check via rebase.

extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/user/UserVerificationService.java
necouchman: Removed `null` check via rebase.

guacamole-ext/src/main/java/org/apache/guacamole/properties/IPAddressListProperty.java
necouchman: @mike-jumper Any further thoughts on this one?

necouchman: @mike-jumper Ping...

WilliamHorner:
I was having a look at this pull request because this functionality will be very useful.  But I am hitting two compile errors:

[ERROR] ..../guacamole-client/extensions/guacamole-auth-duo/src/main/java/org/apache/guacamole/auth/duo/UserVerificationService.java:[178,20] error: variable credentials is already defined in method verifyAuthenticatedUser(AuthenticatedUser)
[ERROR] ..../guacamole-client/extensions/guacamole-auth-duo/src/main/java/org/apache/guacamole/auth/duo/UserVerificationService.java:[179,27] error: variable request is already defined in method verifyAuthenticatedUser(AuthenticatedUser)

Im not sure if this is my mishandling of git, but if I look at the file in the pull request I can see the two definitions there too. 
 I think it is probably just as simple as removing the second set of definitions? ... or fixing my understanding of git.

necouchman:
> I was having a look at this pull request because this functionality will be very useful. But I am hitting two compile errors:
> 
> [ERROR] ..../guacamole-client/extensions/guacamole-auth-duo/src/main/java/org/apache/guacamole/auth/duo/UserVerificationService.java:[178,20] error: variable credentials is already defined in method verifyAuthenticatedUser(AuthenticatedUser) [ERROR] ..../guacamole-client/extensions/guacamole-auth-duo/src/main/java/org/apache/guacamole/auth/duo/UserVerificationService.java:[179,27] error: variable request is already defined in method verifyAuthenticatedUser(AuthenticatedUser)
> 
> Im not sure if this is my mishandling of git, but if I look at the file in the pull request I can see the two definitions there too. I think it is probably just as simple as removing the second set of definitions? ... or fixing my understanding of git.

It should be fixed, now - apparently I didn't verify that it compiled after the last round of changes I made. Oops.

guacamole-ext/src/main/java/org/apache/guacamole/properties/IPAddressListProperty.java
mike-jumper: Yeah, I think you're right. It'd be nice if there were a standard class, but there really isn't anything outside `IPAddress` from this library, and that library seems good.

mike-jumper (approved):

PR #916, opened 2023-10-06
GUACAMOLE-1859: Update to latest compatible Java and JavaScript dependencies.

mike-jumper:
This change updates to the latest compatible versions of both Java and JavaScript dependencies. For JavaScript, this includes an update to WebPack 4.47.0, which required an update to newer Node.js and NPM, which required creation of a utility script to run the same version of NPM as the guac build and minor changes to `dependency-list-plugin.js` that ensure the `.../dist/` directory is automatically created.

There is a new version of the CAS library available, but it's not usable in our current build as it requires Java 17+. Updating to that would mean the CAS extension requires at least Java 17. That might be OK, but the extension subsystem would need some means of performing JVM version sanity checks prior to fully loading the extension. I'll take a look at doing that separately.

Licenses/notices of dependencies have been updated where necessary.


doc/licenses/zxing-3.5.2/AUTHORS
```diff
@@ -73,14 +73,14 @@ Kevin Xue (NetDragon Websoft Inc., China)
 Lachezar Dobrev
 Loránt Gerencsér
 Luiz Silva
-Luka Finžgar
-Łukasz Stefaniak
+Luka Fin?gar
+?ukasz Stefaniak
```
jmuehlner, line 77: It looks like some characters got messed up and replaced with `?` here - is that correct?

mike-jumper: Ah, nope - that's from me copy/pasting the `AUTHORS` file through the meat grinder that is ISO 8859-1 (VNC clipboard). I'll correct this...

mike-jumper:
**NOTE:** There are merge conflicts against `master` if/when these changes are merged back. The conflicts are against `package.json` and `package-lock.json` because of the additional `fuzzysort` dependency on `master`.

The conflicts look mercifully minor. Combining the `fuzzysort` lines with what's already present from these updates resolves the conflict and looks to build without issue.

doc/licenses/zxing-3.5.2/AUTHORS
mike-jumper: Fixed via rebase.

jmuehlner (approved):
LGTM

PR #921, opened 2023-10-19
GUACAMOLE-1868: Add support for new PAM Hostname field.

jmuehlner:


extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmRecordService.java
```diff
@@ -411,6 +412,11 @@ public String getHostname(KeeperRecord record) {
         if (hostsField != null)
             return getSingleStringValue(hostsField.getValue(), Host::getHostName);
 
+        // Next, try a PAM hostname
+        PamHostnames pamHostsField = getField(record, PamHostnames.class, null);
+        if (pamHostsField != null)
+            return getSingleStringValue(pamHostsField.getValue(), Host::getHostName);
```
necouchman, line 418: Is it intended that the present of the "hostname" field will always override this new "PAM Hostname" field?

If so, should the function documentation be updated to both include the alternative field label that is being added, here, and also document that order of preference?

jmuehlner: Sorry, yeah, this change should definitely be documented.

In practice, I wouldn't expect to see both of those fields on the same record - the `PamHostnames` field type only appears in new "PAM Machine" record types, which do not support the `Hosts` field type.

necouchman (approved):

PR #922, opened 2023-10-23
GUACAMOLE-1872: Show the recording playback controls when the mouse is moved.

jmuehlner:


guacamole/src/main/frontend/src/app/player/directives/player.js
```diff
@@ -179,6 +199,12 @@ angular.module('player').directive('guacPlayer', ['$injector', function guacPlay
          */
         var resumeAfterSeekRequest = false;
 
+        /**
+         * A scheduled timer to clean up the mouse activity CSS class, or null
+         * if no timer is scheduled.
+         */
+        var mouseActivityTimer = null;
```
mike-jumper, line 206: I don't see anything assigning a value to this variable. Is this used?

jmuehlner: Oops, yeah, I lost the assignment when I refactored. let me fix that.

jmuehlner: Oops, yeah, I lost the assignment when I refactored. let me fix that.

jmuehlner: Fixed.

guacamole/src/main/frontend/src/app/player/directives/player.js
```diff
@@ -179,6 +199,12 @@ angular.module('player').directive('guacPlayer', ['$injector', function guacPlay
          */
         var resumeAfterSeekRequest = false;
 
+        /**
+         * A scheduled timer to clean up the mouse activity CSS class, or null
+         * if no timer is scheduled.
+         */
```
mike-jumper, line 205: Please add `@type` annotation.

guacamole/src/main/frontend/src/app/player/directives/player.js
```diff
@@ -77,6 +79,24 @@
...
+     * The number of milliseconds after the last detected mouse activity before
+     * the cleanup timer to remove the associated CSS class should be scheduled.
+     */
+    const MOUSE_DEBOUNCE_DELAY = 250;
+
+    /**
+     * The number of milliseconds, after the debounce delay, before the mouse
+     * activity cleanup timer should run.
+     */
+    const MOUSE_CLEANUP_TIMER_DELAY = MOUSE_CLEANUP_DELAY - MOUSE_DEBOUNCE_DELAY;
```
mike-jumper, line 98: Please add `@type` annotations.

guacamole/src/main/frontend/src/app/player/directives/player.js
```diff
@@ -401,6 +427,43 @@ angular.module('player').directive('guacPlayer', ['$injector', function guacPlay
...
+                mouseActivityTimer = null;
+                $element.removeClass('recent-mouse-movement');
+            }, MOUSE_CLEANUP_TIMER_DELAY),
+
+            /*
+             * Only schedule the cleanup task after the mouse hasn't moved
+             * for a reasonable amount of time to ensure that the number of
+             * created cleanup timers remains reasonable.
+             */
+            MOUSE_DEBOUNCE_DELAY);
```
mike-jumper, line 448: > ... to ensure that the number of created cleanup timers remains reasonable.

Is this the case? I see `mouseActivityTimer` as tracking the most recent scheduled cleanup, and that it's cleared prior to scheduling any future cleanup. It looks like there will always be at most one registered timeout for cleanup.

mike-jumper (changes requested):
Just need some `@type`s. Otherwise, LGTM (assuming `debounce()` is indeed necessary).

guacamole/src/main/frontend/src/app/player/directives/player.js
jmuehlner: Yeah, there will be at most once, I was referring more to the number of timers that are scheduled while the mouse is moving. I don't want to be constantly cancelling and rescheduling new timers on every mouse movement. In practice it doesn't seem to cause excessive CPU / memory usage or anything, just seems wasteful.

mike-jumper:
@jmuehlner Think this should be against `staging/1.5.4`?

jmuehlner:
> @jmuehlner Think this should be against `staging/1.5.4`?
Suuuure - it's not quite a bug exactly but I think it makes sense to include. I'll rebase.

jmuehlner:
This change doesn't quite work, due to a different bug that was fixed in master, in https://github.com/apache/guacamole-client/commit/2a2cef918945408da39372d8e3f2499e51528017. I'll need to port just the fix for the control bar hiding to `1.5.4` before this change is ready to go.

jmuehlner:
depends on https://github.com/apache/guacamole-client/pull/923

mike-jumper (approved):

PR #926, opened 2023-11-01
Fixed the problem of incorrect remote host IP in connection history a…

LemonZuo:
Unable to obtain real remote host IP after reverse proxy,If the proxy request header exists, the real IP will be obtained from the request header, otherwise the IP will be obtained through the original method.

### DIFF
--- a/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Credentials.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Credentials.java
@@ -19,6 +19,8 @@
 
 package org.apache.guacamole.net.auth;
 
+import org.apache.guacamole.util.IpUtils;
+
 import java.io.Serializable;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpSession;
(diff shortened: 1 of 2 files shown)


mike-jumper (changes requested):
We cannot use this approach, as it would allow for the source address of a request to be trivially forged in any case where `X-Forwarded-For` is not guaranteed to be set by a trusted service, such as any case where no trusted reverse proxy is involved at all.

This is the main reason no such parsing already exists in the `Credentials` class, and the reason we only perform this kind of parsing when logging things for informational purposes. _The only way to rely on `X-Forwarded-For` is to only rely on it if the administrator explicitly declares that it can be trusted._

In the case of Guacamole (and other web applications running beneath Tomcat), this is done through declaring to Tomcat that you want to trust `X-Forwarded-For` from specific addresses using `RemoteIpValve` in your `server.xml`.

mike-jumper:
For this change to move forward with providing the quality-of-life improvements intended (without introducing a security issue), the change would need to:

1. Effectively re-implement what Tomcat already provides with `RemoteIpValve` via more convenient configuration properties, ensuring the admin has full control over how and whether `X-Forwarded-For` is used.
2. Use a servlet filter or similar to guarantee that `getRemoteAddr()` returns the expected address when the HTTP request is directly inspected (third-party libraries like those used for SAML will otherwise not be affected by the additional configuration and would still not see the expected address without `RemoteIpValve`).

@LemonZuo If you want to give that a shot, the change would need to be made to the webapp, not guacamole-ext. If done correctly, `Credentials` and anything else using standard functions like `getRemoteAddr()` would see the expected address.

PR #927, opened 2023-11-02
GUACAMOLE-1876: Display points of interest heatmap in history recording player.

jmuehlner:
This change adds heatmaps to the in-browser history recording player, one for frames rendered, and one for key events (if the recording has any - otherwise only the frames heatmap will display).

image


mike-jumper (changes requested):

jmuehlner:
OK @mike-jumper all your comments should be addressed now.

mike-jumper (approved):

PR #929, opened 2023-11-13
GUACAMOLE-1872: Hide session recording playback controls even when paused.

jmuehlner:
This change ensures that the session recording player controls can be hidden, even when the playback is paused. The controls will always be shown when the control bar is hovered over.

This is to allow admins to look at the bottom of the screen (which can contain import things like a task bar) when the recording is paused.

### DIFF
--- a/guacamole/src/main/frontend/src/app/settings/styles/history-player.css
+++ b/guacamole/src/main/frontend/src/app/settings/styles/history-player.css
@@ -96,7 +96,7 @@
     background: rgba(0, 0, 0, 0.5);
 }
 
-.settings.connectionHistoryPlayer .guac-player-controls.playing {
+.settings.connectionHistoryPlayer .guac-player-controls {
     opacity: 0;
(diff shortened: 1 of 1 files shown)


guacamole/src/main/frontend/src/app/settings/styles/history-player.css
```diff
@@ -96,23 +96,15 @@
...
     -webkit-transition: opacity 0.25s linear 0.25s;
     -moz-transition: opacity 0.25s linear 0.25s;
     -o-transition: opacity 0.25s linear 0.25s;
     transition: opacity 0.25s linear 0.25s;
 }
 
-.settings.connectionHistoryPlayer .guac-player-controls.paused,
-.settings.connectionHistoryPlayer .guac-player-controls.playing:hover {
-    opacity: 1;
-    -webkit-transition-delay: 0s;
```
jmuehlner, line 110: The default value for these properties should be 0 anyway, so I don't think we need these.

mike-jumper (approved):

PR #931, opened 2023-11-15
GUACAMOLE-1881: Adding new standard token LDAP_DOMAIN by extracting from user credentials

josnabattula:
When user log-in into Guacamole when multiple LDAP's enabled with one of the configuration in ldap-server.yml
` match-usernames:
    - DOMAIN1\\(.*)
    - (.*)@domain1\.example\.net
    ` 
In this pull request we are extracting domain name into custom variable `LDAP_DOMAIN`. this can be used for connections

### DIFF
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
@@ -28,6 +28,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import org.apache.directory.api.ldap.model.entry.Attribute;
 import org.apache.directory.api.ldap.model.entry.Entry;
 import org.apache.directory.api.ldap.model.exception.LdapException;
@@ -64,6 +66,11 @@ public class AuthenticationProviderService {
      * The prefix that will be used when generating tokens.
      */
     public static final String LDAP_ATTRIBUTE_TOKEN_PREFIX = "LDAP_";
+
+    /**
+     * The name of LDAP domain attribute.
+     */
+    public static final String LDAP_DOMAIN_TOKEN = "DOMAIN";
 
     /**
      * Service for creating and managing connections to LDAP servers.
@@ -286,18 +293,18 @@ public class AuthenticationProviderService {
                     CredentialsInfo.USERNAME_PASSWORD);
 
         try {
-        
+
             // Retrieve group membership of the user that just authenticated
             Set<String> effectiveGroups =
                     userGroupService.getParentUserGroupIdentifiers(config, config.getBindDN());
 
             // Return AuthenticatedUser if bind succeeds
             LDAPAuthenticatedUser authenticatedUser = authenticatedUserProvider.get();
+
             authenticatedUser.init(config, credentials,
-                    getAttributeTokens(config), effectiveGroups);
+                    getUserTokens(config, credentials), effectiveGroups);
 
             return authenticatedUser;
-
         }
 
         catch (GuacamoleException | RuntimeException | Error e) {
@@ -308,25 +315,53 @@ public class AuthenticationProviderService {
     }
 
     /**
-     * Returns parameter tokens generated from LDAP attributes on the user
-     * currently bound under the given LDAP connection. The attributes to be
-     * converted into parameter tokens must be explicitly listed in
-     * guacamole.properties. If no attributes are specified or none are
-     * found on the LDAP user object, an empty map is returned.
+     * Returns the current LDAP domain token from the provided user credentials.
+     *
+     * @param credentials
+     *     The credentials used for authentication.
+     *
+     * @return
+     *     Domain name by splitting login username or null if no domain is detected.
+     */
+    private String getDomainToken(Credentials credentials) {
+        String username = credentials.getUsername();
+        //Regex is used to extract the domain from a username
+        //that is in either of these formats: DOMAIN\\username or username@domain.
+        Pattern pattern = Pattern.compile("^(.+)\\\\.*$|^.*@(.+)$");
+        Matcher matcher = pattern.matcher(username);
+        if (matcher.find()) {
+            return matcher.group(1) != null ? matcher.group(1) : matcher.group(2);
+        }
+        return null;
+    }
+
+    /**
+     * Returns parameter tokens generated based on details specific to the user
+     * currently bound under the given LDAP connection. Both LDAP attributes on
+     * the user's associated LDAP object and the credentials submitted by the user
+     * to Guacamole are considered. If any tokens are to be derived from LDAP
+     * attributes, those attributes must be explicitly listed in
+     * guacamole.properties. If no tokens are applicable, an empty map is returned.
      *
      * @param config
      *     The configuration of the LDAP server being queried.
      *
+     * @param credentials
+     *     The credentials to use for authentication.
+     *
      * @return
-     *     A map of parameter tokens generated from attributes on the user
-     *     currently bound under the given LDAP connection, as a map of token
-     *     name to corresponding value, or an empty map if no attributes are
-     *     specified or none are found on the user object.
+     *     A map of parameter tokens. These tokens are generated based on
+     *     the attributes of the user currently bound under the given LDAP connection
+     *     and the user's credentials. The map's keys are the canonicalized attribute
+     *     names with an "LDAP_" prefix, and the values are the corresponding attribute
+     *     values. If the domain name is extracted from the user's credentials, it is added
+     *     to the map with the key "LDAP_DOMAIN". If no applicable tokens are found,
+     *     the method returns an empty map.
      *
      * @throws GuacamoleException
      *     If an error occurs retrieving the user DN or the attributes.
      */
-    private Map<String, String> getAttributeTokens(ConnectedLDAPConfiguration config)
+    private Map<String, String> getUserTokens(ConnectedLDAPConfiguration config, Credentials credentials)
             throws GuacamoleException {
 
         // Get attributes from configuration information
@@ -356,10 +391,16 @@ public class AuthenticationProviderService {
                 tokens.put(TokenName.canonicalize(attr.getId(),
                         LDAP_ATTRIBUTE_TOKEN_PREFIX), attr.getString());
             }
-
         }
         catch (LdapException e) {
             throw new GuacamoleServerException("Could not query LDAP user attributes.", e);
+        }
+
+        // Extracting the domain name from the user's credentials
+        String domainName = getDomainToken(credentials);
+        if (domainName != null) {
+            String tokenName = TokenName.canonicalize(LDAP_DOMAIN_TOKEN, LDAP_ATTRIBUTE_TOKEN_PREFIX);
+            tokens.put(tokenName, domainName);
         }
 
         return tokens;


guacamole/src/main/java/org/apache/guacamole/tunnel/StandardTokenMap.java
```diff
@@ -102,6 +115,13 @@ public StandardTokenMap(AuthenticatedUser authenticatedUser) {
         else
             put(USERNAME_TOKEN, authenticatedUser.getIdentifier());
 
+        if (get(USERNAME_TOKEN).contains("\\")) {
+            put(USERNAME_DOMAIN_TOKEN, get(USERNAME_TOKEN).split("\\\\")[0]);
+            put(USERNAME_ID_TOKEN, get(USERNAME_TOKEN).split("\\\\")[1]);
+        } else {
+            put(USERNAME_DOMAIN_TOKEN, "");
+            put(USERNAME_ID_TOKEN, get(USERNAME_TOKEN));
+        }
```
mike-jumper, line 124: I don't think decisions should be made about the format of the username at the web application level, nor that we should make specific assumptions about the format here when the LDAP handling of username format is so flexible.

What about exposing `authenticatedUser.getIdentifier()` as something like `GUAC_IDENTIFIER`? That would allow the unique identifier that the LDAP support is already parsing out of the username to be exposed automatically without forcing the webapp to be aware of LDAP-specific details.

If it is also necessary to pull the domain of a user authenticating with LDAP and make that available with a token, it would make sense to do that within the LDAP support, and it may be possible to extend the way `match-usernames` works to provide that.

mike-jumper (changes requested):

guacamole/src/main/java/org/apache/guacamole/tunnel/StandardTokenMap.java
necouchman: > I don't think decisions should be made about the format of the username at the web application level...

I agree, and my question would be: Why do these need to be standard tokens at all? We've provided a mechanism within the LDAP module (and most of the other modules, particularly SSO) to define tokens specific to those modules, but that should be usable in connections stored in other modules. Is there any reason not to just do something like:
* LDAP_USER_REALM or LDAP_USER_DOMAIN
* LDAP_USER_IDENTIFIER, or just pull the LDAP attribute with the username?

? Or is there some other piece of this I'm missing?

mike-jumper: If there is to be a new standard token, I think there's some value in having one token (`GUAC_IDENTIFIER`? `GUAC_USER_IDENTIFIER`?) that represents validated identity while the other (`GUAC_USERNAME`) represents the username provided during the auth process, if any.

I'd also have no issue if this were 100% purely LDAP-specific tokens.

> ... or just pull the LDAP attribute with the username?

I think that much is already possible with the functionality provided by the `ldap-user-attributes` property.

necouchman: > If there is to be a new standard token, I think there's some value in having one token (GUAC_IDENTIFIER? GUAC_USER_IDENTIFIER?) that represents validated identity while the other (GUAC_USERNAME) represents the username provided during the auth process, if any.

Sure, that makes sense - it doesn't resolve the issue of getting the domain portion out of the overall username, so if there's a need to have the domain, that would still have to be handled - which seems more appropriate to be LDAP-specific.

> I think that much is already possible with the functionality provided by the ldap-user-attributes property.

Yeah, I was looking at that code, but also did a quick search of my local AD environment and figured out that the domain is not (obviously) present in any attribute without splitting it out of another attribute. So, again, if the domain portion is desired/required, it would need to be something LDAP-specific.

If the desire is just to get the Guacamole identifier, I'm good with the `GUAC_IDENTIFIER` route.

josnabattula: `if the domain portion is desired/required, it would need to be something LDAP-specific` - Yes these variables we are looking are LDAP-specific. `authenticatedUser.getIdentifier()` only gives us which authentication mechanism i have used for my guacamole instance. 
Also In the instance multiple LDAP configurations with `match-usernames` `${GUAC_USERNAME` becomes `domain\username` which is not ideal for RDP connections.  Check below some parts error 
```
    Failure Reason:     Unknown user name or bad password.
    Status:         0xC000006D
    Sub Status:     0xC0000064
 
Process Information:
    Caller Process ID:  0x0
    Caller Process Name:    -

Detailed Authentication Information:
    Logon Process:      NtLmSsp
    Authentication Package: NTLM
    Transited Services: -
    Package Name (NTLM only):   -
    Key Length:     0
```

It would be ideal to have DOMAIN and just user name which are again LDAP sepcific, it makes easy for connection configuration without any hassle.

mike-jumper: > `authenticatedUser.getIdentifier()` only gives us which authentication mechanism i have used for my guacamole instance.

What do you mean? The `AuthenticatedUser` is the object defining the user's authenticated identity, with `getIdentifier()` returning the unique identifier for that identity. In the case of LDAP, that is the username. This is exactly the value that is returned as the username to be exposed in the UI:

https://github.com/apache/guacamole-client/blob/284d1b24f37f06d380312e149d24d96497dd9af4/guacamole/src/main/java/org/apache/guacamole/rest/auth/TokenRESTService.java#L188C1-L194

Providing the identifier as a token would allow the relevant username to be included in the configuration independently of what the user enters in a login form.

fanovilla: You're right @necouchman , main intent is to expose a token (agree should be LDAP specific) containing only the domain (and another containing only the username) that can be used in connection definitions.

> We've provided a mechanism within the LDAP module (and most of the other modules, particularly SSO) to define tokens specific to those modules, but that should be usable in connections stored in other modules.

To do that, I presume we can add tokens to what's already being collected here?https://github.com/apache/guacamole-client/blob/284d1b24f37f06d380312e149d24d96497dd9af4/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java#L355-L358

mike-jumper: @fanovilla Yes, that function (`getAttributeTokens()`) is a reasonable place to add LDAP-specific tokens. If you do so, the name and documentation for that function would need to be adjusted, as it currently states:

> Returns parameter tokens generated from LDAP attributes on the user currently bound under the given LDAP connection. ...

If a standard token for the user identifier is being added, as well, `StandardTokenMap` is correct for that.

fanovilla: thanks @mike-jumper ; @josnabattula not sure how but let me know if you want me to try contribute to your PR branch..

josnabattula: Yep, you can.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -306,6 +311,26 @@ public LDAPAuthenticatedUser authenticateUser(Credentials credentials)
...
+     *     The credentials to use for authentication.
+     *
+     * @return
+     *     Domain name by splitting logged username when multiple LDAP configuration is available
+     *     or null if no such configuration
+     */
+    private String getDomainToken(Credentials credentials) {
+       String ldapDomainName = null;
+        // Creating custom LDAP attribute token - domain name - when configured to multiple LDAP
+        if (credentials.getUsername().contains("\\")) {
```
necouchman, line 329: I think we should probably deal with situations where the username is in the UPN format, which is `USERNAME`@`DOMAIN`?

josnabattula:
@mike-jumper @necouchman could you look into my pull request, as I have move the changes into LDAP module as per the comments.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -356,6 +381,11 @@ private Map<String, String> getAttributeTokens(ConnectedLDAPConfiguration config
                 tokens.put(TokenName.canonicalize(attr.getId(),
                         LDAP_ATTRIBUTE_TOKEN_PREFIX), attr.getString());
             }
+            String domainName = getDomainToken(credentials);
+            if (domainName != null) {
+                String tokenName = TokenName.canonicalize("domain_name", LDAP_ATTRIBUTE_TOKEN_PREFIX);
```
necouchman, line 386: `domain_name` should probably get a constant.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -306,17 +306,42 @@ public LDAPAuthenticatedUser authenticateUser(Credentials credentials)
...
+     *
+     * @return
+     *     Domain name by splitting logged username(domain/username) when multiple LDAP configuration is available
+     *     or null if no such configuration
+     */
+    private String getDomainToken(Credentials credentials) {
+        String ldapDomainName = null;
+        if (credentials.getUsername().contains("\\")) {
+            ldapDomainName = credentials.getUsername().split("\\\\")[0];
+        } else if (credentials.getUsername().contains("@")) {
```
necouchman, line 325: Please follow established style guidelines - in particular, don't "cuddle" the tags.

necouchman (changes requested):
A couple of more tweaks and I think it'll be ready to go.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
josnabattula: @necouchman Could you give us more context, couldn't figure out what your comments is. thanks.

necouchman: @josnabattula Sure:
`} else if (something else) {`

This is known as "cuddling" tags. Code style for this project is:
```
}
else if (something else) {
```

Where the `}` is not on the same line as the `else if` statement.

josnabattula: Thanks for response, figured out :)

josnabattula:
@necouchman  I have resolved your comments.

josnabattula:
@necouchman @mike-jumper I have resolved all the comments. Can I get your tick?

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -306,17 +311,43 @@ public LDAPAuthenticatedUser authenticateUser(Credentials credentials)
         }
 
     }
+    /**
+     * Returns parameter current ldap domain token generated from user credentials
+     * If no multiple LDAP are configured on GUACAMOLE_HOME such ldap-servers.yaml,
+     * a null is returned.
```
necouchman, line 317: Is this actually the way it works - if multiple LDAP servers/domains are not enabled, will it _always_ return `null`? Or are there situations where a single LDAP configuration could still result in a value being returned by this function - such as if the `userPrincipalName` attribute is used when talking to Active Directory?

josnabattula: Yep, I have updated code comments in code to be more clear as it works when there are situations where single LDAP configuration could return domain name if exists otherwise null.

josnabattula:
@necouchman I have resolved your comments. please have a review. thanks.

josnabattula:
@necouchman if you have a minute, could you have look over this again, resolved all comments. Good to get this merged, waiting for this feature for a wee bit of time.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -307,16 +312,39 @@ public LDAPAuthenticatedUser authenticateUser(Credentials credentials)
...
+    private String getDomainToken(Credentials credentials) {
+        String ldapDomainName = null;
+        if (credentials.getUsername().contains("\\")) {
+            ldapDomainName = credentials.getUsername().split("\\\\")[0];
+        }
+        else if (credentials.getUsername().contains("@")) {
+            ldapDomainName = credentials.getUsername().split("@")[1];
+        }
+        return ldapDomainName;
+    }
```
mike-jumper, line 332: Hearkening back to the initial review, I'm still on the fence about this for the same reason:

> I don't think ... that we should make specific assumptions about the format here when the LDAP handling of username format is so flexible.

That said, I do see the utility of having this sort of token just be available, given that this sort of matching is going to be so common.

If we're going to have a `${LDAP_DOMAIN}` token that is always defined for any successful login that follows the `DOMAIN\username` or `username@domain` formats, I think we should do this parsing a bit more efficiently. Currently, the algorithm here is:

1. Search the username string for the first occurrence of a `\` character.
2. If such a character is found, search the username string again, this time splitting the string into an array. Discard everything but the first element in that array.
3. Search the username string yet again, this time for the first occurrence of a `@` character.
4. If such a character is found, search the username string _again_, this time splitting the string into an array. Discard everything but the first element in that array.

This is suboptimal, as:

* It potentially searches the username string 4 times.
* It unnecessarily splits the string into an array, rather than just parsing the interesting part  of the string directly.

Instead, I'd recommend defining and using a `Pattern` that matches _either_ of these formats, extracting the domain. This would reduce things down to a single operation that performs only one search and directly extracts the domain at the same time.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -326,7 +354,7 @@ public LDAPAuthenticatedUser authenticateUser(Credentials credentials)
      * @throws GuacamoleException
      *     If an error occurs retrieving the user DN or the attributes.
      */
-    private Map<String, String> getAttributeTokens(ConnectedLDAPConfiguration config)
+    private Map<String, String> getAttributeTokens(ConnectedLDAPConfiguration config, Credentials credentials)
```
mike-jumper, line 357: This function is specifically documented as:

> Returns parameter tokens generated from LDAP attributes on the user currently bound under the given LDAP connection.

and is specifically named `getAttributeTokens()`. So long as that's the case, it's bad practice to let injection of an entirely different kind of token be a side-effect of this function.

I do see that the documentation here has been updated to clarify this a bit, however that ultimately resulted in the following statement:

> The attributes to be converted into parameter tokens must be explicitly listed in guacamole.properties or domain name of the LDAP connection when multiple auth enabled.

As-written, this does not make sense to me, and the original issues with primary documented scope and the naming of the function remain. If this function is to serve as the basis for injection of the `LDAP_DOMAINS` token (which I agree may make sense), then:

* The function should be renamed to capture its now-broadened scope (`getAttributeTokens()` is no longer correct).
* The function's documentation should be updated to capture that scope ("Returns parameter tokens generated from LDAP attributes" is no longer correct, and other documented aspects may also need adjusting).

Makes sense?

For example, if the function were renamed to something like `getUserTokens()`, the following might be reasonable:

> Returns parameter tokens generated based on details specific to the user currently bound under the given LDAP connection. Both LDAP attributes on the user's associated LDAP object and the credentials submitted by the user to Guacamole are considered. If any tokens are to be derived from LDAP attributes, those attributes must be explicitly listed in guacamole.properties. If no tokens are applicable, an empty map is returned.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -307,16 +312,39 @@ public LDAPAuthenticatedUser authenticateUser(Credentials credentials)
 
     }
 
+    /**
+     * Returns parameter current ldap domain token generated from user credentials
+     * @param credentials
+     *     The credentials used for authentication.
```
mike-jumper, line 318: To comply with JavaDoc standards and the code style we use for comments, this should be formatted as:

```
   /**
     * Returns parameter current ldap domain token generated from user credentials.
     *
     * @param credentials
     *     The credentials used for authentication.
```

That said, the phrase "Returns parameter current ldap domain token generated from user credentials" does not make sense as written. It should probably be rephrased.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -65,6 +65,11 @@ public class AuthenticationProviderService {
      */
     public static final String LDAP_ATTRIBUTE_TOKEN_PREFIX = "LDAP_";
 
+    /**
+     * The name of LDAP domain attribute
```
mike-jumper, line 69: JavaDoc standards require that the bodies of JavaDoc comments end with periods.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -356,6 +384,11 @@ private Map<String, String> getAttributeTokens(ConnectedLDAPConfiguration config
                 tokens.put(TokenName.canonicalize(attr.getId(),
                         LDAP_ATTRIBUTE_TOKEN_PREFIX), attr.getString());
             }
+            String domainName = getDomainToken(credentials);
+            if (domainName != null) {
+                String tokenName = TokenName.canonicalize(LDAP_DOMAIN_TOKEN, LDAP_ATTRIBUTE_TOKEN_PREFIX);
+                tokens.put(tokenName, domainName);
+            }
```
mike-jumper, line 391: This shouldn't be within the `try`/`catch` block that relates to LDAP errors that may be encountered while retrieving attributes, as it's entirely independent of all that. For readability, it should also be visibly separated from the block that precedes it, which is already documented as having a different purpose.

I'd recommend just moving this outside the `try`/`catch`, making sure to add:

* A blank line to visibly separate the block from everything else
* A high-level comment covering the intent of the block.

mike-jumper (changes requested):

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
josnabattula: Agree, I will do that.

josnabattula:
@mike-jumper I have resolved your comments, could you review when you get a minute. Thanks.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -28,6 +28,9 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
```
necouchman, line 33: Please remove this extra line -we do not split `import` lines.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -308,25 +316,49 @@ public LDAPAuthenticatedUser authenticateUser(Credentials credentials)
     }
 
     /**
-     * Returns parameter tokens generated from LDAP attributes on the user
-     * currently bound under the given LDAP connection. The attributes to be
-     * converted into parameter tokens must be explicitly listed in
-     * guacamole.properties. If no attributes are specified or none are
-     * found on the LDAP user object, an empty map is returned.
+     * Returns the current LDAP domain token from the provided user credentials
+     * @param credentials
```
necouchman, line 320: There should be a line between this one and the one before it.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -308,25 +316,49 @@ public LDAPAuthenticatedUser authenticateUser(Credentials credentials)
...
+     *     The credentials used for authentication.
+     *
+     * @return
+     *     Domain name by splitting login username or null if no domain is detected.
+     */
+    private String getDomainToken(Credentials credentials) {
+        String username = credentials.getUsername();
+        Pattern pattern = Pattern.compile("^(.+)\\\\.*$|^.*@(.+)$");
```
necouchman, line 328: Might be nice to document what you're doing, here - regular expressions can be obscure enough that it isn't obvious what this RegEx does, and would benefit from a one line description of it.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -308,25 +316,49 @@ public LDAPAuthenticatedUser authenticateUser(Credentials credentials)
...
+        }
+        return null;
+    }
+
+    /**
+     * Returns parameter tokens generated based on details specific to the user currently bound under the given
+     * LDAP connection. Both LDAP attributes on the user's associated LDAP object and the credentials submitted
+     * by the user to Guacamole are considered. If any tokens are to be derived from LDAP attributes, those
+     * attributes must be explicitly listed in guacamole.properties. If no tokens are applicable, an empty map is
+     * returned.
```
necouchman, line 341: These lines ended up being substantially longer than most of the other lines in the file, well beyond our sort of "soft target" of 80 columns as a wrap point.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/AuthenticationProviderService.java
```diff
@@ -308,25 +316,49 @@ public LDAPAuthenticatedUser authenticateUser(Credentials credentials)
...
-     *     currently bound under the given LDAP connection, as a map of token
-     *     name to corresponding value, or an empty map if no attributes are
-     *     specified or none are found on the user object.
+     *      A map of parameter tokens. These tokens are generated based on
+     *      the attributes of the user currently bound under the given LDAP connection
+     *      and the user's credentials. The map's keys are the canonicalized attribute
+     *      names with an "LDAP_" prefix, and the values are the corresponding attribute
+     *      values. If the domain name is extracted from the user's credentials, it is added
+     *      to the map with the key "LDAP_DOMAIN". If no applicable tokens are found,
+     *      the method returns an empty map.
```
necouchman, line 356: Looks like these lines are all indented by an extra space?

necouchman (changes requested):
Overall I think the code is okay, but there are a few style fix-ups to do, here, to make it consistent with the rest of the code

necouchman:
@josnabattula You need to remove the merge commits from this PR - there are currently two:
ca241ab2b2c17a198f7f312fb2c38d36eb6064c0
26f21166a3883624b5d83fa864a7c17a743fa5dc

Also, you might want to consider squashing the others together into a smaller number - 9 commits seems a bit much for 60 lines of code change ;-).

josnabattula:
@necouchman  I have removed merge commits and squashed other commits also.

necouchman (approved):
I'm good with it. @mike-jumper, any further changes you'd like to see?

fanovilla (approved):
looks good Josna!

josnabattula:
@mike-jumper When you get some time, could you review this pull request? thanks.

josnabattula:
@mike-jumper Can you review this request, we are waiting for this feature for a while. thanks.

josnabattula:
@mike-jumper could you review this pull request, I have resolved your comments. thanks.

mike-jumper:
Yes, I'll take a look.

josnabattula:
@mike-jumper I can close this pull request once superseded by this PR https://github.com/apache/guacamole-client/pull/987 is merged.

mike-jumper:
Closing per above (superseded by #987).

PR #937, opened 2023-12-03
GUACAMOLE-1859: Update webapp dependencies to latest compatible versions.

mike-jumper:


necouchman:
@mike-jumper Did the automatic build run? I don't see a status for that, and don't have a button here to trigger it...

jmuehlner:
> @mike-jumper Did the automatic build run? I don't see a status for that, and don't have a button here to trigger it...

The automated build stuff isn't in the release just yet - only `master`.

PR #938, opened 2023-12-28
GUACAMOLE-1896: Refresh connection parameters when focused client updates.

jmuehlner:


guacamole/src/main/frontend/src/app/client/directives/guacTiledClients.js
```diff
@@ -92,6 +92,11 @@ angular.module('client').directive('guacTiledClients', [function guacTiledClient
             $scope.$emit('guacClientFocused', focusedClient);
         });
 
+        // Notify whenever arguments of currently-focused client changes
+        $scope.$watch('getFocusedClient().arguments', function focusedClientParametersChanged() {
```
jmuehlner, line 96: So the issue is that the arguments get updated in the focused client, but `guacClientFocused` never gets emitted because the above scope watch only checks for object equality.
 
I wish there was a way to pass a custom comparator to `$scope.$watch()` that could check for object equality _or_ changes in the `arguments` field, but there's no way to do that - it's either just checking whether it's the same object or doing a full deep comparison.

jmuehlner: If anyone has a suggestion for a more elegant solution, I'm all ears.

jmuehlner: Updated to use a different event entirely.

necouchman (approved):
LGTM :shipit: 

@mike-jumper Any concerns with this?

mike-jumper:
> LGTM :shipit:
> 
> @mike-jumper Any concerns with this?

Nope. LGTM, as well.

PR #940, opened 2024-01-10
Added alpine based dockerfile and mysql database setup via docker run…

gyurix1968:
A modification has been made during container image creation, allowing for a significantly smaller Docker image to be generated from the source by utilizing the new file (Dockerfile.alpine).

The application's startup script has been adjusted to enable the specification of the Docker runtime, including the first (admin) user's name and corresponding password (via the ADMIN_NAME and ADMIN_PASSWORD variables). Simultaneously, database initialization has been automated (currently supporting only MySQL databases) through the invocation of a separate script. This script checks whether the database server is accessible and if a database has already been created.

### DIFF
--- a/extensions/guacamole-auth-sso/modules/guacamole-auth-sso-ssl/src/main/java/org/apache/guacamole/auth/ssl/SSLClientAuthenticationResource.java
+++ b/extensions/guacamole-auth-sso/modules/guacamole-auth-sso-ssl/src/main/java/org/apache/guacamole/auth/ssl/SSLClientAuthenticationResource.java
@@ -142,6 +142,12 @@ public class SSLClientAuthenticationResource extends SSOResource {
 
     /**
      * Decodes the provided URL-encoded string as UTF-8, returning the result.
+     * <p>
+     * NOTE: The escape() function of the Apache HTTPD server is known to not
+     * encode plus signs, which can appear in the base64-encoded certificates
+     * typically received here. To avoid mangling such certificates, this
+     * function specifically avoids decoding plus signs as spaces (as would
+     * otherwise happen if URLDecoder is used directly).
      *
      * @param value
      *     The URL-encoded string to decode.
@@ -153,6 +159,13 @@ public class SSLClientAuthenticationResource extends SSOResource {
      *     If the provided value is not a valid URL-encoded string.
      */
     private byte[] decode(String value) throws GuacamoleException {
+
+        // Ensure all plus signs are decoded literally rather than as spaces
+        // (the Apache HTTPD implementation of URL escaping that applies to
+        // request headers does not encode plus signs, whereas the Nginx
+        // implementation does)
+        value = value.replace("+", "%2B");
+
         try {
             return URLDecoder.decode(value, StandardCharsets.UTF_8.name())
                     .getBytes(StandardCharsets.UTF_8);


guacamole-docker/bin/start.sh
```diff
@@ -562,11 +562,11 @@ END
 
     set_optional_property                  \
         "sqlserver-auto-create-accounts"  \
-        "$SQLSERVER_AUTO_CREATE_ACCOUNTS"
+        "$SQLSERVERQL_AUTO_CREATE_ACCOUNTS"
 
     set_optional_property      \
         "sqlserver-instance"  \
-        "$SQLSERVER_INSTANCE"
+        "$SQLSERVERQL_INSTANCE"
```
mike-jumper, line 569: Why were these renamed?

guacamole-docker/bin/check_database.sh
```diff
@@ -0,0 +1,101 @@
...
+	sed -i "s/CA458A7D494E3BE824F5E1E175A1556C0F8EEF2C2D7DF3633BEC4A29C4411960/$HASH/" /tmp/002-create-admin-user.sql 
+
+
+	# Check database access
+	for retries in $(seq 0 $((RETRIES+ 1))); do
+	          RESPONSE=$(echo "exit" | timeout $SLEEP_SHORT telnet $DB_HOST $DB_PORT|grep Connected);
+	          echo "RESPONSE: $RESPONSE";
+	          if [ "$RESPONSE" == "" ]; then 
+	              if [[ $retries -le $RETRIES ]] ; then
+	                  echo "Retired $retries";
```
mike-jumper, line 65: What do you mean by "retired"?

guacamole-docker/bin/check_database.sh
```diff
@@ -0,0 +1,101 @@
...
+			      	  echo "Expected $DB_HOST accessed successfuly, continue database init processes."
+	                          break;
+	                      fi  
+	                  done 
+	              fi
+	          else
+		      echo "Expected $DB_HOST accessed successfuly, continue database init processes."
+	              break;
+	          fi
+	done
```
mike-jumper, line 90: 1. The logic here feels overcomplicated for what it is intended to achieve (wait for the database to come up). Why so many nested loops and layers of retrying?
2. I'm on the fence on building this functionality into the Docker image. Perhaps adding automatic initialization into the database support within the webapp itself would make more sense? That would also allow for automatic application of schema updates while maintaining separation of concerns.

guacamole-docker/bin/check_database.sh
```diff
@@ -0,0 +1,101 @@
...
+	DB_ENGINE="/usr/bin/mariadb"
+	
+	
+	UUID=$(openssl rand -hex 16)
+	SALT=$(echo ${UUID:0:8}-${UUID:8:4}-${UUID:12:4}-${UUID:16:4}-${UUID:20:12} | openssl sha256 -hex | awk '{print toupper($2)}');
+	HASH=$(echo -n "$ADMIN_PASSWORD$SALT" | openssl sha256 -hex | awk '{print toupper($2)}')
+	cp /opt/guacamole/mysql/schema/*.sql /tmp/
+	sed -i "s/guacadmin/$ADMIN_NAME/g" /tmp/002-create-admin-user.sql 
+	sed -i "s/FE24ADC5E11E2B25288D1704ABE67A79E342ECC26064CE69C5B3177795A82264/$SALT/" /tmp/002-create-admin-user.sql 
+	sed -i "s/CA458A7D494E3BE824F5E1E175A1556C0F8EEF2C2D7DF3633BEC4A29C4411960/$HASH/" /tmp/002-create-admin-user.sql
```
mike-jumper, line 56: This would need to be implemented for each supported database, not just MySQL/MariaDB.

guacamole-docker/bin/check_database.sh
```diff
@@ -0,0 +1,101 @@
...
+	SLEEP_SHORT=5
+fi
+
+if [ "$SLEEP_LONG" == "" ]; then
+	SLEEP_LONG=10
+fi
+
+if [ "$RETRIES" == "" ]; then
+	RETRIES=5
+fi
```
mike-jumper, line 33: If new environment variables are to be supported by this image, the naming should be more specific. `SLEEP_SHORT`, `SLEEP_LONG`, and `RETRIES` are too broad.

Dockerfile.alpine
mike-jumper, line 1: Why a second Dockerfile vs. modifying the existing Dockerfile?

guacamole-docker/bin/check_database.sh
```diff
@@ -0,0 +1,101 @@
...
+
+	for i in $(ls /tmp/*.sql); do 
+		CREATE_DATABASE=$($DB_ENGINE -u $DB_USER -p$DB_PASSWORD $DB_NAME -h $DB_HOST --port $DB_PORT < $i);
+		if [[ "$(echo $?)" == "1" ]]; then
+			echo "Database already exists";
+			exit 2;                
+		else
+			echo "Database automaticaly initialized." ;
+		fi
+	done
```
mike-jumper, line 100: This is dangerous. Should a user ever manage to write a file to `/tmp` with a `.sql` extension, that file would be automatically run at subsequent startup as a privileged database user. Merely writing files to `/tmp` is not a privileged operation, so this would be privilege escalation.

mike-jumper (changes requested):
Please also see our [contribution guidelines](https://github.com/apache/guacamole-client/blob/master/CONTRIBUTING). In particular, every change needs a corresponding JIRA issue, and that JIRA issue needs to be included in each commit message (see established git history for the formatting convention we use for this).

guacamole-docker/bin/check_database.sh
necouchman: > Perhaps adding automatic initialization into the database support within the webapp itself would make more sense?

+1

PR #945, opened 2024-01-17
GUACAMOLE-1678: Fix the typing issue in the clipboard text in some cases.

myjimmy:
Fixed the typing issue in the clipboard text in some cases.

### DIFF
--- a/guacamole/src/main/frontend/src/app/clipboard/services/clipboardService.js
+++ b/guacamole/src/main/frontend/src/app/clipboard/services/clipboardService.js
@@ -212,7 +212,6 @@ angular.module('clipboard').factory('clipboardService', ['$injector',
 
         // Track the originally-focused element prior to changing focus
         var originalElement = document.activeElement;
-        pushSelection();
 
         // Copy the given value into the clipboard DOM element
         if (typeof data.data === 'string')
@@ -234,11 +233,9 @@ angular.module('clipboard').factory('clipboardService', ['$injector',
         else
             deferred.reject();
 
-        // Unfocus the clipboard DOM event to avoid mobile keyboard opening,
-        // restoring whichever element was originally focused
+        // Unfocus the clipboard DOM event to avoid mobile keyboard opening
         clipboardContent.blur();
         originalElement.focus();
-        popSelection();
 
         return deferred.promise;
     };


necouchman:
@myjimmy I have some concerns with these changes...
* It's unclear to me _why_ removing the `pushSelection()` and `popSelection()` calls fixes the issue described in the Jira ticket? From that, it sounds like this is only an issue on insecure (HTTP://) connections, and others work fine. Why is this particular change necessary to solve that issue?
* Based on the comments around the changes, it seems like the blur was intentionally added for mobile browsers - is it really the push/pop that needs to be removed, or does the blur need to be handled differently (maybe only on mobile browsers)?
* Do these changes have any other impacts to other platforms - particularly the mobile ones mentioned in the comments around this section?

necouchman:
@myjimmy Sorry, I see that you addressed many of these on the Jira ticket. You mentioned, there, that the push/pop functionality was added to support non-text clipboard contents, which Guacamole currently doesn't support - I guess I want to make sure that this really was (supposed to be) fully removed and that these changes don't have any adverse impacts on other clients before pull those calls out.

Also, if those things were added for the purpose of supporting non-text clipboard data, presumably at some point in the future we will want to re-add that support, so not sure if this will become an issue, again, when we go to do that?

myjimmy:
> It's unclear to me why removing the pushSelection() and popSelection() calls fixes the issue described in the Jira ticket? From that, it sounds like this is only an issue on insecure (HTTP://) connections, and others work fine. Why is this particular change necessary to solve that issue?

@necouchman The Asynchronous Clipboard API is available only but in secure contexts (HTTPS) also in some of browsers.
https://developer.mozilla.org/en-US/docs/Web/API/Clipboard#browser_compatibility

> Based on the comments around the changes, it seems like the blur was intentionally added for mobile browsers - is it really the push/pop that needs to be removed, or does the blur need to be handled differently (maybe only on mobile browsers)?

Because the removed code is related to only the clipboard of the non-text content, it have no relation to the blur() and focus() functions.
The blur() and focus() functions are called for performing the deprecated clipboard command.

> Do these changes have any other impacts to other platforms - particularly the mobile ones mentioned in the comments around this section?

I think that these changes have no impacts to other platforms.
I've tested my fix on some kinds of mobile devices (iphone, android) and browsers (firefox, chrome, safari).
It works well.

mike-jumper:
The `pushSelection()` and `popSelection()` functions are not specific to support for images - they prevent the current selection state within the browser from being clobbered by the call to `selectAll()` that has to occur before we invoke `document.execCommand()`.

From the description of GUACAMOLE-1678 and the nature of these changes, it sounds like manipulating the selection state is having unintended side effects, but I'm skeptical that simply removing `pushSelection()`, etc. is correct. I would be concerned that this introduces other strange selection behavior.

PR #946, opened 2024-01-21
Nextcloud JWT extension

pp7En:
Hi, I have built a small extension (based on the JSON extension).

The Nextcloud app "External sites" is required. There will be sent a JWT in the first request. This JWT will be validated (valid signature, valid duration, username) if the JWT is not valid, an exception will be thrown, so no login is possible. I'm using this for a while and maybe it is also interesting and useful for others.

The code could probably be simplified and cleaned up, but it might be interesting as a PoC.

---

Links:

* Repo: https://github.com/nextcloud/external
* Generate Nextcloud JWT: https://github.com/nextcloud/external/blob/master/docs/jwt-sample.php


necouchman:
@pp7En Thanks for this contribution! I think this is a great capability to have. I do have a couple of overall comments for you before even reviewing the code:
* All changes need to have a Jira issue, and the PR and commit messages need to be tagged with that commit message. See other pull requests, particularly merged ones, for guidance. See this page on our website for general guidance: https://guacamole.apache.org/open-source/.
* This seems like the kind of extension that might be best integrated as a sub-module of the guacamole-auth-sso, where things like SAML, OpenID, and CAS support live, particularly as I believe the Nextcloud trusted sites features uses OAuth/OIDC?

pp7En:
Thanks for your quick feedback.

> All changes need to have a Jira issue, and the PR and commit messages need to be tagged with that commit message. See other pull requests, particularly merged ones, for guidance. See this page on our website for general guidance: https://guacamole.apache.org/open-source/.

I will create a Jira ticket and a new pull request and consider these things.

> This seems like the kind of extension that might be best integrated as a sub-module of the guacamole-auth-sso, where things like SAML, OpenID, and CAS support live, particularly as I believe the Nextcloud trusted sites features uses OAuth/OIDC?

Depending on the user, the Nextcloud app only generates a JWT as a URL parameter. The Nextcloud admin can create an "external site" for everyone or groups. The public key must then be added to Guacamole. This page will be added to the main menu and embedded. 

I think the Nextcloud JWT can be used directly for authentication, but these current code changes use the JWT very simply. The Guacamole login credentials are still necessary. If the JWT is valid, the login page will be displayed and if not it will not be displayed because of the exception.

It's a good idea to login directly through the JWT. I will try that.

PR #948, opened 2024-01-30
GUACAMOLE-1906: Bump guac-manifest versions and add 1.5.5 to the extension loader.

necouchman:
Got all the pom.xml files in the last pull request, but forgot the extension loader and the guac-manifest files.

### DIFF
--- a/extensions/guacamole-auth-duo/src/main/resources/guac-manifest.json
+++ b/extensions/guacamole-auth-duo/src/main/resources/guac-manifest.json
@@ -1,6 +1,6 @@
 {
 
-    "guacamoleVersion" : "1.5.4",
+    "guacamoleVersion" : "1.5.5",
 
     "name"      : "Duo TFA Authentication Backend",
     "namespace" : "duo",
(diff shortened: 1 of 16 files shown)


jmuehlner:
You know, while we're at it, it looks like the `guacamole-display-statistics` package still needs to have its version bumped as well. I notice that [your branch](https://github.com/necouchman/guacamole-client/tree/jira/1906/extensions) is so old that the entire extension is missing, but it [exists on master](https://github.com/apache/guacamole-client/blob/master/extensions/guacamole-display-statistics/pom.xml) and should be bumped.

necouchman:
> I notice that [your branch](https://github.com/necouchman/guacamole-client/tree/jira/1906/extensions) is so old that the entire extension is missing, but it [exists on master](https://github.com/apache/guacamole-client/blob/master/extensions/guacamole-display-statistics/pom.xml) and should be bumped.

I _think_ this is because the `guacamole-display-statistics` package isn't in the `1.5.x` series, but is only in the master branch. I actually have a separate Jira issue and [branch](https://github.com/necouchman/guacamole-client/tree/jira/1915) for bumping the versions on those after this one is wrapped up.

jmuehlner:
Oops, yeah, you're totally right.

> I _think_ this is because the `guacamole-display-statistics` package isn't in the `1.5.x` series, but is only in the master branch. I actually have a separate Jira issue and [branch](https://github.com/necouchman/guacamole-client/tree/jira/1915) for bumping the versions on those after this one is wrapped up.

PR #950, opened 2024-01-30
GUACAMOLE-1904: Broadcast events for mouse/touch events in client.

jmuehlner:
A follow on from https://github.com/apache/guacamole-client/pull/942, this change further adds new events to be broadcast when mouse or touch events are handled in the client. This is to allow custom fields that live outside the client menu to de-focus text fields and the like so they don't intercept user input that's intended to go into the client (since clicking inside the client won't automatically defocus input elements rendered outside the client, as users might expect).

### DIFF
--- a/guacamole/src/main/frontend/src/app/client/directives/guacClient.js
+++ b/guacamole/src/main/frontend/src/app/client/directives/guacClient.js
@@ -54,6 +54,7 @@ angular.module('client').directive('guacClient', [function guacClient() {
         const ManagedClient = $injector.get('ManagedClient');
             
         // Required services
+        const $rootScope = $injector.get('$rootScope');
         const $window = $injector.get('$window');
             
         /**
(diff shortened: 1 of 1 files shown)


guacamole/src/main/frontend/src/app/client/directives/guacClient.js
```diff
@@ -220,6 +221,9 @@ angular.module('client').directive('guacClient', [function guacClient() {
             display.showCursor(!localCursor);
             client.sendMouseState(event.state, true);
 
+            // Broadcast the mouse event
+            $rootScope.$broadcast('guacClientMouseEvent', event);
```
mike-jumper, line 225: For these events to provide context for whatever handles those events, they'll need some kind of reference to the client that would receive the event.

mike-jumper, line 225: I think these events should be renamed in line with other established events:
* Omit the redundant "Event" from the name of the event.
* Split into the types of events recognized by the objects firing them (`guacClientMouseDown`, `guacClientMouseUp`, etc.) so that the handler of these events can have that context without needing to implement their own state tracking, similar to the established `guacKeyDown` and `guacKeyUp`.

mike-jumper (changes requested):

guacamole/src/main/frontend/src/app/client/directives/guacClient.js
jmuehlner: How does this look?

mike-jumper: I like it.

mike-jumper (approved):

PR #954, opened 2024-03-03
GUACAMOLE-1905: Update client dependencies to latest compatible versions

necouchman:
Updated the client-side dependencies to the latest compatible versions, or, at least, I think that's what I've done. There were a handful of Java dependencies in the pom.xml files, one angular-translate update, and then version bumps for the JDBC drivers included with the Docker images.

Note, on the JDBC drivers, the Microsoft SQL Server one, in particular, jumped quite a bit. I don't recall if there was a particular reason we were staying with the 9.x version - I vaguely recall perhaps there being some sort of compatibility issue, but if anyone remembers, please let me know.


jmuehlner:
Sadly it looks like 3 NPM packages are now deprecated:
* angular-translate
* angular-translate-interpolation-messageformat
* angular-translate-loader-static-files

Nothing we can really do about that short of upgrading to angular-not-js.

jmuehlner (approved):

jmuehlner:
Oops, I was a bit hasty in merging this - logback 1.5.2 requires Java 11. 

The latest Java 8 compatible version is 1.3.14.

PR #955, opened 2024-03-09
GUACAMOLE-1905: Revert to Java-8-compatible Logback 1.3.14.

jmuehlner:
As discussed in https://github.com/apache/guacamole-client/pull/954, this is the latest Java 8 compatible version of Logback.


necouchman:
Oops.

PR #956, opened 2024-03-11
GUACAMOLE-1239: Configurable username letter case

StephanRichter:
This is [another option](/apache/guacamole-client/pull/902) to make guacamole case-insensitive.

More specific, it can be configured, whether usernames shall be treated as uppercase, lowercase or mixed-case.

If that seems reasonable, but needs more work, don't hesitate to ask.

[1239]: https://issues.apache.org/jira/browse/GUACAMOLE-1239

### DIFF
--- a/guacamole/src/main/java/org/apache/guacamole/rest/auth/TokenRESTService.java
+++ b/guacamole/src/main/java/org/apache/guacamole/rest/auth/TokenRESTService.java
@@ -23,6 +23,7 @@ import com.google.common.io.BaseEncoding;
 import java.io.UnsupportedEncodingException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 import javax.inject.Inject;
 import javax.servlet.http.HttpServletRequest;
 import javax.ws.rs.DELETE;
@@ -36,6 +37,7 @@ import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.MultivaluedMap;
 import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.GuacamoleResourceNotFoundException;
+import org.apache.guacamole.environment.Environment;
 import org.apache.guacamole.net.auth.AuthenticatedUser;
 import org.apache.guacamole.net.auth.Credentials;
 import org.apache.guacamole.net.auth.UserContext;
@@ -55,12 +57,21 @@ public class TokenRESTService {
      * Logger for this class.
      */
     private static final Logger logger = LoggerFactory.getLogger(TokenRESTService.class);
+    private static final String DEFAULT_LETTER_CASE = "mixed-case";
+
 
     /**
      * Service for authenticating users and managing their Guacamole sessions.
      */
     @Inject
     private AuthenticationService authenticationService;
+
+
+    /**
+     * The Guacamole server environment.
+     */
+    @Inject
+    private Environment environment;
 
     /**
      * Returns the credentials associated with the given request, using the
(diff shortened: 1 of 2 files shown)


StephanRichter:
Thanks for your suggestions. I re-created the commit basing it on the current main branch.
Also, I updated the title to reference the JIRA issue.

I am well aware, that these changes would need an update to the documentation, too. If someone could point me towards the place to do this…

necouchman:
@StephanRichter Thanks. One other thing - there's already a PR out there that deals with this - #902. While that is my PR, I'm not opposed to someone else stepping in and working on it; however, there's already been a fair amount of discussion about how to handle it, and the other PR deals more broadly with how to handle username comparison across multiple authentication modules, versus the approach of just forcing the username to a given case and assuming everything will handle it correctly from there. The approach you've taken, here, is much simpler, but I'm not sure that it covers all of the possible scenarios of dealing with multiple modules (JDBC + SSO or JDBC + LDAP, for example).

I'm happy to have further discussion on the merits of one approach versus another, and I know that the previously mentioned PR has languished a bit, so, again, I'm not opposed to others stepping in to help with this.

StephanRichter:
@necouchman Hi! I've seen the other PR. Basically, I don't prefer my "solution" above yours. I just wanted to suggest an easier solution, which was developed with JDBC+LDAP in mind. Also this solution does not alter any behaviour, if left unconfigured. To be honest, I did not study the other solution in detail, as I only spotted it when preparing my PR ;)

guacamole/src/main/java/org/apache/guacamole/rest/auth/TokenRESTService.java
```diff
@@ -122,6 +133,35 @@ private Credentials getCredentials(HttpServletRequest request,
...
+     *                            guacamole.properties, or if the property is
+     *                            not specified.
+     */
+    private void adaptUsernameLetterCase(Credentials credentials) throws GuacamoleException {
+        switch (environment.getProperty(Environment.USERNAME_LETTER_CASE,DEFAULT_LETTER_CASE)){
+            case "lower-case":
+                Optional.ofNullable(credentials.getUsername()).map(String::toLowerCase).ifPresent(credentials::setUsername);
+                break;
+            case "upper-case":
+                Optional.ofNullable(credentials.getUsername()).filter(un -> !"guacadmin".equals(un)).map(String::toUpperCase).ifPresent(credentials::setUsername);
```
jmuehlner, line 158: We shouldn't assume any specific usernames. What is this check intending to accomplish?

guacamole/src/main/java/org/apache/guacamole/rest/auth/TokenRESTService.java
```diff
@@ -55,13 +57,22 @@ public class TokenRESTService {
      * Logger for this class.
      */
     private static final Logger logger = LoggerFactory.getLogger(TokenRESTService.class);
+    private static final String DEFAULT_LETTER_CASE = "mixed-case";
```
jmuehlner, line 60: Constants should be documented.

jmuehlner, line 60: All constants should be documented.

guacamole/src/main/java/org/apache/guacamole/rest/auth/TokenRESTService.java
jmuehlner, line 158: We shouldn't expect any particular username - what is this check intending to accomplish?

mike-jumper:
Closing in favor of #902, which has now been merged.

StephanRichter:
Damn. Right when I managed to assign some time to work on this. The other approach feels a bit complicated. But nevermind, as long as the Issue is resolved. I'm looking forward to the next release ;)

PR #957, opened 2024-03-12
GUACAMOLE-1905: Revert to java-saml-compatible version of woodstox.

jmuehlner:
In https://github.com/apache/guacamole-client/pull/954, the version of woodstox was bumped from `5.4.0` to `6.5.0`, but `6.5.0` is a major version bump past the `5.2.1` version that java-saml expects. }

This change restores the old `5.4.0` version of woodstox, which should remain in place until a newer `5.*` version of woodstox is released, or `java-saml` has a new release that uses a `6.*` version of woodstox.


necouchman (approved):
Thanks for catching these!

PR #964, opened 2024-03-24
GUACAMOLE-1905: Update Kotlin, Jackson, and Jersey to their latest compatible versions.

mike-jumper:
This change updates to the latest versions of dependencies that have been released since we last did this.

I don't think we can safely override the Bouncy Castle dependency of `cas-client-core` without a full understanding of that library's internals, so we may be stuck there until Apereo hopefully eventually issues a new release with updated dependencies. It looks like Apereo should be aware of the fact that their Bouncy Castle is pretty old:

https://github.com/apereo/java-cas-client/pull/628#issuecomment-1659293604

Then again ... Apereo's commit history suggests that they rely pretty heavily on automated dependency updates from bots, and those bots are not going to pick up on these availability of updates since newer Bouncy Castle has different Maven coordinates.


necouchman:
> I don't think we can safely override the Bouncy Castle dependency of cas-client-core without a full understanding of that library's internals, so we may be stuck there until Apereo hopefully eventually issues a new release with updated dependencies.

It is worth noting that there are more recent CAS libraries available (4.x), but the namespace has changed from `org.jasig` to `org.apereo`. When I pulled in the original set of updates I opted not to make that leap given the amount of other code that would have to be changed to complete it - it didn't seem like a "update dependencies without breaking things" sort of a change.

I don't know for sure that this will at all impact the Bouncycastle issue, just throwing it out there.

mike-jumper:
Yep, unfortunately the same issue is there:

https://github.com/apereo/java-cas-client/blob/v4.0.0/pom.xml#L264

I agree with not updating from `org.jasig` to `org.apereo` - worth doing, but probably not as part of the quick-and-easy release finalization updates.

necouchman (approved):

PR #966, opened 2024-03-29
GUACAMOLE-1289: Duo upgrade to API v4 

aleitner:
I continued working off @necouchman's PR https://github.com/apache/guacamole-client/pull/915 to add support for resuming authentication when dealing with the redirect authentication model


necouchman:
Nice @aleitner - thanks for picking this up and taking it over the line!

necouchman:
@stcbus Everything is in this PR - you don't need to pull both of them.

extensions/guacamole-auth-duo/src/main/java/org/apache/guacamole/auth/duo/UserVerificationService.java
```diff
@@ -75,39 +97,72 @@ public void verifyAuthenticatedUser(AuthenticatedUser authenticatedUser)
...
-            // Create an overall description of the additional credentials
-            // required to verify identity
-            CredentialsInfo expectedCredentials = new CredentialsInfo(
-                        Collections.singletonList(signedResponseField));
+        
+        String username = authenticatedUser.getIdentifier();
+
+        try {
+
+        String redirectUrl = confService.getRedirectUrl().toString();
```
jmuehlner, line 105: It looks like this whole block is missing a level of indentation.

guacamole/src/main/java/org/apache/guacamole/rest/auth/ResumableAuthenticationState.java
```diff
@@ -0,0 +1,128 @@
...
+    /**
+     * Checks if this resumable state has expired based on the stored expiration
+     * timestamp and the current system time.
+     *
+     * @return 
+     *     True if the current system time is after the expiration timestamp,
+     *     indicating that the state is expired; false otherwise.
+     */
+    public boolean isExpired() {
+        return System.currentTimeMillis() > expirationTimestamp;
```
jmuehlner, line 92: Technically this should probably be `>=` here since the documentation for the `expirationTimestamp` parameter specifies that it's the time at which the state expires, not after which it expires.

extensions/guacamole-auth-duo/src/main/java/org/apache/guacamole/auth/duo/conf/ConfigurationService.java
```diff
@@ -56,11 +59,11 @@ public class ConfigurationService {
      * key received from Duo for verifying Guacamole users. This value MUST be
      * exactly 20 characters.
      */
-    private static final StringGuacamoleProperty DUO_INTEGRATION_KEY =
```
jmuehlner, line 59: It looks like this block comment should also be updated.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/credentials/GuacamoleInsufficientCredentialsException.java
```diff
@@ -28,6 +28,95 @@
...
+ */
+protected final String state;
+
+    /**
+     * The identifier for the authentication provider that threw this exception.
+     * This is used to link the exception back to the originating source of the
+     * authentication attempt, allowing clients to determine which provider's
+     * authentication process should be resumed.
+     */
+    protected final String providerIdentifier;
```
jmuehlner, line 68: These fields look overindented.

guacamole/src/main/java/org/apache/guacamole/rest/auth/AuthenticationService.java
```diff
@@ -328,6 +354,82 @@ private List<DecoratedUserContext> getUserContexts(GuacamoleSession existingSess
...
+        Credentials resumedCredentials = null;
+
+        // Retrieve signed State from the request
+        HttpServletRequest request = credentials.getRequest();
+        
+        // Retrieve the provider id from the query parameters.
+        String resumableProviderId = request.getParameter(Credentials.RESUME_QUERY);
+        // Check if a provider id is set.
+        if (resumableProviderId == null || resumableProviderId.isEmpty()) {
+            // return if a provider id is not set.
```
jmuehlner, line 380: The formatting of these single-line comments is a bit funky - comments like this should always start with a capital letter (this one doesn't), and generally do not end with a period.

This comment should definitely be capitalized. I'd also consider removing the periods to match existing comment style.

extensions/guacamole-auth-duo/src/main/java/org/apache/guacamole/auth/duo/conf/ConfigurationService.java
```diff
@@ -69,26 +72,38 @@ public class ConfigurationService {
...
+    private static final StringGuacamoleProperty DUO_CLIENT_SECRET =
             new StringGuacamoleProperty() {
 
         @Override
-        public String getName() { return "duo-secret-key"; }
+        public String getName() { return "duo-client-secret"; }
 
     };
-
+
```
jmuehlner, line 82: No bogus whitespace please.

extensions/guacamole-auth-duo/src/main/java/org/apache/guacamole/auth/duo/conf/ConfigurationService.java
```diff
@@ -110,51 +125,65 @@ public String getAPIHostname() throws GuacamoleException {
...
+    public String getClientId() throws GuacamoleException {
+        return environment.getRequiredProperty(DUO_CLIENT_ID);
     }
 
     /**
-     * Returns the secret key received from Duo for verifying Guacamole users,
-     * as defined in guacamole.properties by the "duo-secret-key" property. This
-     * value MUST be exactly 20 characters.
+     * Returns the client secert received from Duo for verifying Guacamole users,
+     * as defined in guacamole.properties by the "duo-client-secert" property.
```
jmuehlner, line 144: Typo: `secert`

stcbus:
Not sure if this is the correct place to mention this or not, but using this branch works perfectly for me, except on the redirect to Duo, I just see a "LOGIN.INFO_DUO_REDIRECT_PENDING" text. I assume it is missing the translation text for that variable in translations?

mike-jumper:
> Not sure if this is the correct place to mention this or not, but using this branch works perfectly for me, except on the redirect to Duo, I just see a "LOGIN.INFO_DUO_REDIRECT_PENDING" text. I assume it is missing the translation text for that variable in translations?

Yes, that string will need to be added here. I don't see any changes present that define that string.

If you still see that text after the string is added, you probably just need to clear browser cache.

stcbus:
Also, just to note (I'm sure it's known!), the docker `start.sh` script needs updated with the new names of the DUO keys and to incorporate the new variables for redirect URL and timeout.

aleitner:
Added the missing translations and revised the guacamole.properties setup for duo

stcbus:
Thanks @aleitner  ! In the interest of accuracy, I believe the application key is no longer used with v4, so that part can be removed from docker/config as well. Please correct, if wrong (I certainly don't pass it in my testing).

aleitner:
> Thanks @aleitner ! In the interest of accuracy, I believe the application key is no longer used with v4, so that part can be removed from docker/config as well. Please correct, if wrong (I certainly don't pass it in my testing).

Yup, it's no longer necessary so I just removed it. Thanks!

jmuehlner (approved):
LGTM

stcbus:
Let me know if I should report a bug now, but since it's not released quite yet, we just ran into 2 issues I think that maybe are things that need fixed for the extension:

1. I am getting failures after successful logins, and that can trigger the ban extension now with lots of testing. I don't believe we had that before, so I think maybe somehow Duo extension doesn't report correctly to the ban extension?

2. In a certain scenario, we get a generic guacamole error occured message. This is 100% reproducible in the following scenario:

- Log into guac, pass through duo
- after log in, log out
- click "relogin" button

Re-loading the page is fine, so something with the relogin button or flow from there, we are rejected and get the below log message:

`01:35:12.356 [http-nio-8080-exec-5] WARN  o.a.g.event.EventLoggingListener - Authentication attempt from 128.146.93.48 for user "USERNAME failed: Duo Client error. (rejected by "duo")
01:35:12.356 [http-nio-8080-exec-5] ERROR o.a.g.rest.RESTExceptionMapper - Request could not be processed: Duo Client error.`

stcbus:
From a quick troubleshooting session, it seems like the re-login button sends you back with the duo token stuff still in the URL field? Maybe that is why?

aleitner:
> From a quick troubleshooting session, it seems like the re-login button sends you back with the duo token stuff still in the URL field? Maybe that is why?
> 
> edit: Also confirmed, the duo token and other query parmeters are there right on log on as well. Positive this is it.

Looking into a fix and I'll add it to my other PR https://github.com/apache/guacamole-client/pull/973

PR #967, opened 2024-04-03
GUACAMOLE-1916: Jump to top of page after clone.

necouchman:


aleitner:
Awesome I was just thinking about needing this

guacamole/src/main/frontend/src/app/manage/controllers/manageConnectionController.js
```diff
@@ -318,6 +318,7 @@ angular.module('manage').controller('manageConnectionController', ['$scope', '$i
      */
     $scope.cloneConnection = function cloneConnection() {
         $location.path('/manage/' + encodeURIComponent($scope.selectedDataSource) + '/connections').search('clone', identifier);
+        $window.scrollTo(0,0);
```
jmuehlner, line 321: Doesn't the $window service need to be injected for this to work? I don't see it in this file.

necouchman: Well, it works without it just fine - I _think_ `$window` is a sort of global variable in jQuery; however, for consistency, I've added it.

jmuehlner: [$window](https://docs.angularjs.org/api/ng/service/$window) is an angularjs service, and while it _might_ be automagically included somehow from something else (like jQuery), it's probably a bad idea to just assume that it'll always be implicitly present. Much better to explicitly use the angularjs service IMO.

PR #970, opened 2024-04-07
GUACAMOLE-1866: Allow recent connections to be removed.

necouchman:
This is an additional change to the "Recent Connections" functionality that allows for a connection to be removed from the list. A small red "X" is added to the upper-right corner of the connection thumbnail, as shown in the image below, and, when clicked, the connection will be removed.

It's worth noting that I had some trouble positioning the "X" exactly where I wanted it (directly over the thumbnail image), while still allowing the click functionality to work correctly. If anyone has suggestions on how to update the CSS to position, but allow it to work, I'm open to suggestions. The main challenge was that I cannot place the `<div></div>` tags that contain the image inside of the `<a></a>` tags, so positioning over the `` was a bit challenging.

image

image

### DIFF
--- a/guacamole/src/main/frontend/src/app/history/services/guacHistory.js
+++ b/guacamole/src/main/frontend/src/app/history/services/guacHistory.js
@@ -43,6 +43,22 @@ angular.module('history').factory('guacHistory', ['$injector',
     service.recentConnections = [];
 
     /**
+     * Remove from the list of connection history the item having the given
+     * identfier.
+     * 
+     * @param {String} id
+     *     The identifier of the item to remove from the history list.
+     *     
+     * @returns {boolean}
+     *     True if the removal was successful, otherwise false.
+     */
+    service.removeEntry = function removeEntry(id) {
+    
+        return _.remove(service.recentConnections, entry => entry.id === id).length > 0;
+    
+    };
+
+    /**
      * Updates the thumbnail and access time of the history entry for the
      * connection with the given ID.
      * 
--- a/guacamole/src/main/frontend/src/app/home/styles/home.css
+++ b/guacamole/src/main/frontend/src/app/home/styles/home.css
@@ -49,6 +49,7 @@ div.recent-connections div.connection {
     text-align: center;
     max-width: 75%;
     overflow: hidden;
+    position: relative;
 }
 
 a.home-connection, .empty.balancer a.home-connection-group {
@@ -79,4 +80,31 @@ a.home-connection, .empty.balancer a.home-connection-group {
 .header-app-name {
     font-size: 0.85em;
     box-shadow: none;
+}
+
+.recent-connections .connection .remove-recent::after {
+    content: '';
+    display: block;
+    height: 100%;
+    width: 100%;
+    background-repeat: no-repeat;
+    background-size: contain;
+    background-position: center center;
+    background-image: url('images/x.svg');
+}
+
+.recent-connections .connection .remove-recent {
+    background-color: red;
+    height: 1em;
+    width: 1em;
+    position: absolute;
+    top: 10px;
+    z-index: 10;
+    float: right;
+    right: 10px;
+    opacity: .2;
+}
+
+.recent-connections .connection .remove-recent:hover {
+    opacity: 1.0;
 }
--- a/guacamole/src/main/frontend/src/app/home/directives/guacRecentConnections.js
+++ b/guacamole/src/main/frontend/src/app/home/directives/guacRecentConnections.js
@@ -60,6 +60,21 @@ angular.module('home').directive('guacRecentConnections', [function guacRecentCo
             $scope.recentConnections = [];
             
             /**
+             * Remove the connection from the recent connection list having the
+             * given identifier.
+             * 
+             * @param {!RecentConnection} recentConnection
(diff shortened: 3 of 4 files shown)


guacamole/src/main/frontend/src/app/history/services/guacHistory.js
```diff
@@ -42,6 +42,31 @@ angular.module('history').factory('guacHistory', ['$injector',
...
+     *     
+     * @returns {boolean}
+     *     True if the removal was successful, otherwise false.
+     */
+    service.removeEntry = function removeEntry(id) {
+    
+        var i;
+        
+        for (i = 0; i < service.recentConnections.length; i++) {
+            if (service.recentConnections[i].id === id) {
```
jmuehlner, line 60: This definitely works fine. That said, we do have access to lodash in the webapp, which makes these sorts of operations a bit less clunky. 

For example, the contents of this function could be replaced by 

```
return _.remove(service.recentConnections, entry => entry.id === id).length > 0;
```

necouchman: Ah, cool, I'm happy to give it a shot with that.

guacamole/src/main/frontend/src/app/home/styles/home.css
```diff
@@ -79,4 +80,21 @@ a.home-connection, .empty.balancer a.home-connection-group {
 .header-app-name {
     font-size: 0.85em;
     box-shadow: none;
+}
+
+.recent-connections .connection .x-thumbnail {
+    background-color: red;
+    display: flex;
```
jmuehlner, line 87: Hmm, why `flex`?

necouchman: I'm not entirely sure, but if remove it, or set it to something else, it displays like this (white "X" outside of the red box):

image

guacamole/src/main/frontend/src/app/history/services/guacHistory.js
necouchman: Thanks for the hint - this seems to work well!

guacamole/src/main/frontend/src/app/home/styles/home.css
```diff
@@ -79,4 +80,21 @@ a.home-connection, .empty.balancer a.home-connection-group {
 .header-app-name {
     font-size: 0.85em;
     box-shadow: none;
+}
+
+.recent-connections .connection .x-thumbnail {
```
jmuehlner, line 85: `x-thumbnail` doesn't seem quite right for a CSS class name - it's describing what the element looks like, not what it _is_.

necouchman: Renamed via re-base - how's `remove-recent`?

jmuehlner: That works for me.

guacamole/src/main/frontend/src/app/home/styles/home.css
jmuehlner: I think it's because the size only gets inherited due to the `flex` property. I was able to get this to work without using `display: flex` (and using an `::after` element instead of an `img` child), which seems more semantically correct:

```
diff --git a/guacamole/src/main/frontend/src/app/home/styles/home.css b/guacamole/src/main/frontend/src/app/home/styles/home.css
index ef3178952..1a4ed9cfa 100644
--- a/guacamole/src/main/frontend/src/app/home/styles/home.css
+++ b/guacamole/src/main/frontend/src/app/home/styles/home.css
@@ -82,9 +82,19 @@ a.home-connection, .empty.balancer a.home-connection-group {
     box-shadow: none;
 }
 
+.recent-connections .connection .x-thumbnail::after {
+    content: '';
+    display: block;
+    height: 100%;
+    width: 100%;
+    background-repeat: no-repeat;
+    background-size: contain;
+    background-position: center center;
+    background-image: url('images/x.svg');
+}
+
 .recent-connections .connection .x-thumbnail {
     background-color: red;
-    display: flex;
     height: 1em;
     width: 1em;
     position: absolute;
@@ -97,4 +107,4 @@ a.home-connection, .empty.balancer a.home-connection-group {
 
 .recent-connections .connection .x-thumbnail:hover {
     opacity: 1.0;
-}
\ No newline at end of file
+}
diff --git a/guacamole/src/main/frontend/src/app/home/templates/guacRecentConnections.html b/guacamole/src/main/frontend/src/app/home/templates/guacRecentConnections.html
index 97207dbd1..50e07e623 100644
--- a/guacamole/src/main/frontend/src/app/home/templates/guacRecentConnections.html
+++ b/guacamole/src/main/frontend/src/app/home/templates/guacRecentConnections.html
@@ -20,7 +20,6 @@
         </a>
         <!-- Remove thumbnail -->
         <div class="x-thumbnail" ng-click="removeRecentConnection(recentConnection)">
-            <img alt="X" src="images/x.svg"/>
         </div>
     </div>
```

necouchman: Cool - reworked it with your suggest CSS code.

jmuehlner (approved):

PR #972, opened 2024-04-10
GUACAMOLE-1325: Migrate from javax to jakarta namespace.

necouchman:
Okay, well, here we go. This PR takes care of the migration from the `javax.*` namespaces to the new `jakarta.*` counterparts, updates dependencies, and updates the licenses. I've got it all in a single commit, now, because I couldn't find very clean places to break up the commit - dependency management at this level reminds me of the cat-in-bathtub gif.

The code does compile, I'm able to install it into Tomcat 10.1, and run it, log in (JDBC), and start Guacamole session. It's worth noting that this will be a big change in terms of compatibility:
* Right now I'm compiling and running with Java 11.x. I have not gone back to test 8.x, yet - I will, here, soon - but there's a reasonable chance that 8.x will no longer work with some of the dependency updates. 
* There were a couple of dependencies that I could not fully update because they required Java 17, and I wasn't willing to push quite to that level of breakage.
* This completely removes the Websocket support for Jetty8, Jetty9, and Tomcat 7, relying on the standard WebSocket support that should be built in to anything recent.
* I have not tested with anything besides Tomcat 10.x, yet. I intend to try Tomcat 11, JBoss, and Wildfly, at least.
* There is likely a decent amount of cleanup to be done, still, both in pom.xml files, and in the `doc/licenses` directory. I suspect I've left a lot of unused cruft around in getting things updated.
* I have not tested all of the extensions, yet - just JDBC. I'll try to get to several of them, soon, but there will likely be some I cannot test.
* I'm unsure about the future of the RADIUS module - I haven't even tried to compile that one, yet, and, given the age of the jRADIUS library, I'm not sure it can live on as it is. We'll can either find a different RADIUS library, write our own (ick, but not too difficult), drop RADIUS support altogether, or see if there are other options for maintaining compatibility with class loading and the deep magic of Java.

I'm going to put this in Draft mode for now - I welcome comments, requests for changes, and testing.

PR #973, opened 2024-04-11
GUACAMOLE-1289: Remove spring dependency and update dependencies with vulnerabilities.

aleitner:


aleitner:
Added a fix for the relogin bug mentioned in the previous Pull request https://github.com/apache/guacamole-client/pull/966#issuecomment-2050816886

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Credentials.java
```diff
@@ -84,9 +90,9 @@ public class Credentials implements Serializable {
 
     /**
      * Construct a Credentials object with the given username, password,
-     * and HTTP request.  The information is assigned to the various
-     * storage objects, and the remote hostname and address is parsed out
-     * of the request object.
+     * HTTP request. The information is assigned to 
+     * the various storage objects, and the remote hostname and address is
```
jmuehlner, line 94: The formatting on this comment is a bit wonky now - mind fixing it?

jmuehlner:
Please update the PR title to reflect that it's also fixing the re-login bug.

doc/licenses/spring-web-5.3.33/README
```diff
@@ -1,7 +1,7 @@
 Spring Framework (https://spring.io/projects/spring-framework)
 --------------------------------------------------------------
 
-    Version: 5.3.25
+    Version: 5.3.33
```
jmuehlner, line 4: [Dependabot ](https://github.com/apache/guacamole-client/security/dependabot/64) recommends 6.0.18. Why not use that version?

necouchman: I ran across this when working #972 - I believe it bumps minimum Java version to 17, and I'm reasonably certain it also requires the move to jakarta.* namespaces.

jmuehlner: Ah ok. Makes sense. It won't make dependabot fully happy, but I guess it's the best we can do.

jmuehlner:
> Remove spring dependency

Which spring dependency was removed? I see version updates but I don't think I see anything being fully removed.

doc/licenses/spring-web-5.3.33/README
mike-jumper: I think we can avoid Spring entirely here - I see it used in guacamole-auth-duo only for `UriComponentsBuilder`:

```java
String builtUrl = UriComponentsBuilder
        .fromUriString(redirectUrl)
        .queryParam(Credentials.RESUME_QUERY, DuoAuthenticationProvider.PROVIDER_IDENTIFER)
        .build()
        .toUriString();
```

I _think_ this can be rewritten to just use `UriBuilder`: https://javaee.github.io/javaee-spec/javadocs/javax/ws/rs/core/UriBuilder.html

doc/licenses/spring-web-5.3.33/README
jmuehlner, line 1: Why is this spring dependency still here, and updated? It looks like it's not even used anymore.

aleitner: Whoops. I was updating all the dependencies with vulnerabilities. Removed

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/Credentials.java
```diff
@@ -98,20 +104,43 @@ public class Credentials implements Serializable {
...
-        // If session exists get it, but don't create a new one.
-        this.session = request.getSession(false);
+    }
+
+    /**
+     * Checks if the current authentication process is a resumed one.
+     *
+     * @return True if authentication is resumed, otherwise false.
+     */
+    public Boolean isAuthenticationResumed() {
```
jmuehlner, line 132: It'd be nice if this could avoid going in `Credentials`, because it's pretty specific to Duo, and it's also not even the only way of resuming sessions - see the SSO (SAML, etc) extensions, which have their own [completely-different mechanism](https://github.com/apache/guacamole-client/blob/main/extensions/guacamole-auth-sso/modules/guacamole-auth-sso-base/src/main/java/org/apache/guacamole/auth/sso/SSOAuthenticationEventListener.java) for resuming interrupted auth attempts. 

Maybe the DUO extension could do something similar? (using a URL parameter to track state). If there could be some shared functionality for this, it would be ideal, though that seems unlikely considering how different the implementations are.

At the bare minimum, I think that the SSO extensions should at least _set_ this flag.

aleitner: The idea was to migrate sso to this method

jmuehlner: It looks like the DUO extension is already[ tracking state through URL parameters](https://github.com/apache/guacamole-client/blob/0e12f48c29bb872b2b73873427430286aa27202d/extensions/guacamole-auth-duo/src/main/java/org/apache/guacamole/auth/duo/UserVerificationService.java#L123).

jmuehlner: > The query strings are defined by the authentication provider modules so that we can avoid accidental overlap

FWIW, the ULR parameter for SAML is also called "state". Likely things won't work if people try to use DUO + SAML together. Might want to consider changing one or both of those.

Were you planning on migrating the SSO extensions to use this new flag in this PR?

aleitner: You're correct that the SAML extension also uses a `state` parameter, which could lead to confusion when both `DUO` and `SAML` are used concurrently, but I also added the `provider_id` as a required query parameter for resuming authentication. This acts as a namespacing mechanism to prevent collision. When redirecting from duo the query parameters will look like such: `provider_id=duo&state=abc123&duo_code=abc123`. 

This PR isn't meant for migrating the SSO extensions but their was the plan to eventually migrate them to use the same code for resuming authentication.

jmuehlner: It definitely feels like we shouldn't have the SSO extensions unmigrated for very long ... what do you think @mike-jumper ? Should migration of the SSO extensions to this new system be part of this ticket?

necouchman: I would agree that the SSO extension migration should be with this ticket - either this PR, or another PR to follow quickly after this one...

mike-jumper:
Closing in favor of recently-merged #980.

PR #974, opened 2024-04-12
GUACAMOLE-1068: Generate random key at each login if key is unconfirmed.

necouchman:
A simple take at randomizing the key for users who have not confirmed enrollment. I'm not sure if this introduces any other gotchas or security issues - it almost seems too simple?

### DIFF
--- a/extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/user/UserVerificationService.java
+++ b/extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/user/UserVerificationService.java
@@ -112,18 +112,9 @@ public class UserVerificationService {
 
         // If no key is defined, attempt to generate a new key
         String secret = attributes.get(TOTPUser.TOTP_KEY_SECRET_ATTRIBUTE_NAME);
-        if (secret == null || secret.isEmpty()) {
-
-            // Generate random key for user
-            TOTPGenerator.Mode mode = confService.getMode();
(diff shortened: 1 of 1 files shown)


mike-jumper:
Hm ... I'd think this would result in the TOTP key getting regenerated when the user attempts to enroll:

1. User attempts to log in.
2. Key is generated and presented to user for enrollment.
3. To confirm enrollment, the user resubmits credentials plus the TOTP code for _that_ key.
4. That's a login attempt and the user isn't confirmed yet, so the key is regenerated.
5. The TOTP code doesn't match the newly-generated key and the user is blocked from enrolling.

necouchman:
> Hm ... I'd think this would result in the TOTP key getting regenerated when the user attempts to enroll:
> 
> 1. User attempts to log in.
> 2. Key is generated and presented to user for enrollment.
> 3. To confirm enrollment, the user resubmits credentials plus the TOTP code for _that_ key.
> 4. That's a login attempt and the user isn't confirmed yet, so the key is regenerated.
> 5. The TOTP code doesn't match the newly-generated key and the user is blocked from enrolling.

Ah, okay - I thought it was probably too simple...and haven't thoroughly tested it, yet. Will convert to draft and take another look.

necouchman:
Okay, this change seems to work as expected, now.

jmuehlner (approved):
LGTM. I also tested it real quick and it does indeed seem to work as expected.

necouchman:
> LGTM. I also tested it real quick and it does indeed seem to work as expected.

Thanks!

PR #976, opened 2024-04-17
GUACAMOLE-1760: Add client-side options for compression and quality levels

necouchman:
Client-side changes to match apache/guacamole-server#503.


guacamole-ext/src/main/resources/org/apache/guacamole/protocols/vnc.json
```diff
@@ -58,6 +58,14 @@
                     "type"    : "BOOLEAN",
                     "options" : [ "true" ]
                 },
+                {
+                    "name" : "compress-level",
+                    "type" : "NUMERIC"
```
jmuehlner, line 63: The only valid options here are integers from 1-9, yeah? I wonder if these should be dropdowns instead of freeform fields.

necouchman: Yeah, I went back and forth about that, and then thought maybe it's a time to think about creating an option for the `NUMERIC` field that would allow placing limits on the ranges of the values.

jmuehlner: That's a neat idea, though I think it might make more sense when there's more options than this. This is quite a bit fewer options than e.g. font-size

```
{
    "name"  : "font-size",
    "type"  : "ENUM",
    "options" : [ "", "8", "9", "10", "11", "12", "14", "18", "24", "30", "36", "48", "60", "72", "96" ]
},```

necouchman: Yeah, I'm not sure this particular use-case warrants it, just something we might do for the future. Could limit things like port numbers, font sizes, timeout values, etc., to sane values without having to enumerate them the way the enum requires.

necouchman: Okay, converted to `ENUM` fields and added the English translations for the options.

PR #988, opened 2024-05-17
Added possibility to use a proxy configuration with the simpleconnect…

SteMu86:
Added possibility to use a proxy configuration with the simpleconnect…

### DIFF
--- a/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleConnection.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleConnection.java
@@ -39,6 +39,8 @@ import org.apache.guacamole.protocol.ConfiguredGuacamoleSocket;
 import org.apache.guacamole.protocol.GuacamoleClientInformation;
 import org.apache.guacamole.protocol.GuacamoleConfiguration;
 import org.apache.guacamole.token.TokenFilter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * A Connection implementation which establishes the underlying connection to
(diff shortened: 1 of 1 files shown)


guacamole-ext/src/main/java/org/apache/guacamole/net/auth/simple/SimpleConnection.java
```diff
@@ -201,13 +243,28 @@ public void setAttributes(Map<String, String> attributes) {
...
+                this.proxyConfig = LocalEnvironment.getInstance().getDefaultGuacamoleProxyConfiguration();
+            }
+            else {
+                this.proxyConfig = new GuacamoleProxyConfiguration(
+                            this.fullConfig.getParameter("proxy-host"),
+                            Integer.parseInt(this.fullConfig.getParameter("proxy-port")),
+                            GuacamoleProxyConfiguration.EncryptionMethod.NONE
+                    );
+                logger.debug("missplaced proxy configuration found. use it anyway.");
+            }
```
necouchman, line 261: I'm not really in favor of doing this - in this case, checking for "misplaced attributes in connection parameters" just encourages people to intentionally "misplace" attributes in (as) connection parameters. I don't think we should enable a mistaken behavior that 1) we don't really want, and 2) up to now hasn't been allowed.

necouchman (commented):
I've a couple of issues with this PR:
* See comment in the code below, but I'm not in favor of allowing people to use connection parameters as substitutes for attributes.
* The changes to the `SimpleConnection` class, minus the "misplaced" attributes, are largely a duplicate of #353 and will be implemented when that gets merged into the code, hopefully, soon, for the 1.6.0 release.
* If you'd like to find a way to extend `SimpleConnection` to allow for specifying the proxy configuration, and you think it should be done outside the scope of #353 and its associated Jira issue, then you'll need to create a Jira issue for that and to properly tag both the pull request and the commit messages with that issue.

PR #990, opened 2024-06-01
Fix SAML read Certificate and private key

kalaxcortez:
Please validate the reading of the certificate and private key files for SAML, it is only one line and it solves the entire signing issue, please validate and if you do not like the structure, change it but solve it, validate it locally and everything works, Greetings

### DIFF
--- a/extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/src/main/java/org/apache/guacamole/auth/saml/conf/ConfigurationService.java
+++ b/extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/src/main/java/org/apache/guacamole/auth/saml/conf/ConfigurationService.java
@@ -371,7 +371,13 @@ public class ConfigurationService {
      *     If the X.509 certificate cannot be parsed.
      */
     public File getCertificateFile() throws GuacamoleException {
-        return environment.getProperty(SAML_X509_CERT_PATH);
+    	File certificate = null;
+    	try {
+			certificate = environment.getProperty(SAML_X509_CERT_PATH).getCanonicalFile();
+		} catch (IOException | GuacamoleException e) {			
+			e.printStackTrace();
+		}
+        return certificate;
     }
 
     /**
@@ -387,7 +393,13 @@ public class ConfigurationService {
      *     If the private key file cannot be parsed.
      */
     public File getPrivateKeyFile() throws GuacamoleException {
-        return environment.getProperty(SAML_PRIVATE_KEY_PATH);
+    	File privateKey = null;
+    	try {
+    		privateKey = environment.getProperty(SAML_PRIVATE_KEY_PATH).getCanonicalFile();
+		} catch (IOException | GuacamoleException e) {			
+			e.printStackTrace();
+		}
+        return privateKey;
     }
 
     /**
@@ -480,7 +492,7 @@ public class ConfigurationService {
                     readFileContentsIntoString(privateKeyFile, "Private Key"));
 
         // If a certificate file is set, load the value into the builder now
-        File certificateFile = getCertificateFile();
+        File certificateFile = getCertificateFile();       
         if (certificateFile != null)
             samlMap.put(SettingsBuilder.SP_X509CERT_PROPERTY_KEY,
                     readFileContentsIntoString(certificateFile, "X.509 Certificate"));


necouchman:
@kalaxcortez Thank you for contributing, here. Before we can accept this contribution, a number of things need to be done:
* You need to request a Jira account and open a Jira issue, and your pull request and commit messages need to be tagged with the Jira issue. You can look at any of the existing closed, and many of the open pull requests for examples of this.
* You need to follow established style within the code. For example, you're cuddling braces (`} catch () {`) - they should be on new lines. Please adjust your style to conform to the rest of the code.
* I think the `IOException` handling would actually be better placed in the `FileGuacamoleProperty` class (in `guacamole-ext`), where it could easily be used by any and all users of the class, rather than just these two properties. It does not seem to be a situation unique to these particular usages of that property.
* You're catching both `IOException` and `GuacamoleException`, here, and printing the stack trace - this defeats the purpose of having this particular function throw the `GuacamoleException`. If you're going to catch `IOException`, it should be re-thrown as a `GuacamoleException`, and the throwing of `GuacamoleException` should be left alone.

> please validate and if you do not like the structure, change it but solve it

I believe your suggested changes have merit, and, with some tweaks, would be quite valuable to have added to the project. I'd highly encourage you to take the time to do the steps mentioned above and clean this up for inclusion - and, if you're so inclined, find other ways to contribute to the project.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/src/main/java/org/apache/guacamole/auth/saml/conf/ConfigurationService.java
```diff
@@ -371,7 +371,13 @@ public int getAuthenticationTimeout() throws GuacamoleException {
      *     If the X.509 certificate cannot be parsed.
      */
     public File getCertificateFile() throws GuacamoleException {
-        return environment.getProperty(SAML_X509_CERT_PATH);
+    	File certificate = null;
+    	try {
+			certificate = environment.getProperty(SAML_X509_CERT_PATH).getCanonicalFile();
```
mike-jumper, line 376: This will throw a `NullPointerException` if `getProperty()` returns `null` (if the property is not present). Same for the other call to `getCanonicalFile()`.

mike-jumper, line 376: What is the problem you're encountering that is fixed by calling `getCanonicalFile()`?

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/src/main/java/org/apache/guacamole/auth/saml/conf/ConfigurationService.java
```diff
@@ -371,7 +371,13 @@ public int getAuthenticationTimeout() throws GuacamoleException {
...
+		} catch (IOException | GuacamoleException e) {			
+			e.printStackTrace();
```
mike-jumper, line 378: If an error should be logged, it should be logged by a `Logger` instance with the stack trace logged only at the debug level. We typically do this by logging things twice, with an admin-facing message at the error level and a developer-facing message at the debug level.

That said, I think the proper thing to do here would be to just rethrow the `IOException` as a `GuacamoleException`, with the `GuacamoleException` adding a meaningful message.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/src/main/java/org/apache/guacamole/auth/saml/conf/ConfigurationService.java
```diff
@@ -480,7 +492,7 @@ public Saml2Settings getSamlSettings() throws GuacamoleException {
                     readFileContentsIntoString(privateKeyFile, "Private Key"));
 
         // If a certificate file is set, load the value into the builder now
-        File certificateFile = getCertificateFile();
+        File certificateFile = getCertificateFile();
```
mike-jumper, line 495: It looks like this line was changed only by adding several spaces to the end.

mike-jumper (changes requested):

adb014:
Two years without feedback on "why" this change is needed, it should probably be rejected to clear up the list of pull requests

PR #991, opened 2024-06-03
GUACAMOLE-1956: Update to latest compatible dependency versions.

jmuehlner:


guacamole/src/main/frontend/package.json
jmuehlner, line 1: Packages that are not upgraded:
* Webpack along with all associated plugins and loaders
  `angular-templatecache-webpack-plugin` requires webpack 4, and we are already on the latest version of all webpack-4 compatible packages.
* `@simonwep/pickr`
     Version 1.9.1 is available, but fails to compile/parse: 
     ```
     [INFO] ERROR in ./node_modules/@simonwep/pickr/dist/pickr.min.js 2:9541
    [INFO] Module parse failed: Unexpected token (2:9541)
    [INFO] You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
    [INFO] | /*! Pickr 1.9.1 MIT | https://github.com/Simonwep/pickr */
    [INFO] > !function(t,e){"object"==typeof exports&&"object"==typeof module?module.exports=e():"function"==typeof define&&define.amd?define([],e):"object"==typeof exports?exports.Pickr=e():t.Pickr=e()}(self,(()=>(()=>{"use strict";var t={d:(e,o)=>{for(var n in o)t.o(o,n)&&!t.o(e,n)&&Object.defineProperty(e,n,{enumerable:!0,get:o[n]})},o:(t,e)=>Object.prototype.hasOwnProperty.call(t,e),r:t=>{"undefined"!=typeof Symbol&&Symbol.toStringTag&&Object.defineProperty(t,Symbol.toStringTag,{value:"Module"}),Object.defineProperty(t,"__esModule",{value:!0})}},
    ...
     ```

necouchman:
@jmuehlner : I think this should go to the `patch` branch, correct?

jmuehlner:
> @jmuehlner : I think this should go to the `patch` branch, correct?

Yep. I just haven't retargeted it yet.

doc/licenses/bouncycastle-fips-1.0.2.4/README
mike-jumper, line 1: The directory containing this should be renamed to match the version.

doc/licenses/mybatis-guice-4.0.0/README
```diff
@@ -1,7 +1,7 @@
 MyBatis-Guice (http://www.mybatis.org/guice/)
 ---------------------------------------------
 
-    Version: 3.18
+    Version: 4.0.0
```
mike-jumper, line 4: This is a pretty big jump. Checking the dependencies for MyBatis-Guice 4.0.0, it looks like they may have migrated things from `javax.*` to `jakarta.*`, which might not be compatible if any of those classes will be provided by Tomcat. Can you confirm whether this is the case?

I haven't yet checked whether there are any dependencies that would break compatibility with Java 8, but that would be another thing to verify here.

mike-jumper (changes requested):

doc/licenses/mybatis-guice-4.0.0/README
jmuehlner: Ah yeah, good call. I'll revert this one.

mike-jumper (approved):

PR #993, opened 2024-06-15
GUACAMOLE-1290: Add parameter for SSH public key.

necouchman:


guacamole-ext/src/main/resources/org/apache/guacamole/protocols/ssh.json
```diff
@@ -38,6 +38,10 @@
                 {
                     "name"  : "passphrase",
                     "type"  : "PASSWORD"
+                },
+                {
+                    "name"  : "public-key"
```
mike-jumper, line 43: Is there a corresponding translation for the field header?

mike-jumper (changes requested):

guacamole-ext/src/main/resources/org/apache/guacamole/protocols/ssh.json
necouchman: Yeah, should probably include that. Fixed via rebase.

guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -741,6 +742,7 @@
         "FIELD_HEADER_PASSPHRASE"    : "Passphrase:",
         "FIELD_HEADER_PORT"          : "Port:",
         "FIELD_HEADER_PRIVATE_KEY"   : "Private key:",
+	"FIELD_HEADER_PUBLIC_KEY"    : "Public key:",
```
mike-jumper, line 745: Eek! A tab!

Otherwise, LGTM.

mike-jumper (changes requested):

guacamole/src/main/frontend/src/translations/en.json
necouchman: Whoa, not sure how that snuck in there. Fixed via rebase.

mike-jumper (approved):

PR #995, opened 2024-07-06
GUACAMOLE-1957: Support granular permissions management.

necouchman:
I've taken a stab at making the permissions management a bit more granular, allowing the four possible object-level permissions to be managed individually for connections, connection groups, and sharing profiles. As usual, I could use suggestions for how the GUI might be better set up to support this - I was originally going to try to create a table; however, going that route would require reworking the entire permissions management system, possibly implementing some sort of `guac-group-table` directive or something like that. So, I opted for a little simpler route.

Here's how it looks in practice:
image


mike-jumper:
Interesting!

necouchman:
> Interesting!

"Interesting" can be such an ambiguous word. :rofl:

necouchman:
> > > Interesting!
> > 
> > 
> > "Interesting" can be such an ambiguous word. 🤣
> 
> Heh - true. I think I like this approach, though. I'm curious how well it scales for larger numbers of connections.

Definitely a good point - I don't have all that many to test with in a test environment at the moment, so not sure how well the pagination works with it at the moment.

PR #996, opened 2024-07-11
GUAC-1967: Remove duplicate wheel handler registrations.

neandrake:
Modern-day browsers abide the standardized `wheel` event (typed `WheelEvent`), however to support older browsers the legacy `DOMMouseScroll` (FireFox) and `mousewheel` (all others) events are also being registered.

However modern browsers support both the `wheel` event as well as the legacy, which results in the wheel handler being registered multiple times. This results in the wheel handler firing twice instead of once, causing excessive scrolling.

This change separates the event registration such that either the modern or the legacy events are registered, and not both.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Mouse.js
+++ b/guacamole-common-js/src/main/webapp/modules/Mouse.js
@@ -258,9 +258,16 @@ Guacamole.Mouse = function Mouse(element) {
 
     }
 
-    element.addEventListener('DOMMouseScroll', mousewheel_handler, false);
-    element.addEventListener('mousewheel',     mousewheel_handler, false);
-    element.addEventListener('wheel',          mousewheel_handler, false);
+    if (window.WheelEvent) {
+        // All modern browsers support wheel events.
+        element.addEventListener('wheel', mousewheel_handler, false);
+    }
+    else {
+        // Legacy FireFox wheel events.
+        element.addEventListener('DOMMouseScroll', mousewheel_handler, false);
+        // Legacy Chrome/IE/other wheel events.
+        element.addEventListener('mousewheel', mousewheel_handler, false);
+    }
 
     /**
      * Whether the browser supports CSS3 cursor styling, including hotspot


neandrake:
Oops my editor removed trailing spaces automatically. Let me know if that's a problem and I'll see to get that reverted.

guacamole-common-js/src/main/webapp/modules/Mouse.js
```diff
@@ -299,7 +305,7 @@ Guacamole.Mouse = function Mouse(element) {
      * hotspot coordinates. This affects styling of the element backing this
      * Guacamole.Mouse only, and may fail depending on browser support for
      * setting the mouse cursor.
-     *
```
jmuehlner, line 302: Please remove these whitspace-only changes. I know that removing trailing whitespace improves the formatting, but it clutters up the git history. 

Formatting changes like this should only be included if the lines in question are actually being meaningfully modified.

guacamole-common-js/src/main/webapp/modules/Mouse.js
```diff
@@ -258,9 +258,15 @@ Guacamole.Mouse = function Mouse(element) {
 
     }
 
-    element.addEventListener('DOMMouseScroll', mousewheel_handler, false);
-    element.addEventListener('mousewheel',     mousewheel_handler, false);
-    element.addEventListener('wheel',          mousewheel_handler, false);
+    if (window.WheelEvent) {
+        // All modern browsers support wheel events.
+        element.addEventListener('wheel',          mousewheel_handler, false);
+    } else {
```
necouchman, line 264: Please don't cuddle the braces...
```
{
else {
```

guacamole-common-js/src/main/webapp/modules/Mouse.js
```diff
@@ -258,9 +258,15 @@ Guacamole.Mouse = function Mouse(element) {
 
     }
 
-    element.addEventListener('DOMMouseScroll', mousewheel_handler, false);
-    element.addEventListener('mousewheel',     mousewheel_handler, false);
-    element.addEventListener('wheel',          mousewheel_handler, false);
+    if (window.WheelEvent) {
+        // All modern browsers support wheel events.
+        element.addEventListener('wheel',          mousewheel_handler, false);
```
necouchman, line 263: Because these lines are now separated by `if/else` blocks and comments, I don't know that there's any reason to have all of the extra spaces between the `'wheel',` and the `mousewheel_handler` - both this line, and below.

necouchman (changes requested):
Just a couple of style changes, and I think it's good.

necouchman (approved):
@jmuehlner Any further concerns?

jmuehlner:
> @jmuehlner Any further concerns?

Nope, LGTM.

necouchman:
@neandrake Actually, one more request - can you please rebase against the `apache:patch` branch (instead of `main`)?

neandrake:
> @neandrake Actually, one more request - can you please rebase against the `apache:patch` branch (instead of `main`)?

I've no idea how to do that. I opened a new pull request on the patch branch, #997

PR #997, opened 2024-07-11
GUACAMOLE-1967: Remove duplicate wheel handler registrations.

neandrake:
Modern-day browsers abide the standardized wheel event (typed `WheelEvent`), however to support older browsers the legacy `DOMMouseScroll` (FireFox) and `mousewheel` (all others) events are also being registered.

However modern browsers support both the modern event as well as the legacy, which results in the wheel handler being registered multiple times. This results in the wheel handler firing twice instead of once, causing excessive scrolling.

This change separates the event registration such that either the modern or the legacy events are registered, and not both.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Mouse.js
+++ b/guacamole-common-js/src/main/webapp/modules/Mouse.js
@@ -258,9 +258,16 @@ Guacamole.Mouse = function Mouse(element) {
 
     }
 
-    element.addEventListener('DOMMouseScroll', mousewheel_handler, false);
-    element.addEventListener('mousewheel',     mousewheel_handler, false);
-    element.addEventListener('wheel',          mousewheel_handler, false);
+    if (window.WheelEvent) {
+        // All modern browsers support wheel events.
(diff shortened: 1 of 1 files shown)


necouchman:
Looks good - one more thing - the title of the Pull Request and commit messages should include "GUACAMOLE-1967" - please don't shorten it to GUAC-1967.

neandrake:
Oh could I have "rebased" the earlier PR by editing directly in GitHub to target the `patch` version here? I just noticed that. Sorry for the GitHub fumbling.

necouchman:
> Oh could I have "rebased" the earlier PR by editing directly in GitHub to target the `patch` version here? I just noticed that. Sorry for the GitHub fumbling.

No worries. Rebasing in git can be challenging.

The commit message still needs to be fixed for this PR - you can "git rebase -i <COMMIT BEFORE YOUR COMMIT>" and then use the "reword" option to change the message. Then "git push -f".

neandrake:
The commit is also fixed now. Thank you.

necouchman (approved):

PR #998, opened 2024-07-22
changes

Saran-Madhivanan:


necouchman:
@Saran-Madhivanan : If you're wanting to make changes to the main repositories of the Guacamole source code, you need to open a Jira ticket documenting what changes you're wanting to make and why, and then you need to tag the pull request with those changes.

Also, if you look at the rest of the Guacamole code, you will see that we do not use `System.out.print*` methods - we use the SLF4J logging classes `Logger` and `LoggerFactory`.

https://guacamole.apache.org/open-source/

stcbus:
lol looks like we got another "Jia Tan" around. Nice try.

necouchman:
@stcbus Let's not be too hasty to cast blame like that - looking at Saran-Madhivanan's fork of the guacamole-client repo, it looks like they are working on legitimate changes to the code to fit into some application or variant of Guacamole that they are developing. I suspect that opening the pull request to this repo was just a mistake after making the first of those changes (it was the first commit on top of the existing `main` branch).

I just want to set expectations early about how changes should be integrated into the code, in case there is legitimate interest in pushing something into the source repo :-).

PR #1001, opened 2024-07-30
GUACAMOLE-1974: Allow deferring received pipe streams for later consumption by name-specific handlers.

jmuehlner:
This change modifies the `ManagedClient` class such that pipe streams prior to an `onpipe` handler being set can still be processed. 

The idea is that you can register a pipe stream handler for a particular pipe stream name, and that handler will be reliably invoked when a new pipe stream with that name is received, OR if such a pipe stream was already received prior to registration.

### DIFF
--- a/guacamole/src/main/frontend/src/app/client/types/ManagedClient.js
+++ b/guacamole/src/main/frontend/src/app/client/types/ManagedClient.js
@@ -63,11 +63,48 @@ angular.module('client').factory('ManagedClient', ['$rootScope', '$injector',
     var THUMBNAIL_UPDATE_FREQUENCY = 5000;
 
     /**
+     * A deferred pipe stream, that has yet to be consumed, as well as all
+     * axuilary information needed to pull data from the stream.
+     *
+     * @constructor
+     * @param {DeferredPipeStream|Object} [template={}]
+     *     The object whose properties should be copied within the new
+     *     DeferredPipeStream.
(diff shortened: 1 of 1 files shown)


aleitner:
Looks solid to me! But I'd like another set of eyes on it

necouchman:
I'm also good with the changes - looks pretty straight-forward to me, though admittedly JavaScript is not my strong point.

necouchman (approved):

guacamole/src/main/frontend/src/app/client/types/ManagedClient.js
```diff
@@ -1004,6 +1076,72 @@ angular.module('client').factory('ManagedClient', ['$rootScope', '$injector',
...
+     *     The client for which the deferred pipe stream handler should be
+     *     detached.
+     */
+    ManagedClient.detachDeferredPipeHandler = function detachDeferredPipeHandler(
+        managedClient, name, handler) {
+
+        // Remove the handler if found
+        if (managedClient.deferredPipeStreamHandlers[name] === handler)
+            delete managedClient.deferredPipeStreamHandlers[name];
+    }
```
mike-jumper, line 1143: Missing semicolon here.

jmuehlner: Fixed via rebase.

PR #1002, opened 2024-07-31
GUACAMOLE-1642: Add translation string for encoding field.

necouchman:


guacamole/src/main/frontend/src/translations/en.json
```diff
@@ -911,6 +911,7 @@
         "FIELD_HEADER_DISABLE_SERVER_INPUT" : "Disable server input when client is connected:",
         "FIELD_HEADER_ENABLE_AUDIO"     : "Enable audio:",
         "FIELD_HEADER_ENABLE_SFTP"      : "Enable SFTP:",
+	"FIELD_HEADER_ENCODINGS"	: "Display encodings:",
```
mike-jumper, line 914: Oops - a wild tab appears!

mike-jumper (changes requested):

guacamole/src/main/frontend/src/translations/en.json
necouchman: Oops...that's what I get for switching editors for a simple change. Should be fixed, now.

PR #1004, opened 2024-08-09
GUACAMOLE-1976: Add OPTIONAL token modifier.

necouchman:


guacamole-ext/src/main/java/org/apache/guacamole/token/TokenFilter.java
```diff
@@ -225,15 +225,21 @@ private String filter(String input, boolean strict)
                 // strict mode is enabled
                 if (tokenValue == null) {
 
+                    // Token marked as optional, so just skip it and update
+                    // last match.
+                    if (modifier != null && modifier.equals("OPTIONAL")) {
```
jmuehlner, line 230: I wonder if it's worth putting in a warning if one of the provided tokens contains `"OPTIONAL"` in it somewhere. There's nothing that would stop a caller from doing that, and it might cause some unexpected behavior.

necouchman: You mean just a `Logger` warning, to the log file? Or you thinking some sort of on-screen warning?

jmuehlner: Just the standard logging system seems fine - it wouldn't be a very common occurence, and the consequences are not so dire.

necouchman: Added via rebase.

jmuehlner: Hmm, on second thought, maybe this should be at debug level? If used for its intended purpose this would make the logs pretty spammy. Sorry for the trouble.

necouchman: Yeah, that's probably a good call - depending on the number of connections and number of times this occurs, a `warn` message could get quite frequent. Changed via rebase.

PR #1007, opened 2024-08-22
GUACAMOLE-1979: Allow setting required properties for connecting to MySQL 8.4 and later.

jmuehlner:
These properties allow connecting to a MySQL 8.4+ database server. 

Either the public key can be retrieved from the server and set using the `mysql-server-rsa-public-key-file` property, or the public key can be automatically retrieved by setting the `mysql-allow-public-key-retrieval` property.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/MySQLAuthenticationProviderModule.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-mysql/src/main/java/org/apache/guacamole/auth/mysql/MySQLAuthenticationProviderModule.java
@@ -76,6 +76,10 @@ public class MySQLAuthenticationProviderModule implements Module {
         myBatisProperties.setProperty("mybatis.pooled.pingEnabled", "true");
         myBatisProperties.setProperty("mybatis.pooled.pingQuery", "SELECT 1");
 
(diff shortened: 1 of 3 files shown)


necouchman:
Changes look fine to me. Do older drivers and/or MySQL/MariaDB servers that don't support these properties react okay if/when they are set?

jmuehlner:
> Changes look fine to me. Do older drivers and/or MySQL/MariaDB servers that don't support these properties react okay if/when they are set?

Yep, confirmed that all of these new properties don't do anything when connecting to a database that doesn't support them. I tested with MySQL 5.3 and tried setting `allowPublicKeyRetrieval` to both "true" and "false", as well as setting `serverRsaPublicKeyFile` to both valid and invalid file paths, and everything was ignored.

I also tried setting various invalid/unknown properties with both the MySQL and MariaDB drivers and they were simply ignored.

necouchman:
@jmuehlner One other question - this is currently targeted at the `main` branch - wasn't sure if the aim is to get this into the 1.6.0 release, or you're good with it waiting until the next one? Don't know if you're considering it a bug or not?

jmuehlner:
> @jmuehlner One other question - this is currently targeted at the `main` branch - wasn't sure if the aim is to get this into the 1.6.0 release, or you're good with it waiting until the next one? Don't know if you're considering it a bug or not?

Yeah, I was thinking it might be better to avoid increasing the scope of 1.6.0 further. This isn't really a bug exactly - it's more of an improvement enabling the use of later MySQL versions.

PR #1008, opened 2024-08-25
GUACAMOLE-1866: Clean up style of recent connection removal button.

mike-jumper:
This change restyles the recent connection removal button such that:

* It's a gray-colored "X" that becomes black when hovered.
* The button itself is not visible unless the associated recent connection is being hovered.


mike-jumper:
Closing in favor of broader UI cleanup around the newly configurable "Recent Connections" section.

PR #1013, opened 2024-08-28
GUACAMOLE-1980: Update client versions to 1.6.0.

jmuehlner:


mike-jumper (changes requested):
Need to also bump:

https://github.com/apache/guacamole-client/blob/addac2a263190d028501d191fb455d60d71c6a14/guacamole-common-js/src/main/webapp/modules/Version.js#L30

jmuehlner:
> Need to also bump:
> 
> https://github.com/apache/guacamole-client/blob/addac2a263190d028501d191fb455d60d71c6a14/guacamole-common-js/src/main/webapp/modules/Version.js#L30

It's done

mike-jumper (approved):

PR #1014, opened 2024-08-30
GUACAMOLE-61: New JDBC tokens should follow the standard naming conventions.

necouchman:
This tweaks the recently-added JDBC tokens to following the naming conventions that the other extensions use.


mike-jumper (approved):
Very nice.

PR #1022, opened 2024-10-03
GUACAMOLE-1020: Perform some additional null checks in restriction verification service.

necouchman:
I've added a few null checks and re-arranged a couple of other checks to address some issues that Coverity flagged in the recently-committed login restriction code.

### DIFF
--- a/extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/RestrictionVerificationService.java
+++ b/extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/RestrictionVerificationService.java
@@ -135,9 +135,6 @@ public class RestrictionVerificationService {
     public static RestrictionType allowedByHostRestrictions(String allowedHostsString,
             String deniedHostsString, String remoteAddress) {
         
-        // Convert the string to a HostName
-        HostName remoteHostName = new HostName(remoteAddress);
-        
         // If attributes do not exist or are empty then the action is allowed.
         if ((allowedHostsString == null || allowedHostsString.isEmpty()) 
                 && (deniedHostsString == null || deniedHostsString.isEmpty()))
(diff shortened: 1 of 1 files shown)


extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/RestrictionVerificationService.java
```diff
@@ -152,19 +149,42 @@ public static RestrictionType allowedByHostRestrictions(String allowedHostsStrin
...
+        // if a match is found, or if an error occurs in processing a host in
+        // the list.
         List<HostName> deniedHosts = HostRestrictionParser.parseHostList(deniedHostsString);
         for (HostName hostName : deniedHosts) {
+
+            if (hostName == null) {
+                LOGGER.warn("Host name is null, denying this action.");
+                LOGGER.debug("Received NULL host object, returning an implicit deny.");
+                return RestrictionType.IMPLICIT_DENY;
+            }
```
mike-jumper, line 175: I'm not sure this is possible either - the value of `hostName` here would come from the result of `HostRestrictionParser.parseHostList()`, but that function returns a `List<HostName>` containing nothing but non-null `HostName` objects.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/RestrictionVerificationService.java
```diff
@@ -182,14 +202,21 @@ public static RestrictionType allowedByHostRestrictions(String allowedHostsStrin
         // action if there are any matches.
         List<HostName> allowedHosts = HostRestrictionParser.parseHostList(allowedHostsString);
         for (HostName hostName : allowedHosts) {
+
+            if (hostName == null) {
+                LOGGER.warn("Host name from allow list is null, skipping and moving to the next.");
+                LOGGER.debug("Receive a null value from the list of allowed hosts, skipping the entry.");
+                continue;
+            }
```
mike-jumper, line 210: Same here - I'm not sure it's possible for `parseHostList()` to return a `List<HostName>` containing `null` entries.

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/RestrictionVerificationService.java
```diff
@@ -152,19 +149,42 @@ public static RestrictionType allowedByHostRestrictions(String allowedHostsStrin
...
+                if (hostName.isAddress()
+                        && hostName.toAddress().contains(remoteHostName.asAddress())) {
                     return RestrictionType.EXPLICIT_DENY;
+                }
 
-                else
+                else {
                     for (IPAddress currAddr : hostName.toAllAddresses())
-                        if (currAddr.matches(remoteHostName.asAddressString()))
+                        if (currAddr != null && currAddr.matches(remoteHostName.asAddressString()))
```
mike-jumper, line 185: Checking the documentation of `toAllAddresses()`, it doesn't look like this function produces `null` results. It's documented as similar to `asAddress()`, except that it returns an empty array instead of `null`:

https://seancfoley.github.io/IPAddress/IPAddress/apidocs/inet/ipaddr/HostName.html#toAllAddresses--

extensions/guacamole-auth-restrict/src/main/java/org/apache/guacamole/auth/restrict/RestrictionVerificationService.java
```diff
@@ -182,14 +202,21 @@ public static RestrictionType allowedByHostRestrictions(String allowedHostsStrin
...
+
             try {
                 // If the entry is an IP or Subnet, check the remote address against it directly
                 if (hostName.isAddress() && hostName.toAddress().contains(remoteHostName.asAddress()))
                     return RestrictionType.EXPLICIT_ALLOW;
                 
                 // Entry is a hostname, so resolve to IPs and check each one
                 for (IPAddress currAddr : hostName.toAllAddresses())
-                    if (currAddr.matches(remoteHostName.asAddressString()))
+                    if (currAddr != null && currAddr.matches(remoteHostName.asAddressString()))
```
mike-jumper, line 219: Same here - I think these may all be false positives.

necouchman:
> Looking these changes, I'm not sure where Coverity saw the potential for `null` values. Could these be all false positives?

They seemed a little false-positive to me, but who am I to question Coverity?!

mike-jumper (approved):

PR #1024, opened 2024-10-07
GUACAMOLE-1989: Add the ability to open guacamole as an independent application.

corentin-soriano:
The current logos are too low quality, so I sourced the svg from the Guacamole website. If you have larger ones, I can modify my pull request.


necouchman:
@corentin-soriano I think the logos from the website repo should be fine.

guacamole/src/main/frontend/src/manifest.json
```diff
@@ -0,0 +1,24 @@
...
+      "type": "image/png",
+      "sizes": "512x512"
+    },
+    {
+      "src": "./images/logo-192.png",
+      "type": "image/png",
+      "sizes": "192x192"
+    }
+  ]
+}
```
necouchman, line 24: Any reason this is spaces with 2 character indentations, instead of the 4 characters that is common across (most of) the rest of the Guacamole source code?

corentin-soriano: Bad configuration of my IDE, I corrected it!

necouchman (approved):
Looks good. Waiting for the automated build check to complete, and then can merge the code.

PR #1026, opened 2024-10-11
GUACAMOLE-1239: Correct regressions introduced by case sensitive checks for usernames.

necouchman:
This corrects the regressions introduced by #902 in implementing case-sensitive usernames, correcting the `@Parameter` annotations in a couple of places and moving the `caseSensitive` parameter up to the base mapper class instead of overriding the `UserMapper` class. It also moves the warning for the case sensitivity check for MySQL and SQL Server to the Guice module class so that it is only shown once when the module is loaded and not repeatedly for each query (as discussed on the Jira issue).

Note that I left the documentation for the `caseSensitive` parameter generic in the base class, and fairly ambiguous - if someone thinks I should put more verbage there around the fact that it currently only applies to usernames, I'm happy to do that, just trying to leave it as generic as possible sense it is in the base class.


extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
```diff
@@ -510,7 +530,7 @@ public void deleteObject(ModeledAuthenticatedUser user, String identifier)
         beforeDelete(user, identifier);
         
         // Delete object
-        getObjectMapper().delete(identifier);
+        getObjectMapper().delete(identifier, environment.getCaseSensitiveUsernames());
```
mike-jumper, line 533: Since this is the base class, I think this should be `getCaseSensitiveIdentifiers()`, not `environment.getCaseSensitiveUsernames()`.

mike-jumper (changes requested):
Everything else looks good and I like the generic rewording.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/base/ModeledDirectoryObjectService.java
necouchman: Thanks for catching that - I've corrected this in a rebase.

necouchman:
> Everything else looks good and I like the generic rewording.

I need to take one more look at the `ObjectRelationMapper` and related interfaces to see if I need to move those changes down to the base class and do something similar there. Then it should be ready for review/merge.

necouchman:
@mike-jumper Okay, ready for final review, I believe...

mike-jumper (approved):

PR #1040, opened 2024-12-28
Added Russian keyboard layout

VAGurko:


.gitignore
```diff
@@ -1,2 +1,5 @@
 *~
 target/
+.gigaide/
+.idea/
+guacamole-client.iml
```
necouchman, line 5: These changes should not be included as part of this change. If we need to add entries to .gitignore, those should be handled via a specific Jira issue and pull request.

necouchman (changes requested):
@VAGurko : Thank you for contributing this code. There are a couple of things that need addressing outside of the one comment below:
* All changes need a Jira issue associated with them, which means you'll need to go to our Jira page and request an account. Once that's created you'll need to create a Jira ticket for the request and tag both the Pull Request and the commit messages with that Jira issue. You can see this in action in the other pull requests opened for the project.
* You have four commit messages, including a merge commit, for what is a relatively  minor change. The merge commit needs to be removed, and the other commits should probably just be squashed together.
* The commit messages need to be meaningful and unique. If you're going to have multiple commits, having them all say, "Added Russian keyboard layout" doesn't help to determine what each commit is actually accomplishing. The commit messages should be descriptive of what is happening in the change, for example: "Added Russian keyboard layout to the RDP client settings."

For more complete guidance on contributing to the project, please see: https://guacamole.apache.org/open-source/.

PR #1041, opened 2025-01-02
GUACAMOLE-1956: (Re-)update to latest compatible dependencies.

mike-jumper:
This change applies further updates to the Java and JavaScript dependencies, while also rolling back some of the updates that were made to SAML (these resulted in Java 11 being required).


extensions/guacamole-auth-sso/modules/guacamole-auth-sso-cas/pom.xml
```diff
@@ -77,9 +77,20 @@
                     <artifactId>commons-codec</artifactId>
                 </exclusion>
 
+                <exclusion>
```
necouchman, line 80: Might be useful to document the reason for the exclusions, here?

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/pom.xml
```diff
@@ -143,14 +143,19 @@
                     <artifactId>commons-codec</artifactId>
                 </exclusion>
 
+                <exclusion>
```
necouchman, line 146: Might be useful to have a comment documenting the reason for the exclusions?

necouchman (changes requested):
Overall not many concerns - I think i'd be good to add some comments documenting the reason for the added exclusions in a couple of the pom.xml files. Should be ready to :shipit: after that.

mike-jumper:
Sure - I'll add the additional comments.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-saml/pom.xml
mike-jumper: Fixed via rebase and simply removing the exclusion (no longer necessary - xmlsec actually directly depends on this version of Woodstox).

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-cas/pom.xml
mike-jumper: Fixed via rebase.

necouchman (approved):

PR #1045, opened 2025-01-13
GUACAMOLE-377: Revert to synchronous flush (asynchronous is slower).

mike-jumper:
This partly reverts the changes introduced by d6db8fac7ea01d8a6a4a8c2b8bc84b66be4fb471 (#681), restoring the previous client-side frame handling.

The new asynchronous flush mechanism leveraging `requestAnimationFrame()` does not perform as well as the old synchronous flush. This appears to be due to delays in when the browser actually allows the frame to proceed, which cause the client to lag behind.

A future change that could improve things further might be to process graphical changes in a WebWorker when available.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Display.js
+++ b/guacamole-common-js/src/main/webapp/modules/Display.js
@@ -186,19 +186,6 @@ Guacamole.Display = function() {
     var frames = [];
 
     /**
-     * The ID of the animation frame request returned by the last call to
-     * requestAnimationFrame(). This value will only be set if the browser
-     * supports requestAnimationFrame(), if a frame render is currently
-     * pending, and if the current browser tab is currently focused (likely to
-     * handle requests for animation frames). In all other cases, this will be
-     * null.
-     *
-     * @private
-     * @type {number}
-     */
-    var inProgressFrame = null;
-
-    /**
      * Flushes all pending frames synchronously. This function will block until
      * all pending frames have rendered. If a frame is currently blocked by an
      * asynchronous operation like an image load, this function will return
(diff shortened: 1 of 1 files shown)


mike-jumper:
I've added some additional changes here that should significantly help with latency. By using the new `ImageDecoder` object where available, we can decode image streams as they are received without having to wait for the whole image contents to become available.

guacamole-common-js/src/main/webapp/modules/InputStream.js
```diff
@@ -76,4 +76,66 @@ Guacamole.InputStream = function(client, index) {
         client.sendAck(guac_stream.index, message, code);
     };
 
+    /**
+     * Creates a new ReadableStream that receives the data sent to this stream
+     * by the Guacamole server. This function may be invoked at most once per
+     * stream, and invoking this function will overwrite any installed event
+     * handlers on this stream.
+     * 
+     * A ReadableStream is a JavaScript object defined the "Streams" standard.
```
necouchman, line 85: Missing either `in` or `by` in this comment: `defined in the "Streams" standard.`

guacamole-common-js/src/main/webapp/modules/InputStream.js
```diff
@@ -76,4 +76,66 @@ Guacamole.InputStream = function(client, index) {
...
+     * It is supported by most browsers, but not necessarily all browsers.
+     * The caller should verify this support is present before invoking this
+     * function. The behavior of this function then the browser does not
```
necouchman, line 88: `then` -> `when`

DanielMcAssey:
Additionally, another future improvement is use of [OffscreenCanvas](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas) with a Worker

necouchman (changes requested):
Just the two comment issues identified above, and I think this is ready to go.

necouchman (approved):
@mike-jumper - I'm good with it. Not sure if there's anything that @DanielMcAssey mentioned that's worth addressing, or if you want to keep it as-is?

mike-jumper:
I think it's find to keep as-is. I've been experimenting with using a WebWorker for rendering (and, implicitly, OffscreenCanvas), but I don't think there's any need to hold off on these improvements.

If the WebWorker approach does improve things further, I think it's OK for those improvements to be after 1.6.0. The current approach is already significantly faster.

PR #1053, opened 2025-01-22
GUACAMOLE-2021: Fix session recording playback heatmap for short videos.

eugen-keeper:
I believe we can simply limit minimum `maxPossibleBucketValue` to 1 instead of having variable `NUM_BUCKETS`.


corentin-soriano (commented):
I think this fix could be added to the patch branch since the related functionality will be present in 1.6.0.

corentin-soriano (approved):

PR #1055, opened 2025-01-26
GUACAMOLE-2025: Add a button to close the recording video player.

corentin-soriano:
Currently the only way to close the recording video player is to use the browser's "back" button.
Add a button that automatically hides with the control bar:
image

### DIFF
--- a/guacamole/src/main/frontend/src/app/player/styles/player.css
+++ b/guacamole/src/main/frontend/src/app/player/styles/player.css
@@ -83,6 +83,7 @@ guac-player .guac-player-controls {
     margin: 0;
 }
 
+.guac-player-controls .guac-player-close:hover,
 .guac-player-controls .guac-player-play:hover,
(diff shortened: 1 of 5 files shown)


guacamole/src/main/frontend/src/app/player/styles/player.css
```diff
@@ -113,6 +114,31 @@ guac-player .guac-player-controls {
...
+    background: transparent;
+    border: 0;
+    box-shadow: none;
+    margin: 0;
+    min-width: 0;
+    padding: 0;
+    position: fixed;
+    right: 0;
+    top: 0;
+}
```
corentin-soriano, line 127: z-index needs to be fixed so that it does not go over the fullscreen button:
Capture d'écran 2025-04-09 193138

corentin-soriano: Size adjusted so that the close button is always visible:
image

PR #1057, opened 2025-02-07
GUACAMOLE-2036: Implement the GuacamoleParser within the ReaderGuacamoleReader.

aleitner:
Implement the GuacamoleParser within the ReaderGuacamoleReader to support multi byte characters when reading instructions.

### DIFF
--- a/guacamole-common/src/main/java/org/apache/guacamole/io/ReaderGuacamoleReader.java
+++ b/guacamole-common/src/main/java/org/apache/guacamole/io/ReaderGuacamoleReader.java
@@ -24,19 +24,24 @@ import java.io.IOException;
 import java.io.Reader;
 import java.net.SocketException;
 import java.net.SocketTimeoutException;
-import java.util.Deque;
-import java.util.LinkedList;
+import java.util.Arrays;
 import org.apache.guacamole.GuacamoleConnectionClosedException;
 import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.GuacamoleServerException;
 import org.apache.guacamole.GuacamoleUpstreamTimeoutException;
 import org.apache.guacamole.protocol.GuacamoleInstruction;
+import org.apache.guacamole.protocol.GuacamoleParser;
 
 /**
  * A GuacamoleReader which wraps a standard Java Reader, using that Reader as
  * the Guacamole instruction stream.
  */
 public class ReaderGuacamoleReader implements GuacamoleReader {
+
+    /**
+     * The GuacamoleParser instance for parsing instructions.
+     */
+    private GuacamoleParser parser = new GuacamoleParser();
 
     /**
      * Wrapped Reader to be used for all input.
@@ -57,10 +62,10 @@ public class ReaderGuacamoleReader implements GuacamoleReader {
      * The location within the received data buffer that parsing should begin
      * when more data is read.
      */
-    private int parseStart;
+    private int parseStart = 0;
 
     /**
-     * The buffer holding all received, unparsed data.
+     * The buffer holding all received data.
      */
     private char[] buffer = new char[20480];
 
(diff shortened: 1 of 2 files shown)


aleitner:
We need to benchmark this to make sure there's no degradation of performance

mike-jumper:
Well, I've thrown together a quick benchmark, and the new version leveraging `GuacamoleParser` actually seems significantly faster:

```console
[mjumper@mdesktop ~]$ java -cp ./guacamole-common-old-reader.jar:. ReaderBenchmark desktop-vnc.guac 
desktop-vnc.guac ... 18.79 ms / run
[mjumper@mdesktop ~]$ java -cp ./guacamole-common-new-reader.jar:. ReaderBenchmark desktop-vnc.guac 
desktop-vnc.guac ... 8.54 ms / run
[mjumper@mdesktop ~]$
```

This is just with calls to `readInstruction()`. The bad news is that things fail when using `read()` instead:

```console
[mjumper@mdesktop ~]$ java -cp ./guacamole-common-old-reader.jar:. ReaderBenchmark --buffers desktop-vnc.guac 
desktop-vnc.guac ... 13.9 ms / run
[mjumper@mdesktop ~]$ java -cp ./guacamole-common-new-reader.jar:. ReaderBenchmark --buffers desktop-vnc.guac 
desktop-vnc.guac ... FAIL
org.apache.guacamole.GuacamoleException: No instruction available to read.
        at org.apache.guacamole.io.ReaderGuacamoleReader.read(ReaderGuacamoleReader.java:93)
        at ReaderBenchmark.readFile(ReaderBenchmark.java:25)
        at ReaderBenchmark.main(ReaderBenchmark.java:57)
[mjumper@mdesktop ~]$ 
```

The test file that I'm using is a recording of a VNC session that's about 19M in size. The benchmark just times how long it takes to fully read through the recording, averaged over 100 trials.

Here's the hacky, quick-and-dirty benchmark: https://gist.github.com/mike-jumper/d0939f0c1a4688e9217062788a3ebfba

mike-jumper (changes requested):
A few things:

* **Good news:** it's faster. **Bad news:** `read()` seems to fail (see [the comment regarding benchmark results](https://github.com/apache/guacamole-client/pull/1057#issuecomment-2679632871))
* This should be against `patch`.
* The [GUACAMOLE-615](https://issues.apache.org/jira/browse/GUACAMOLE-615) issue was part of a past release (1.5.2), so it's now strictly historical. Any new changes will need to be associated with a new JIRA issue.

aleitner:
> A few things:
> 
>     * **Good news:** it's faster. **Bad news:** `read()` seems to fail (see [the comment regarding benchmark results](https://github.com/apache/guacamole-client/pull/1057#issuecomment-2679632871))
> 
>     * This should be against `patch`.
> 
>     * The [GUACAMOLE-615](https://issues.apache.org/jira/browse/GUACAMOLE-615) issue was part of a past release (1.5.2), so it's now strictly historical. Any new changes will need to be associated with a new JIRA issue.

Big relief that it's faster! I just added tests for reader.read and rebased to patch

mike-jumper (approved):
These changes do fix the issue with `read()`, though `read()` is now taking a bit of a performance hit:

`ReaderGuacamoleReader` throughput for the old, new, and proposed implementations.

I think the performance hit is due to the need to regenerate the protocol representation of the instruction from scratch. The above graph shows:

* **Old**: The `ReaderGuacamoleReader` from before this PR.
* **New**: The `ReaderGuacamoleReader` from this PR.
* **Revised**: The `ReaderGuacamoleReader` from this PR, with additional modifications to avoid regenerating what we've parsed from scratch (a672229dada96f8036aa6f2443f53b718528cc72).

I'm good with the changes as they stand and think it makes sense to consider further revisions like a672229dada96f8036aa6f2443f53b718528cc72 in a separate PR.

Unfortunately, it seems impossible to maintain the significant performance improvement of the `readInstruction()` implementation here without sacrificing the ability to efficiently pass received instructions on to the client.

PR #1058, opened 2025-02-10
GUACAMOLE-2030: Fix KSM static token mapping for per-user config.

eugen-keeper:
Fixes the issue plus changes log messages for this case:
`16:52:41.850 [http-nio-8080-exec-10] WARN  o.a.g.vault.ksm.secret.KsmClient - "keeper://XXX/field/login" could not be resolved: Record 'XXX' not found`
`16:52:42.442 [http-nio-8080-exec-10] WARN  o.a.g.vault.ksm.secret.KsmClient - "keeper://XXX/field/password" could not be resolved: Record 'XXX' not found`

### DIFF
--- a/extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
+++ b/extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
@@ -651,25 +651,19 @@ public class KsmClient {
 
         }
 
-        // Unfortunately, the notation parser within the Keeper SDK throws
-        // plain Errors for retrieval failures ...
-        catch (Error e) {
-            logger.warn("Record \"{}\" does not exist.", notation);
+        // Unfortunately, the notation parser within the Keeper SDK
+        // only throws plain Errors and Exceptions.
+        // There is no way to differentiate if an error is caused by
+        // a non-existing record or a pure parse failure.
+        catch (Error | Exception e) {
+            logger.warn("Keeper notation \"{}\" could not be resolved "
+                    + "to a record: {}", notation, e.getMessage());
             logger.debug("Retrieval of record by Keeper notation failed.", e);
 
             // If the secret is not found, invoke the fallback function
             if (fallbackFunction != null)
                 return fallbackFunction.get();
 
-            return CompletableFuture.completedFuture(null);
-        }
-
-        // ... and plain Exceptions for parse failures (no subclasses)
-        catch (Exception e) {
-            logger.warn("\"{}\" is not valid Keeper notation. Please check "
-                    + "the documentation at {} for valid formatting.",
-                    notation, KEEPER_NOTATION_DOC_URL);
-            logger.debug("Provided Keeper notation could not be parsed.", e);
             return CompletableFuture.completedFuture(null);
         }
         finally {


extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
```diff
@@ -664,12 +664,28 @@ public Future<String> getSecret(
...
+            // failures only. In practice you may receive a "record not found"
+            // exception here (no subclasses).
+            String message = e.getMessage();
+            if (message.contains("not found")) {
+                logger.warn("\"{}\" could not be resolved: {}", notation, message);
+            }
+            else {
+                logger.warn("\"{}\" may be not valid Keeper notation. Please check "
                     + "the documentation at {} for valid formatting.",
                     notation, KEEPER_NOTATION_DOC_URL);
```
necouchman, line 678: Minor style nitpick, but I think these two lines should be further indented to match up with the indentation of the `logger.warn()` line above?

eugen-keeper: Fixed. Correct?

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
```diff
@@ -664,12 +664,28 @@ public Future<String> getSecret(
...
-            logger.debug("Provided Keeper notation could not be parsed.", e);
+            // The original implementation considered plain exceptions as parse
+            // failures only. In practice you may receive a "record not found"
+            // exception here (no subclasses).
+            String message = e.getMessage();
+            if (message.contains("not found")) {
+                logger.warn("\"{}\" could not be resolved: {}", notation, message);
+            }
+            else {
+                logger.warn("\"{}\" may be not valid Keeper notation. Please check "
```
mike-jumper, line 676: This should read either "may not be valid" or "may be invalid".

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
```diff
@@ -664,12 +664,28 @@ public Future<String> getSecret(
...
 
-        // ... and plain Exceptions for parse failures (no subclasses)
         catch (Exception e) {
-            logger.warn("\"{}\" is not valid Keeper notation. Please check "
-                    + "the documentation at {} for valid formatting.",
-                    notation, KEEPER_NOTATION_DOC_URL);
-            logger.debug("Provided Keeper notation could not be parsed.", e);
+            // The original implementation considered plain exceptions as parse
+            // failures only. In practice you may receive a "record not found"
+            // exception here (no subclasses).
```
mike-jumper, line 670: With this in mind, perhaps this whole section should be unified with the `Error` section above?

Rather than looking for keywords in the message, I think it might be better to just represent the failure more generically while also logging the message we receive, for example, something like:

```java
catch (Error | Exception e) {
    logger.warn("Keeper notation \"{}\" could not be resolved to a record: {}", notation, e.getMessage());
    logger.debug("Retrieval of record by Keeper notation failed.", e);

    ...
}

**Review by mike-jumper — CHANGES_REQUESTED** [review]

File: extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java (hunk 3 above)
Comment (eugen-keeper) [reply extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java]: You are right. Changed.

File: extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java [hunk 4]
```diff
@@ -653,8 +653,9 @@ public Future<String> getSecret(
 
         // Unfortunately, the notation parser within the Keeper SDK throws
         // plain Errors for retrieval failures ...
```
mike-jumper, line 655: This should be updated accordingly.

mike-jumper (changes requested):
Comment needs updating, otherwise LGTM.

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
eugen-keeper: My bad. Is it better now?

mike-jumper (approved):

mike-jumper:
Before I merge this, @eugen-keeper is there any reason this shouldn't be on `patch`?

extensions/guacamole-vault/modules/guacamole-vault-ksm/src/main/java/org/apache/guacamole/vault/ksm/secret/KsmClient.java
mike-jumper: Yep!

mike-jumper (changes requested):
Should be good to go once rebased against `patch`.

mike-jumper (approved):

PR #1060, opened 2025-02-15
GUACAMOLE-2032: Fix hasClientGroups() to no longer take into account the currently attached client.

corentin-soriano:
Currently, the function systematically returns true and leaves this small rectangle visible:
image

The fix allows this element to only be displayed when there is at least one **unattached** client.


guacamole/src/main/frontend/src/app/client/directives/guacClientPanel.js
```diff
@@ -84,7 +84,10 @@ angular.module('client').directive('guacClientPanel', ['$injector', function gua
              *     panel, false otherwise.
              */
             $scope.hasClientGroups = function hasClientGroups() {
-                return $scope.clientGroups && $scope.clientGroups.length;
+                return $scope.clientGroups && $scope.clientGroups.filter(function(group) {
+                    // Filter out attached group as it is not shown in the panel
+                    return !group.attached;
+                }).length > 0;
```
mike-jumper, line 90: Rather than `filter()`, which here will create a new array that is ultimately discarded, it would be better to use `some()`.

mike-jumper (changes requested):

guacamole/src/main/frontend/src/app/client/directives/guacClientPanel.js
corentin-soriano: Yes it's a good idea, I modified it.

mike-jumper (approved):

PR #1067, opened 2025-03-05
GUACAMOLE-2036: Increase the number of max elements to match guacamole server.

aleitner:
In guacamole-server `GUAC_INSTRUCTION_MAX_ELEMENTS` is defined as 128. In guacamole-client, `INSTRUCTION_MAX_ELEMENTS` is still 64.


mike-jumper (changes requested):
LGTM, but please rebase against `patch`.

mike-jumper (approved):

PR #1085, opened 2025-05-14
GUACAMOLE-2069: Fix order of select field items.

eugen-keeper:


guacamole/src/main/frontend/src/app/form/templates/selectField.html
```diff
@@ -4,5 +4,5 @@
             ng-disabled="disabled"
             guac-focus="focused"
             ng-model="model"
-            ng-options="option as getFieldOption(option) | translate for option in field.options | orderBy: value"></select>
+            ng-options="option as getFieldOption(option) | translate for option in field.options"></select>
```
necouchman, line 7: I guess I'm curious what the rationale is behind making this change? Of course I'm fine having the source code sorted by default - it helps, not just for the web UI, but for the readable of the source code itself. But why remove the sorting in the code, here?

eugen-keeper: I am not sure if I understood your question correctly. Anyway, the primary issue is that this thing breaks order of enum items with numbers. For example, if we have something like "256KB", "1MB", "10MB", the **orderBy** option will make it shown as 1MB, 10MB, 256KB. We discussed this issue with Mike yesterday and it looks like we already specify the order we want manually everywhere. We do not need any type of automatic sorting.

necouchman: Sorry, I actually read the Jira ticket after putting that comment on there, so I understand, now :-).

PR #1087, opened 2025-05-28
GUACAMOLE-2051: Allow Guacamole properties to be either comma or semi-colon or comma delimited.

necouchman:
Took a quick stab at fixing this issue with adding the comma back in as a delimiter. Not tested, yet, but I think the RegEx is sound.

### DIFF
--- a/guacamole-ext/src/main/java/org/apache/guacamole/properties/FileGuacamoleProperty.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/properties/FileGuacamoleProperty.java
@@ -23,6 +23,7 @@ import java.io.File;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.regex.Pattern;
 import org.apache.guacamole.GuacamoleException;
 
 /**
@@ -30,6 +31,14 @@ import org.apache.guacamole.GuacamoleException;
  */
 public abstract class FileGuacamoleProperty implements GuacamoleProperty<File> {
 
+    /**
+     * A pattern which matches against the delimiters between values. This is
+     * currently either a comma or a semicolon and any following whitespace.
+     * Parts of the input string which match this pattern will not be included
+     * in the parsed result.
+     */
+    static final Pattern DELIMITER_PATTERN = Pattern.compile("[,;]\\s*");
+    
     @Override
     public File parseValue(String value) throws GuacamoleException {
 
--- a/guacamole-ext/src/main/java/org/apache/guacamole/properties/GuacamoleProperty.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/properties/GuacamoleProperty.java
@@ -21,7 +21,6 @@ package org.apache.guacamole.properties;
 
 import java.util.Collection;
 import java.util.Collections;
-import java.util.regex.Pattern;
 import org.apache.guacamole.GuacamoleException;
 
 /**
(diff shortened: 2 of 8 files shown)


guacamole-ext/src/main/java/org/apache/guacamole/properties/GuacamoleProperty.java
```diff
@@ -39,7 +39,7 @@ public interface GuacamoleProperty<Type> {
      * input string which match this pattern will not be included in the parsed
      * result.
      */
-    static final Pattern DELIMITER_PATTERN = Pattern.compile(";\\s*");
+    static final Pattern DELIMITER_PATTERN = Pattern.compile("[;,]\\s*");
```
mike-jumper, line 42: I'm a little worried about this vs. the rationale cited in #490:

> Also, note that I switched over to semicolon delimiting for these - that's worth some discussion, but my rationale for that is that some URIs - LDAP ones, for instance - rely on the ability to use commas to separate LDAP DNs within the URI, so, one way or the other we need to address that.

Looking at this again now with fresh eyes, it also seems a bit odd to have this declared here on the base `GuacamoleProperty` interface - that should probably either be part of an abstract base class or defined as needed within implementations.

Perhaps `URIGuacamoleProperty` should use semicolons, while `StringGuacamoleProperty` (and everything else) should use commas? Extensions can provide their own more specific `GuacamoleProperty` implementations if the default delimiters don't work.

Based on the changes from #490, I think the established properties that we need to consider when deciding this are:

* `json-trusted-networks`
* `ldap-group-name-attribute`
* `ldap-user-attributes`
* `ldap-username-attribute`
* `openid-attributes-claim-type`
* `quickconnect-allowed-parameters`
* `quickconnect-denied-parameters`
* `ssl-subject-username-attribute`

necouchman:
```
json-trusted-networks
ldap-group-name-attribute
ldap-user-attributes
ldap-username-attribute
openid-attributes-claim-type
quickconnect-allowed-parameters
quickconnect-denied-parameters
ssl-subject-username-attribute
```

I think all of these will be good with comma-separated values, or perhaps both comma and semicolon-separated values. I suspect the URIs along with any of the ones that take a LDAP DNs can use semicolons.

mike-jumper: Sounds reasonable. I think all of the above (except `ssl-subject-username-attribute`) are already established and documented as accepting comma-delimited lists of values.

necouchman: Looking further at this, it seems like `URIGuacamoleProperty` is really the only base one of those classes that needs any adjusting. Most of the LDAP-related ones that are used for DNs don't really support multi-value delimiting, they seem to just use the default behavior of providing a `singletonList()` for the specified value.

Anything I'm missing?

mike-jumper: (And `DELIMITER_PATTERN`, an implementation detail, shouldn't be defined at the `interface` level.)

necouchman: These issues should be addressed.

guacamole-ext/src/main/java/org/apache/guacamole/properties/FileGuacamoleProperty.java
```diff
@@ -23,13 +23,22 @@
...
 import org.apache.guacamole.GuacamoleException;
 
 /**
  * A GuacamoleProperty whose value is a filename.
  */
 public abstract class FileGuacamoleProperty implements GuacamoleProperty<File> {
 
+    /**
+     * A pattern which matches against the delimiters between values. This is
+     * currently simply a semicolon and any following whitespace. Parts of the
```
mike-jumper, line 36: > ... This is currently simply a semicolon and any following whitespace ...

This needs to be updated to include the comma among the characters used as delimiters (same for other comments).

mike-jumper (changes requested):
Minor change needed for docs, otherwise LGTM.

guacamole-ext/src/main/java/org/apache/guacamole/properties/FileGuacamoleProperty.java
necouchman: Updated!

PR #1091, opened 2025-06-23
GUACAMOLE-954: Add LDAP support for nested user groups

stcbus:
This is the start of a pull request to address GUACAMOLE-954 in a performant way, but unfortunately this feature only works with Active Directory. 

I added a new config boolean due to this called 'nested-groups' to enable/disable the AD-specific query.

It makes use of the LDAP_MATCHING_RULE_IN_CHAIN mentioned in the Jira with a specific OID.

I understand if the team is not keen on merging a feature that is only for a specific vendor's LDAP implementation, but this is performant and very useful for us in our AD environment so I would like to try and get it added for the others who requested it and likely using AD as well.

### DIFF
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/JacksonLDAPConfiguration.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/JacksonLDAPConfiguration.java
@@ -205,6 +205,13 @@ public class JacksonLDAPConfiguration implements LDAPConfiguration {
     private String memberAttributeType;
 
     /**
+     * The raw YAML value of {@link LDAPGuacamoleProperties#LDAP_NESTED_GROUPS}.
+     * If not set within the YAML, this will be null.
+     */
+    @JsonProperty("nested-groups")
+    private Boolean nestedGroups;
+
+    /**
      * The default configuration options for all parameters.
      */
     private LDAPConfiguration defaultConfig = new DefaultLDAPConfiguration();
@@ -432,6 +439,11 @@ public class JacksonLDAPConfiguration implements LDAPConfiguration {
     @Override
     public String getMemberAttribute() throws GuacamoleException {
         return withDefault(memberAttribute, defaultConfig::getMemberAttribute);
+    }
+
+    @Override
+    public boolean getNestedGroups() throws GuacamoleException {
+        return withDefault(nestedGroups, defaultConfig::getNestedGroups);
     }
 
     @Override
--- a/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
+++ b/extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
@@ -27,10 +27,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import org.apache.directory.api.ldap.model.entry.Entry;
+import org.apache.directory.api.ldap.model.entry.Value;
 import org.apache.directory.api.ldap.model.exception.LdapInvalidAttributeValueException;
 import org.apache.directory.api.ldap.model.filter.AndNode;
 import org.apache.directory.api.ldap.model.filter.EqualityNode;
 import org.apache.directory.api.ldap.model.filter.ExprNode;
+import org.apache.directory.api.ldap.model.filter.ExtensibleNode;
 import org.apache.directory.api.ldap.model.filter.NotNode;
 import org.apache.directory.api.ldap.model.name.Dn;
 import org.apache.guacamole.auth.ldap.conf.MemberAttributeType;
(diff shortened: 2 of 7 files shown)


necouchman:
> I understand if the team is not keen on merging a feature that is only for a specific vendor's LDAP implementation, but this is performant and very useful for us in our AD environment so I would like to try and get it added for the others who requested it and likely using AD as well.

Personally I think it would be fine to have this feature included, provided we clearly document when it is expected to work and when it probably will not work. Many other software applications out there that leverage LDAP have ways of configuring the type of LDAP server you're pointing to in order to make use of specific features, and I agree that it would be useful for folks pointing at AD-based LDAP servers.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
```diff
@@ -227,14 +229,36 @@ public List<Entry> getParentUserGroupEntries(ConnectedLDAPConfiguration config,
...
+            filterValue = null;
+
+            // Explicitly set the matching rule ID and dnAttributes
+            node.setMatchingRuleId("1.2.840.113556.1.4.1941");
+            node.setDnAttributes(false);
+            node.setValue(new Value(userIDorDN));
+            groupFilter = new AndNode(
+                    groupFilter, node
+            );
+        }
```
necouchman, line 259: I have a few concerns for this:
* Why is `filterValue` set to `null`, and what is the implication of passing through `null` to the `queryService.search()` method?
* The OID for LDAP_MATCHING_RULE_IN_CHAIN should probably be defined as a `final String` above - if the Apache LDAP API doesn't already have that defined somewhere?

stcbus: @necouchman I believe in my testing the group filter applied with the filtervalue as userIDorDN did not work. It seemed like one or the other and I was trying to avoid duplicate code calling the function twice.

Looks like from the doc, it probably was so the groups could be found outside of where the user's account object was located:

> The value that should be searched search for within the attributes of objects within the LDAP directory. **If null, the search will test only for the presence of at least one of the given attributes on each object, regardless of the value of those attributes.**

I am not aware of any constant for that value, but I agree it makes sense to add it to a variable.

necouchman: Okay - yes, please define that OID as a constant wherever it makes sense.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/group/UserGroupService.java
```diff
@@ -227,14 +229,37 @@ public List<Entry> getParentUserGroupEntries(ConnectedLDAPConfiguration config,
         groupAttributes.add(memberAttribute);
 
         // Get all groups the user is a member of starting at the groupBaseDN,
-        // excluding guacConfigGroups
+        // excluding guacConfigGroups and evaluating nested groups 
+        // (if enabled).
+
+        ExprNode groupFilter = config.getGroupSearchFilter();
+        String filterValue = userIDorDN;
+        final String matchingOID = "1.2.840.113556.1.4.1941";
```
necouchman: I was thinking this would be more appropriate as a class-level variable, defined as a `public` or `private` `final String` - something like:

`private final String LDAP_GROUP_NESTED_MATCHING_OID = "1.2.840.113556.1.4.1941";`

defined up in the top section of the class.

stcbus: Apologies @necouchman , I am not very familiar with Java, I think I have it how you want it, now.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/JacksonLDAPConfiguration.java
```diff
@@ -204,6 +204,13 @@ public class JacksonLDAPConfiguration implements LDAPConfiguration {
     @JsonProperty("member-attribute-type")
     private String memberAttributeType;
 
+    /**
+     * The raw YAML value of {@link LDAPGuacamoleProperties#LDAP_NESTED_GROUPS}.
+     * If not set within the YAML, this will be null.
```
necouchman: This may be copy-pasta, but I think, if not set within the yaml, this will be `false`, correct, as opposed to `null`?

necouchman (changes requested):
One more minor comment update, and I think this will be ready to merge.

extensions/guacamole-auth-ldap/src/main/java/org/apache/guacamole/auth/ldap/conf/JacksonLDAPConfiguration.java
stcbus: good catch, fixed.,

necouchman:
Thanks @stcbus! If you could also update the apache/guacamole-manual with a PR, that would be helpful to make sure documentation is up-to-date.

PR #1096, opened 2025-07-21
GUACAMOLE-2099: Update jdbc drivers to latest.

aleitner:
* Update mysql jdbc driver 8.3.0 => 9.3.0
* Update postgresql jdbc driver 42.7.2 => 42.7.7
* Fix typos within sqlserver mybatis mapper files.


mike-jumper (changes requested):
LGTM, but this should be against the `patch` branch.

aleitner:
> LGTM, but this should be against the `patch` branch.

rebased for patch

mike-jumper (approved):

PR #1099, opened 2025-08-01
GUACAMOLE-2114: Custom base for CATALINA_BASE and GUACAMOLE_HOME

tiborauer:
Allowing a custom base, rather than the ones randomly created in /tmp, would enable storing the program's data in a user-defined location. This feature would be particularly useful on Kubernetes, where the lifecycle of the program's data independent from that of the program would improve robustness. Also, the program's data could be shared with sidecar containers to extend functionality.


guacamole-docker/entrypoint.d/100-generate-guacamole-home.sh
```diff
@@ -59,8 +59,25 @@ is_property_set() {
 # Start with a fresh GUACAMOLE_HOME
 #
 
-rm -rf /tmp/guacamole-home.*
-GUACAMOLE_HOME="`mktemp -p /tmp -d guacamole-home.XXXXXXXXXX`"
+if [[ -z "$APACHE_BASE" ]]; then
```
necouchman: I'd request changing `APACHE_BASE` to `TOMCAT_BASE` or, perhaps, `WEBAPPS_BASE` throughout the code - Apache is the name of a large, non-profit software organization, so I'm not really sure it makes sense to use that as part of the variable.

tiborauer: Thank you for the suggestion!

mike-jumper:
> Allowing a custom base, rather than the ones randomly created in /tmp, would enable storing the program's data in a user-defined location. This feature would be particularly useful on Kubernetes, where the lifecycle of the program's data independent from that of the program would improve robustness. Also, the program's data could be shared with sidecar containers to extend functionality.

@tiborauer Can you clarify what sort of use case requires storing `CATALINA_HOME` and `GUACAMOLE_HOME` data in a user-defined location? My original thinking behind the randomized bases beneath `/tmp` was to:

* Ensure the container entrypoint always has control over the contents of these locations (ie: avoid breaking the expectations of the entrypoint).
* Avoid downstream reliance on storing anything in these locations (ie: the container shouldn't rely on state persisting within the container itself).

PR #1109, opened 2025-08-31
Add Traditional Chinese (zh_TW) translation

jasoncheng7115:
I have added the zh_TW.json translation file for Traditional Chinese, 
but it does not appear as an option in the language selection menu.

### DIFF
--- a/guacamole/src/main/frontend/src/translations/zh_TW.json
+++ b/guacamole/src/main/frontend/src/translations/zh_TW.json
@@ -1,0 +1,1241 @@
+{
+    
+    "NAME" : "繁體中文",
+    
+    "APP" : {
+
+        "NAME"    : "Apache Guacamole",
+        "VERSION" : "${project.version}",
+
+        "ACTION_ACKNOWLEDGE"        : "確定",
+        "ACTION_CANCEL"             : "取消",
+        "ACTION_CLEAR"              : "清除",
+        "ACTION_CLONE"              : "複製",
+        "ACTION_CONTINUE"           : "繼續",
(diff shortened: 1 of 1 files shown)


guacamole/src/main/frontend/src/translations/zh_TW.json
```diff
@@ -0,0 +1,1241 @@
+{
+    
+    "NAME" : "繁體中文",
+    
+    "APP" : {
+
+        "NAME"    : "Apache Guacamole",
+        "VERSION" : "${project.version}",
```
necouchman, line 8: Please do not include these values in the translation - they will be pulled automatically from the English translation.

necouchman (changes requested):
@jasoncheng7115 Thanks for contributing this. There's a couple of things that need to be done:
* You need to request a Jira account and create a Jira issue for this, and tag both the PR and the commit messages with the GUACAMOLE-XXXX issue.
* Please make sure not to include anything that you're not translating in the translation file.

See: https://guacamole.apache.org/open-source/

As far as why the translation is not loading, I'd say the first thing to check is clearing your browser cache and reloading the page to see if that gets the language to load. Other than that, maybe increase the log level of the web application and see if there are any messages about the new language file in there?

https://guacamole.apache.org/doc/gug/configuring-guacamole.html#logging-within-the-web-application

jasoncheng7115:
Thank you for your detailed explanation. I will review and learn more about the Guacamole contribution process and then make the necessary corrections.

adb014:
I propose closing this pull request in favour of #1183 that being against a more recent Guacamole probably has additional translations...

PR #1111, opened 2025-09-03
GUACAMOLE-2130: Support for Linked Records in KSM.

sschiffli:
To support new linked records in KSM, this PR add new "[CRITERIA]" names to be automatically injected. 

1. SERVER_ADMIN - Identical to SERVER but explicitly uses "admin" credentials.
2. SERVER_LAUNCH - Identical to SERVER but explicitly uses "launch" credentials.
3. GATEWAY_ADMIN - Identical to GATEWAY but explicitly uses "admin" credentials.
4. GATEWAY_LAUNCH - Identical to GATEWAY but explicitly uses "launch" credentials.


doc/licenses/kotlinx-serialization-1.9.0/README
```diff
@@ -1,7 +1,7 @@
 Kotlin Serialization (https://github.com/Kotlin/kotlinx.serialization)
 ----------------------------------------------------------------------
 
-    Version: 1.7.2
+    Version: 1.9.0
```
necouchman, line 4: Do we know that this version of Kotlin is still at our overall Java 8 compatibility level?

sschiffli: Good question, I did not check before.  But as far as I can tell this version does retain Java 8 compatibility.  I don't know if there is a more definitive way to check, but here is this output

```
$ javap -verbose -cp kotlinx-serialization-core-jvm-1.9.0.jar kotlinx/serialization/modules/SerializersModule | grep "major version"
  major version: 52
```

mike-jumper: It looks like Kotlin is Java 8 for all versions: https://kotlinlang.org/docs/faq.html#which-versions-of-jvm-does-kotlin-target

mike-jumper (commented):
LGTM ... @necouchman, are you OK with the scope of this being on `patch`, given that some documentation updates will be needed? Or should it be against `main`?

mike-jumper (approved):

mike-jumper:
@sschiffli FYI, these are the relevant docs that would need updating:

https://github.com/apache/guacamole-manual/blob/patch/src/vault.md.j2

PR #1113, opened 2025-09-03
GUACAMOLE-2099: Add missing "AND" in SQL Server query for permission deletion.

mike-jumper:


necouchman (approved):
Oops. My bad. :grimacing:

PR #1116, opened 2025-09-16
GUACAMOLE-2137: Introduce HashiCorp Vault token handler (based on the KSM module)

TdlQ:
This PR introduces a new module to handle HashiCorp Vault tokens. It is heavily inspired by and reuses a significant amount of code from the existing KSM module.

The main goal is to provide a dedicated, lightweight solution for fetching secrets from HashiCorp Vault for use in Guacamole connection parameters. This allows for replacing static credentials with dynamic, centrally managed secrets.

### Key Features & Implementation Details
- Token Format: The module uses a new token format, ${HASHIVAULT:path/to/secret/key}, to reference secrets stored in Vault. For example: Password: ${HASHIVAULT:path/to/my/server/guacamole_connection/password}.
- Centralized Configuration: Vault configuration is managed through a base64-encoded JSON object (vault_url, vault_token, cache_lifetime), which is stored in the HV_CONFIG parameter and can be overridden at connection groups level.
- Efficient Caching: The module is optimized for performance. When multiple tokens reference the same Vault path (e.g., username and password from the same secret), it performs only a single HTTP query to Vault. Subsequent requests for keys within the same path are served directly from a concurrent, time-based cache.
- Asynchronous Handling: All Vault queries are performed asynchronously to prevent blocking the connection process. This is achieved using CompletableFuture and a "in-flight" request caching pattern to handle concurrent requests for the same secret efficiently.

### Notable Differences and Design Choices (vs KSM)
- Simplicity: This module is designed to be a simpler, more lightweight alternative to the KSM module, focusing exclusively on basic token handling. It intentionally lacks more advanced features.
- Execution Order: The setAttributes() method now directly calls processAttributes() to ensure correct execution order, which was an issue observed during development.
- User Custom Configuration: The user-defined configuration part is currently a placeholder. It mimics KSM's design but might be simplified or removed in the future if a clear use case for it does not emerge.

### DIFF
--- a/extensions/guacamole-vault/modules/guacamole-vault-hv/src/main/java/org/apache/guacamole/vault/hv/conf/HvAttributeService.java
+++ b/extensions/guacamole-vault/modules/guacamole-vault-hv/src/main/java/org/apache/guacamole/vault/hv/conf/HvAttributeService.java
@@ -1,0 +1,155 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.guacamole.vault.hv.conf;
+
+import com.google.inject.Singleton;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.apache.guacamole.GuacamoleException;
+import org.apache.guacamole.form.BooleanField;
+import org.apache.guacamole.form.Form;
+import org.apache.guacamole.form.TextField;
+import org.apache.guacamole.language.TranslatableGuacamoleClientException;
+import org.apache.guacamole.vault.conf.VaultAttributeService;
+
+@Singleton
+public class HvAttributeService implements VaultAttributeService {
+
+    /**
+     * The name of the attribute which can contain a HV configuration blob
+     * associated with either a connection group or user.
+     */
+    public static final String HV_CONFIGURATION_ATTRIBUTE = "hv-config";
+
+    /**
+     * The HV configuration attribute contains sensitive information, so it
+     * should not be exposed through the directory. Instead, if a value is
+     * set on the attributes of an object, the following value will be exposed
+     * in its place, and correspondingly the underlying value will not be
+     * changed if this value is provided to an update call.
+     */
+    public static final String HV_ATTRIBUTE_PLACEHOLDER_VALUE = "**********";
+
+    /**
+     * All attributes related to configuring the HV vault on a
+     * per-connection-group or per-user basis.
+     */
+    public static final Form HV_CONFIGURATION_FORM = new Form(
+        "hv-config",
+        Arrays.asList(new TextField(HV_CONFIGURATION_ATTRIBUTE))
+    );
+
+    /**
+     * All HV-specific attributes for users, connections, or connection groups,
+     * organized by form.
+     */
+    public static final Collection<Form> HV_ATTRIBUTES = Collections.unmodifiableCollection(Arrays.asList(HV_CONFIGURATION_FORM));
+
+    /**
+     * The name of the attribute which can controls whether a HV user configuration
+     * is enabled on a connection-by-connection basis.
+     */
+    public static final String HV_USER_CONFIG_ENABLED_ATTRIBUTE = "hv-user-config-enabled";
+
+    /**
+     * The string value used by HV attributes to represent the boolean value "true".
+     */
+    public static final String TRUTH_VALUE = "true";
+
+    /**
+     * All attributes related to configuring the HV vault on a per-connection basis.
+     */
+    public static final Form HV_CONNECTION_FORM = new Form(
+        "hv-config",
+        Arrays.asList(new BooleanField(HV_USER_CONFIG_ENABLED_ATTRIBUTE, TRUTH_VALUE))
+    );
+
+    /**
+     * All HV-specific attributes for connections, organized by form.
+     */
+    public static final Collection<Form> HV_CONNECTION_ATTRIBUTES = Collections.unmodifiableCollection(Arrays.asList(HV_CONNECTION_FORM));
+
+    @Override
+    public Collection<Form> getConnectionAttributes() {
+        return HV_CONNECTION_ATTRIBUTES;
+    }
+
+    @Override
+    public Collection<Form> getUserAttributes() {
+        return HV_ATTRIBUTES;
+    }
+
+    @Override
+    public Collection<Form> getUserPreferenceAttributes() {
+        return getUserAttributes();
+    }
+
+    @Override
+    public Collection<Form> getConnectionGroupAttributes() {
+        return HV_ATTRIBUTES;
+    }
+
+    /**
+     * Sanitize the value of the provided HV config attribute. If the provided
+     * config value is non-empty, it will be replaced with the placeholder
+     * value to avoid leaking sensitive information. If the value is empty, it
+     * will be replaced by `null`.
+     *
+     * @param hvAttributeValue
+     *    The HV configuration attribute value to sanitize.
+     *
+     * @return
+     *    The sanitized HV configuration attribute value, stripped of any
+     *    sensitive information.
+     */
+    public static String sanitizeHvAttributeValue(String hvAttributeValue) {
+
+        // Any non-empty values may contain sensitive information, and should
+        // be replaced by the safe placeholder value
+        if (hvAttributeValue != null && !hvAttributeValue.trim().isEmpty())
+            return HV_ATTRIBUTE_PLACEHOLDER_VALUE;
+
+        // If the configuration value is empty, expose a null value
+        else
+            return null;
+
+    }
+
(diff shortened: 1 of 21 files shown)


corentin-soriano (commented):
I think you need to add `guacamole-vault/hv` in the `map_extensions` to resolve the CI build error:
```
#20 143.2 Mapped: guacamole-vault/ksm -> KSM_
..........
#20 144.3 ERROR: Unmapped extension: /opt/guacamole/extensions/guacamole-vault/hv/guacamole-vault-hv.jar
```
After this line:
https://github.com/apache/guacamole-client/blob/0336e2c7e1be7d5510bdad9505dc093e76a54444/guacamole-docker/build.d/010-map-guacamole-extensions.sh#L117

adb014:
Shouldn't This merge with the PR #1143   as Hashicorp and OpenBao use essentially the same API

extensions/guacamole-vault/modules/guacamole-vault-hv/src/main/java/org/apache/guacamole/vault/hv/conf/HvAttributeService.java
```diff
@@ -0,0 +1,155 @@
...
+import org.apache.guacamole.vault.conf.VaultAttributeService;
+
+@Singleton
+public class HvAttributeService implements VaultAttributeService {
+
+    /**
+     * The name of the attribute which can contain a HV configuration blob
+     * associated with either a connection group or user.
+     */
+    public static final String HV_CONFIGURATION_ATTRIBUTE = "hv-config";
```
adb014, line 44: Wouldn't it be better to use roleID and secretID that use an opaque base64 encoded json blob containing the token ?

extensions/guacamole-vault/modules/guacamole-vault-hv/src/main/java/org/apache/guacamole/vault/hv/secret/HvClient.java
```diff
@@ -0,0 +1,270 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.inject.Inject;
+import com.google.inject.assistedinject.Assisted;
+import com.google.inject.assistedinject.AssistedInject;
+import java.io.IOException;
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
```
adb014, line 31: Wouldn't it be better to use org.springframework.vault and delegate all of http communication and reuse and concurrency issues to spring ? See my propsed code block in #1143

extensions/guacamole-vault/modules/guacamole-vault-hv/src/main/java/org/apache/guacamole/vault/hv/secret/HvClient.java
```diff
@@ -0,0 +1,270 @@
...
+    /**
+     * Name of the HTTP header for the token, as specified in documentation:
+     * https://developer.hashicorp.com/vault/docs/auth/token
+     */
+    static final String HASHICORP_VAULT_HTTP_HEADER_TOKEN = "X-Vault-Token";
+
+    /**
+     * API version.
+     */
+    static final String HASHICORP_VAULT_HTTP_VERSION = "/v1/";
```
adb014, line 60: If allowing both Hashicorp and OpenBao to use the same jar file, this should allow /v2/ as well

extensions/guacamole-vault/modules/guacamole-vault-hv/src/main/java/org/apache/guacamole/vault/hv/secret/HvClient.java
```diff
@@ -0,0 +1,270 @@
...
+    /**
+     * The HV configuration associated with this client instance.
+     */
+    private final Map<String, String> hvConfig;
+
+    /**
+     * The maximum amount of time that an entry will be stored in the cache
+     * before being refreshed, in milliseconds.
+     */
+    private long cacheLifetime;
```
adb014, line 92: If delegated to org.springframework.vault, cache is withn the spring framework. I also has an issue with caching values within Guacamole in that highly sensitive data might end up swapped to disk and not in memory and more easily attackable. If caching is used, need to think about how to avoid cache leaks

adb014 (commented):
I'm not a member of this projet but a user wanting openboa/hv support, so I'd like to see an openbao/hv jar file in a future version of Guacamole. You can take or leave my comments as you want :-)

extensions/guacamole-vault/modules/guacamole-vault-hv/src/main/java/org/apache/guacamole/vault/hv/secret/HvClient.java
adb014: One idea to avoid cache leaks is to use a weak cache like

```java
com.github.benmanes.caffeine.cache.Caffeine;
com.github.benmanes.caffeine.cache.Cache;

...
private final Cache<String, Object> cachedSecrets = 
        Caffeine.newBuilder().weakKeys().build();
```

The keys and values are garbage collected under memory pressure, meaning there will be more cache misses. But better that than a cache leak.

extensions/guacamole-vault/modules/guacamole-vault-hv/src/main/java/org/apache/guacamole/vault/hv/conf/HvConfigurationService.java
```diff
@@ -0,0 +1,204 @@
...
+     * each connection parameter token and the name of the secret from which
+     * the value for that token should be read, while the properties file is an
+     * alternative to guacamole.properties where each property value is the
+     * name of a secret containing the actual value.
+     */
+    public HvConfigurationService() {
+        super(TOKEN_MAPPING_FILENAME, PROPERTIES_FILENAME);
+    }
+
+    /**
```
adb014, line 120: Should probably always return false here. Hashicorp is not really designed to allow users to store their own secrets in it. Yes its technically possible to create a mount path for a user and therir secrets and a dedicated token for this mount path, it practical terms its too painful to do it really... Removing this will simplify the code enormously

extensions/guacamole-vault/modules/guacamole-vault-hv/src/main/java/org/apache/guacamole/vault/hv/secret/HvSecretService.java
```diff
@@ -0,0 +1,335 @@
...
+     * Factory for creating HV client instances.
+     */
+    @Inject
+    private HvClientFactory hvClientFactory;
+
+    /**
+     * A map of base-64 encoded JSON HV config blobs to associated HV client instances.
+     * A distinct HV client will exist for every HV config.
+     */
+    private final ConcurrentMap<String, HvClient> hvClientMap = new ConcurrentHashMap<>();
```
adb014, line 93: This is only needed to allow users to have their own secrets stored in hashcorp on a dedicated mount path with a dedicated token.. Probably not needed, allowing the code to be very much simplified

extensions/guacamole-vault/modules/guacamole-vault-hv/src/main/java/org/apache/guacamole/vault/hv/conf/HvAttributeService.java
adb014, line 1: If this are no user attributes and we don't allow different vaults, mount pathes, etc on a per connection basis. As far as I can see this enitire is not needed.

BTW, I tried your module on a 1.6.0 guacmole client and the forms defined in this file didn't appear in the Admin UI. Is this expected ?

adb014:
I’ve been looking at the today and I like your scheme

```
HASHIVAULT:<mount path>/<path>/<key>/<value>
```

that you have used. But I’d like to see it completed with other secret engines than just the key-value secret engine. It seems to me only the KV, LDAP and SSH secret engines make sense for guacamole connections, but if we want to store guacamole’s database secret in a vault we’ll also need the database engine as well. I propose including the engine to use in the prefix and removing HASHI from the prefix to make it shorter after adding the engine and agnostic wrt to openbao. 

So 4 prefixes

VAULT_KV: for the key-value engine 
VAULT_SSH: for the ssh engine
VAULT_LDAP: for the ldap engine (also used with Active Directory)
VAULT_DB: for the database engine

So for a static-role in the ldap engine the token to use would be `${VAULT_LDAP:<mount path>/static-role/<role>/<value>}` where `<value>` would be “password” or “username”. This works for dynamic roles as well, replacing “static-role” with “role” as long as we don’t deal with the TTL within guacamole. The same for service accounts

For the ssh secret engine using certificates we’ll need to create the ssh keys in guacamole and use the secret engine to sign them. The account needs to exist on the ssh host so the username should be taken from the connection username or a key-value otherwise. However, no path or value is needed for these tokens. So I’d propose a generic token like `${VAULT_SSH:<mount path>/public}` and `${VAULT_SSH:<mount path>/private}` for the two ssh certificates. However, there is a quack.  The signed ssh certificate can contain restrictions on the ssh parameters like port forwarding that could be used to stop these short lifed certificates from being abused

For SSH one-time passwords they are created for a particular username and hostname, that can be recovered from the connection information itself. Only the password token makes sense in that case which might be returned with the token `${VAULT_SSH:<mount path>/otp}`

The database tokens are a bit like the KV tokens with seperate username and password values. So the token in that case would be `${VAULT_DB:<mount path>/database/creds/<value>}`. Though the “database/creds” part of the path always there so they could be dropped in the token and added by guacamole in the query to the vault. 

There are many edge cases in implementing all of these tokens types. I’ll take a try at it based on your code and the other PR for openbao, but testing it extensively will be difficult

mldmld68:
Hi there, I'm interessed by Hashicorp Vault support. We found all the ssh keys & windows passwords in the SQL database in clear text. We want to be sure theses information could not easily read.
BTW, the secrets could also be cyphered in the DB and Vault only store a key to unencrypt all of them. Thanks

adb014:
> Hi there, I'm interessed by Hashicorp Vault support. We found all the ssh keys & windows passwords in the SQL database in clear text. We want to be sure theses information could not easily read. BTW, the secrets could also be cyphered in the DB and Vault only store a key to unencrypt all of them. Thanks

That’s a Vault configuration issue and nothing to do with Guacamole. If the Vault is using a database for storage it always stores “secrets” encrypted. It might store some metadata non encrypted however. 

By “ssh keys” do you mean static keys stored in a key-value secret engine of the Vault or temporary signed keys using the vault ssh secret engine? If the second, look at #1214

PR #1117, opened 2025-09-16
GUACAMOLE-2138: Add connection-timeout parameter

tkuhlengel:
# Summary

In my company's business use case for Guacamole, for security and auditing purposes, we need to be able to ensure that any idle user is disconnected and logged off within a set period of time of idleness. 

In an *ideal* version of such a system, we would do the following. 

1.  Check if the user has interacted with the connection in the last X minutes, or if an active SFTP transfer is happening on the connection.
2. If not, terminate the user connection after X minutes of inactivity. 
3. If they remain idle, the Guacamole idle timer will log them out after the pre-configured login idle timeout.

This requires a lot of conditionals and would be more difficult to implement and maintain in an ongoing project like Guacamole. 

A *more practical*, yet sufficient, version is:
 1. The Administrator sets a maximum duration for any connection, specified in minutes. 
 2. Any connection that exceeds that duration, regardless of activity, is terminated while the user remains logged in.  They are free to reconnect if the user is still active. 
 3. The login idle timeout starts when the connection ends. 

This second option meets our business needs, and we would like to share it with others. 

# Features

* Add parameter `connection-timeout` in `guacamole.properties`, disconnecting users after `connection-timeout` minutes. Defaults to 0, disabling the feature. 
* Implement the connection timeout using a Map that stores the creation time of the connection.
* Integrate the connection timeout check into the existing idle timeout function.

### DIFF
--- a/guacamole/src/main/java/org/apache/guacamole/GuacamoleSession.java
+++ b/guacamole/src/main/java/org/apache/guacamole/GuacamoleSession.java
@@ -19,6 +19,7 @@
 
 package org.apache.guacamole;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
@@ -60,6 +61,11 @@ public class GuacamoleSession {
      * All currently-active tunnels, indexed by tunnel UUID.
      */
     private final Map<String, UserTunnel> tunnels = new ConcurrentHashMap<>();
+
+    /**
+     * Creation times for all tunnels, indexed by tunnel UUID.
+     */
+    private final Map<String, Long> tunnelCreationTimes = new ConcurrentHashMap<>();
 
     /**
      * Service for dispatching events to registered event listeners.
@@ -246,7 +252,9 @@ public class GuacamoleSession {
      */
     public void addTunnel(UserTunnel tunnel) {
         this.access();
-        tunnels.put(tunnel.getUUID().toString(), tunnel);
+        String tunnelId = tunnel.getUUID().toString();
+        tunnels.put(tunnelId, tunnel);
+        tunnelCreationTimes.put(tunnelId, System.currentTimeMillis());
     }
 
     /**
@@ -262,6 +270,7 @@ public class GuacamoleSession {
      */
     public boolean removeTunnel(String uuid) {
         this.access();
+        tunnelCreationTimes.remove(uuid);
         return tunnels.remove(uuid) != null;
     }
 
@@ -282,6 +291,94 @@ public class GuacamoleSession {
      */
     public long getLastAccessedTime() {
         return lastAccessedTime;
+    }
+
+    /**
+     * Returns the creation time of the specified tunnel, as the number of
+     * milliseconds since midnight January 1, 1970 GMT. Invoking this function
+     * does not affect the last access time of this session.
+     *
+     * @param uuid
+     *     The UUID of the tunnel to get the creation time for.
+     *
+     * @return 
+     *     The time the tunnel was created, or 0 if the tunnel doesn't exist.
+     */
+    public long getTunnelCreationTime(String uuid) {
+        Long creationTime = tunnelCreationTimes.get(uuid);
+        return creationTime != null ? creationTime : 0;
+    }
+
+    /**
+     * Returns the age of the oldest tunnel in this session, in milliseconds.
+     * If no tunnels exist, returns 0. Invoking this function does not affect 
+     * the last access time of this session.
+     *
+     * @return 
+     *     The age of the oldest tunnel in milliseconds, or 0 if no tunnels exist.
+     */
+    public long getOldestTunnelAge() {
+        if (tunnelCreationTimes.isEmpty()) {
+            return 0;
+        }
+        
+        long currentTime = System.currentTimeMillis();
+        long oldestCreationTime = tunnelCreationTimes.values().stream()
+                .mapToLong(Long::longValue)
+                .min()
+                .orElse(currentTime);
+        
+        return currentTime - oldestCreationTime;
+    }
+
+    /**
+     * Closes and removes any tunnels in this session that exceed the specified
+     * age limit. Invoking this function does not affect the last access time
+     * of this session.
+     *
+     * @param maxAge
+     *     The maximum allowed age of tunnels in milliseconds. Tunnels older
+     *     than this will be closed and removed.
+     *
+     * @return 
+     *     The number of tunnels that were closed and removed.
+     */
+    public int closeExpiredTunnels(long maxAge) {
+        if (maxAge <= 0 || tunnelCreationTimes.isEmpty()) {
+            return 0;
+        }
+        
+        long currentTime = System.currentTimeMillis();
+        int closedCount = 0;
+        
+        // Find tunnels that exceed the age limit
+        List<String> expiredTunnelIds = new ArrayList<>();
+        for (Map.Entry<String, Long> entry : tunnelCreationTimes.entrySet()) {
+            long tunnelAge = currentTime - entry.getValue();
+            if (tunnelAge >= maxAge) {
+                expiredTunnelIds.add(entry.getKey());
+            }
+        }
+        
+        // Close and remove expired tunnels
+        for (String tunnelId : expiredTunnelIds) {
+            UserTunnel tunnel = tunnels.get(tunnelId);
+            if (tunnel != null) {
+                try {
+                    tunnel.close();
+                    logger.debug("Closed tunnel \"{}\" due to connection timeout.", tunnelId);
+                }
+                catch (GuacamoleException e) {
+                    logger.debug("Unable to close expired tunnel \"" + tunnelId + "\".", e);
+                }
+                // Remove from both maps regardless of whether close succeeded
+                tunnels.remove(tunnelId);
+                tunnelCreationTimes.remove(tunnelId);
+                closedCount++;
+            }
+        }
+        
+        return closedCount;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


guacamole/src/main/java/org/apache/guacamole/GuacamoleSession.java
```diff
@@ -61,6 +62,11 @@ public class GuacamoleSession {
      */
     private final Map<String, UserTunnel> tunnels = new ConcurrentHashMap<>();
 
+    /**
+     * Creation times for all tunnels, indexed by tunnel UUID.
+     */
+    private final Map<String, Long> tunnelCreationTimes = new ConcurrentHashMap<>();
+
```
necouchman: Thoughts on whether it makes sense to do it this way, or to just add a way within the various `Tunnel` interfaces/classes, to track tunnel creation time? The `GuacamoleTunnel` interface could get a new method `getCreationTime()` (or whatever makes sense), and then one of the implementing classes could get either a `Long` or perhaps one of the Java `Time`/`DateTime` fields to track this?

tkuhlengel: I'm okay with changing it to use the GuacamoleTunnel as backing for the creation time.  I just need a hint on how to iterate through the currently open connections/Tunnels

necouchman: Well, the `tunnels` `Map`, directly above this, has all currently active tunnels indexed by UUID. So, you should just be able to iterate through `tunnels` and, for each tunnel, grab the creation time and close it if it has expired. I think the only real substantive change over what you're already doing is that, instead of tracking tunnel creation time in a separate `Map` you're just using the already-implemented map and iterating through that?

guacamole/src/main/java/org/apache/guacamole/GuacamoleSession.java
```diff
@@ -24,6 +24,7 @@
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+
```
necouchman: Please remove this extra line - our current style does not separate the `import` statements like this, and I don't think this is the place to start that :-).

guacamole-common/src/main/java/org/apache/guacamole/net/GuacamoleTunnel.java
```diff
@@ -121,4 +121,11 @@ public interface GuacamoleTunnel {
      */
     boolean isOpen();
 
+    /**
+     * Returns the creation time of this GuacamoleTunnel.
+     *
+     * @return The creation time of this GuacamoleTunnel.
```
necouchman: I know you're just following the style used in this particular Java file, but, overall, we've moved to the following, and code additions follow this style:
```
/**
 * Returns the creation time of this GuacamoleTunnel
 *
 * @return
 *     The creation time of this GuacamoleTunnel.
 */
```

guacamole/src/main/java/org/apache/guacamole/rest/auth/HashTokenSessionMap.java
```diff
@@ -66,6 +66,18 @@ public class HashTokenSessionMap implements TokenSessionMap {
...
+     * If 0, connections will not be automatically terminated based on age.
+     */
+    private final IntegerGuacamoleProperty CONNECTION_TIMEOUT =
+            new IntegerGuacamoleProperty() {
+
+        @Override
+        public String getName() { return "connection-timeout"; }
+
+    };
+
```
necouchman, line 80: I'm thinking that we may want to name this slightly differently. "timeout" carries with it the notion that the connection is waiting for something - user input, or a response from a remote system, etc. - and what you've really implemented, here, is a maximum connection duration, at which point the connection(s) will be forcibly dropped and users will have to reconnect. My thought is that maybe this code - both internally within the code and for exposed things like properties - should call this something different - `maximum-session-duration` or `session-limit` or something similar?

guacamole/src/main/java/org/apache/guacamole/rest/auth/HashTokenSessionMap.java
```diff
@@ -85,10 +98,25 @@ public HashTokenSessionMap(Environment environment) {
...
         }
+
+        // Read connection timeout from guacamole.properties
+        try {
+            connectionTimeoutValue = environment.getProperty(CONNECTION_TIMEOUT, 0); // Disabled by default
+        }
+        catch (GuacamoleException e) {
+            logger.error("Unable to read guacamole.properties: {}", e.getMessage());
+            logger.debug("Error while reading connection timeout value.", e);
+            connectionTimeoutValue = 0;
```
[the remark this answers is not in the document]
necouchman: Even though I hate CoPilot with hell fury, it's right, here, there should not be an extra space.

guacamole/src/main/java/org/apache/guacamole/rest/auth/HashTokenSessionMap.java
```diff
@@ -85,10 +98,25 @@ public HashTokenSessionMap(Environment environment) {
...
+            logger.debug("Error while reading connection timeout value.", e);
+            connectionTimeoutValue = 0;  
+        }
         
         // Check for expired sessions every minute
         logger.info("Sessions will expire after {} minutes of inactivity.", sessionTimeoutValue);
-        executor.scheduleAtFixedRate(new SessionEvictionTask(sessionTimeoutValue * 60000l), 1, 1, TimeUnit.MINUTES);
+        if (connectionTimeoutValue > 0) {
+            logger.info("Connections will be terminated after {} minutes regardless of activity.", connectionTimeoutValue);
+        } else {
```
necouchman: Please don't cuddle these lines...
```
}

else {
```

guacamole/src/main/java/org/apache/guacamole/rest/auth/HashTokenSessionMap.java
```diff
@@ -104,16 +132,26 @@ private class SessionEvictionTask implements Runnable {
...
         /**
          * Creates a new task which automatically evicts sessions which are
          * older than the specified timeout, or are marked as invalid by an
          * extension.
          * 
          * @param sessionTimeout The maximum age of any session, in
          *                       milliseconds.
+         * @param connectionTimeout The maximum age of any connection, in
+         *                         milliseconds. If 0, connections will not be
+         *                         terminated based on age.
```
necouchman: As with the note in the above file, this should follow the newer style that most additions and updates are using:
```
* @param sessionTimeout
*     The maximum age of any session, in milliseconds
```

guacamole/src/main/java/org/apache/guacamole/GuacamoleSession.java
```diff
@@ -244,9 +246,11 @@ public Map<String, UserTunnel> getTunnels() {
      *
      * @param tunnel The tunnel to associate with this session.
      */
+
```
necouchman: There should not be a space between function documentation and the start of the function itself.

guacamole/src/main/java/org/apache/guacamole/GuacamoleSession.java
```diff
@@ -284,6 +293,94 @@ public long getLastAccessedTime() {
...
+                }
+                catch (GuacamoleException e) {
+                    logger.debug("Unable to close expired tunnel \"" + tunnelId + "\".", e);
+                }
+                // Remove from both maps regardless of whether close succeeded
+                tunnels.remove(tunnelId);
+                tunnelCreationTimes.remove(tunnelId);
+                closedCount++;
+            }
+        }
```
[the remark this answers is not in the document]
necouchman: I'm not sure I agree with CoPilot, here; however, is there any reason not to use the `removeTunnel()` method?

guacamole/src/main/java/org/apache/guacamole/rest/auth/HashTokenSessionMap.java
tkuhlengel: `maximum-session-duration` might work as a descriptive setting name. I don't know if that is a similar name to a user login session, which is different. 

* `maximum-connection-duration` might work.  I'll put that in for now and we can iterate on it.

guacamole/src/main/java/org/apache/guacamole/GuacamoleSession.java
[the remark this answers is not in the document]
tkuhlengel: I'm not the expert in this particular case. 

My understanding of removeTunnel is that it doesn't close an active tunnel, only removes it from the list of active tunnels.  I think that's the equivalent of disassociating it from one user? 

I could use it on line `+375` in this chunk, but I'm not sure how much difference it will make to make it smaller or meaningfully better other than consistency. I still need to close the tunnel first.

tkuhlengel:
In terms of communication with guacd: 
* Would it be okay to use the `SESSION_TIMEOUT` error code or should I make my own? Or possibly just use the 'disconnect' handler on the guacd side?
* If the client needs to send information back to `guacd` when tunnels are closed, why isn't a message isn't sent `tunnel.close()` implementations by default?

necouchman:
> In terms of communication with guacd:
> 
> * Would it be okay to use the `SESSION_TIMEOUT` error code or should I make my own? Or possibly just use the 'disconnect' handler on the guacd side?

I'm tempted to say make another one, because of things I've mentioned above - the implementation, here, isn't so much a "timeout" as it is a limit. To me, "timeout" carries a notion of waiting for something - user interaction, a network response, etc. - and not getting that. In this case, you're actually putting a cap on the session duration regardless of whether the user is active or not, the server is responsive or not, etc.

> * If the client needs to send information back to `guacd` when tunnels are closed, why isn't a message isn't sent `tunnel.close()` implementations by default?

I would guess there just hasn't been a need for it in the past. It could probably be added - my only caution would be to make sure that adding it doesn't break compatibility with clients that don't send a message.

PR #1119, opened 2025-09-30
Guacamole 2086: Add Tomcat HealthCheckValve

dmunozv04:
This PR allows to set Tomcat's HealthCheckValve with environment variables. The implementation is taken from the RemoteIpValve script. 

The second commit uses these new variables to configure docker's HEALTHCHECK. This could be removed if it shouldn't be enabled by default. 

I'm happy to make any changes necessary. Thanks for reviewing this PR.


Dockerfile
```diff
@@ -99,8 +99,11 @@ USER guacamole
...
 ENV BAN_ENABLED=true \
     ENABLE_FILE_ENVIRONMENT_PROPERTIES=true \
-    GUACAMOLE_HOME=/etc/guacamole
+    GUACAMOLE_HOME=/etc/guacamole \
+    HEALTH_CHECK_VALVE_ENABLED=true
 
 # Start Guacamole under Tomcat, listening on 0.0.0.0:8080
 EXPOSE 8080
+HEALTHCHECK --interval=60s --timeout=10s --start-period=30s \
+    CMD curl --fail --silent --show-error http://localhost:8080/health || exit 1
```
necouchman: What happens if the user sets the `HEALTH_CHECK_VALVE_ENABLED` property to `false`?

dmunozv04: Docker will show the container as `Up About a minute (health: starting)` and after the start-period it'll show it as `Up 3 minutes (unhealthy)` 
This health status is just a visual indication, but will not cause the container to fail. 

However, it looks like docker swarm will attempt to restart an unhealthy container to try to get it running ([source](https://statusq.org/archives/2022/02/01/10830/)), and [Podman](https://www.redhat.com/en/blog/podman-edge-healthcheck) can be configured to restart, stop or kill an unhealthy container. 

Kubernetes just ignores this HEALTHCHECK and relies on its own [probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/)

Considering this, and the low impact of having an http route that just replies with `200 OK`, maybe it could be better to always force the health check to enabled or not have it set on the Dockerfile like it's been until now. I don't know what option is better

necouchman: Hmmm....I'm not sure about adding something like this that is known to cause problems. I also did a quick look to see if it's possible to do a conditional `HEALTHCHECK`, and that didn't turn up anything obvious. Maybe in the `|| exit 1` portion a conditional statement could be put to check for the status of the the variable that enables or disables the health check?

Alternatively, I think maybe the way to go is just not to make it configurable and simply enable the Tomcat `HealthCheckValve` always, as you suggest.

@mike-jumper Thoughts or preferences?

mike-jumper: We could move the healthcheck implementation into a script and run _that_ instead of directly running `curl`:

```Dockerfile
HEALTHCHECK --interval=60s --timeout=10s --start-period=30s \
    CMD [ "/opt/guacamole/bin/healthcheck.sh" ]
```

That would allow the script to deal with pinging `http://localhost:8080/health` only if `HEALTH_CHECK_VALVE_ENABLED` is `true`, simply succeeding if not. It would also avoid invoking `curl` unnecessarily, and would give a future point of expansion if we implement additional checks.

dmunozv04: Sounds good to me, I'll implement it soon.
This way the health check will also work if a user changes the `HEALTH_CHECK_VALVE_PATH` variable to something different than `/health`

Thanks for taking your time to review this PR!

dmunozv04: I added the script, let me know if you want any other changes

guacamole-docker/bin/healthcheck.sh
```diff
@@ -0,0 +1,40 @@
...
+## it will simply return success and log that health checks are disabled.
+##
+
+
+# Check if health check valve is enabled
+if [ "$HEALTH_CHECK_VALVE_ENABLED" = "true" ]; then
+    # Default health check path (/health)
+    HEALTH_CHECK_PATH="${HEALTH_CHECK_VALVE_PATH:-/health}"
+    # Perform actual health check via curl
+    curl --fail --silent --show-error "http://localhost:8080${HEALTH_CHECK_PATH}" || exit 1
```
necouchman, line 35: A couple of concerns about the `HEALTH_CHECK_VALVE_PATH` variable:
* It looks like it assumes that an absolute path will be provided?
* What happens if the user provides a relative path (no leading /), both here and in the server.xml file?

dmunozv04: I assumed an absolute path would be provided since it looks like tomcat only supports absolute paths:
https://github.com/apache/tomcat/blob/eaea7503163e44ba7260c9c7b3af3f4ba387d072/java/org/apache/catalina/valves/HealthCheckValve.java#L82

If the user provides a relative path, the health check won't be reachable and curl will complain: `"curl: (3) URL rejected: Port number was not a decimal number between 0 and 65535\n"`

Maybe a check can be added when loading the variables to warn the user if the path is not absolute.

necouchman: Yeah, I would think checking the variable would be the right way to go...

PR #1138, opened 2025-12-16
Add Displayname in web.xml

scal-csimon:


necouchman:
@scal-csimon Thanks for the contribution. In order to consider and move forward with this, you will first need to:
* Get a [Jira](https://issues.apache.org/jira/browse/GUACAMOLE) account and create a Jira issue for this, describing the motivation behind this change.
* Tag both your commit message and the pull request with the Jira issue (`GUACAMOLE-0000:`).

adb014:
@necouchman Do you want this trivial change ? If so I'll create the jira ticket, rebase it against staging1.6.1 and try to get it closed

necouchman:
@adb014 I don't particularly care much one way or the other. If it's a useful thing to do, then, by all means, put in the Jira issue and the updated PR. It isn't something that I've ever run into needing, so I'm not entirely sure what the need is and we can also probably just close it and move on.

adb014:
In a Tomcat application, I'm not even sure where "<display-name>" is even used. Briefly, testing this it didn't seem to change anything... So, I'd say unless the original author steps up and says why they want this, this should be dropped.

PR #1147, opened 2026-01-06
GUACAMOLE-1170: Support sized IntegerPool to maximize time between stream index reuse.

sschiffli:
Adding ability for IntegerPool to accept a minimum pool size, so new integers are preferred over reused integers until the minimum size is reached (just like the guacamole-sever pool).

This fixes a race condition with stream index reuse outlined in GUACAMOLE-1170.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/IntegerPool.js
+++ b/guacamole-common-js/src/main/webapp/modules/IntegerPool.js
@@ -22,9 +22,16 @@ var Guacamole = Guacamole || {};
 /**
  * Integer pool which returns consistently increasing integers while integers
  * are in use, and previously-used integers when possible.
- * @constructor 
+ *
+ * When a minimum pool size is specified, new integers are preferred over reused
+ * integers until the minimum size is reached.
+ *
+ * @constructor
+ * @param {number} [min_size]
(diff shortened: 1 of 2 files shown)


guacamole-common-js/src/main/webapp/modules/IntegerPool.js
```diff
@@ -51,13 +58,17 @@ Guacamole.IntegerPool = function() {
...
+     *
      * @return {!number}
      *     The next available integer.
      */
     this.next = function() {
 
-        // If free'd integers exist, return one of those
+        // If min_size is specified, prefer new integers until we reach it
+        if (min_size && guac_pool.next_int < min_size)
+            return guac_pool.next_int++;
```
mike-jumper: Rather than having `guac_pool.next_int++` in two distinct spots, I think it would be better to absorb the inverse of this test into the `if (pool.length > 0)` block below.

mike-jumper (changes requested):

guacamole-common-js/src/main/webapp/modules/IntegerPool.js
sschiffli: Good call, that is much easier to read

PR #1153, opened 2026-01-25
GUACAMOLE-1286: Add support for providing a custom IV.

necouchman:
Based on another situation that I've run into where I need to be able to provide an `IV` for the crypto module, I've tweaked the code for the JSON extension slightly to allow this value to be passed in as a query parameter along with the encrypted and signed JSON data. I've no idea if this is actually safe or not (from what I've read it seems that passing the IV in the open with the encrypted data is okay?), and am opening as a draft since I haven't even tested it, yet.

### DIFF
--- a/extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/CryptoService.java
+++ b/extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/CryptoService.java
@@ -19,6 +19,7 @@
 
 package org.apache.guacamole.auth.json;
 
+import com.google.common.io.BaseEncoding;
 import java.security.InvalidAlgorithmParameterException;
 import java.security.InvalidKeyException;
 import java.security.Key;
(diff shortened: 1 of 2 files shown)


extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/CryptoService.java
```diff
@@ -133,13 +137,25 @@ public SecretKey createSignatureKey(byte[] keyBytes) {
...
+    public byte[] decrypt(Key key, byte[] cipherText, String ivStr)
+            throws GuacamoleException {
 
         try {
 
+            // Use NULL_IV by default, but, if an IV has been provided as
+            // part of the request, use that, instead.
+            IvParameterSpec iv = NULL_IV;
+            if (ivStr != null && !ivStr.isEmpty()) {
+                byte[] ivBytes = BaseEncoding.base64().decode(ivStr);
```
eugen-keeper, line 149: This can throw `IllegalArgumentException` for malformed `ivStr` base64, but that exception is not handled here. It propagates to `UserDataService.fromCredentials()`, where the existing catch (`IllegalArgumentException`) logs `Submitted data is not proper base64`, which is misleading when only iv is invalid. Maybe consider catching `IllegalArgumentException` here and rethrowing as `GuacamoleServerException("Invalid IV provided.")`.

extensions/guacamole-auth-json/src/main/java/org/apache/guacamole/auth/json/CryptoService.java
eugen-keeper, line 122: The method now accepts an optional `ivStr` parameter for a custom IV. Please consider updating the Javadoc to reflect the new behavior.

eugen-keeper (changes requested):
Please consider updating the README to document the new behavior `extensions\guacamole-auth-json\doc\README.md`

PR #1155, opened 2026-01-27
GUACAMOLE-2207: Add support for Cmd(Meta)-based click modifiers

sschiffli:
Forward guacClientMouseDown events from the Mouse module to the Keyboard module via a new handleMouseEvent() method. When a mouse click occurs with the Meta modifier pressed and there's a pending Meta keydown event, the mouse event is added to the keyboard event log to provide the context needed to resolve the deferred Meta key. This enables the Meta key to be sent before the mouse click, properly delivering Cmd+Click to the backend.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Keyboard.js
+++ b/guacamole-common-js/src/main/webapp/modules/Keyboard.js
@@ -294,8 +294,10 @@ Guacamole.Keyboard = function Keyboard(element) {
             this.keysym = keysym_from_key_identifier(this.keyIdentifier, this.location, this.modifiers.shift);
 
         // If a key is pressed while meta is held down, the keyup will
-        // never be sent in Chrome (bug #108404)
-        if (this.modifiers.meta && this.keysym !== 0xFFE7 && this.keysym !== 0xFFE8)
+        // never be sent in Chrome (bug #108404). Modifier keys are excluded
+        // from this workaround as they have reliable keyup events and need
+        // to be held down simultaneously with Meta.
+        if (this.modifiers.meta && !isModifierKey(this.keysym))
             this.keyupReliable = false;
 
         // We cannot rely on receiving keyup for lock keys on certain platforms
@@ -594,6 +596,23 @@ Guacamole.Keyboard = function Keyboard(element) {
     };
 
     /**
+     * All modifier key keysyms for easy lookup.
+     *
+     * @private
+     * @type {!Object.<number, boolean>}
+     */
+    var modifierKeysyms = (function() {
+        var lookup = {};
+        for (var modifier in modifierKeysymsByType) {
+            var keysyms = modifierKeysymsByType[modifier];
+            for (var i = 0; i < keysyms.length; i++) {
+                lookup[keysyms[i]] = true;
+            }
+        }
+        return lookup;
+    })();
+
+    /**
      * All keysyms that represent each supported toggle modifier
      * type.
      *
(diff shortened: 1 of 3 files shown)


guacamole-common-js/src/main/webapp/modules/Keyboard.js
```diff
@@ -285,8 +285,10 @@ Guacamole.Keyboard = function Keyboard(element) {
             this.keysym = keysym_from_key_identifier(this.keyIdentifier, this.location, this.modifiers.shift);
 
         // If a key is pressed while meta is held down, the keyup will
-        // never be sent in Chrome (bug #108404)
-        if (this.modifiers.meta && this.keysym !== 0xFFE7 && this.keysym !== 0xFFE8)
+        // never be sent in Chrome (bug #108404). Modifier keys are excluded
+        // from this workaround as they have reliable keyup events and need
+        // to be held down simultaneously with Meta.
+        if (this.modifiers.meta && !isModifierKey(this.keysym))
```
sschiffli, line 300: I could see an argument where this should be a separate issue, as this is needed to support CMD+Shift+Click (not CMD+Click), so let me know what you think or if a rewording of the JIRA/PR would be sufficient.

necouchman: I think it's fine in here. You can reword the Jira issue if you want to make it more clear that it's working to support both Shift+CMD an Cmd, but no major concerns from me.

guacamole/src/main/frontend/src/app/index/controllers/indexController.js
```diff
@@ -183,6 +183,11 @@ angular.module('index').controller('indexController', ['$scope', '$injector',
     var keyboard = new Guacamole.Keyboard($document[0]);
     keyboard.listenTo(sink.getElement());
 
+    // Resolve deferred Meta key events on mouse click to enable Cmd+Click
+    $scope.$on('guacBeforeClientMouseDown', function(event, mouseEvent, client) {
```
bbennett-ks: Will unused argument (client) cause lint errors?

bbennett-ks (approved):

guacamole-common-js/src/main/webapp/modules/Keyboard.js
```diff
@@ -926,6 +987,39 @@ Guacamole.Keyboard = function Keyboard(element) {
...
+    /**
+     * Handles a mouse event to resolve deferred Meta key events. When a Meta
+     * key is pressed, it is deferred to determine if it's being used as a
+     * modifier or as a standalone key. A mouse click with Meta held provides
+     * the context needed to resolve this, enabling Cmd+Click functionality.
+     *
+     * @param {Guacamole.Mouse.Event} mouseEvent
+     *     The mouse event that occurred.
+     */
+    this.handleMouseEvent = function(mouseEvent) {
```
mike-jumper: I think we should consider renaming this. To me, "handle" implies that the entirety of event processing is dealt with here, but this is specific to updating the keyboard state with any changes that are visible only within the mouse event.

I'd suggest something like `updateModifierState()`, `syncModifierState()`, etc.

guacamole-common-js/src/main/webapp/modules/Mouse.js
```diff
@@ -510,6 +510,81 @@ Guacamole.Mouse.State.Buttons = {
...
+/**
+ * Returns the modifier state applicable to the mouse event given.
+ *
+ * @param {!MouseEvent} e
+ *     The mouse event to read.
+ *
+ * @returns {!Guacamole.Mouse.ModifierState}
+ *     The current state of keyboard modifiers.
+ */
+Guacamole.Mouse.ModifierState.fromMouseEvent = function(e) {
```
mike-jumper: Rather than duplicate `Guacamole.Keyboard.ModifierState` onto `Guacamole.Mouse`, maybe we should add `Guacamole.Keyboard.ModifierState.fromMouseEvent()`?

mike-jumper (changes requested):

guacamole-common-js/src/main/webapp/modules/Keyboard.js
sschiffli: Yeah I struggled with this naming.  There are already `syncModifierStates` and `updateModifierState` in this file, so maybe `updateModifiersFromMouse` is a good canidate?

guacamole-common-js/src/main/webapp/modules/Mouse.js
sschiffli: I considered this as well, I originally opted for duplication since MouseEvent and KeyboardEvent didn't share any common base class, but practically speaking this does feel a bit unnecessary.

mike-jumper (approved):

PR #1158, opened 2026-01-31
GUACAMOLE-2211: Add example JavaScript encryption.

necouchman:


extensions/guacamole-auth-json/doc/encrypt-json.js
```diff
@@ -0,0 +1,183 @@
...
+}
+
+/**
+ * Take the encoded array and generate an HMAC signature for it,
+ * combining the HMAC signature with the data and returning the
+ * resultant array.
+ *
+ * @param key
+ *     The 128-bit key used to sign the data, provided as a
+ *     16-character hexadecimal value.
```
eugen-keeper, line 81: Probably 16 bytes and 32-character hexadecimal value.

eugen-keeper (changes requested):

PR #1170, opened 2026-02-20
GUACAMOLE-2224: Fix effective user group expansion for SAML/SSO groups

tkuhlengel:
PR corresponding to https://issues.apache.org/jira/browse/GUACAMOLE-2224

## Problem

When a user authenticates via an external SSO provider (e.g. SAML with Azure AD), the identity provider asserts group membership by name. These names are stored on the `AuthenticatedUser` and surfaced through `getEffectiveUserGroups()`.

The previous implementation in `ModeledAuthenticatedUser` resolved this with a flat union:

```java
return Sets.union(user.getEffectiveUserGroups(), super.getEffectiveUserGroups());
```

This combined the user's database group memberships with the raw SSO claims, but made no attempt to walk the database group hierarchy. If `child-group` (matched by SAML claim) was a member of `parent-group` in the database, `parent-group` was never included — and any permissions granted only to `parent-group` were invisible to the SSO user.

Database-native users were unaffected because their memberships are stored as `entity_id` references already handled by the existing recursive expansion in `EntityService`.

### `ModeledPermissions.expandEffectiveGroups()` — guacamole-auth-jdbc-base

A new method delegates to `EntityService.retrieveEffectiveGroups()`, passing caller-provided external group names as additional seeds for recursive DB group expansion:

```java
public Set<String> expandEffectiveGroups(Collection<String> externalEffectiveGroups) {
    return entityService.retrieveEffectiveGroups(this, externalEffectiveGroups);
}
```

This exposes the existing recursive CTE logic as a callable entry point on the permissions object.

### `ModeledAuthenticatedUser.getEffectiveUserGroups()` — guacamole-auth-jdbc-base

The flat union is replaced with a call to `expandEffectiveGroups()`, passing the raw SSO claims as seeds:

```java
@Override
public Set<String> getEffectiveUserGroups() {
    return user.expandEffectiveGroups(super.getEffectiveUserGroups());
}
```

`super.getEffectiveUserGroups()` returns the SSO group name claims from the originating `AuthenticatedUser`. These are now fed into the recursive expansion, so any ancestor groups of those claims in the database are included in the result — making inherited permissions visible to SSO users.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledAuthenticatedUser.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledAuthenticatedUser.java
@@ -19,7 +19,6 @@
 
 package org.apache.guacamole.auth.jdbc.user;
 
-import com.google.common.collect.Sets;
 import java.util.Collections;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
@@ -170,11 +169,21 @@ public class ModeledAuthenticatedUser extends RemoteAuthenticatedUser {
     public void setIdentifier(String identifier) {
         user.setIdentifier(identifier);
     }
-    
+
+
+    /**
+     * Expands a user's groups through the parents in database group hierarchy
+     * so that parent groups of external groups (e.g. SAML/SSO group claims)
+     * are included. This also covers the user's own direct
+     * DB memberships (via entity_id) and skeleton users with null entity_id.
+     *
+     * @return
+     *     The set of effective groups for this user, whether inherited or
+     *     direct.
+     */
     @Override
     public Set<String> getEffectiveUserGroups() {
-        return Sets.union(user.getEffectiveUserGroups(),
-                super.getEffectiveUserGroups());
+        return user.expandEffectiveGroups(super.getEffectiveUserGroups());
     }
 
     /**
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledAuthenticatedUser.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledAuthenticatedUser.java
@@ -23,6 +23,7 @@ import java.util.Collections;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import org.apache.guacamole.GuacamoleException;
+import org.apache.guacamole.auth.jdbc.base.EntityService;
 import org.apache.guacamole.net.auth.AuthenticatedUser;
 import org.apache.guacamole.net.auth.AuthenticationProvider;
 import org.apache.guacamole.net.auth.Credentials;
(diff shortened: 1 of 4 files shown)


tkuhlengel:
This is ready for review.

necouchman:
Thanks @tkuhlengel, I'll try to take a look. FYI, your "Typo fix" commit needs to either be properly tagged with the GUACAMOLE-2224 issue, or needs to be squashed into the other one.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledAuthenticatedUser.java
```diff
@@ -170,11 +169,21 @@ public String getIdentifier() {
...
+     *
+     * @return
+     *     The set of effective groups for this user, whether inherited or
+     *     direct.
+     */
     @Override
     public Set<String> getEffectiveUserGroups() {
-        return Sets.union(user.getEffectiveUserGroups(),
-                super.getEffectiveUserGroups());
+        return user.expandEffectiveGroups(super.getEffectiveUserGroups());
```
necouchman, line 186: Rather than creating a new `expandEffectiveGroups()` method in the `ModeledUser` class, is there any reason not to just call the `entityService.retrieveEffectiveGroups()` method, here, and add it to the `Sets.union()` call?

tkuhlengel: I think `ModeledAuthenticatedUser` doesn't have access to `entityService`.  It's injected in `ModeledPermissions`, which `ModeledUser` inherits from. The thin `expandEffectiveGroups()` wrapper keeps the service access encapsulated without changing any permissions or inheritance.

To call `entityService` directly here I'd need to either inject it into `ModeledAuthenticatedUser` or add a getter method on `ModeledUser`.

I can try to go that route if you want. The wrapper `user.expandEffectiveGroups()` seemed like the most compact option.

tkuhlengel:
Fixed the commit issue.

Replied to the review point.

tkuhlengel:
Ready for followup.

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledAuthenticatedUser.java
necouchman: I think injecting `entityService` into the `ModeledAuthenticatedUser` class would be the way I'd prefer to go - it seems to me that would result, overall, in easier-to-follow code, rather than having to jump from one class to another and then on to another.

tkuhlengel: Addressed in the latest commit, though it is a bit bigger of a changeset.

necouchman:
Thanks, @tkuhlengel - this looks good. Could you rebase against `staging/1.6.1` and we'll get it into this next release?

tkuhlengel:
@necouchman I think I finished the rebase correctly. Please let me know if you need something different from the rebase or if it needs to be squashed.  

I assumed that any changes that weren't mine were intended to be as they were from `staging/1.6.1`

corentin-soriano:
> @necouchman I think I finished the rebase correctly. Please let me know if you need something different from the rebase or if it needs to be squashed.
> 
> I assumed that any changes that weren't mine were intended to be as they were from `staging/1.6.1`

Could you please drop the commits coming from the main branch?

tkuhlengel:
> > @necouchman I think I finished the rebase correctly. Please let me know if you need something different from the rebase or if it needs to be squashed.
> > I assumed that any changes that weren't mine were intended to be as they were from `staging/1.6.1`
> 
> Could you please drop the commits coming from the main branch?

Done

PR #1173, opened 2026-02-23
GUACAMOLE-2228: Write stream blobs asynchronously.

eugen-keeper:
The PR contains some changes to the core functionality, so I would appreciate it if you could review it in detail. Thanks!


necouchman:
> The PR contains some changes to the core functionality, so I would appreciate it if you could review it in detail. Thanks!

If this changes core functionality, then it should probably be based against the `main` branch rather than the `staging/1.6.1` or even the `patch` branch. Changing core functionality sounds to me like something that should be in the next feature release (1.7.0) rather than a patch release - unless, of course, it breaks compatibility with anything, in which case it should go to `next`.

PR #1176, opened 2026-02-25
GUACAMOLE-2232: On-Screen Keyboard: CapsLock gets out-of-sync on macOS.

bbennett-ks:
## Summary
- `OnBoardKeyBoard.js`: 
   - Added `Key` object  `toggle` property.
   - Updated logic to send `<keyup><keydown>` for each key press for keys with `toggle == true`.
 - `*erty.json`:
    - Add `toggle = true` for `<Caps>` for all keyboard layouts.

## Testing
- macOS Tahoe (26.3): native browser (Chrome) and Rocky 9 VM  Chrome browser
- Windows 11: native browser (Chrome) and Rocky 9 VM  Chrome browser

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/OnScreenKeyboard.js
+++ b/guacamole-common-js/src/main/webapp/modules/OnScreenKeyboard.js
@@ -306,28 +306,36 @@ Guacamole.OnScreenKeyboard = function(layout) {
                 // Retrieve originally-pressed keysym, if modifier was already pressed
                 var originalKeysym = modifierKeysyms[key.modifier];
 
-                // Activate modifier if not pressed
                 if (originalKeysym === undefined) {
-                    
+                    // Activate modifier if not pressed
                     addClass(keyboard, modifierClass);
                     modifierKeysyms[key.modifier] = key.keysym;
-                    
-                    // Send key event only if keysym is meaningful
-                    if (key.keysym && osk.onkeydown)
-                        osk.onkeydown(key.keysym);
-
                 }
-
-                // Deactivate if not pressed
                 else {
-
+                    // Deactivate modifier if pressed
                     removeClass(keyboard, modifierClass);
                     delete modifierKeysyms[key.modifier];
-                    
-                    // Send key event only if original keysym is meaningful
-                    if (originalKeysym && osk.onkeyup)
-                        osk.onkeyup(originalKeysym);
+                }
 
+                // Toggle modifiers send both keydown and keyup on each press
+                if (key.toggle) {
+                    if (key.keysym && osk.onkeydown)
+                        osk.onkeydown(key.keysym);
+                    if (key.keysym && osk.onkeyup)
+                        osk.onkeyup(key.keysym);
+                }
+
+                // Sticky modifiers send keydown on activate, keyup on deactivate
+                else {
+                    if (originalKeysym === undefined) {
+                        if (key.keysym && osk.onkeydown)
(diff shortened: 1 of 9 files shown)


guacamole-common-js/src/main/webapp/modules/OnScreenKeyboard.js
```diff
@@ -58,6 +58,23 @@ Guacamole.OnScreenKeyboard = function(layout) {
...
+     * @param {!number} keysym
+     *     The keysym to test.
+     *
+     * @returns {!boolean}
+     *     true if the keysym is a lock key, false otherwise.
+     */
+    var isToggleLockKeysym = function isToggleLockKeysym(keysym) {
+        return keysym === 0xFFE5  // Caps Lock
+            || keysym === 0xFF7F  // Num Lock
+            || keysym === 0xFF14; // Scroll Lock
```
mike-jumper: The on-screen keyboard implementation intentionally avoids making any assumptions about how the keys that send particular keysyms behave, relying instead on the JSON describing the layout to define this:

https://github.com/apache/guacamole-client/blob/4245fbb7e2224e43141e4e5833647a38988edd0b/guacamole/src/main/frontend/src/layouts/en-us-qwerty.json#L55-L94

Keeping with that pattern, I don't think we should do this here. We should instead allow the JSON to define whether a modifier is a toggle modifier.

mike-jumper (changes requested):

guacamole-common-js/src/main/webapp/modules/OnScreenKeyboard.js
bbennett-ks: Reverted previous change; add new `toggle` parameter to `key` object.

necouchman:
@bbennett-ks You appear to have either multiple Jira issues or some extra commits, here. I see two for GUACAMOLE-2225 and one for GUACAMOLE-2232.

bbennett-ks:
> @bbennett-ks You appear to have either multiple Jira issues or some extra commits, here. I see two for [GUACAMOLE-2225](https://issues.apache.org/jira/browse/GUACAMOLE-2225) and one for [GUACAMOLE-2232](https://issues.apache.org/jira/browse/GUACAMOLE-2232).

Fixed. Thanks!

guacamole-common-js/src/main/webapp/modules/OnScreenKeyboard.js
```diff
@@ -306,28 +306,35 @@ Guacamole.OnScreenKeyboard = function(layout) {
...
-                    // Send key event only if keysym is meaningful
-                    if (key.keysym && osk.onkeydown)
-                        osk.onkeydown(key.keysym);
-
                 }
-
-                // Deactivate if not pressed
                 else {
-
+                    // Deactivate if not pressed
```
eugen-keeper: `Deactivate if pressed` ?

guacamole-common-js/src/main/webapp/modules/OnScreenKeyboard.js
```diff
@@ -375,7 +382,7 @@ Guacamole.OnScreenKeyboard = function(layout) {
         }
 
     };
-
+// Deactivate if not pressed
```
eugen-keeper: Probably a typo which should be removed.

eugen-keeper (changes requested):

bbennett-ks:
Fixed merge conflict.

PR #1185, opened 2026-03-15
fix(build): disable tests, minification, and use token instead of id_…

dnoliver:
NOTE: the objective of this PR is to share what changes I did to get this working to "my" Cognito configuration, no guarantees that this works for other Cognito instances

…token

- The changes allows building on OSX due to a problem in closure compiler
- Skip Maven tests in Docker build by default
- Change OpenID response_type from id_token to token for AWS Cognito compatibility
- Disable Google Closure Compiler minification in webpack config

These changes are to be temporary workarounds for compatibility issues with AWS Cognito's OpenID implementation.


mike-jumper:
Closing as PR description indicates this is not intended to be merged.

PR #1190, opened 2026-03-26
GUACAMOLE-2002: Deduplicate Kubernetes clipboard configuration section.

eugen-keeper:
Only the **main** branch has the duplication issue. So I assume this should be merged to there directly.


mike-jumper (approved):

mike-jumper:
LGTM - confirmed that this applies only to `main`/`next` (1.7.0+) and not `patch` (1.6.2) or `staging/1.6.1`, and that the metadata of [GUACAMOLE-2002](https://issues.apache.org/jira/browse/GUACAMOLE-2002) matches this.

PR #1195, opened 2026-04-01
GUACAMOLE-2252: Preserve Ctrl+Alt when AltGr is genuinely held.

amswiatkowski:
On Windows, the browser sends a synthetic Left Ctrl keydown before the AltGraph keydown. The release_simulated_altgr function checks whether Ctrl+Alt are pressed and, for printable keysyms, releases them. This breaks AltGr character input because Windows RDP targets require both Ctrl and Alt scancodes to remain active for AltGr resolution.

Add a guard that checks whether AltGr (0xFE03) or Right Alt (0xFFEA) is genuinely held in guac_keyboard.pressed before allowing the release. When either key is active, Ctrl+Alt are preserved, and the correct AltGr character is produced on the target.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Keyboard.js
+++ b/guacamole-common-js/src/main/webapp/modules/Keyboard.js
@@ -1339,6 +1339,13 @@ Guacamole.Keyboard = function Keyboard(element) {
         if (!guac_keyboard.modifiers.ctrl || !guac_keyboard.modifiers.alt)
             return;
 
+        // Do not release Ctrl+Alt when AltGr is genuinely held. On Windows,
+        // the browser sends a synthetic Left Ctrl before AltGraph. RDP
+        // targets require both Ctrl and Alt scancodes to remain active for
+        // AltGr character resolution.
+        if (guac_keyboard.pressed[0xFE03] || guac_keyboard.pressed[0xFFEA])
+            return;
+
         // Assume [A-Z] never require AltGr
         if (keysym >= 0x0041 && keysym <= 0x005A)
             return;


guacamole-common-js/src/main/webapp/modules/Keyboard.js
```diff
@@ -1339,6 +1339,13 @@ Guacamole.Keyboard = function Keyboard(element) {
         if (!guac_keyboard.modifiers.ctrl || !guac_keyboard.modifiers.alt)
             return;
 
+        // Do not release Ctrl+Alt when AltGr is genuinely held. On Windows,
+        // the browser sends a synthetic Left Ctrl before AltGraph. RDP
+        // targets require both Ctrl and Alt scancodes to remain active for
+        // AltGr character resolution.
+        if (guac_keyboard.pressed[0xFE03] || guac_keyboard.pressed[0xFFEA])
+            return;
```
mike-jumper, line 1347: I'm not sure about this. It's true that Windows emulates AltGr with Ctrl+Alt, but:

* Windows also accepts the dedicated AltGr key, which the RDP support already sends.
* If RDP servers in general do require different key events for AltGr, continuing to represent AltGr as AltGr within the Guacamole protocol is still the correct thing to do, as it preserves fidelity of user intent. It's the RDP support itself that should make any necessary protocol- or platform-specific translations.

The proposed logic would also mean that the:

```c
guac_keyboard.release(0xFFE3); // Left ctrl
...
guac_keyboard.release(0xFFEA); // Right alt
```

lines later on in this function will never have any effect (as the only way these lines will be reached is if neither of these keys are pressed).

mike-jumper (changes requested):

PR #1196, opened 2026-04-01
GUACAMOLE-2253: Fix Shift key race condition causing random capitaliz…

amswiatkowski:
When a user releases Shift before a character key, the browser reports the unshifted keysym on keyup, creating a mismatched press/release pair where the shifted keysym is never explicitly released. Over time, this leaves orphaned pressed-key state and causes seemingly random uppercase characters in the remote session.

Track non-modifier keys pressed while Shift is held in a dedicated map. On Shift keyup, force-release all tracked keys using their original (shifted) keysym before releasing Shift itself, ensuring every press has a matching release.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Keyboard.js
+++ b/guacamole-common-js/src/main/webapp/modules/Keyboard.js
@@ -644,6 +644,18 @@ Guacamole.Keyboard = function Keyboard(element) {
     var lastToggleKeydownState = {};
 
     /**
+     * Map of non-modifier keys pressed while Shift is held, keyed by keyCode,
+     * storing the keysym used at press time. When Shift is released before the
+     * character key, the keyup event resolves to the unshifted keysym, creating
+     * a mismatched press/release pair. This map allows force-releasing tracked
+     * keys with their original (shifted) keysym on Shift keyup.
+     *
+     * @private
+     * @type {!Object.<number, number>}
+     */
+    var shiftedKeys = {};
+
+    /**
      * The state of every key, indexed by keysym. If a particular key is
      * pressed, the value of pressed for that keysym will be true. If a key
      * is not currently pressed, it will not be defined. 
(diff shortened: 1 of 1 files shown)


guacamole-common-js/src/main/webapp/modules/Keyboard.js
```diff
@@ -1459,8 +1480,22 @@ Guacamole.Keyboard = function Keyboard(element) {
...
+                // pressed while Shift was held, using their original (shifted)
+                // keysym. This prevents mismatched press/release pairs when
+                // Shift keyup arrives before the character keyup.
+                if (keysym === 0xFFE1 || keysym === 0xFFE2) {
+                    for (var keyCode in shiftedKeys) {
+                        guac_keyboard.release(shiftedKeys[keyCode]);
+                        delete recentKeysym[keyCode];
+                    }
+                    shiftedKeys = {};
+                }
```
mike-jumper: Rather than track specifically which keys are held while particular modifiers are active, I think a more robust approach would be to entirely `reset()` the keyboard state if we receive a keyup for a keysym that isn't recorded as pressed.

This is essentially we already do when an unrecognized keyup is received from the browser:

https://github.com/apache/guacamole-client/blob/4245fbb7e2224e43141e4e5833647a38988edd0b/guacamole-common-js/src/main/webapp/modules/Keyboard.js#L1467-L1471

mike-jumper:
> When a user releases Shift before a character key, the browser reports the unshifted keysym on keyup, creating a mismatched press/release pair where the shifted keysym is never explicitly released. Over time, this leaves orphaned pressed-key state and causes seemingly random uppercase characters in the remote session.

@amswiatkowski I'm not sure this is true. The following takes effect when the `KeyupEvent` is initially created:

https://github.com/apache/guacamole-client/blob/4245fbb7e2224e43141e4e5833647a38988edd0b/guacamole-common-js/src/main/webapp/modules/Keyboard.js#L377-L380

If a user releases a key that isn't already known to be pressed, the current behavior is already to maintain consistency with the keysym used most recently for the key having that JavaScript keycode.

PR #1197, opened 2026-04-07
GUACAMOLE-2250: Add an option to create groups on connection import.

eugen-keeper:


guacamole/src/main/frontend/src/app/import/controllers/importConnectionsController.js
```diff
@@ -345,6 +358,170 @@ angular.module('import').controller('importConnectionsController', ['$scope', '$
...
+                        // Create the new group under the parent's REAL identifier
+                        const newGroup = new ConnectionGroup({
+                            name: segment,
+                            parentIdentifier: identifierByPath[parentPath] 
+                        });
+
+                        return connectionGroupService.saveConnectionGroup(dataSource, newGroup)
+                            .then(() => {
+                                identifierByPath[nextPath] = newGroup.identifier;
+                            });
```
bbennett-ks, line 448: My understanding is that we can't create a multi-level group in a single REST API call, since we need the parent identifier before we can create each child group.

Since we don't catch failures here, a failed group creation will abort the import, potentially leaving behind full and intermediate groups that were created earlier... Should we track and delete those groups on failure, or is it better to document that partial group creation can occur?

guacamole/src/main/frontend/src/app/import/controllers/importConnectionsController.js
```diff
@@ -345,6 +358,170 @@ angular.module('import').controller('importConnectionsController', ['$scope', '$
...
+                    });
+                });
+            });
+
+            return sequence.then(() => {
+                parseResult.connectionObjects.forEach(connectionObject => {
+                    if (!connectionObject.parentIdentifier) {
+                        connectionObject.parentIdentifier =
+                            identifierByPath[connectionObject.group] || identifierByPath['ROOT'];
```
bbennett-ks: If the group path can't be (somehow) resolved, we'll silently place the connection under ROOT. Is that intentional, or should we abort importing that connection instead? And add a warning, either way?

guacamole/src/main/frontend/src/app/import/controllers/importConnectionsController.js
bbennett-ks: Just noticed this:
```
 If not, any errors will be displayed and any already-created entities
     * will be rolled back.
```

guacamole/src/main/frontend/src/app/import/controllers/importConnectionsController.js
```diff
@@ -368,57 +545,64 @@ angular.module('import').controller('importConnectionsController', ['$scope', '$
...
+                                        $location.url('/settings/' + dataSource + '/connections');
+                                    }
+                                }]
+                            });
+                        });
+                    })
+                    // If an error occurs while trying to create users or groups, 
+                    // display the error to the user.
+                    .catch(error => {
+                        $scope.processing = false;
```
bbennett-ks, line 626: Add `console.error(error)` here, so this can be seen on the DevTools console?? Or maybe better, add to `handleError()`? Multiple instances in file.

guacamole/src/main/frontend/src/app/import/services/connectionParseService.js
bbennett-ks: Existing nit: I noticed `lookups.groupPathsByIdentifier` names are reversed. `lookups.groupIdentifiersByPath' May be worth renaming correctly to prevent a future bug. 

I searched the file for use to see how they were indexed, and found a couple discrepancies it may be worth looking into.
**groupPathsByIdentifier** (mapped by path):
```
                    && !groupPathsByIdentifier[providedIdentifier]) {    <<<
                group = groupPathsByIdentifier[providedIdentifier];      <<<
```

guacamole/src/main/frontend/src/app/import/templates/connectionImport.html
```diff
@@ -58,6 +58,14 @@ <h2>{{'IMPORT.SECTION_HEADER_CONNECTION_IMPORT' | translate}}</h2>
                 </label>
                 <span ng-attr-title="{{'IMPORT.HELP_EXISTING_PERMISSION_MODE' | translate}}" class="help"></span>
             </li>
+            <li>
+                <input type="checkbox"
```
bbennett-ks, line 62: You may want to look into `guacamole-manual/src/batach-import.md` and update it with the new UI setting.

bbennett-ks (changes requested):

guacamole/src/main/frontend/src/app/import/controllers/importConnectionsController.js
eugen-keeper: Yes — that’s correct. Guacamole’s connection group API creates one group per POST, and each child needs its parent’s identifier. Only ROOT is known upfront; every other parent gets a server-assigned identifier when it’s created, so a multi-level hierarchy has to be built level by level, not in a single call.

Documented (lines 373 - 377). A rollback would be complex (track only new groups, delete deepest-first, handle delete errors) and wouldn’t make the whole import atomic anyway — groups can also remain if connection import fails later.

guacamole/src/main/frontend/src/app/import/controllers/importConnectionsController.js
eugen-keeper: Fixed! Unresolved paths now show in the import error table and block the import.

guacamole/src/main/frontend/src/app/import/controllers/importConnectionsController.js
eugen-keeper: Fixed

guacamole/src/main/frontend/src/app/import/controllers/importConnectionsController.js
eugen-keeper: Added it into `handleError`

guacamole/src/main/frontend/src/app/import/services/connectionParseService.js
eugen-keeper: Fixed

guacamole/src/main/frontend/src/app/import/templates/connectionImport.html
eugen-keeper: I will submit a PR to there.

eugen-keeper:
https://github.com/apache/guacamole-manual/pull/294

guacamole/src/main/frontend/src/app/import/controllers/importConnectionsController.js
bbennett-ks: I think there is still an issue here. I **Import** (Automatically create groups) the following.
```
[
    { "name": "repro-5", "protocol": "vnc", "group": "ROOT//Repro5" }
]
```
But no error is reported:
image

And the group was created.

bbennett-ks:
I noticed 1 other thing (unrelated to this PR): the CONNECTION IMPORT screen: if no file is selected, the Cancel button is grayed out & there's no way to exit the screen. Besides selecting another menu option.

I'll file s separate issue for this as I think the fix is non-trivial.

image

guacamole/src/main/frontend/src/app/import/controllers/importConnectionsController.js
eugen-keeper: The issue was in the double slash after root. I added some normalization to fix  the issue. Also I improved the error watch a little to catch such cases in the future.

bbennett-ks (approved):

bbennett-ks:
> I noticed 1 other thing (unrelated to this PR): the CONNECTION IMPORT screen: if no file is selected, the Cancel button is grayed out & there's no way to exit the screen. Besides selecting another menu option.
> 
> I'll file s separate issue for this as I think the fix is non-trivial.
> 
> image

https://issues.apache.org/jira/browse/GUACAMOLE-2286

guacamole/src/main/frontend/src/app/import/directives/connectionImportErrors.js
```diff
@@ -254,62 +304,47 @@ angular.module('import').directive('connectionImportErrors', [
...
 
                     // If the error is not a known translatable type, add the
                     // message directly to the error array
-                    else
-                         connectionError.errors.getArray()[errorIndex] = (
-                             error.message ? error.message : error);
+                    else {
+                        connectionError.errors.getArray()[errorIndex] = (
+                                error.message ? error.message : error);
+                        }
```
necouchman: Some reason for the extra indention, here? The braces...well...I can go either way on those...

guacamole/src/main/frontend/src/app/import/directives/connectionImportErrors.js
```diff
@@ -254,62 +304,47 @@ angular.module('import').directive('connectionImportErrors', [
...
                         // Fetch the translation and update it when it's ready
                         translationPromises.push($translate(
-                            error.key, error.variables)
-                            .then(translatedError => {
-                                connectionError.errors.getArray()[errorIndex] = translatedError;
-                            }));
+                                error.key, error.variables).then(translatedError => {
+                            connectionError.errors.getArray()[errorIndex] = translatedError;
+                        }));
+                    }
```
necouchman: The updated indentation, here, seems...strange.

guacamole/src/main/frontend/src/app/import/directives/connectionImportErrors.js
```diff
@@ -131,37 +138,36 @@ angular.module('import').directive('connectionImportErrors', [
...
-                name: connection.name,
+                name: connectionObject.name,
                 group: parseResult.groupPaths[index],
-                protocol: connection.protocol,
+                protocol: connectionObject.protocol,
 
                 // The human-readable error messages
-                errors: new DisplayErrorList(
-                        [ ...(parseResult.errors[index] || []) ])
+                errors: new DisplayErrorList([])
```
necouchman, line 160: Why are not passing through the errors anymore?

guacamole/src/main/frontend/src/app/import/directives/connectionImportErrors.js
```diff
@@ -114,7 +121,7 @@ angular.module('import').directive('connectionImportErrors', [
 
         /**
          * Generate a ImportConnectionError representing any errors associated
-         * with the row at the given index within the given parse result.
+         * with the row at the given index within the given patch result.
```
necouchman, line 124: Well, we're still passing in the `parseResult`, along with the `patches`, so is this correct?

necouchman (changes requested):
Just a few curiosities, should be pretty much ready to go.

guacamole/src/main/frontend/src/app/import/directives/connectionImportErrors.js
eugen-keeper: Good catch. I’ve reverted this block to the original style.

guacamole/src/main/frontend/src/app/import/directives/connectionImportErrors.js
eugen-keeper: Thanks. Reverted,

guacamole/src/main/frontend/src/app/import/directives/connectionImportErrors.js
eugen-keeper: this was intentional as part of splitting the old shared `generateConnectionError` into separate parse vs. patch helpers.

`generatePatchConnectionError` is only used when `patchFailure` is set, which can only happen after a successful parse and group resolution (`handleParseSuccess` returns early if `parseResult.hasErrors`). In that scenario, parse-time errors are always empty, so passing `parseResult.errors[index]` through would never add anything.

Parse errors are now handled exclusively by `generateParseConnectionError` and the `parseResult.hasErrors` watcher. API errors for the patch path are added separately via getAPIError() / removeError in the reduce loop below.

guacamole/src/main/frontend/src/app/import/directives/connectionImportErrors.js
eugen-keeper: Yes - `parseResult` is still needed here for `groupPaths[index]`. We only stopped passing through `parseResult.errors[index]` because this path runs when `hasErrors` is `false`.

bbennett-ks (approved):
LGTM!

necouchman (approved):

PR #1198, opened 2026-04-07
GUACAMOLE-2258 : OpenID code flow, client secret and PKCE

adb014:
The existing OpenID extension only supports OpenID implicit flow. Knowing that this method is deprecated in OAuth2 due to its poor security, Guacamole really needs to add code flow authorization to its OpenID extension.

This pull request add code flow, but also allows for pkce challenges to be passed to the authorization endpoint and pkce verifiers and client secrets to be passed to the token endpoint. This options for this are 

```
openid-flow-type: implicit | code
openid-authorization-endpoint: https://identity-provider/auth
openid-token-endpoint:  https://identity-provider/token
openid-jwks-endpoint: https://identity-provider/jwks
openid-issuer: identity-provider
openid-client-id: my-client-id
openid-client-secret: my-client-secret | null
openid-pkce-required: true | false
openid-redirect-uri: https://example.net/guacamole
```

Allowing the existing implicit flow code to be used by default without changes to existing configurations. 

One problem I had is that the getLoginURI method didn't seem to allow me to save the PKCE verifier and in any case I didn't want the code returned by the identity provider to appear in the browser navigation bar. So I created a local REST redirect api to talk to the identity provider and receive the response. If you know how to use the AuthenticationSessionManager within getLoginURI and  hide the returned code in the navigation bar this might be done differently. 

There is a draft JAR built against guacamole 1.6.0 on my fork of guacamole-client


adb014:
I've figured out how to use a SessionManager directly in getLoginURI to store the PKCE verifier. The local REST redirect and callback for the session are no longer needed and have be removed. Making the implementation cleaner

adb014:
Looking at the discussion in #517, I finally decided to add the well-known endpoint and an enum for the response_type to limit the values allowed. Given this it was also easy to add the use of "token" response_types as discussed in #517 for AWS Cognito. 

A minimal guacamole.properties can then be

```
openid-response-type: id_token | token | code
openid-well-known-endpoint: https://identity-provider/well-known
openid-client-id: my-client-id
openid-client-secret: my-client-secret | null
openid-pkce-required: true | false
openid-redirect-uri: https://example.net/guacamole
```

The issuer, authorization, token and jwks endpoints returned by the well-known endpoint can be overridden with the previous endpoints in which case it’s possible to not have a well-known endpoint. 

Can't see anything else to add to this patch. Can someone review it ?

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/ConfigurationService.java
```diff
@@ -20,12 +20,17 @@
 package org.apache.guacamole.auth.openid.conf;
 
 import com.google.inject.Inject;
+
```
necouchman: We generally don't break between `import` blocks, you can remove this line.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/ConfigurationService.java
```diff
@@ -110,6 +120,66 @@ public class ConfigurationService {
...
+     * try again.
+     */
+    private static final IntegerGuacamoleProperty OPENID_AUTH_TIMEOUT =
+            new IntegerGuacamoleProperty() {
+        @Override
+        public String getName() { return "openid-auth-timeout"; }
+
+    };
+
+
```
necouchman, line 185: This block of code needs to be consistent in its style, both internal (within this block) and also with established style throughout the code:
* There are varying occurrences of blank lines before the `@Override` and after the functions.
* Sometimes you put the braces and `return` statement on a single line, sometimes you used multiple lines.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/OpenIDWellKnown.java
```diff
@@ -0,0 +1,214 @@
...
+/**
+ * Class for retrieving well-known endpoint data.
+ */
+@Singleton
+public class OpenIDWellKnown {
+
+    /**
+     * Logger for this class.
+     */
+    private final Logger logger = LoggerFactory.getLogger(OpenIDWellKnown.class);
```
necouchman: This should be static.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/OpenIDAuthenticationSession.java
```diff
@@ -0,0 +1,56 @@
...
+
+package org.apache.guacamole.auth.openid;
+
+import org.apache.guacamole.net.auth.AuthenticationSession;
+
+/**
+ * Representation of an in-progress OpenID authentication attempt.
+ */
+public class OpenIDAuthenticationSession extends AuthenticationSession {
+    /**
```
necouchman, line 29: Should be a blank line above this one.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/OpenIDAuthenticationProviderModule.java
```diff
@@ -42,7 +45,7 @@ protected void configure() {
         bind(ConfigurationService.class);
         bind(NonceService.class).in(Scopes.SINGLETON);
         bind(TokenValidationService.class);
-        
+        bind(OpenIDAuthenticationSessionManager.class);
```
necouchman, line 48: The blank line can stay, and this new line should be above it - it separates the OpenID-specific binds from the generic one.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/PKCEUtil.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Utility class for generating PKCE parameters.
+ *
+ * Supports:
+ *   - code_verifier (random Base64URL)
+ *   - code_challenge (S256)
+ */
+public final class PKCEUtil {
+
+    private static final SecureRandom RANDOM = new SecureRandom();
```
necouchman, line 36: Please document this.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/PKCEUtil.java
```diff
@@ -0,0 +1,64 @@
...
+
+    private PKCEUtil() {}
```
necouchman, line 41: Please document this constructor.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/PKCEUtil.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Generates a high-entropy PKCE code_verifier.
+     */
```
necouchman, line 48: Please complete the documentation for this method (at the very least needs an `@return` block).

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/PKCEUtil.java
```diff
@@ -0,0 +1,64 @@
...
+    public static String generateCodeVerifier() {
+        byte[] bytes = new byte[64];
+        RANDOM.nextBytes(bytes);
+        return base64Url(bytes);
+    }
+
+    /**
+     * Computes the PKCE code_challenge = BASE64URL(SHA256(code_verifier)).
+     */
```
necouchman, line 64: Please complete documentation for this method.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/PKCEUtil.java
```diff
@@ -0,0 +1,64 @@
...
+    public static String generateCodeChallenge(String verifier) throws Exception {
+        MessageDigest sha = MessageDigest.getInstance("SHA-256");
+        byte[] hash = sha.digest(verifier.getBytes("US-ASCII"));
+        return base64Url(hash);
+    }
+
+    /**
+     * Base64URL encoding without padding.
+     */
```
necouchman, line 79: Please complete documentation for this method.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/JsonUrlReader.java
```diff
@@ -0,0 +1,98 @@
...
+import org.jose4j.json.JsonUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/*
+ * Utility class to open a http connection to a URL, send a body
+ * and receive a response in the form of a parsed JSON
+ */
+public final class JsonUrlReader {
+    /**
```
necouchman, line 41: Please add a blank line above this line.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/JsonUrlReader.java
```diff
@@ -0,0 +1,98 @@
...
+     * Logger for this class.
+     */
+    private static final Logger logger = LoggerFactory.getLogger(JsonUrlReader.class);
+
+    private JsonUrlReader() {}
```
necouchman, line 50: Please document this constructor.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/JsonUrlReader.java
```diff
@@ -0,0 +1,98 @@
...
+
+    public static Map<String,Object> fetch(String method, URL url, String body) throws GuacamoleException {
```
necouchman: Please document this method.

Also, if this is intended to be a more generic utility class, then maybe it should throw something other than a `GuacamoleException`?

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/JsonUrlReader.java
```diff
@@ -0,0 +1,98 @@
...
+        if (url == null) {
```
necouchman: What if the URL is an empty string?

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/JsonUrlReader.java
```diff
@@ -0,0 +1,98 @@
...
+            throw new GuacamoleException("JsonUrlReader : Missing URL");
```
necouchman: If you are going to throw `GuacamoleException` types, it should probably be something more specific than just the generic `GuacamoleException`.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/JsonUrlReader.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+        
+        try {
+            // Open connection, using HttpURLConnection
+            HttpURLConnection conn = (HttpURLConnection) url.openConnection();
+
+            conn.setRequestMethod(method);
+            conn.setRequestProperty("Content-Type", "application/x-www-form-urlencoded; charset=UTF-8");
+
+            if (! method.equalsIgnoreCase("GET")) {
```
necouchman: Do we really expect the case of the HTTP method to vary? I see it fairly consistently as "GET" (in capital letters) - are there situations where we think this won't be the case?

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/JsonUrlReader.java
```diff
@@ -0,0 +1,98 @@
...
+import org.apache.guacamole.GuacamoleException;
+import org.jose4j.json.JsonUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/*
+ * Utility class to open a http connection to a URL, send a body
+ * and receive a response in the form of a parsed JSON
+ */
+public final class JsonUrlReader {
```
necouchman, line 40: The Apache Commons IO project has some classes that might already do this. Maybe it would be better to use one of those than to re-invent it?

https://www.baeldung.com/java-read-json-from-url

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/token/TokenValidationService.java
```diff
@@ -131,6 +138,119 @@ public JwtClaims validateToken(String token) throws GuacamoleException {
...
+                       .append(URLEncoder.encode(value, StandardCharsets.UTF_8))
+                       .toString();
+     }
+
+    /**
+     * Exchanges the authorization code for tokens.
+     *
+     * @param code
+     *     The authorization code received from the IdP.
+     * @param codeVerifier
```
necouchman, line 179: Please add a blank line in between these `@param` blocks.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/token/TokenValidationService.java
```diff
@@ -131,6 +138,119 @@ public JwtClaims validateToken(String token) throws GuacamoleException {
...
+                bodyBuilder.append("&").append(urlencode("code_verifier", verifier));
+            }
+
+            Map<String,Object> json = 
+                    JsonUrlReader.fetch("POST", confService.getTokenEndpoint().toURL(),
+                                        bodyBuilder.toString());
+
+            return (String) json.get("id_token");
+
+        } catch (Exception e) {
```
necouchman: Please don't cuddle the braces.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/token/TokenValidationService.java
```diff
@@ -131,6 +138,119 @@ public JwtClaims validateToken(String token) throws GuacamoleException {
...
+            // Validate JWT
+            return jwtConsumer.processToClaims(token);
+        }
+        // Log any failures to validate/parse the JWT
+        catch (InvalidJwtException e) {
+            logger.info("Rejected invalid OpenID token: {}", e.getMessage(), e);
+        }
+
+        return null;
+    }
```
necouchman: It seems like this new method does almost exactly the same thing as the existing `validateToken()` method, save the call to the `exchangeCode()` method. Surely we can avoid the duplication, here.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/token/TokenValidationService.java
```diff
@@ -20,17 +20,24 @@
...
+import java.net.URI;
+import java.nio.charset.StandardCharsets;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+
```
necouchman: Please remove this additional blank line.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/OpenIDWellKnown.java
```diff
@@ -0,0 +1,214 @@
...
+    @Inject
+    private Environment environment;
+
+    /*
+     * Creates an OpenIDWellKnown class that reads the json from an OIDC
+     * well-known endpoint and saves these values for later use. Use Guice
+     * to ensure environment exists before initializing.
+     */
+    public OpenIDWellKnown() {
+    }
```
necouchman: Having a variable and a constructor this far down, while not explicitly violating any style guideline, is a bit unconventional and not how we tend to do things in the existing code base. The order tends to be (more or less): variables, constructors, and then methods.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/OpenIDWellKnown.java
```diff
@@ -0,0 +1,214 @@
...
+     *     The well-known endpoint of the OIDC service, as configured with
+     *     guacamole.properties.
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed, or if the authorization
+     *     endpoint property is missing.
+     */
+    public URI getWellKnownEndpoint() throws GuacamoleException {
+        return environment.getProperty(OPENID_WELL_KNOWN_ENDPOINT);
+    }
```
necouchman: Aside from being called inside this class in the `init()` method, the only other call that I see is in the `ConfigurationService`, which, as I mention in the comment above, may not be necessary. I think that call can be removed and this can be changed to `private`.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/OpenIDWellKnown.java
```diff
@@ -0,0 +1,214 @@
...
+        
+        Runnable task = new Runnable() {
+            int attempts = 0;
+
+            @Override
+            public void run() {
+                attempts++;
+
+                try {
+                    Map<String,Object> json = JsonUrlReader.fetch("GET", getWellKnownEndpoint().toURL(), "");
```
necouchman: What happens, here, if the administrator has not provided a well-known endpoint? The call, above, to `getProperty()` in the `getWellKnownEndpoint()` implementation, will return `null` if the property has not been provided. My guess is that this should be checked before kicking off this scheduler? Am I missing where that check happens?

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/ConfigurationService.java
```diff
@@ -334,7 +419,130 @@ public String getIssuer() throws GuacamoleException {
...
+     *     The well-known endpoint of the OIDC service, as configured with
+     *     guacamole.properties.
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed, or if the well-known
+     *     endpoint property is missing.
+     */
+    public URI getWellKnownEndpoint() throws GuacamoleException {
+        return confWellKnown.getWellKnownEndpoint();
+    }
```
necouchman: Is this method really necessary? I don't see this being actually called anywhere in the code? It seems like the well-known endpoint is only ever used inside the `OpenIDWellKnown` class.

necouchman (changes requested):
@adb014 I've done an initial review, here, and, generally speaking, I believe what you've got is pretty sound. Most of the comments I have made have to do with style - being consistent about the style you've implemented, and matching that to the existing code style. If you take a look at those comments I think you'll get the gist of what to pay attention to.

From a functional perspective, my main concern is that I believe you've written/tested this assuming that the OpenID well-known endpoint will be provided, and there are some places where it seems like the code, as implemented, would error out or at least behave erratically if the well-known endpoint happened to be left out of the configuration. So, a little work to make sure that the changes you're making don't break existing configurations would be great.

Thanks for picking this up and running with it - it seems like there's a growing community of folks who will be very happy to have these additional OpenID functions available.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/token/TokenValidationService.java
adb014: There is a nonce in the existing method in the call to jwtConsumer. Treating that with an if/then in the jwtConsumerBuilder seemed that I end up making the code sufficiently different that a new validateToken method was better. Pretty much a matter of choice. Ok I,ll dedup it.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/JsonUrlReader.java
adb014: If I only wanted the JsonUrlReader class for the well-known endpoint, you’d be right. I also use it in the exchange of a code for a token, which must use a POST with the parameters in the body. With a quick look around not finding something appropriate I implemented it myself. 

If you have an idea that can read json from a url with both POST and GET, I’ll get rid of this function.

adb014:
I’m on my phone at the moment, so will look at this later. Though I tested with 7 scenarios 

- id_token + well-known 
- id_token + manual endpoint and issuer 
- token + well-known 
- token + manual endpoint and issuer 
- code + well-known 
- code + manual endpoint and issuer 
- code + manual endpoint and issuer + pkce

The PKCE path seemed sufficiently independent that it only needed testing once. All of these were tested against a local keycloak only. 

If the well-known endpoint is not given then the calls like confService.getIssuer() will see the manual value in priority and return it. If only the well-known endpoint is set, then the confWellKnown.getIssuer() method is called from within confService.gerIssuer(). This call might fail initially if the well-known data is not available yet. I tested in docker and the guacamole container starts faster than keycloak 😆. This seemed like a likely scenario so I protected the detection of the well-known data in a runnable called till valid data is returned or it times out. This does mean that the identity provider must be available at least once in the first 2 minutes of starting guacamole, and the well-known data won’t be looked for later. 

Even if the call to fails initially on start up following calls with later connections will work once the well-known data is found. 

Should I allow the init function of confWellKnown to run later during calls to the getters ? Seemed that the failure scenario was sufficiently rare that it’s not needed

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/ConfigurationService.java
adb014: It’s not called anywhere. It’s implemented for completeness of the confService class, so all configuration variables have a getter. It might be used later. 

It can be removed if you prefer

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/OpenIDWellKnown.java
adb014: At the moment JsonUrlReader will be called 24 times failing each time and eventually stop. Yes I could have a fast return from init() if the well-known endpoint is missing

The missing well-known endpoint just means the getters of confWellKnown will always return null, which is what we want in confService

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/OpenIDWellKnown.java
adb014: See comment for the getter in confService. It’s as you prefer.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/JsonUrlReader.java
adb014: I thought my JsonUrlReader might end up used elsewhere and the case insensitive compare was just a bit of future idiot proofing. As you prefer

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/JsonUrlReader.java
adb014: You’re of cause right. It should probably return an IOException

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/JsonUrlReader.java
adb014: Good point. I’ll fix it when I’m not on my phone

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/OpenIDAuthenticationProviderModule.java
adb014: Ok done

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/PKCEUtil.java
adb014: Ok. done

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/PKCEUtil.java
adb014: Done

adb014:
Ok, I've treated all of your comments. Thinking about it, I still have one issue. Guacamole displays the data returned by keycloak in the navigation bar, including the "code" field for code flow... If the user refreshs the page, as they often do if there is a problem, authenticateUser will be called again with the stale code and this will be posted to the identity provider. The identity provider detects this as a code "reuse" and probably an attack. This is bad

An idea would be in authenticateUser we do a redirect to the guacamole root instead of returning authenticateUser. Though this currently breaks things. The only other choice I see is the use of code like

```javascript
(function guacOpenIDTransformToken() {
    if (window.history && history.replaceState)
        history.replaceState(null, "", window.location.pathname + window.location.hash);
    if (/^#(?![?\/])(.*&)?id_token=/.test(location.hash))
        location.hash = '/?' + location.hash.substring(1);
})();
```

client side in transformToken.js to hide the non URL fragment query parameters returned by the identity provider from the user.. Still needs testing, so I'll submit another patch soon

adb014:
> Thinking about it, I still have one issue. Guacamole displays the data returned by keycloak in the navigation bar, including the "code" field for code flow... If the user refreshs the page, as they often do if there is a problem, authenticateUser will be called again with the stale code and this will be posted to the identity provider. The identity provider detects this as a code "reuse" and probably an attack. This is bad

Forget it. I'm an idiot.. The existing js code converts the query parameters into a URL fragment so they won't be posted to the server on a refresh.. The code as it stands is ok.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/JsonUrlReader.java
necouchman: It was not intended to be a rhetorical question/remark - I was legitimately asking if there are situations where those requests come back in varying cases. Just because I've never seen it that way doesn't mean that it doesn't happen :-).

adb014 (commented):
My comments were in a self review for some reason !

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/JsonUrlReader.java
adb014: I had commented on this Friday, but it seems I did it in a self-review that I deleted. So here is my comment again

Going to the[ RFC 7231 for HTTP/1.1  ](https://datatracker.ietf.org/doc/html/rfc7231#section-4.1) we see the text

```
  The method token is case-sensitive because it might be used as a gateway to object-based systems with case-sensitive method names.
```

so any use of non uppercase method names is against the standard and should be ignored

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/JsonUrlReader.java
adb014: I'd written a huge amount of comments in my deleted self review for this point. Basically thinking out loud. So here you'll get the condensed version.

Basically doing something like

```
$ find . -name "*.java" -exec grep -H "httpClient" {} \;
```

and all other possible httpclients I could think of, I found no example of an outbound HTTP connexion from guacamole. So as far as I can see we are free to implement this how we want. However it needs to

- don’t introduce any new large dependencies to guacamole
- treat both POST and GET requests
- Be relatively clean

java.net.httpUrlConnection is a bit old and ugly, but I couldn't find any apache commons I/O function that treated both POST and GET. Using Jackson to deserialize to an Object seemed also to be overkill as I only needed a couple of values. 

The slightly more modern  java.net.http.httpClient will make the code cleaner. Don't see the point of using org.apache.http.impl.client.CloseableHttpClient as the http post/get operations are one off each time. Though using org.apache.http.impl.client.HttpClient is an option that might be attractive as it supplies functions to more cleanly treat the posted form data. 

I've updated the code using java.net.http.httpClient and included asynchronous operation to release control back to the Tomcat Servlet so as not to impact the rest of guacamole if the operation takes some time.

If you prefer a different http client, there is a plethora to choose from, then sure I'll change it. Now is the time to complain as this is the first outbound connection in guacamole, so getting it right before introducing code we might regret is the right thing to do. :-)

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/JsonUrlReader.java
```diff
@@ -0,0 +1,101 @@
...
+        if (uri == null || uri.toString().isEmpty()) {
+            throw new IOException("JsonUrlReader: Missing URL");
+        }
+
+        try {
+            HttpClient client = HttpClient.newHttpClient();
+            HttpRequest request = HttpRequest.newBuilder()
+                    .uri(uri)
+                    .header("Content-Type", "application/x-www-form-urlencoded; charset=UTF-8")
+                    .method(method,
```
adb014: On my machine using "GET" here doesn't crash here with an IllegalArgumentException though the documentation makes me think it should. Should I protect this in a if/else block ? Or is 

```
.method(method, (method == "GET" ? HttpRequest.BodyPublishers.NoBody() :  HttpRequest.BodyPublishers.ofString(body == null ? "" : body, StandardCharsets.UTF_8)))
```

just too ugly ?

adb014:
@necouchman anticipating that you’ll want a patch to the manual, I started think what is needed but ran into a question. At the moment the configuration is split into required and optional configuration options. However, with the proposed changes only openid-client-id and openid-redirect-uri remain required. The other existing « required » options can either remain required or be replaced with openid-well-known-endpoint. 

To me the split required/optional make less sense now. What I propose is a discussion implicit flow versus code flow and another section on the well-known endpoint and the fact it can replace up to 4 (3 with implicit flow) other required configuration options. I then propose to combine the existing required and optional sections into a single section. 

How do you want this addressed ?

adb014:
> @necouchman anticipating that you’ll want a patch to the manual, I started think what is needed but ran into a question. At the moment the configuration is split into required and optional configuration options. However, with the proposed changes only openid-client-id and openid-redirect-uri remain required. The other existing « required » options can either remain required or be replaced with openid-well-known-endpoint.
> 
> To me the split required/optional make less sense now. What I propose is a discussion implicit flow versus code flow and another section on the well-known endpoint and the fact it can replace up to 4 (3 with implicit flow) other required configuration options. I then propose to combine the existing required and optional sections into a single section.
> 
> How do you want this addressed ?

I went ahead and implemented a possible change to the manual in a seperate pull request

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/JsonUrlReader.java
adb014: Hi @necouchman, thinking again about what you said above I might have misunderstood your ultimate goal. Do you mean you want to "outsource" potentially risky functions to a known and trusted source ? 

If that is the case rather than do this at the Url2Json level, we could do it directly wiith the Spring IDC/OAuth2 code "org.springframework.security.oauth2". This would be a relatively easy change and comes with the advantage of the security risks in the OIDC code are delegated to Spring ;-) ... Though if we do that for the code flow authorization in OIDC, then the existing Jwt claims verification code should be replaced with the equivalent code in Spring.

Frankly I have no hangups about the desired implementation , and I'll reimplement it however you want

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/ConfigurationService.java
```diff
@@ -260,11 +339,16 @@ public class ConfigurationService {
      *     guacamole.properties.
      *
      * @throws GuacamoleException
-     *     If guacamole.properties cannot be parsed, or if the authorization
-     *     endpoint property is missing.
+     *     If guacamole.properties cannot be parsed, or it authorization and
+     *     well-known endpoints are property is missing.
```
necouchman: There are a couple of typos, here - I think the original `or if the` language was correct, and the change should just be to change the second line to  `and well-known endpoint properties are missing.`

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/ConfigurationService.java
necouchman: There are a few more minor tweaks that should be made, here - adding some newlines after the `getName()` overrides in a couple of places.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/ConfigurationService.java
```diff
@@ -260,11 +339,16 @@ public class ConfigurationService {
...
      */
     public URI getAuthorizationEndpoint() throws GuacamoleException {
-        return environment.getRequiredProperty(OPENID_AUTHORIZATION_ENDPOINT);
+        URI authorization_endpoint = environment.getProperty(OPENID_AUTHORIZATION_ENDPOINT);
+        authorization_endpoint = authorization_endpoint == null ? confWellKnown.getAuthorizationEndpoint() : authorization_endpoint;
```
necouchman: This is a nice succinct line of code, but I don't know that it makes for terribly readable code. Also, there is a form of the `environment.getProperty()` call that lets you specify a default value, so you could really just combine this line with the above:
```
URI authorization_endpoint = environment.getProperty(OPENID_AUTHORIZATION_ENDPOINT, confWellKnown.getAuthorizationEndpoint());
```

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/ConfigurationService.java
```diff
@@ -313,10 +397,15 @@ public URI getRedirectURI() throws GuacamoleException {
      *
      * @throws GuacamoleException
      *     If guacamole.properties cannot be parsed, or if the issuer property
-     *     is missing.
+     *     and well-known endpoints are property is missing.
      */
     public String getIssuer() throws GuacamoleException {
-        return environment.getRequiredProperty(OPENID_ISSUER);
+        String issuer = environment.getProperty(OPENID_ISSUER);
+        issuer = issuer == null ? confWellKnown.getIssuer() : issuer;
```
necouchman: Similar to the above in `getAuthorizationEndpoint()`, I'd suggest this line use the `getProperty()` call that uses `confWellKnown.getIssuer()` as the default value.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/ConfigurationService.java
```diff
@@ -313,10 +397,15 @@ public URI getRedirectURI() throws GuacamoleException {
      *
      * @throws GuacamoleException
      *     If guacamole.properties cannot be parsed, or if the issuer property
-     *     is missing.
+     *     and well-known endpoints are property is missing.
```
necouchman: -> `and well-known endpoint properties are missing.`

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/ConfigurationService.java
```diff
@@ -330,11 +419,117 @@ public String getIssuer() throws GuacamoleException {
      *     guacamole.properties.
      *
      * @throws GuacamoleException
-     *     If guacamole.properties cannot be parsed, or if the JWKS endpoint
-     *     property is missing.
+     *     If guacamole.properties cannot be parsed, or if the JWKS and
+     *     well-known endpoints are property is missing.
```
necouchman: -> `and well-known endpoint properties are missing.`

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/ConfigurationService.java
```diff
@@ -330,11 +419,117 @@ public String getIssuer() throws GuacamoleException {
...
      */
     public URI getJWKSEndpoint() throws GuacamoleException {
-        return environment.getRequiredProperty(OPENID_JWKS_ENDPOINT);
+        URI jwks_uri =  environment.getProperty(OPENID_JWKS_ENDPOINT);
+        jwks_uri = jwks_uri == null ? confWellKnown.getJWKSEndpoint() : jwks_uri;
```
necouchman: This can also be changed to make the well-known endpoint call the default value.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/ConfigurationService.java
```diff
@@ -330,11 +419,117 @@ public String getIssuer() throws GuacamoleException {
...
+     * Returns the token endpoint (URI) of the OIDC service as
+     * configured with guacamole.properties.
+     *
+     * @return
+     *     The token endpoint of the OIDC service, as configured with
+     *     guacamole.properties.
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed, or if the token and
+     *     well-known endpoints are property is missing.
```
necouchman: -> `well-known endpoint properties are missing.`

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/ConfigurationService.java
```diff
@@ -330,11 +419,117 @@ public String getIssuer() throws GuacamoleException {
...
+     */
+    public URI getTokenEndpoint() throws GuacamoleException {
+        URI token_endpoint = environment.getProperty(OPENID_TOKEN_ENDPOINT);
+        token_endpoint = token_endpoint == null ? confWellKnown.getTokenEndpoint() : token_endpoint;
```
necouchman: Also can use default value of `getProperty()`, here.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/ConfigurationService.java
```diff
@@ -330,11 +419,117 @@ public String getIssuer() throws GuacamoleException {
...
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed.
+     */
+    public OpenIDResponseType getResponseType() throws GuacamoleException {
+        return environment.getProperty(OPENID_RESPONSE_TYPE, DEFAULT_RESPONSE_TYPE);
+    }
+
+    /**
+     * Returns true if the response type defines an implict flow
```
necouchman: ...otherwise false, as configured in guacamole.properties.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/ConfigurationService.java
```diff
@@ -330,11 +419,117 @@ public String getIssuer() throws GuacamoleException {
...
+     *
+     * @return
+     *     The whether implicit flow is used or not, as configured with guacamole.properties.
```
necouchman: This should be something more like:
```
true if implicit flow should be used, otherwise false.
```

And the mention of guacamole.properties can be in the main body of the comment above.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/ConfigurationService.java
```diff
@@ -330,11 +419,117 @@ public String getIssuer() throws GuacamoleException {
...
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed.
+     */
+    public boolean isImplicitFlow() throws GuacamoleException {
+        OpenIDResponseType response_type = environment.getProperty(OPENID_RESPONSE_TYPE, DEFAULT_RESPONSE_TYPE);
+        return response_type != OpenIDResponseType.CODE;
+    }
+
+    /**
+     * Returns the OIDC client secret used for toen validation, as configured
```
necouchman: `toen` -> `token`

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/ConfigurationService.java
```diff
@@ -330,11 +419,117 @@ public String getIssuer() throws GuacamoleException {
...
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed.
+     */
+    public String getClientSecret() throws GuacamoleException {
+       return environment.getProperty(OPENID_CLIENT_SECRET);
+    }
+
+    /**
+     * Returns  a boolean value of whether "Proof Key for Code Exchange (PKCE)"
```
necouchman: There's an extra space between `Returns` and `a`, here.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/ConfigurationService.java
```diff
@@ -330,11 +419,117 @@ public String getIssuer() throws GuacamoleException {
...
+     * will be used, as configured with guacamole.properties. The choice of
+     * whether to use PKCE is up to you, but the OIDC service must support
+     * it. By default will be false.
+     *
+     * @return
+     *     The whether to use PKCE or not, as configured with guacamole.properties.
```
necouchman: Similar to above, this should be something more like:
```
true if PKCE should be used, otherwise false.
```

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/ConfigurationService.java
```diff
@@ -330,11 +419,117 @@ public String getIssuer() throws GuacamoleException {
...
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed.
+     */
+    public boolean isPKCERequired() throws GuacamoleException {
+        Boolean enabled = environment.getProperty(OPENID_PKCE_REQUIRED);
+        return enabled != null && enabled;
```
necouchman: Two things:
* I don't think there's any reason to use a `Boolean` here when you're just returning a `boolean`.
* Again, make use of the default value option - if false is the default, this can be simplified to:
```
return environment.getProperty(OPENID_PKCE_REQUIRED, false);
```

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/OpenIDResponseType.java
```diff
@@ -0,0 +1,67 @@
...
+     /**
+      * Response type "code" used for code flow authentication
+      */
+     @PropertyValue("code")
+     CODE("code");
+     
+     /*
+      * The string value of the response type used
+      */
+     public final String STRING_VALUE;
```
necouchman: The comment blocks in this section each use a slightly different style. The style used elsewhere in the code is as follows:
```
/**
 * Documentation of the enum type, variable, etc.
 */
```

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/OpenIDWellKnown.java
```diff
@@ -0,0 +1,227 @@
...
+     * @return
+     *     The well-known endpoint of the OIDC service, as configured with
+     *     guacamole.properties.
+     *
+     * @throws GuacamoleException
+     *     If guacamole.properties cannot be parsed, or if the well-known
+     *     endpoint property is missing.
+     */
+    private URI getWellKnownEndpoint() throws GuacamoleException {
+        return environment.getProperty(OPENID_WELL_KNOWN_ENDPOINT);
```
necouchman: Note that this call to `getProperty()` will *not* throw a `GuacamoleException` if the `OPENID_WELL_KNOWN_ENDPOINT` property is missing. It will return null. If you want the behavior you've described in the `@throws` block of the method documentation, you'll need to use `getRequiredProperty()`, instead.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/OpenIDWellKnown.java
```diff
@@ -0,0 +1,227 @@
...
+
+    /**
+     * The detected token endpoint
+     */
+    private static URI token_endpoint = null;
+
+    /**
+     * The detected jwks_uri
+     */
+    private static URI jwks_uri = null;
```
necouchman: Two issues, here:
* I'm not sure these should be `static`? I know this is a `Singleton` class, but you're still expecting that these are private and would be accessed via the get methods defined below, so I think they should not be static?
* If they're static, the generally-accepted style is `UPPER_CASE_WITH_UNDERSCORES`. If they're not static, the generally-accepted style used elsewhere in the code is `camelCaseWithoutUnderscores`.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/OpenIDWellKnown.java
```diff
@@ -0,0 +1,227 @@
...
+        return environment.getProperty(OPENID_WELL_KNOWN_ENDPOINT);
+    }
+
+    /**
+     * Returns the issuer to expect for all received ID tokens, as configured
+     * from the well_known endpoint.
+     *
+     * @return
+     *     The issuer to expect for all received ID tokens, as returned by the
+     *     well-known endpoint.
```
necouchman: These comments should be consistent about using `well-known` vs. `well_known`. I think `well-known` is the way to go.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/OpenIDWellKnown.java
```diff
@@ -0,0 +1,227 @@
...
+     */
+    public URI getJWKSEndpoint() {
+        return jwks_uri;
+    }
+
+    /**
+     * The Guacamole server environment.
+     */
+    @Inject
+    private Environment environment;
```
necouchman: This injected variable should be at the top with the other class variables, before any methods are defined.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/OpenIDWellKnown.java
```diff
@@ -0,0 +1,227 @@
...
+
+    /*
```
necouchman: This is missing the extra `*` that defines this as a JSDoc-style comment block.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/token/TokenValidationService.java
```diff
@@ -70,20 +73,26 @@ public class TokenValidationService {
...
+     *     an id_token and then validate.
+     *
+     * @param verifier
+     *     A PKCE verifier or null if not used. Only used with code flow
      *
      * @return
-     *     The JWT claims contained within the given ID token if it passes tests,
-     *     or null if the token is not valid.
+     *     The JWT claims contained within the id_token if it passes tests,
+     *     or null if the id_token is not valid.
```
necouchman: Would it make more sense, here to throw an exception instead of returning `null` if the id_token is invalid? Presumably an invalid id_token constitutes an authentication failure, which should then be passed back to the user, perhaps using `GuacamoleInvalidCredentialsException`? Or is there some reason within the code flow that `null` is preferred for this?

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/token/TokenValidationService.java
```diff
@@ -98,26 +107,36 @@ public JwtClaims validateToken(String token) throws GuacamoleException {
                 .setExpectedAudience(confService.getClientID())
                 .setVerificationKeyResolver(resolver)
                 .build();
-
+                
+        /* Exchange code → token */
+        if (! confService.isImplicitFlow()) {
```
necouchman: You can remove the space between the `!` and `confService`.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/token/TokenValidationService.java
```diff
@@ -131,6 +150,73 @@ public JwtClaims validateToken(String token) throws GuacamoleException {
...
+            Map<String,Object> json = 
+                    JsonUrlReader.fetch("POST", confService.getTokenEndpoint(),
+                                        bodyBuilder.toString());
+
+            return (String) json.get("id_token");
+
+        }
+        catch (Exception e) {
+            logger.info("Rejected invalid OpenID code exchange: {}", e.getMessage(), e);
+        }
```
necouchman: Catching `Exception` here is probably not a good idea:
* This method `throws GuacamoleException`, but if you catch the overall `Exception` class, here, that will never actually happen, and this method will just return null every time. I suspect we actually want the `GuacamoleException` exceptions to be passed on to the callers.
* By catching `Exception` this makes the very generic logger message here both misleading and unhelpful. The OpenID code exchange isn't necessarily rejected, here, as it states - it could be that a property wasn't properly configured or there was an issue retrieving it. You are passing through the underlying exception and message, which is good, but the initial log message doesn't really help.
* Also, `logger.info()` is probably not the correct level, here - this constitutes an error, not an informational message, so it should be `logger.error()`.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/JsonUrlReader.java
necouchman: I think there are multiple things to think about:
* Risk is certainly one of them, though it's a bit of a trade. You either write your own functions and risk that you've done something bad and it will have to be corrected, or you rely on someone else. There's risk either way.
* Maintainability is more what I had in mind - if someone else already has a class and set of functions for reading JSON from a URL, and they're actively maintaining it, it allows the Guacamole project to use that and not have to maintain a set of classes for reading JSON URLs, which is not the goal of this project. It's why we aren't writing our own JDBC drivers, SAML client, OpenID client, LDAP client, etc.

To your point, as long as the Spring OAuth2 code is licensed in a way that allows us to leverage it, I have no problem with considering using a different OAuth2 client or some code from that. I don't think there's anything specific that ties Guacamole to the currently-implemented OpenID client.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/PKCEUtil.java
```diff
@@ -0,0 +1,85 @@
...
+     *      in Base64 URL format.
+     */
+    public static String generateCodeChallenge(String verifier) throws Exception {
+        MessageDigest sha = MessageDigest.getInstance("SHA-256");
+        byte[] hash = sha.digest(verifier.getBytes("US-ASCII"));
+        return base64Url(hash);
+    }
+
+    /**
+     * Base64URL encoding without padding.
```
necouchman: This should be a little more descriptive...
```
A method that Base64URL-encodes a provided byte array without padding...
```

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/OpenIDAuthenticationSession.java
necouchman: There still should be a blank line above this one.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/OpenIDAuthenticationSession.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+package org.apache.guacamole.auth.openid;
+
+import org.apache.guacamole.net.auth.AuthenticationSession;
+
+/**
+ * Representation of an in-progress OpenID authentication attempt.
+ */
+
```
necouchman: No blank line needed, here.

necouchman (changes requested):
Thanks for continuing to work on this @adb014. I've got the next round of requested changes - several more style issues that need to be addressed to make sure that the style conforms to the generally-accepted and used style throughout the code, along with some exception-handling issues and a couple of other minor things.

Also, your commit messages need some tweaks:
* Make sure they are descriptive of the changes.
* Make sure they are tagged with the `GUACAMOLE-2258:` prefix.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/OpenIDAuthenticationSession.java
adb014: I think this is a display issue, in the git there is a line

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/OpenIDWellKnown.java
adb014: Sure if its Singleton making them static is redundant. I removed the "static"

adb014: Though  I've noticed a race condition in OpenIDWellKnown.java.. It's possible to call configService.getAuthorizationEndpoint() and the other getters that depend on the Well Known endpoint before OpenIDWellKnown.init() has finished setting its variables. It this case the module initialization completely fails!! It depends on Guice for when things are injected...

At this point I think the getters in OpenIDWellKnown should block until the init function has completed, failed or if there is non WellKnown endpoint set. I'll fix this and propose a modification soon

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/token/TokenValidationService.java
adb014: If the JsonUrlReader is replaced with "org.springframework.security.oauth2" this function should be replaced as well 

Though your question is independent of this. This function is called from authenticateUser, where a null value results in the username being set to null. authenticateUser itself then throws the GuacamoleInvalidCredentialsException. So in this case I think the return value still remain null

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/token/TokenValidationService.java
adb014: If "org.springframework.security.oauth2" is used this function needs to be rethought.. Yeah I shouldn't catch all exceptions, only the one that represents a failure to exchange a code for a token.

I'll revisit this after looking at using "org.springframework.security.oauth2"

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/OpenIDWellKnown.java
adb014: It can be null, so it shouldn't throw an exception

adb014: Hang on.. Looking at getProperty n Environment.java it throws a GuacamoleException ! So not treating this one; or at least only the comment ;-)

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/conf/OpenIDWellKnown.java
adb014: I added a CountDownLatch to this class and blocked the getters till the call to the Well Known endpoint suceeds, fails or the endpoint is absent. So resolved

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/JsonUrlReader.java
adb014: Ok I wrote a version based on "org.springframework.security.oauth2" and here are the list of new depedencies identified by their lack of license, even after exclusion of the maximum number of transitive dependencies 

```
ERROR: License information missing for org.springframework.security:spring-security-oauth2-jose:jar:6.2.4
ERROR: License information missing for org.springframework.security:spring-security-core:jar:6.2.4
ERROR: License information missing for org.springframework.security:spring-security-crypto:jar:6.2.4
ERROR: License information missing for org.springframework:spring-aop:jar:6.1.6
ERROR: License information missing for org.springframework:spring-context:jar:6.1.6
ERROR: License information missing for org.springframework:spring-expression:jar:6.1.6
ERROR: License information missing for org.springframework.security:spring-security-oauth2-core:jar:6.2.4
ERROR: License information missing for org.springframework:spring-core:jar:6.1.6
ERROR: License information missing for org.springframework:spring-jcl:jar:6.1.6
ERROR: License information missing for com.nimbusds:nimbus-jose-jwt:jar:9.24.4
ERROR: License information missing for com.github.stephenc.jcip:jcip-annotations:jar:1.0-1
ERROR: License information missing for org.springframework.security:spring-security-oauth2-client:jar:6.2.4
ERROR: License information missing for org.springframework.security:spring-security-web:jar:6.2.4
ERROR: License information missing for com.nimbusds:oauth2-oidc-sdk:jar:9.43.3
ERROR: License information missing for com.nimbusds:content-type:jar:2.2
ERROR: License information missing for net.minidev:json-smart:jar:2.4.10
ERROR: License information missing for net.minidev:accessors-smart:jar:2.4.9
ERROR: License information missing for org.ow2.asm:asm:jar:9.3
ERROR: License information missing for com.nimbusds:lang-tag:jar:1.7
ERROR: License information missing for org.springframework:spring-web:jar:6.1.6
ERROR: License information missing for org.springframework:spring-beans:jar:6.1.6
```

Is it a problem if pull in that many additional dependencies ? The springframwork OAuth2 code needs springframework core to be at least version 4 and higher if possible.

adb014: As far as I can see there is no way to get rid of an httpclient in some form without upgrading springframework core. What do you want to do ?

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/token/TokenValidationService.java
adb014: As it should always return null, rexolving this comment

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/token/TokenValidationService.java
adb014: As Exception replace with IOException in the catch in the git, resolve this comment, leave only the JsonUrlreader comment open

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/util/JsonUrlReader.java
adb014: Ok, a "mvn dependency:tree" ahow me that the version 3.0.5 of springframework core referenced in doc/licenses is not in fact used anywhere. However, looking further the springframework OAuth2 package needs at least springframework core version 6 and both need Java17+.  I'd thought that Guacamole was still trying to keep Java11 compatibility, so this change would mean that at leaset Java17 would be needed for this module. See https://issues.apache.org/jira/browse/GUACAMOLE-1806 where compatability with Java8/Java11 seems to be required

adb014: Thinking of a half way solution where we keep Jose4j for the JWT claims but use springframework to read the json.. We coud get rid of JsonUrlReader in OpenBaoWellKnown.java like 

```
import org.springframework.web.client.restTemplate;
import java.util.Map;

RestTemplate restTemplate = new RestTemplate();
Map<String,Object> response = restTemplate.getForObject(configService.getWellKnownEndpoint().toString(), Map.Class);

issuer = (String) response.getBody().get("issuer");
authorization_endpoint =  (String) response.getBody().get("authorization_endpoint");
token_endpoint =  (String) response.getBody().get("token_endpoint");
jwks_uri =  (String) response.getBody().get("jwks_uri");
``` 
and in TokenValidationServic.java like

```
import org.springframework.http.HttpEntity;import org.springframework.http.HttpType;
import org.springframework.util.LinkedMultiValueMap;
import org.springframework.util.MultiValueMap;
import org.springframework.web.client.RestTemplate;
import org.springframework.http.HttpHeaders;
import java.util.Map;

RestTemplate restTemplate = new RestTemplate();
MultiValueMap<String, String> body = new LinkedMultiValueMap<String, String>();
body.add("grant_type", "authorization_code");
body.add("code", code);
body.add("redirect_uri", confService.getRedirectURI().toString());
body.add("client_id", confService.getClientID());
body.add("scope", confService.getScope());

String clientSecret = confService.getClientSecret();
if (clientSecret != null && !clientSecret.trim().isEmpty()) {
    body.add("client_secret", clientSecret);
}
if (confService.isPKCERequired()) {
    body.add("code_verifier, verifier);
}

HttpHeaders headers = new HttpHeaders();
headers.setContentType(MediaType.APPLICATION_FORM_URLENCODED);
HttpEntity<MultiValueMap<String, String>> request =
        new HttpEntity<MultiValueMap<String, String>>(body, headers);
String response =  restTemplate.postForObject(tokenUri, request, String.class);

return  (String) response.getBody().get("iid_token");
``` 

Though Frankly it doesn't seem much simplier than what I already have

adb014:
> Thanks for continuing to work on this @adb014. I've got the next round of requested changes - several more style issues that need to be addressed to make sure that the style conforms to the generally-accepted and used style throughout the code, along with some exception-handling issues and a couple of other minor things.
> 
> Also, your commit messages need some tweaks:
> 
> * Make sure they are descriptive of the changes.
> * Make sure they are tagged with the `GUACAMOLE-2258:` prefix.

Aren't all the commits warpped up into a single commit against the guacamole-client repo when the PR is accepted with a seperate description ?  I've just given my commits a bit of description so that I can personally find the right commit if needed. If the commit messages are all imported in to guacamole, yes I understand that they'll need to be more descriptive and tagged

necouchman:
> Aren't all the commits warpped up into a single commit against the [guacamole-client](https://issues.apache.org/jira/browse/GUACAMOLE-client) repo when the PR is accepted with a seperate description ? 

No - a merge commit is added, but each of your commits will still be there, and thus they need to be descriptive, not just to you, but to the entire community.

adb014:
> > Aren't all the commits warpped up into a single commit against the [guacamole-client](https://issues.apache.org/jira/browse/GUACAMOLE-client) repo when the PR is accepted with a seperate description ?
> 
> No - a merge commit is added, but each of your commits will still be there, and thus they need to be descriptive, not just to you, but to the entire community.

Do you want me to recommit them against a clean guacamole-client tree with a single more descriptive commit message and the JIRA ticket in a new pull-request ?

adb014:
> to recommit them

@necouchman Ok I reworked all of the commit messages and force pushed them to my repo

adb014:
Looking at the CI build failure there is an error

```
#20 49.15 [INFO] 02 05 2026 12:26:17.271:INFO [launcher]: Starting browser Firefox
#20 49.15 [INFO] 02 05 2026 12:26:17.279:ERROR [launcher]: Cannot start Firefox
#20 49.15 [INFO] 	
#20 49.15 [INFO] Command '/usr/bin/firefox' requires the firefox snap to be installed.
#20 49.15 [INFO] Please install it with:
#20 49.15 [INFO] 
#20 49.15 [INFO] snap install firefox
#20 49.15 [INFO] 
#20 49.15 [INFO] 
#20 49.15 [INFO] 02 05 2026 12:26:17.279:ERROR [launcher]: Firefox stdout: 
#20 49.15 [INFO] 02 05 2026 12:26:17.279:ERROR [launcher]: Firefox stderr: 
```

in the build of guacamole-common-js. I rebased against the latest main branch and tested again and it build here. Can you relaunch the workflow, as this looks like the build problem is not with my code ?

adb014:
From what I understand in my other PR this should be rebased against staging/1.6.1. However, I didn't create a personal branch for these change and so as far as I can see any forced push is going to mess up this PR...

@necouchman can to resolve any outsatnding issues you have with this PR, I'll correct for the requested changes and after rebase against staging/1.6.1 and generate a new PR

necouchman:
> From what I understand in my other PR this should be rebased against staging/1.6.1. However, I didn't create a personal branch for these change and so as far as I can see any forced push is going to mess up this PR...
> 
> @necouchman can to resolve any outsatnding issues you have with this PR, I'll correct for the requested changes and after rebase against staging/1.6.1 and generate a new PR

I believe this should stay based against the `main` branch - it's enough of a "feature" change, and a pretty large one at that, that I think it should be in the 1.7.0 release and not in a patch release.

adb014:
> > From what I understand in my other PR this should be rebased against staging/1.6.1. However, I didn't create a personal branch for these change and so as far as I can see any forced push is going to mess up this PR...
> > @necouchman can to resolve any outsatnding issues you have with this PR, I'll correct for the requested changes and after rebase against staging/1.6.1 and generate a new PR
> 
> I believe this should stay based against the `main` branch - it's enough of a "feature" change, and a pretty large one at that, that I think it should be in the 1.7.0 release and not in a patch release.

After reading, https://cwiki.apache.org/confluence/display/GUAC/Version+Numbering+and+Branching+Scheme, and particularly the phrase 

```
Changes that require substantial updates to the documentation are never part of patch releases, ...
```

and this PR definitely does require major documentation updates.. Need to re-rebase #1216 as well ;-)

PR #1200, opened 2026-04-10
GUACAMOLE-2260: Text input paste issues: extra blank lines (Chrome) and Ctrl-modified characters.

bbennett-ks:


bbennett-ks:
The "extra blank lines" issue was caused by the fix for GUACAMOLE-2237; the fix for both of these issues will be included in that PR.

PR #1201, opened 2026-04-10
GUACAMOLE-2237: Pasting multiline text via browser menu produces sing…

bbennett-ks:
Inline comments explain the core issue and the fix.

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Keyboard.js
+++ b/guacamole-common-js/src/main/webapp/modules/Keyboard.js
@@ -1071,6 +1071,19 @@ Guacamole.Keyboard = function Keyboard(element) {
     };
 
     /**
+     * Releases any modifier keys that are currently held, so that synthesized
+     * text input is not treated as modified or shortcut input on the remote
+     * side. Lock keys are excluded because they toggle state rather than being
+     * held.
+     */
+    this.releaseModifiersForSyntheticInput = function releaseModifiersForSyntheticInput() {
+        Guacamole.Keyboard.MODIFIER_KEYSYMS.forEach(function(keysym) {
+            if (guac_keyboard.pressed[keysym])
+                guac_keyboard.release(keysym);
+        });
+    };
+
+    /**
      * Resets the state of this keyboard, releasing all keys, and firing keyup
      * events for each released key.
      */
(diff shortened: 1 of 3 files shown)


bbennett-ks:
I've included the fixes for GUACAMOLE-2260 in this PR:
- The `Keyboard.js` fix causes the "extra blank lines" part of the issue.
- The other part of the 2260 issue is in the same code block.

guacamole-common-js/src/main/webapp/modules/Keyboard.js
```diff
@@ -1060,7 +1061,31 @@ Guacamole.Keyboard = function Keyboard(element) {
...
+                    ? (str.codePointAt ? str.codePointAt(i + 1) : str.charCodeAt(i + 1))
+                    : 0;
+
+                if (nextCodepoint === 0x0A)
+                    i++;
+            }
+            else if (codepoint === 0x0A) {
+                // keysym_from_charcode(0x0A) would return 0xFF0A which is not
+                // a keyboard character and is ignored by protocol input
+                // handlers.
```
mike-jumper: `0xFF0A` is, in fact, a valid X11 keysym (see https://cgit.freedesktop.org/xorg/proto/x11proto/tree/keysymdef.h).

guacamole/src/main/frontend/src/app/textInput/directives/guacTextInput.js
```diff
@@ -286,6 +286,35 @@ angular.module('textInput').directive('guacTextInput', [function guacTextInput()
...
+                    return;
+
+                // Take over paste handling before the browser inserts anything.
+                e.preventDefault();
+                e.stopPropagation();
+
+                var content = e.clipboardData.getData('text/plain');
+
+                // Normalize line endings to \n.
+                content = content.replace(/\r\n/g, '\n').replace(/\r/g, '\n');
```
mike-jumper: This is the proper place for normalization (rather than assuming `type()` should not send linefeed keysyms), but would `/\r\n|\r/g` not work here (instead of multiple calls to `replace()`)?

guacamole-common-js/src/main/webapp/modules/Keyboard.js
bbennett-ks: I meant a physical keyboard key rather than X11 key validity. Enter/Return corresponds to `CR` (`0x0D`), while `LF` (`0x0A`) does not have a corresponding standard keyboard key, even though some terminal/emulator paths may treat LF similarly to Enter/Return. That said, this comment was imprecise and isn’t needed anymore. I removed it and added more specific comments elsewhere describing why LF is remapped to Return in these paste/input paths.

guacamole/src/main/frontend/src/app/textInput/directives/guacTextInput.js
bbennett-ks: Thanks! I've moved the Keyboard paste normalization to the proper place and have optimized normalization regex as well. 

I've also added more comments to explain the issues and design. I normalize a paste differently for Keyboard (normalize to `CR`) vs guacTextInput (normalize to `LF`) which seems to be the best thing to do. Please let me know if otherwise.

PR #1207, opened 2026-05-06
GUACAMOLE-2220: Kubernetes: add support for terminal-type.

bbennett-ks:
### Related PRs
[GUACAMOLE-2220 (server 637)](https://github.com/apache/guacamole-server/pull/637)
[GUACAMOLE-2220 (manual 291)](https://github.com/apache/guacamole-manual/pull/291)

### Overview
Unlike SSH and Telnet, Kubernetes sessions do not automatically pass a terminal type. Kubernetes exposes stdin, stdout, and stderr, but has no built-in way to propagate the TERM environment variable.

Kubernetes supports two connection modes:
- **Attach**: connect stdin/stdout/stderr to the main process already running in the container.
- **Exec**: start a new process in the container and connect to that process.

Because Kubernetes does not pass the terminal type itself, this change adds a server-side substitution mechanism for Exec sessions: Any **$TERM** in the Exec command line is replaced with the configured terminal type when the session starts. This lets users decide whether and how to pass the terminal type, for example:
```
env TERM=$TERM /bin/bash
```
Attach mode has no command line, so the terminal type is not propagated automatically.

- Adds support for `terminal-type` for Kubernetes sessions.
- Adds `conditionalLabel` for `Field`, so a form field can replace its label based on another field's value.
- `conditionalLabel`  is used for the Kubernetes `terminal-type` label:
   - If `exec-command` is empty,  the label uses the default `Terminal type (set manually in session):`.
image
   - Else, (a command was specified), the label switches to `Terminal type (available as $TERM in command):`.
image

### DIFF
--- a/guacamole-ext/src/main/java/org/apache/guacamole/form/Field.java
+++ b/guacamole-ext/src/main/java/org/apache/guacamole/form/Field.java
@@ -22,6 +22,7 @@ package org.apache.guacamole.form;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import java.util.Collection;
+import java.util.Map;
 
 /**
  * Represents an arbitrary field, such as an HTTP parameter, the parameter of a
@@ -148,6 +149,31 @@ public class Field {
     private Collection<String> options;
 
     /**
+     * Conditionally overrides this field's label with an alternate
+     * label based on the value of another field.
+     *
+     * This value is a map of string keys to values, similar to a
+     * JSON object. For example:
+     *
+     * {
+     *     "field"    : "exec-command",
+     *     "equals"   : "",
+     *     "labelKey" : "FIELD_HEADER_TERMINAL_TYPE_ATTACH"
+     * }
+     *
+     * Supported entries include:
+     * - "field"      : (String) name of the field to evaluate
+     * - "equals"     : (Object) value the field must equal (optional)
+     * - "notEquals"  : (Object) value the field must not equal (optional)
+     * - "labelKey"   : (String) label key to use if the condition
+     *                           matches
+     *
+     * Exactly one of "equals" or "notEquals" should be specified. If both
+     * are specified, "equals" takes precedence and "notEquals" is ignored.
+     */
+    private Map<String, Object> conditionalLabel;
+
+    /**
      * Creates a new Parameter with no associated name or type.
      */
     public Field() {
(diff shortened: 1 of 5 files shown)


guacamole-ext/src/main/java/org/apache/guacamole/form/Field.java
```diff
@@ -147,6 +148,22 @@ public static class Type {
...
+     * - "field"      : (String) name of the field to evaluate
+     * - "equals"     : (Object) value the field must equal (optional)
+     * - "notEquals"  : (Object) value the field must not equal (optional)
+     * - "labelKey"   : (String) label key to use if the condition
+     *                           matches
+     *
+     * Exactly one of "equals" or "notEquals" should be specified. If both
+     * are specified, "equals" takes precedence and "notEquals" is ignored.
+     */
+    private Map<String, Object> conditionalLabel;
```
necouchman, line 174: This comment is a bit unclear to me - I don't understand how a two-field `Map` is supposed to contain three things: 1) the field to evaluate, 2) `equals` or `notEquals`, and 3) the label key? Maybe I'm just being dense...it happens quite frequently, but can you clarify this?

bbennett-ks: From a C++ background, this seemed bit counterintuitive to me as well.

Map<String, Object> conditionalLabel defines a dictionary/map of string keys to object values. In this case, it is used as a convenient JSON-like structure containing entries such as:
```
{
     "field"      : (String) name of the field to evaluate
     "equals"     : (Object) value the field must equal (optional)
     "notEquals"  : (Object) value the field must not equal (optional)
     "labelKey"   : (String) label key to use if the condition matches
}
```
This provides a convenient way to define and pass around conditional label configuration for consumption.

I'll update the comment so it's more clear.

bbennett-ks: @necouchman I did the English translations. How are other translation done for other languages? Should I try AI? Or, do we periodically have translation professionally translated periodically?

necouchman: English translations are fine - in general we provide the ones we can, and the community provides the updates and corrections. I suppose we could look into AI - actually, I wonder if there are any services out there, maybe even ones for which ASF has a subscription/license, that would automate that translation for us - take the en.json files and then generate/update the matching entries in the other language files. Seems like a really good job for AI.

Anyway, I won't hold up merging this for something like that, just a good thought of something we might look at automating in the future.

necouchman (approved):

PR #1210, opened 2026-05-18
GUACAMOLE-2275: Automatically release virtual keyboard SHIFT keys after character input.

eugen-keeper:


bbennett-ks (approved):
LGTM! Wondering if we should apply the same logic to Ctrl/Alt/Windows/Meta as well? That can be a different issue.

eugen-keeper:
> LGTM! Wondering if we should apply the same logic to Ctrl/Alt/Windows/Meta as well? That can be a different issue.

I believe SHIFT is special here - it’s mainly for one-shot capitalization, while Ctrl/Alt/Meta are for chord sequences where the latched state is intentional. Caps is a lock and shouldn’t auto-release either. I’d keep this change SHIFT-only.

eugen-keeper:
Documentation update: https://github.com/apache/guacamole-manual/pull/295

PR #1212, opened 2026-05-20
GUACAMOLE-2277: On-Screen Keyboard: incorrect Meta keysym generated.

bbennett-ks:
Currently, the On-Screen Keyboard Meta key generates Super_L keysym. This works fine for RDP sessions:
- Linux, treats it as the Meta key.
- Windows interprets as Meta which is treated as the Windows key.

With VNC, inconsistent browser key events are generated
- Linux, pressing Meta emits an Alt/Shift sequence.
- macOS: pressing Meta emits AltLeft.

Emitting Meta_L instead:
- RDP, Windows and Linux: change in behavior.
- VNC, Linux macOS: Meta browser key events are generated.

See the Jira for more details: https://issues.apache.org/jira/browse/GUACAMOLE-2277

### DIFF
--- a/guacamole/src/main/frontend/src/layouts/de-de-qwertz.json
+++ b/guacamole/src/main/frontend/src/layouts/de-de-qwertz.json
@@ -118,7 +118,7 @@
         "Meta" : [{
             "title"    : "Meta",
             "modifier" : "meta",
-            "keysym"   : 65511
+            "keysym"   : 65515
         }],
 
         "^" : [
--- a/guacamole/src/main/frontend/src/layouts/en-us-qwerty.json
+++ b/guacamole/src/main/frontend/src/layouts/en-us-qwerty.json
@@ -90,7 +90,7 @@
         "Meta" : [{
             "title"    : "Meta",
             "modifier" : "meta",
-            "keysym"   : 65511
+            "keysym"   : 65515
         }],
 
         "`" : [
--- a/guacamole/src/main/frontend/src/layouts/es-es-qwerty.json
+++ b/guacamole/src/main/frontend/src/layouts/es-es-qwerty.json
@@ -118,7 +118,7 @@
         "Meta" : [{
             "title"    : "Meta",
             "modifier" : "meta",
-            "keysym"   : 65511
+            "keysym"   : 65515
         }],
 
         "º" : [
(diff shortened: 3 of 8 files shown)


necouchman:
Looking back at the history of this file, it seems we've gone back and forth about it:

https://github.com/apache/guacamole-client/commit/8b5cdbb3b513225cdb6900332ac34250094cce79

We should probably discuss which one really makes the most sense so we're not swapping back and forth every few years...

@mike-jumper ?

bbennett-ks:
> Looking back at the history of this file, it seems we've gone back and forth about it:
> 
> [8b5cdbb](https://github.com/apache/guacamole-client/commit/8b5cdbb3b513225cdb6900332ac34250094cce79)
> 
> We should probably discuss which one really makes the most sense so we're not swapping back and forth every few years...
> 
> @mike-jumper ?

Yep, it's definitely not black & white. Though I think `Super_L` is lightest shade of gray. :) I have a little more detail in the Jira (https://issues.apache.org/jira/browse/GUACAMOLE-2277). Also, we could make the behavior configurable. Or make it's behavior based on session type?

But without doing something, the Meta key is useless on macOS so no Meta short cuts.

mike-jumper:
I'm unsure about this. The Meta key _is_ the Mac Command key, and the opposite of this was part of 1.4.0:

https://issues.apache.org/jira/browse/GUACAMOLE-1386

mike-jumper:
Looking at the [GUACAMOLE-2277](https://issues.apache.org/jira/browse/GUACAMOLE-2277) issue, it looks like there may be some confusion: the Meta key is not Alt on a PC keyboard, it's the Windows key.

Some Linux editors refer to Alt as Meta, but that's just the terminology used by those editors. Windows key is Meta, Alt is Alt.

bbennett-ks:
> Looking at the [GUACAMOLE-2277](https://issues.apache.org/jira/browse/GUACAMOLE-2277) issue, it looks like there may be some confusion: the Meta key is not Alt on a PC keyboard, it's the Windows key.
> 
> Some Linux editors refer to Alt as Meta, but that's just the terminology used by those editors. Windows key is Meta, Alt is Alt.

Apologies, I was referring to key events were generated and how they were interpreted by the OS. ! I've provided more detail and clarity (hopefully!).

bbennett-ks:
@mike-jumper @necouchman I updated https://issues.apache.org/jira/browse/GUACAMOLE-2277 with more details, including before and after this fix for OSes & protocols.

Based on these results, I believe this is the correct fix.

PR #1213, opened 2026-05-21
GUACAMOLE-2276: Add and update Czech localization entries

Vict0rC:
Update `cs.json` for the 1.6.1 release line.

**Added translations**
- `CLIENT.ERROR_CLIENT_205` - connection resource sharing error
- `PLAYER.ACTION_CLOSE` - recording player close action

**Updated translations**
- `CLIENT.HELP_MOUSE_MODE_ABSOLUTE` - remove mixed English wording
- `PLAYER.INFO_NUMBER_OF_RESULTS` - Czech plural forms (`shoda` / `shody` / `shod`)
- `FIELD_HEADER_FONT_NAME` / `FIELD_HEADER_FONT_SIZE` - standardized to "Název písma" / "Velikost písma" (Kubernetes, SSH, Telnet)

### DIFF
--- a/guacamole/src/main/frontend/src/translations/cs.json
+++ b/guacamole/src/main/frontend/src/translations/cs.json
@@ -84,6 +84,7 @@
         "ERROR_CLIENT_201"     : "Synchronizační server je zaneprázdněn, zkuste to prosím znovu později.",
         "ERROR_CLIENT_202"     : "Guacamole server zavřel spojení protože vzdálený počítač příliš dlouho neodpovídal. Zkuste to prosím později, nebo kontaktujte správce.",
         "ERROR_CLIENT_203"     : "Chyba vzdáleného serveru, spojení bylo uzavřeno. Zkuste to prosím později, nebo kontaktujte správce.",
+        "ERROR_CLIENT_205"     : "Toto připojení vyžaduje prostředek, který je právě používán, ale sdílený přístup k tomuto prostředku buď není možný, nebo není povolen. Zkuste to prosím později.",
         "ERROR_CLIENT_207"     : "Server vzdálené plochy je aktuálně nedostupný. Pokud problém přetrvává, informujte prosím správce systému nebo zkontrolujte systémové protokoly.",
         "ERROR_CLIENT_208"     : "Server vzdálené plochy není aktuálně k dispozici. Pokud problém přetrvává, informujte prosím správce systému nebo zkontrolujte systémové protokoly.",
         "ERROR_CLIENT_209"     : "Server vzdálené plochy ukončil připojení, protože je v konfliktu s jiným připojením. Prosím zkuste to znovu později.",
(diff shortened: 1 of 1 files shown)


guacamole/src/main/frontend/src/translations/cs.json
```diff
@@ -43,7 +43,7 @@
         
         "FIELD_HEADER_PASSWORD"       : "Heslo:",
         "FIELD_HEADER_PASSWORD_AGAIN" : "Heslo znovu:",
-        "FIELD_HEADER_RECORDING_WRITE_EXISTING"  : "Povolit zápis do existujícího záznamového souboru:",
```
necouchman, line 46: Why is this being removed?

guacamole/src/main/frontend/src/translations/cs.json
```diff
@@ -153,7 +154,7 @@
         "SECTION_HEADER_FILE_TRANSFERS"  : "Přenos souborů",
         "SECTION_HEADER_INPUT_METHOD"    : "Metoda vstupu",
         
-        "SECTION_HEADER_MOUSE_MODE"      : "Mód emulace myši",
```
necouchman, line 156: Why is this being removed?

necouchman (changes requested):
Looks much better - just a couple of minor things.

Vict0rC:
anything else wrong?

necouchman:
> anything else wrong?

I left a couple of comments above that I do not see having been addressed or commented on?

Vict0rC:
I tried to answer everything right away. Did I miss something?

guacamole/src/main/frontend/src/translations/cs.json
necouchman: I see the line is no longer being removed, but the spacing that aligns it with the line below has now been removed. It looks like that's the only change to this line - the translation appears identical, so there's no reason to change this line at all, correct?

guacamole/src/main/frontend/src/translations/cs.json
necouchman: Similar to the comment above - the only change to this line appears to be removal of the space that aligns the spacing with the lines above it. There's no reason to change this, as far as I can tell?

necouchman (changes requested):
Overall looks good, but the two lines that I had previously called out, I'm not sure why they're being modified at all? The lines were spaced such that they lined up with surrounding lines, which is consistent with the style in the translation files. As far as I can tell, there's no reason at all to modify those lines?

Sorry, it probably seems like I'm being pedantic or nit-picky, but we do have established style guidelines for reasons, and it'd be great if those two lines could just be restored to how they were before, as they're not really changing.

Vict0rC:
Hello Nick, is it ok now? :-)

mike-jumper (changes requested):
The `FIELD_HEADER_RECORDING_INCLUDE_CLIPBOARD` strings were introduced as part of GUACAMOLE-1969, which is part of 1.7.0 (`main`). To include this in a patch release or within the upcoming 1.6.1, the changes that define that string will need to be moved to a different PR specific to 1.7.0.

Otherwise, LGTM, and should be good to base off `staging/1.6.1`.

necouchman (approved):

PR #1218, opened 2026-06-02
GUACAMOLE-2281: Add WebAuthn passthrough relay over auth-challenge.

aleitner:
Adds a WebAuthn passthrough relay that runs ceremonies on the user's local authenticator on behalf of a remote session. The server opens an inbound auth-challenge stream carrying the ceremony request body; the relay runs the WebAuthn ceremony and replies on an outbound auth-response stream.

guacamole-common-js (Client.js)

sendAuthChallenge and sendAuthResponse returning a Guacamole.OutputStream the caller writes the body to.
onauthchallenge and onauthresponse events firing with (stream, mimetype, challengeId).
Matching auth-challenge and auth-response instruction handlers.
Angular client

ManagedClient routes inbound auth-challenge streams to ManagedWebAuthn based on mimetype (application/x-webauthn-create+json or application/x-webauthn-get+json), acking unsupported mimetypes.
ManagedWebAuthn assembles the challenge body off the stream, runs the ceremony via the existing webAuthnService under an AbortController keyed by challenge_id, and writes the credential or error JSON back on an auth-response stream.
On tunnel CLOSED, abortAll dismisses the local authenticator UI for any ceremonies left in flight.

[Guacamole Server PR](https://github.com/apache/guacamole-server/pull/675)

### DIFF
--- a/guacamole-common-js/src/main/webapp/modules/Client.js
+++ b/guacamole-common-js/src/main/webapp/modules/Client.js
@@ -23,7 +23,7 @@ var Guacamole = Guacamole || {};
  * Guacamole protocol client. Given a {@link Guacamole.Tunnel},
  * automatically handles incoming and outgoing Guacamole instructions via the
  * provided tunnel, updating its display using one or more canvas elements.
- * 
+ *
  * @constructor
  * @param {!Guacamole.Tunnel} tunnel
  *     The tunnel to use to send and receive Guacamole instructions.
@@ -540,6 +540,35 @@ Guacamole.Client = function(tunnel) {
     };
 
     /**
+     * Sends an auth-response instruction over the tunnel, announcing an
+     * output stream carrying the response body for a previously-received
+     * auth-challenge identified by the same challengeId.
+     *
+     * @param {!string} mimetype
+     *     The mimetype of the data which will be sent over the returned
+     *     stream.
+     *
+     * @param {!string} challengeId
+     *     The challenge identifier of the originating auth-challenge
+     *     being responded to.
+     *
+     * @return {Guacamole.OutputStream}
+     *     The created stream, or null if not connected.
+     */
+    this.sendAuthResponse = function(mimetype, challengeId) {
+
+        // Do not send if not connected
+        if (!isConnected())
+            return null;
+
+        var stream = guac_client.createOutputStream();
+        tunnel.sendMessage("auth-response", stream.index, mimetype,
+                challengeId);
+        return stream;
+
+    };
+
+    /**
      * Opens a new clipboard object for writing, having the given mimetype. The
      * instruction necessary to create this stream will automatically be sent.
      *
@@ -917,7 +946,26 @@ Guacamole.Client = function(tunnel) {
      *     The name of the pipe.
      */
     this.onpipe = null;
-    
+
+    /**
+     * Fired when an "auth-challenge" instruction is received from the
+     * server, announcing a stream carrying the challenge body for a
+     * pending authentication exchange. The handler is responsible for
+     * consuming the body from the given input stream and replying via
+     * sendAuthResponse() with the same challengeId.
+     *
+     * @event
+     * @param {!Guacamole.InputStream} stream
+     *     The input stream carrying the challenge body.
+     *
+     * @param {!string} mimetype
+     *     The mimetype of the challenge body. Identifies the auth flavor.
+     *
+     * @param {!string} challengeId
+     *     The challenge identifier carried by the auth-challenge.
+     */
+    this.onauthchallenge = null;
+
     /**
      * Fired when a "required" instruction is received. A required instruction
      * indicates that additional parameters are required for the connection to
@@ -1557,7 +1605,7 @@ Guacamole.Client = function(tunnel) {
             var mimetype = parameters[1];
             var name = parameters[2];
 
-            // Create stream 
+            // Create stream
             if (guac_client.onpipe) {
                 var stream = streams[stream_index] = new Guacamole.InputStream(guac_client, stream_index);
                 guac_client.onpipe(stream, mimetype, name);
@@ -1566,6 +1614,23 @@ Guacamole.Client = function(tunnel) {
             // Otherwise, unsupported
             else
                 guac_client.sendAck(stream_index, "Named pipes unsupported", 0x0100);
+
+        },
+
+        "auth-challenge": function(parameters) {
+
+            var stream_index = parseInt(parameters[0]);
+            var mimetype = parameters[1];
+            var challengeId = parameters[2];
+
+            if (guac_client.onauthchallenge) {
+                var stream = streams[stream_index] =
+                        new Guacamole.InputStream(guac_client, stream_index);
+                guac_client.onauthchallenge(stream, mimetype, challengeId);
+            }
+            else
+                guac_client.sendAck(stream_index,
+                        "Auth challenges unsupported", 0x0100);
 
         },
 
--- a/guacamole/src/main/frontend/src/app/client/types/ManagedClient.js
+++ b/guacamole/src/main/frontend/src/app/client/types/ManagedClient.js
@@ -36,6 +36,7 @@ angular.module('client').factory('ManagedClient', ['$rootScope', '$injector',
     const ManagedFilesystem      = $injector.get('ManagedFilesystem');
     const ManagedFileUpload      = $injector.get('ManagedFileUpload');
     const ManagedShareLink       = $injector.get('ManagedShareLink');
+    const ManagedWebAuthn        = $injector.get('ManagedWebAuthn');
 
     // Required services
     const $document               = $injector.get('$document');
@@ -293,6 +294,17 @@ angular.module('client').factory('ManagedClient', ['$rootScope', '$injector',
          */
         this.deferredPipeStreamHandlers = template.deferredPipeStreamHandlers || {};
 
+        /**
+         * State tracking the WebAuthn passthrough relay for this client.
+         * Ceremony requests arrive as inbound "auth-challenge" streams
+         * with WebAuthn-flavored mimetypes; responses are sent back as
+         * outbound "auth-response" streams carrying the credential or
+         * error.
+         *
+         * @type ManagedWebAuthn
+         */
+        this.managedWebAuthn = template.managedWebAuthn || null;
+
     };
 
     /**
@@ -443,6 +455,25 @@ angular.module('client').factory('ManagedClient', ['$rootScope', '$injector',
             tunnel : tunnel
         });
 
+        // Set up WebAuthn passthrough and dispatch inbound auth-challenge
+        // streams to it. Other mimetypes are not currently expected.
+        managedClient.managedWebAuthn = ManagedWebAuthn.getInstance(managedClient);
+
+        client.onauthchallenge = function clientAuthChallengeReceived(
+                stream, mimetype, challengeId) {
+            if (ManagedWebAuthn.handlesMimetype(mimetype))
+                ManagedWebAuthn.handleChallenge(
+                        managedClient.managedWebAuthn,
+                        stream, mimetype, challengeId);
+            else {
+                console.warn('[ManagedClient] no handler for auth-challenge '
+                        + 'mimetype "' + mimetype + '" (id=' + challengeId
+                        + ')');
+                stream.sendAck('Unsupported auth-challenge mimetype',
+                        Guacamole.Status.Code.UNSUPPORTED);
+            }
+        };
+
         // Fire events for tunnel errors
         tunnel.onerror = function tunnelError(status) {
             $rootScope.$apply(function handleTunnelError() {
@@ -485,6 +516,11 @@ angular.module('client').factory('ManagedClient', ['$rootScope', '$injector',
 
                     // Connection has closed
                     case Guacamole.Tunnel.State.CLOSED:
+                        // Dismiss any local authenticator UI left stranded
+                        // by the tunnel dropping mid-ceremony.
+                        if (managedClient.managedWebAuthn)
+                            ManagedWebAuthn.abortAll(
+                                    managedClient.managedWebAuthn);
                         ManagedClientState.setConnectionState(managedClient.clientState,
                             ManagedClientState.ConnectionState.DISCONNECTED);
                         break;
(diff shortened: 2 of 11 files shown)


guacamole/src/main/frontend/src/app/client/types/ManagedWebAuthn.js
```diff
@@ -0,0 +1,204 @@
...
+            const client = managedWebAuthn.client && managedWebAuthn.client.client;
+            if (client)
+                client.sendDone(id, 'error', JSON.stringify({
+                    name: 'DataError',
+                    message: 'Malformed WebAuthn request payload'
+                }));
+            return;
+        }
+
+        managedWebAuthn.inFlight++;
```
eugen-keeper: Probably this should be done inside of `$rootScope.$apply` because `ManagedWebAuthn.handleRequest` is called from outside of the Angular world: `client.onreq` -> Guacamole `tunnel.oninstruction`.

guacamole/src/main/frontend/src/app/client/types/ManagedWebAuthn.js
```diff
@@ -0,0 +1,204 @@
...
+     *     The response payload object; will be JSON-encoded for the
+     *     "done" instruction.
+     */
+    function sendDone(managedWebAuthn, id, status, payload) {
+
+        const client = managedWebAuthn.client && managedWebAuthn.client.client;
+        if (client)
+            client.sendDone(id, status, JSON.stringify(payload));
+
+        managedWebAuthn.inFlight = Math.max(0, managedWebAuthn.inFlight - 1);
```
eugen-keeper: As far as I understand `sendDone` can be executed from inside Angular UI update and from pure browser/Guacamole API, so we should wrap the code on lines 194 - 198 into `$rootScope.$apply`. Maybe something like the following for this and line 144:

`function applyPanelStateChange(updateFn) {`
`   if ($rootScope.$$phase)`
`     updateFn();`
`   else`
`     $rootScope.$apply(updateFn);`
`}`


and then

`applyPanelStateChange(() => { managedWebAuthn.inFlight++; });`

guacamole/src/main/frontend/src/app/client/types/ManagedClient.js
```diff
@@ -443,6 +465,44 @@ angular.module('client').factory('ManagedClient', ['$rootScope', '$injector',
             tunnel : tunnel
         });
 
+        // Dispatch inbound "req" instructions by method name to whichever
+        // handler the application registered via
+        // ManagedClient.registerRequestHandler(). Unknown methods are
+        // rejected with a NotSupportedError "done".
+        client.onreq = function clientReqReceived(id, name, payload) {
```
eugen-keeper: `onreq` handling is set up, but there’s no matching `client.oncancel` handler. Should we have it?

guacamole/src/main/frontend/src/app/webauthn/services/webAuthnService.js
```diff
@@ -0,0 +1,267 @@
...
+                            request.sameOriginWithAncestors !== false
+                }
+            });
+
+            const op = request.kind === 'create'
+                    ? navigator.credentials.create({ publicKey })
+                    : navigator.credentials.get({ publicKey });
+
+            op.then(function ceremonyResolved(credential) {
+                deferred.resolve(serializeCredential(credential));
```
eugen-keeper, line 267: Probably there should be **null** check and processing here because according to MSDN **navigator.credentials.get** -> `The API always fulfills with a single credential or null.`. So it can be null for at least a couple of cases.

guacamole/src/main/frontend/src/app/webauthn/services/webAuthnService.js
```diff
@@ -0,0 +1,267 @@
...
+         * @param {Boolean} [request.sameOriginWithAncestors=true]
+         *     Whether the remote page that initiated the ceremony is
+         *     same-origin with all its ancestor frames.
+         *
+         * @returns {!Promise.<!Object>}
+         *     A promise that resolves with the serialized credential, or
+         *     rejects with an Error decorated with a "policyBlocked" property
+         *     set to true if the browser refused the override due to policy.
+         */
+        service.performCeremony = function performCeremony(request) {
```
eugen-keeper: What happens if the Guacamole session disconnects while a **WebAuthn** ceremony is still in progress? Will it be handled correctly?

guacamole/src/main/frontend/src/app/client/types/ManagedWebAuthn.js
```diff
@@ -0,0 +1,204 @@
...
+         */
+        this.lastStatus = template.lastStatus || null;
+
+        /**
+         * The error message from the most recently-failed ceremony, or null
+         * if there have been no failures.
+         *
+         * @type {String}
+         */
+        this.lastError = template.lastError || null;
```
eugen-keeper: `lastError` is updated when a ceremony fails, but I don’t see it used anywhere in the templates or other code

eugen-keeper (changes requested):

guacamole/src/main/frontend/src/app/client/types/ManagedWebAuthn.js
aleitner: Fixed by adding an `applyPanelStateChange` helper like you suggested. Used for the `inFlight` bump and at the `sendDone` site from the other comment about `$rootScope.$apply`

guacamole/src/main/frontend/src/app/client/types/ManagedWebAuthn.js
aleitner: Same helper applied here. The `sendDone` path is already inside a digest in practice (resolution comes through `$q.defer().promise.then`, which auto-schedules via `$evalAsync`), but the wrap covers future non-`$q` callers

guacamole/src/main/frontend/src/app/client/types/ManagedClient.js
aleitner: Good catch. Added `client.oncancel` and a `tunnel.onstatechange CLOSED` listener. Both abort an `AbortSignal` threaded through `webAuthnService.performCeremony`. Also fixes the other comment about disconnect during ceremony: killing guacd mid ceremony dismisses the dialog

guacamole/src/main/frontend/src/app/webauthn/services/webAuthnService.js
aleitner: Reject with `NotAllowedError`. We don't forward mediation today so this isn't reachable, but worth guarding

guacamole/src/main/frontend/src/app/webauthn/services/webAuthnService.js
aleitner: Same `AbortController` plumbing as the other comment about the missing `oncancel` handler: the `CLOSED` branch of `tunnel.onstatechange` calls `ManagedWebAuthn.abortAll`, firing every controller in the table. Verified by killing guacd while the YubiKey was waiting for a touch, the local dialog dismissed

guacamole/src/main/frontend/src/app/client/types/ManagedWebAuthn.js
aleitner: Good catch. Unused (no template binding, no controller read). Dropped the field, the constructor copy, and the `sendDone` assignment

eugen-keeper (approved):

guacamole-common-js/src/main/webapp/modules/Client.js
```diff
@@ -1557,7 +1730,46 @@ Guacamole.Client = function(tunnel) {
...
+                var chunks = [];
+                var totalLength = 0;
+                var aborted = false;
+
+                var reader = new Guacamole.StringReader(stream);
+
+                reader.ontext = function(text) {
+                    if (aborted)
+                        return;
+                    totalLength += text.length;
```
bbennett-ks: Depending on the MIME type, `text.length` is not necessarily the size of the text in bytes. For example, a UTF-8 character can be as large as 4 bytes, so we could potentially send oversize packets (which would be rejected by the server).

bbennett-ks: Maybe rename `RPC_MAX_BODY_LENGTH`  to `RPC_MAX_BODY_SIZE`?

bbennett-ks (commented):
Noticed this reviewing the server side.

bbennett-ks (changes requested):

guacamole-common-js/src/main/webapp/modules/Client.js
aleitner: Switched to `ArrayBufferReader` so the cap is counted in bytes rather than characters.
Renamed `RPC_MAX_BODY_LENGTH` to `RPC_MAX_BODY_SIZE`

bbennett-ks (approved):
LGTM!

guacamole/src/main/frontend/src/app/client/types/ManagedWebAuthn.js
```diff
@@ -0,0 +1,350 @@
...
+     * no live session behind it.
+     *
+     * @param {!ManagedWebAuthn} managedWebAuthn
+     *     The ManagedWebAuthn whose ceremonies should be canceled.
+     */
+    ManagedWebAuthn.abortAll = function abortAll(managedWebAuthn) {
+        const controllers = managedWebAuthn._abortControllers;
+        for (const id in controllers) {
+            if (Object.prototype.hasOwnProperty.call(controllers, id))
+                controllers[id].abort();
```
bbennett-ks, line 225: Should `controllers[id]` be deleted as well?

guacamole/src/main/frontend/src/app/client/types/ManagedWebAuthn.js
```diff
@@ -0,0 +1,350 @@
...
+
+            console.warn('[WebAuthn] ceremony id=' + challengeId
+                    + ' kind=' + kind
+                    + (error && error.policyBlocked ? ' (policy-blocked)' : '')
+                    + ' failed:', error);
+
+            sendResponse(managedWebAuthn, mimetype, challengeId, 'error', {
+                name: (error && error.name) || 'Error',
+                message: (error && error.message) || String(error)
+            });
```
bbennett-ks, line 288: `managedWebAuthn._abortControllers[challengeId]` needs to be deleted on failure.

guacamole/src/main/frontend/src/app/client/types/ManagedWebAuthn.js
```diff
@@ -0,0 +1,350 @@
...
+
+        const client = managedWebAuthn.client && managedWebAuthn.client.client;
+        if (client) {
+            const stream = client.sendAuthResponse(mimetype, challengeId);
+            if (stream) {
+                const writer = new Guacamole.StringWriter(stream);
+                writer.sendText(JSON.stringify(Object.assign(
+                        { status: status }, payload)));
+                writer.sendEnd();
+            }
```
bbennett-ks, line 332: Log a warning?

guacamole/src/main/frontend/src/app/client/types/ManagedWebAuthn.js
```diff
@@ -0,0 +1,350 @@
...
+        }
+
```
bbennett-ks: If an exception isn't caught, `applyPanelStateChange()` won't be called. Is that OK?

guacamole/src/main/frontend/src/app/webauthn/services/webAuthnService.js
```diff
@@ -0,0 +1,302 @@
...
+         * @returns {!String}
+         *     The base64 encoding of the given buffer.
+         */
+        function base64Encode(buf) {
+            const bytes = buf instanceof ArrayBuffer
+                    ? new Uint8Array(buf)
+                    : new Uint8Array(buf.buffer, buf.byteOffset, buf.byteLength);
+            let str = '';
+            for (let i = 0; i < bytes.length; i++)
+                str += String.fromCharCode(bytes[i]);
```
bbennett-ks, line 83: If credentials are large, this could be pretty expensive (O(n2)), since a new string is created for every byte. It seems like the JavaScript framework should have something for this.

guacamole/src/main/frontend/src/app/client/types/ManagedWebAuthn.js
```diff
@@ -0,0 +1,350 @@
...
+                    Guacamole.Status.Code.UNSUPPORTED);
+            return;
+        }
+
+        stream.sendAck('Ready', Guacamole.Status.Code.SUCCESS);
+
+        let payload = '';
+        const reader = new Guacamole.StringReader(stream);
+        reader.ontext = function challengeChunk(text) {
+            payload += text;
```
bbennett-ks: This is another instance of O(n^2) new string for every byte. In JS, use an array?

bbennett-ks (changes requested):
I reviewed the changes between `bf1a524` and current (`bdf19be`).

guacamole/src/main/frontend/src/app/client/types/ManagedWebAuthn.js
aleitner: sendResponse handles it

guacamole/src/main/frontend/src/app/client/types/ManagedWebAuthn.js
aleitner: sendResponse deletes it on both branches

guacamole/src/main/frontend/src/app/client/types/ManagedWebAuthn.js
aleitner: Added the warn

guacamole/src/main/frontend/src/app/client/types/ManagedWebAuthn.js
aleitner: Wrapped in try/finally

guacamole/src/main/frontend/src/app/webauthn/services/webAuthnService.js
aleitner: Payloads are bounded (a few KB per ceremony) and there's no framework helper for this. Can switch to chunked String.fromCharCode.apply if you'd prefer

guacamole/src/main/frontend/src/app/client/types/ManagedWebAuthn.js
aleitner: Switched to Guacamole.JSONReader

guacamole/src/main/frontend/src/app/webauthn/services/webAuthnService.js
bbennett-ks: Nope, what every you think is best.

bbennett-ks (approved):
LGTM!

PR #1221, opened 2026-06-25
GUACAMOLE-2290: Treat invalid localStorage JSON as missing

pv2b:


guacamole/src/main/frontend/src/app/storage/services/localStorageService.js
```diff
@@ -121,8 +121,14 @@ angular.module('storage').provider('localStorageService', [function localStorage
...
         // Pull and parse value from internal storage, if present
         var data = storedItems[key];
-        if (data)
-            return JSON.parse(data);
+        if (data) {
+            try {
+                return JSON.parse(data);
+            }
+            catch (ignore) {
+                return null;
```
necouchman, line 130: @pv2b Am I right in assuming that, in the case you've run into as describe in the Jira ticket, the only implication of doing this - returning `null` and avoiding passing on the exception - is that, in the case where the `GUAC_AUTH_TOKEN` value in `localStorage` ends up corrupt or missing and this `JSON.parse()` call fails, you get kicked back to the login screen?

Are there any other places where `getItem()` is used where altering this behavior - returning null instead of passing back the exception - would cause issues?

pv2b: In my testing, without this patch, if GUAC_AUTH_TOKEN in localStorage is set to invalid JSON (for example, the string "undefined") then the exception will end up never being caught and handled, which causes the entire UI to fail loading. The result is a completely blank page.

By altering the behaviour to simply return null in case the JSON is invalid, the Guacamole app will act as if the value in localStorage never existed in the first place. For the case of authentication tokens, that means kicking the user back to the login screen.

After you made the comment, I quickly re-checked the codebase for all uses of getItem(), and I didn't find any obvious cases that are relying on getItem() throwing an exception in case it contains invalid JSON.

Of course, it would be a slightly more conservative fix to instead wrap the call to getItem() in getCurrentToken() in authenticationService.js, or perhaps even in request(). But even if it's more conservative, I'm not sure it's really "better".

The issue is that the implicit "usage contract" for getItem() seems to be to simply just check for a null result - none of the callers except getItem() to throw an exception. If we were to start using exceptions for flow control, then it would probably be better for getItem() to also throw an exception in case the value is missing, no? Because it's pretty strange that it returns null for some exceptional cases, and throws an exception for others. Better to pick one way to signal to the caller something went wrong, and stick with it. That's what my patch does.

That said, it could also be worth adding a console.log if this exception is hit. It's not expected that this would ever be hit in a stock Guacamole deployment - the issue is that we are running a second authentication layer on the outside of Guacamole that, when it times out, will return error pages to API calls. This causes Guacamole to store invalid JSON in LocalStorage, because of poor error handling on the write side as well. That could also be fixed, of course, but I'm trying to keep the change as minimal as possible.

necouchman: Sounds good - I'm aligned with your overall fix, here. I do think it would be good to log an `error` to the JS console before returning null.

pv2b: I've added an error output as you requested

necouchman (approved):

PR #1224, opened 2026-07-04
Add a draft security threat model (THREAT_MODEL.md) + SECURITY.md + discoverability wiring

potiuk:
**This is a proposal for the Guacamole PMC to review — please correct, reject, or discuss as needed.** The PMC owns the document; nothing here is a requirement.

This adds a draft **umbrella** `THREAT_MODEL.md` for the Apache Guacamole family, a `SECURITY.md` (the code repos don't currently carry one — the security policy lives only at guacamole.apache.org/security), and an `AGENTS.md` wiring `AGENTS.md -> SECURITY.md -> THREAT_MODEL.md` so an automated scanner can mechanically discover it. Path 3 as agreed on the list (we draft the v0, the PMC reviews). It lands on apache/guacamole-client as the canonical home; guacamole-server and guacamole-website get small pointer PRs to it once you're happy with the shape.

Generated from Guacamole's public artefacts (the security page, docs, advisory history) via the [`threat-model-producer`](https://gist.github.com/scovetta/2dc9a0695c7cbcc32e23799e00d2ced3) rubric. Provenance-tagged throughout; every *(inferred)* claim routes to a numbered §14 question (15 of them).

The model's spine is the guacd(C) / client(Java) / website(static) split:

- **guacd + libguac + the C protocol plugins** are modeled as the primary **memory-safety** surface — untrusted data from the remote desktop server flowing back into C parsers (anchored to the real CVE history: the 2020 "Reverse RDP" set, the VNC/RDP/SSH parsing CVEs);
- **guacamole-client** as the authn/authz + tunnel surface;
- **guacamole-website** is explicitly **out of model for runtime findings**, in scope only for discoverability.

Two calls we'd especially value your confirmation on (both in §14 wave 1): (1) **guacd is not an auth boundary** and must be kept off untrusted networks — a report against an exposed `:4822` is a misconfiguration, not an in-model bug; (2) the plaintext/unauthenticated webapp↔guacd default (`guacd-ssl` off on a trusted segment) as the supported production posture.

What's most useful: walk the §14 questions and confirm / correct / strike each — a one-liner each is enough. This PR only adds files; it edits no existing content.

Context: the ASF Security team is preparing projects for an automated agentic security scan we're piloting; discoverability is the one hard prerequisite.

### DIFF
--- a/THREAT_MODEL.md
+++ b/THREAT_MODEL.md
@@ -1,0 +1,581 @@
+<!--
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+-->
+
+# Apache Guacamole — Threat Model
+
+## §1 Header
+
+- **Project:** Apache Guacamole (the clientless remote-desktop gateway family)
+- **Repositories modeled (scope locked by the PMC):**
+  `apache/guacamole-client`, `apache/guacamole-server`, `apache/guacamole-website`.
+- **Version/commit binding:** Written against the `1.6.x` line / `main` of both
+  code repositories as of the date below. This is an *umbrella* model covering all
+  three repositories; a report against release *N* is triaged against the model as
+  it stood at *N*, not at HEAD.
+- **Date:** 2026-07-04
+- **Author:** ASF Security team, drafted via the threat-model-producer (Scovetta)
+  rubric at the Guacamole PMC's request (path 3 — the Security team drafts, the
+  PMC reviews and ratifies).
+- **Status:** **v0 DRAFT — for Guacamole PMC review.** Nothing here is ratified;
+  every *(inferred)* claim is a proposal for the PMC to confirm, correct, or
+  strike (see §14). No maintainer answers have been folded in yet.
+- **Reporting cross-reference:** Findings that violate a property we claim in §8
+  should be reported privately per our security policy
+  (<https://guacamole.apache.org/security/>, `security@guacamole.apache.org`, or
+  the ASF Security Team) *(documented)*. Findings that fall under §3 (out of
+  scope) or §9 (properties we do not provide) will be closed citing this
+  document.
+- **Provenance legend** — every non-trivial claim carries exactly one tag:
+  - *(documented)* — stated in our own docs (the security page, the manual, the
+    configuration reference). Cited inline.
+  - *(maintainer)* — stated by a Guacamole maintainer in response to this
+    process. **None yet — this is v0.**
+  - *(inferred)* — reasoned from architecture, code structure, or domain
+    knowledge; not yet confirmed. Every such tag has a matching open question in
+    §14.
+- **Draft confidence:** ~30 documented citations / 0 maintainer / 18 distinct
+  *(inferred)* claims (I1–I18, each with a §14 question). The bulk of the model's
+  *reasoning* is still hypothesis pending PMC review — expected for a v0.
+
+**What Guacamole is.** Apache Guacamole is a *clientless, protocol-agnostic remote
+desktop gateway* delivered as an HTML5 web application *(documented —
+[introduction](https://guacamole.apache.org/doc/gug/introduction.html))*. A user
+opens a browser, authenticates to the Guacamole web application, and is connected
+to a remote machine over VNC, RDP, SSH, telnet, or Kubernetes — with no client
+software installed beyond the browser. The system has two moving parts that must
+be modeled separately because they live at very different trust levels and are
+written in different languages: **guacd**, a C daemon (`apache/guacamole-server`)
+that actually speaks the remote-desktop protocols and translates them to and from
+the text-based *Guacamole protocol*; and the **web application**, a Java servlet
+plus JavaScript/HTML5 front-end (`apache/guacamole-client`) that authenticates
+users, brokers connections, and tunnels the Guacamole protocol between the browser
+and guacd *(documented —
+[architecture](https://guacamole.apache.org/doc/gug/guacamole-architecture.html))*.
+The third repository, `apache/guacamole-website`, is the static project website and
+carries essentially no runtime trust surface (see §2/§3).
+
+---
+
+## §2 Scope and intended use
+
+**Primary intended use.** A centrally-administered gateway that lets authenticated
+browser users reach remote desktops/shells they are authorized for, over a
+firewall-friendly HTTPS front door, without installing native RDP/VNC/SSH clients
+*(documented — introduction)*. Guacamole is a *deployed network service operated
+by an administrator*, not an in-process library.
+
+**Caller roles.** There is no single "caller"; the roles split as follows and each
+gets separate treatment in §6/§7:
+
+- **Browser user (client)** — authenticated to the web app, otherwise untrusted.
+  Interacts only through the tunnel and the web UI *(inferred — I1)*.
+- **Unauthenticated web attacker** — can reach the web app's HTTP surface but has
+  no valid session *(inferred — I1)*.
+- **Operator / administrator** — deploys guacd and the web app, writes
+  `guacamole.properties`, configures extensions and connection definitions.
+  Trusted for the instance *(documented — configuring-guacamole)*.
+- **Remote desktop server (peer)** — the VNC/RDP/SSH/telnet/Kubernetes endpoint
+  guacd connects out to. **Authenticated to guacd by address only, and adversarial
+  in this model** — its protocol responses are parsed by C code (see §7)
+  *(inferred — I2)*.
+- **The web app, as seen by guacd** — guacd treats whatever connects to its socket
+  and speaks the Guacamole protocol as authoritative (see §4, §9) *(inferred —
+  I3)*.
+
+**Component-family table.** The single most important orienting fact is the
+**guacd (C) vs. web-app (Java) vs. website (static)** split:
+
+| Family | Repo | Language | Representative entry point | Touches outside the process? | In model? |
+| --- | --- | --- | --- | --- | --- |
+| **guacd core + libguac** | guacamole-server | C | TCP `:4822`, Guacamole-protocol parser, plugin loader | Yes — outbound TCP to remote servers; loads `.so` plugins | **Yes** — primary memory-safety surface |
+| **Protocol plugins** (VNC/RDP/SSH/telnet/Kubernetes) | guacamole-server | C | per-protocol client `.so`; parses remote-server wire data, images, audio, clipboard | Yes — network sockets to the remote peer | **Yes** — the untrusted-remote-input surface |
+| **Web application / servlet** | guacamole-client | Java | HTTP endpoints, the tunnel servlet, REST API | Yes — HTTP in, TCP out to guacd | **Yes** |
+| **Auth / extension framework** | guacamole-client | Java | LDAP, SAML, CAS, OpenID, DB auth (MySQL/Postgres/SQL Server/MariaDB), TOTP, Duo, RADIUS, header/JSON auth | Yes — LDAP/HTTP/DB to identity sources | **Yes** — authn/authz surface |
+| **JavaScript / HTML5 front-end** | guacamole-client | JS/HTML | browser DOM, canvas renderer, tunnel client | Runs in the user's browser | **Yes** — client-side (XSS) surface |
+| **Project website** | guacamole-website | static HTML/Jekyll | published pages | No runtime; build-time only | **Out of model for runtime findings** — see §3 |
+
+Anything marked out of model here reappears in §3 with the reason.
+
+---
+
+## §3 Out of scope (explicit non-goals)
+
+- **The project website (`apache/guacamole-website`) as a runtime target.** It is a
+  static, published site with no server-side execution, no authentication surface,
+  and no user data. It is *in scope only for discoverability* — i.e. it is where
+  our published security policy and manual live, and a scan should be able to find
+  those — but it has no runtime attack surface, so runtime memory-safety / injection
+  / authz findings against it are `OUT-OF-MODEL: unsupported-component`
+  *(inferred — I4)*.
+- **The security of the remote desktops themselves.** Guacamole is a gateway; the
+  hardening, patch level, and credentials of the VNC/RDP/SSH/telnet/Kubernetes
+  targets are the operator's concern, not ours *(inferred — I5)*.
+- **The operator / administrator as an adversary.** An operator who can edit
+  `guacamole.properties`, define connections, or run code on the guacd or web-app
+  hosts already controls the instance; we do not model them as an attacker
+  *(inferred — I6)*.
+- **guacd exposed directly to untrusted networks.** guacd is designed to sit behind
+  the web app on a trusted network segment; a deployment that publishes `:4822` to
+  the Internet is a misconfiguration, not a scenario we defend (see §9, §10)
+  *(inferred — I3)*.
+- **Third-party dependency internals.** libVNCserver, FreeRDP, libssh2, Cairo,
+  etc., and the front-end's bundled JS libraries are modeled only at the boundary
+  where guacd/the web app hands them data; CVEs internal to those projects are
+  triaged upstream *(inferred — I7)*.
+- **Test code, examples, and build tooling** across all three repos are separately
+  authored and not covered by the core guarantees *(inferred — I8)*.
+
+---
+
+## §4 Trust boundaries and data flow
+
+Guacamole is a service, so it has several distinct trust boundaries rather than one
+API surface. The end-to-end path is:
+
+```
+browser  ──HTTPS/tunnel──▶  web app (Java)  ──TCP :4822──▶  guacd (C)  ──RDP/VNC/SSH──▶  remote server
+   ▲                                                                                          │
+   └───────────── Guacamole protocol (rendered images/audio/clipboard) ◀──────────────────────┘
+```
+
+Trust transitions, each of which a finding must cross to be in-model:
+
+1. **Browser → web app (the HTTP boundary).** The primary authentication and
+   session boundary. Everything arriving here is untrusted until a session is
+   established and authorized *(inferred — I1)*.
+2. **Web app → guacd (the socket boundary).** The web app connects to guacd on TCP
+   4822 and sends connection instructions plus the Guacamole protocol stream. **By
+   default this link is unauthenticated and unencrypted**; `guacd-ssl` and network
+   placement are the only controls *(documented — configuring-guacamole)*. guacd
+   trusts this peer (see §9).
+3. **guacd → remote server (the outbound protocol boundary — CRITICAL).** guacd's C
+   protocol plugins open a connection to the remote VNC/RDP/SSH/etc. endpoint. **The
+   remote server's responses — framebuffer/image data, audio, clipboard, virtual
+   channels, terminal/console escape sequences — flow back into C parsing code and
+   then, as Guacamole-protocol instructions, all the way back to the browser.** This
+   is the highest-value boundary in the whole system and the source of the
+   project's memory-safety CVE history (see §7, §11a) *(documented — the security
+   page lists VNC/RDP/SSH memory-corruption CVEs originating from remote-server or
+   session data; inferred as to it being the load-bearing boundary — I2)*.
+4. **guacd/web app → identity & storage backends.** LDAP/SAML/DB/RADIUS lookups
+   during authentication, and connection-credential storage/retrieval *(documented
+   — introduction lists the extensions; inferred as to trust treatment — I9)*.
+
+**Reachability precondition per family.** A finding in a guacd protocol plugin is
+in-model only if it is reachable from data the **remote server or the connected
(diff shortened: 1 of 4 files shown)


THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+- **The web app, as seen by guacd** — guacd treats whatever connects to its socket
+  and speaks the Guacamole protocol as authoritative (see §4, §9) *(inferred —
+  I3)*.
+
+**Component-family table.** The single most important orienting fact is the
+**guacd (C) vs. web-app (Java) vs. website (static)** split:
+
+| Family | Repo | Language | Representative entry point | Touches outside the process? | In model? |
+| --- | --- | --- | --- | --- | --- |
+| **guacd core + libguac** | guacamole-server | C | TCP `:4822`, Guacamole-protocol parser, plugin loader | Yes — outbound TCP to remote servers; loads `.so` plugins | **Yes** — primary memory-safety surface |
```
corentin-soriano: Should we specify the incoming connection from guacamole-client?

necouchman: I'm not sure it should call out the inbound connection specifically from guacamole-client, since we do advertise/support it as a more generic Guacamole Protocol proxy that can support custom front-ends, but inbound from _a_ client is probably worth noting.

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+**What Guacamole is.** Apache Guacamole is a *clientless, protocol-agnostic remote
+desktop gateway* delivered as an HTML5 web application *(documented —
+[introduction](https://guacamole.apache.org/doc/gug/introduction.html))*. A user
+opens a browser, authenticates to the Guacamole web application, and is connected
+to a remote machine over VNC, RDP, SSH, telnet, or Kubernetes — with no client
+software installed beyond the browser. The system has two moving parts that must
+be modeled separately because they live at very different trust levels and are
+written in different languages: **guacd**, a C daemon (`apache/guacamole-server`)
+that actually speaks the remote-desktop protocols and translates them to and from
+the text-based *Guacamole protocol*; and the **web application**, a Java servlet
```
necouchman: Is `text-based` the best wording, here? I ask because, while we do deliver most of the protocol over text, we also deliver images??

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+
+---
+
+## §2 Scope and intended use
+
+**Primary intended use.** A centrally-administered gateway that lets authenticated
+browser users reach remote desktops/shells they are authorized for, over a
+firewall-friendly HTTPS front door, without installing native RDP/VNC/SSH clients
+*(documented — introduction)*. Guacamole is a *deployed network service operated
+by an administrator*, not an in-process library.
```
necouchman, line 88: While this captures the overall scope of the project, I wonder how this statement applies to folks who use the `libguac` portion of the guacamole-server project, or the `guacamole-common` and `guacamole-common-js` portions of the guacamole-client project? How does the fact that we do provide those as libraries that can be used by third party projects relate to our documentation of the thread model?

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+- **The security of the remote desktops themselves.** Guacamole is a gateway; the
+  hardening, patch level, and credentials of the VNC/RDP/SSH/telnet/Kubernetes
+  targets are the operator's concern, not ours *(inferred — I5)*.
+- **The operator / administrator as an adversary.** An operator who can edit
+  `guacamole.properties`, define connections, or run code on the guacd or web-app
+  hosts already controls the instance; we do not model them as an attacker
+  *(inferred — I6)*.
+- **guacd exposed directly to untrusted networks.** guacd is designed to sit behind
+  the web app on a trusted network segment; a deployment that publishes `:4822` to
+  the Internet is a misconfiguration, not a scenario we defend (see §9, §10)
```
necouchman: I would replace `the Internet` with `any un-trusted network` or something similar - while the Internet may be the ultimate example of an un-trusted network, it is certainly not the only example, and publishing guacd port 4822 to any users or systems that should not have direct proxy access through guacd to your remote systems is a violation of the design principles of the software.

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+
+## §4 Trust boundaries and data flow
+
+Guacamole is a service, so it has several distinct trust boundaries rather than one
+API surface. The end-to-end path is:
+
+```
+browser  ──HTTPS/tunnel──▶  web app (Java)  ──TCP :4822──▶  guacd (C)  ──RDP/VNC/SSH──▶  remote server
+   ▲                                                                                          │
+   └───────────── Guacamole protocol (rendered images/audio/clipboard) ◀──────────────────────┘
```
necouchman: This is not quite accurate - I don't know how much it matters, but, taken at face value, this states that the remote server will speak Guacamole protocol directly to the user's browser, which is not how it works. I'm not entirely sure how to diagram this best in a Github comment, but:
* All traffic from the user's browser passes through the HTTP(S)/WS(S) tunnel back to the Java application, which handles both API calls used for authentication, authorization, and auditing, but also handles the establishment and maintenance of the HTTP(S)/WS(S) tunnel back to guacd. The only traffic that the user's browser will ever see will be between it and the Web Application host(s).
* The traffic to/from the user's browser, tunneled over HTTP(S)/WS(S) by the web application will land at the guacd application, which does all of the translation between the Guacamole protocol and the remote protocols.
* The remote server will only ever see traffic from and send traffic to the guacd instance - there is no direct path from the remote server back to the user's browser. That traffic traverses the same path back - from remote server to guacd, then from guacd back through the tunnel maintenance by the web application, where it is then delivered to the user's browser.

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+   remote server's responses — framebuffer/image data, audio, clipboard, virtual
+   channels, terminal/console escape sequences — flow back into C parsing code and
+   then, as Guacamole-protocol instructions, all the way back to the browser.** This
+   is the highest-value boundary in the whole system and the source of the
+   project's memory-safety CVE history (see §7, §11a) *(documented — the security
+   page lists VNC/RDP/SSH memory-corruption CVEs originating from remote-server or
+   session data; inferred as to it being the load-bearing boundary — I2)*.
+4. **guacd/web app → identity & storage backends.** LDAP/SAML/DB/RADIUS lookups
+   during authentication, and connection-credential storage/retrieval *(documented
+   — introduction lists the extensions; inferred as to trust treatment — I9)*.
```
necouchman: Slight clarification, here - guacd does not store anything outside of memory. The only credential lookups and storage are done by the web application, which 1) authenticates the users, 2) provides authorization for connections, 3) documents the connections, 4) stores and looks up connection details, potentially including credentials, 5) requests additional credentials from the user, when required, and then 6) delivers the connection information to guacd. guacd only "stores" any of this information in-memory, and only for the duration of the connection, after which it is discarded.

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+
+**Reachability precondition per family.** A finding in a guacd protocol plugin is
+in-model only if it is reachable from data the **remote server or the connected
+session** controls (framebuffer, audio, clipboard, channel, console bytes). A
+finding in libguac's Guacamole-protocol parser is in-model if reachable from the
+bytes crossing boundary 2. A finding in the web app is in-model if reachable from
+an **unauthenticated** request or from an **authenticated but unprivileged** user's
+input. A finding reachable only from `guacamole.properties`, a connection
```
necouchman: I'm wondering if this is complete enough? Does `authenticated but unprivileged` also cover situations where the user is 1) authenticated, 2) allowed to access a resource, but 3) able to do something unintended. For example, user is authorized to log in, allowed to access a connection, but is supposed to be blocked from downloading data from the remote server, but finds a way to do that?

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+- **guacd host.** A POSIX/Linux host with a conformant C toolchain and allocator;
+  guacd runs as a long-lived daemon, typically one process forking a child per
+  connection *(inferred — I11)*.
+- **Network placement.** guacd and the web app are co-located or on the same
+  trusted network segment; guacd's `:4822` is **not** reachable by untrusted
+  parties *(inferred — I3, see §10)*.
+- **Web-app host.** A standard servlet container (e.g. Tomcat) fronted by a
+  reverse proxy or the container itself terminating TLS *(inferred — I12)*.
+- **TLS termination.** HTTPS for the browser boundary is provided by the operator's
+  container/reverse proxy, not by Guacamole itself *(inferred — I12)*.
```
necouchman: It may be worth a little bit more direct language, here, to explicitly state that the TLS termination is the responsibility of the operator/administrator, and that "vulnerabilities" arising from failure to properly secure (TLS terminate) the web application may be considered invalid?

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+  multi-user and multi-session concurrently *(inferred — I11)*.
+- **What we do *not* do to the host (negative inventory — high-priority to
+  confirm).** We believe guacd does not read ambient environment variables as a
+  trust input, does not phone home, and confines its outbound connections to the
+  operator-defined targets; the web app does not execute operator-supplied code
+  beyond the extensions an operator installs *(inferred — I13)*.
+
+---
+
+## §5a Build-time and configuration variants
```
necouchman: I'm not sure if `Build-time` is standard language, or might be a bit misleading, here? I think of "build" as compiling the software, versus "deploy" - all of the items below are, as mentioned, part of the deployment.

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+| Web app: file browser / SFTP names | filenames rendered in UI | **yes** — via remote session content | output encoding (cf. CVE-2016-1566 stored XSS) *(documented)* |
+| Web app → guacd | connection instructions (protocol, hostname, port, credentials) | **no** — constructed by the trusted web app from operator-defined connections | `guacd-ssl` + network isolation of `:4822` *(inferred — I3)* |
+| libguac | Guacamole-protocol element lengths/opcodes on the socket | at boundary 2, **yes** if the socket is exposed | length/element validation (cf. CVE-2023-30575, CVE-2012-4415) *(documented)* |
+| guacd RDP plugin | virtual-channel PDUs, audio-input buffers, static channels | **yes** — malicious/compromised RDP server | bounds/UAF safety in C (cf. CVE-2020-9497/9498, CVE-2023-30576) *(documented)* |
+| guacd VNC plugin | framebuffer/image geometry & pixel data | **yes** — malicious VNC server | integer-overflow safety on image buffers (cf. CVE-2023-43826) *(documented)* |
+| guacd SSH/telnet plugin | terminal/console escape sequences | **yes** — remote shell / session content | console-code validation (cf. CVE-2024-35164) *(documented)* |
```
necouchman, line 293: Should probably include Kubernetes in this list, too?

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+
+## §9 Security properties the project does *not* provide
+
+- **guacd is not an authentication or authorization boundary.** guacd trusts
+  whatever connects to its socket and speaks the Guacamole protocol; it executes
+  the connection instructions it is given (which protocol, which host, which
+  credentials). Authorization is entirely the web app's job. A report that "guacd
+  connected to an arbitrary host / used given credentials" is by design — the web
+  app is responsible for only ever sending authorized instructions *(inferred —
+  I3)*.
```
necouchman, line 398: I don't know if it's also worth specifically calling out, here, that guacd also puts no restrictions on what valid connection parameters can be used by anyone who is accessing it - that is, it isn't just that users can connect, if they have access to guacd, but they can perform any operation that any of the remote protocols supports - file transfers, clipboard data transfers, etc.

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+answer** for the PMC to confirm, correct, or strike, and the section it lands in.
+Grouped into waves.
+
+### Wave 1 — scope, the guacd trust posture, and insecure defaults (highest leverage)
+
+1. **(I3) guacd trusts its connecting web app and is not an auth boundary; it must
+   be kept off untrusted networks.** Proposed: correct — guacd executes the
+   Guacamole-protocol instructions it receives, authorization lives in the web app,
+   and exposing `:4822` is a misconfiguration, not an in-model attack. *Lands in
+   §4/§7/§9/§10/§11a.*
```
necouchman: This looks correct to me.

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+2. **(I2) The primary modeled guacd adversary is a malicious/compromised remote
+   desktop server whose data reaches C parsers; remote-data-driven memory
+   corruption is security-critical.** Proposed: correct — consistent with the CVE
+   history. Confirm this is the intended framing and that clipboard/audio/printer
+   channels are included. *Lands in §6/§7/§8.3/§9.*
```
necouchman: Agreed.

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+3. **(I14) Is plaintext, unauthenticated webapp↔guacd (the `guacd-ssl`=off /
+   `guacd-hostname`=localhost default) the *supported production posture on a
+   trusted segment*, making a report against cleartext `:4822`
+   `OUT-OF-MODEL: non-default-build`?** Or should such a report be `VALID`?
+   Proposed: supported-on-trusted-segment; `guacd-ssl` required once the link is
+   not fully trusted. *Lands in §5a/§9/§13.*
```
necouchman: I believe this should be `OUT-OF-MODEL: <something>` - it is the default configuration, and I believe our circumstantial experience is that _most_ people tend to run guacd on the same host as the client application, whether natively or in containers, so it tends to be safe for most people. In any case, it is a deliberate design and clearly up to the operator/administrator to configure if they are uncomfortable with the plaintext nature of the default. This also follows with the premise that it is up to the operator/administrator to secure/TLS-terminate the web application.

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+4. **(I4/I8) `guacamole-website` and all test/example/build code are out of model
+   for runtime findings (in scope only for discoverability of the security policy
+   and docs).** Proposed: correct. *Lands in §2/§3/§11a.*
```
necouchman: Correct.

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+5. **(I5/I6) The remote desktops, their credentials, and the operator/admin are out
+   of the adversary model.** Proposed: correct. *Lands in §3/§7.*
```
necouchman: Agreed.

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+
+### Wave 2 — web-app authn/authz surface
+
+6. **(I1) The modeled web adversaries are the unauthenticated web attacker and the
+   authenticated-but-unprivileged user; the claimed properties are user
+   authentication (§8.1) and per-user connection/tunnel/history isolation (§8.2).**
+   Proposed: correct — consistent with CVE-2021-43999/41767/2020-11997. *Lands in
+   §6/§7/§8.*
```
necouchman: Not sure if anything should be added to this that deals with connection parameters - being able to bypass security guards that the parameters are meant to inhibit (for example, being able to transfer files if file transfer is disabled)?

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+7. **(I15) With no auth extension installed there is no production authentication;
+   operators must install one (and MFA for sensitive deployments), and a no-auth
+   install is `OUT-OF-MODEL: non-default-build`.** Proposed: correct. *Lands in
+   §5a/§8.1/§9/§10.*
```
necouchman: Agreed.

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+8. **(I16) Output encoding of session/remote-derived strings (§8.5) is a claimed
+   property (CVE-2016-1566), but we make no blanket CSRF/clickjacking guarantee
+   beyond framework mechanisms.** Proposed: correct. *Lands in §8/§9.*
```
necouchman: Agreed.

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+9. **(I9) Identity/storage backends (LDAP/SAML/DB/RADIUS) are trusted infrastructure
+   configured by the operator; compromise of those backends is out of model.**
+   Proposed: correct. *Lands in §4/§6.*
```
necouchman: Agreed.

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+
+### Wave 3 — environment, transport, and resource guarantees
+
+10. **(I12) TLS termination and secure-cookie handling for the browser boundary are
+    the operator's responsibility (cf. CVE-2018-1340); absent TLS, §8.6 does not
+    hold.** Proposed: correct. *Lands in §5/§8.6/§10.*
```
necouchman: Agreed

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+11. **(I18) `guacd-ssl` provides transport encryption but not, by itself, mutual
+    authentication of the web app to guacd (false friend).** Proposed: correct —
+    confirm whether any mutual-auth mechanism exists on this link. *Lands in §9.*
```
necouchman: This is correct - there is currently no mutual authentication implemented native to the web app <-> guacd components. It can be implemented with, for example, SSH or SSL tunneling configured with additional components, but that is out of scope of this model.

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+12. **(I11/I13) guacd runs as a forking POSIX daemon, does not treat environment
+    variables as a trust input, does not phone home, and confines outbound
+    connections to operator-defined targets; guacd and the web app are expected to
+    run matching patched versions.** Proposed: correct — this negative-side-effects
+    inventory is the one most needing maintainer confirmation. *Lands in §5.*
```
necouchman: Correct...mostly...I think. We do _support_ running different versions of web application and guacd, and have built in some mechanisms to handle negotiation of features between those versions; however, this is strictly designed to handle feature differences, and no guarantee is provided that doing so will be "safe".

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+13. **(I17) Resource guarantees: is there *any* categorical line on DoS from a
+    hostile remote stream or a client opening many tunnels — e.g. "a crash/UAF is a
+    bug, but sustained CPU/bandwidth from a connected server is not"?** Proposed: no
+    categorical resource guarantee at this layer; memory-safety failures remain
+    bugs. *Lands in §8/§9/§11a.*
```
necouchman: I do not believe there are any resource guarantees at this time - it is up to the operator to monitor and throttle resource utilization.

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+14. **(I7/I10) Third-party protocol libraries (FreeRDP/libVNCserver/libssh2/Cairo,
+    bundled JS) are modeled only at the boundary where we hand them data; internal
+    CVEs are triaged upstream. Reachability preconditions per §4 are the triage
+    test.** Proposed: correct. *Lands in §3/§4.*
```
necouchman: Agreed.

THREAT_MODEL.md
```diff
@@ -0,0 +1,581 @@
...
+
+### Wave 4 — meta
+
+15. **Document ownership & coexistence.** There is currently **no `SECURITY.md` in
+    either code repository** — the sole published security artifact is the website
+    security page. Proposed: this threat model becomes the canonical model that the
+    security page (and new `SECURITY.md` files in both repos) link to. Confirm the
+    venue and whether the PMC wants the machine-readable §15 sidecar. *Meta — no
+    body claim.*
```
necouchman: I'm okay with the proposed solution, but ultimately will leave it to @mike-jumper to decide on that.

necouchman (commented):
Okay, several comments from me - apologies in advance if any of these are nonsensical.

necouchman (commented):
Okay, several comments from me - apologies in advance if any of these are nonsensical.

potiuk:
@necouchman thanks — this was a genuinely useful review, and several of your points corrected the draft rather than just confirming it. All folded in as of `1652ba4`. @corentin-soriano your inbound-connection point is in there too.

**The §4 diagram was wrong** and you were right to call it out. As drawn it implied the remote server speaks the Guacamole protocol straight to the browser. Redrawn so every byte in both directions passes through the web app, with the return path (`remote → guacd → web app → browser`) spelled out and the REST API's role noted.

Other corrections:

- **guacd persists nothing outside memory** — boundary 4 said "guacd/web app" for credential storage. Now the six-step web-app sequence you described, with guacd holding parameters only for the life of the connection.
- **guacd puts no restriction on connection parameters** — so reaching guacd grants any capability the protocol supports (file transfer, clipboard, printing), not just session access. That's now explicit in §9.
- **"any untrusted network"**, not "the Internet", in both §3 and §10.
- **The trust-boundary precondition was incomplete.** Your file-transfer-disabled example is exactly the gap: an authenticated *and authorized* user exceeding what the authorization meant to permit. Connection parameters are themselves a guard, and bypassing one is in-model (§4, §8.2).
- **libguac / guacamole-common / guacamole-common-js as libraries**, and guacd as a supported generic protocol proxy — §2 said "not an in-process library" flatly, which ignored that audience.
- Protocol is text-based **but carries binary payloads**; **Kubernetes** added to the §6 ingress table; **TLS termination** now says a report about an un-terminated deployment is `DOWNSTREAM`; **§5a renamed** "Build-time" → "Deployment".
- **Version mismatch is supported** with feature negotiation — the correction being that no safety guarantee comes with it.

Rulings recorded as the PMC's: **I14** cleartext `:4822` on a trusted segment is `OUT-OF-MODEL` (both "pending the I14 ruling" placeholders resolved); **I18** no native mutual auth on the web app ↔ guacd link; **I17** no resource guarantees.

Left open: the document-ownership/venue question, since you're content with the proposal but defer to @mike-jumper — and the §15 sidecar with it.

PR #1225, opened 2026-07-06
GUACAMOLE-2296: Broadcast guacLoginUpdated event on token renewal

arulrajnet:
When a server re-issues the same authentication token (token renewal/extension scenario), emit the new 'guacLoginUpdated' event to notify listeners that an existing session has been renewed. This distinguishes token renewal from new login events and enables downstream applications to detect session re-authentication.

Changes:
- Add guacLoginUpdated broadcast in authenticate() when token remains unchanged
- Document new event in service javadoc
- Update comment to reflect event emission behavior

### DIFF
--- a/guacamole/src/main/frontend/src/app/auth/service/authenticationService.js
+++ b/guacamole/src/main/frontend/src/app/auth/service/authenticationService.js
@@ -36,6 +36,12 @@
  *      Authentication was successful and a new token was created. This event
  *      receives the authentication token as its sole parameter.
  *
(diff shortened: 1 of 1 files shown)


necouchman:
@arulrajnet I'm not sure I understand the point of just adding this definition - it doesn't look like it actually gets used or consumed anywhere?? The other events that are broadcast - for example, `guacLoginPending` - have matching `$rootScope.on()` calls that then actually do something when the event is broadcast. Do you have some follow-on changes that will make use of this?

PR #1226, opened 2026-07-10
GUACAMOLE-2132: Update dependencies to latest compatible versions.

mike-jumper:
Another set of updates, presumably the last before release. I've omitted the dependency update tooling that I've been working on for now (in the interest of time) - it needs more work.


mike-jumper:
We'll have to do regression tests, of course, but I used a binary Java linkage checking tool to at least verify that these updates don't break ABI expectations.

PR #1231, opened 2026-07-19
GUACAMOLE-2302: Add test coverage for TokenFilter parameter token grammar

ciroiriarte:
Adds characterization tests for the parameter token grammar accepted by `TokenFilter`.

`TokenFilterTest` covers basic substitution, escaping, and the `LOWER`/`UPPER` modifiers. What isn't covered is the boundary of the grammar itself:

- which candidate token names the pattern accepts (currently `[A-Za-z0-9_]*`)
- how an unrecognized modifier is handled (currently falls through to `default:` and appends the value unchanged)
- how strict mode treats text that resembles a token but is not one (currently not matched at all, so passed through literally rather than reported undefined)

That last point is load-bearing. `filterStrict()` has a single caller in the tree — `VaultUserContext.getTokens()` — which resolves secret names from `vault-token-mapping.yml` and treats `GuacamoleTokenUndefinedException` as a signal to skip that secret:

```java
catch (GuacamoleTokenUndefinedException e) {
    logger.debug("Secret for token \"{}\" will not be retrieved. ...");
    continue;
}
```

So a change in which strings the token pattern matches can change whether a vault secret is retrieved at all, with only a debug-level log. These tests pin down the current behavior so that any future change to the pattern is evaluated against it rather than discovered later.

No production code is changed — this is test coverage only.

The tests also record two existing behaviors that are easy to alter unintentionally:

- `$${TOKEN}` escaping, including in combination with modifiers
- directly adjacent tokens (`${A}${B}`) substitute only the first, because the pattern matches one character ahead of each token to detect escaping and that character is consumed by the preceding match

For context, this came out of reviewing #1216, which proposes widening the token pattern. The tests are independent of that PR and apply to `main` as it stands.

### DIFF
--- a/guacamole-ext/src/test/java/org/apache/guacamole/token/TokenFilterGrammarTest.java
+++ b/guacamole-ext/src/test/java/org/apache/guacamole/token/TokenFilterGrammarTest.java
@@ -1,0 +1,256 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.guacamole.token;
+
+import org.apache.guacamole.GuacamoleException;
+import org.junit.Test;
+import static org.junit.Assert.*;
+
+/**
+ * Tests which characterize the existing parameter token grammar accepted by
+ * TokenFilter. These tests are intended to pin down current behavior so that
+ * any change to the token pattern can be evaluated for backwards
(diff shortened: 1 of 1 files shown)


necouchman (approved):

ciroiriarte:
Gentle nudge on this one — it's approved and test-only (adds CUnit coverage for the `TokenFilter` parameter-token grammar, no production changes), so it should be low-risk to land. Could a committer merge it when convenient? Happy to rebase if it's drifted from `main`.

necouchman:
@ciroiriarte Yeah, I was about to merge this one when Github had some issues the other day that caused the builds to fail. Unfortunately this build never recovered, so I'm going to have to manually build it just to make sure it builds prior to merging.

necouchman:
FYI - I've run the manual build without issues, so proceeding with the merge.

adb014:
I can confirm that using the TokenFilter 

```diff
      * entire token itself.
      */
-    private final Pattern tokenPattern = Pattern.compile("(.*?)(^|.)(\\$\\{([A-Za-z0-9_]*)(\\:(.*))?\\})");
+    private final Pattern tokenPattern = Pattern.compile("(.*?)(^|.)(\\$\\{([A-Za-z0-9_]*|[a-z]+:\\/(?:\\/[a-zA-Z0-9_-]+)+(?:\\?(?:[^=&?#]+=[^&#:]*(?:&[^=&?#]+=[^&#:]*)*)?)?)(\\:([^:]*))?\\})");
 
     /**
```

that adds a second token format that accepts tokens like

```
vault://ssh/guacamole/server1/sign/public?key_type=ec&key_bits=521
```

instead of my originally proposed TokenFilter both lets all of these tests pass and allows the Hashicorp extension in #1216 to run correctly... I'll rebase and update my pull request to both use and pass these tests

PR #1232, opened 2026-07-20
GUACAMOLE-2305: Do not reflect sensitive parameter values streamed via argv.

ciroiriarte:
Resolves [GUACAMOLE-2305](https://issues.apache.org/jira/browse/GUACAMOLE-2305).

When the server streams a connection parameter back to the client through an `argv`
stream, `ManagedClient` reflects the received value into the UI-visible arguments
model, which backs the connection parameters panel.

There is no filtering by parameter name, so when the streamed parameter holds a secret
— a password, a passphrase, a private key — the value becomes readable by any user
able to view that connection. This matters most where a connection is configured
centrally with credentials the user is not otherwise meant to see.

This is a display concern only. Mutating an immutable argument toward guacd is already
rejected server-side, so the issue is what a user can *read*, not what they can send.

**The change:** skip reflection for parameter names matching
`password|passphrase|private-key|client-key|secret|token`, case-insensitive. Public
material (`host-key`, `public-key`, `pubkey`) is deliberately not matched, since it is
not secret and is useful to display. One file, 20 lines added.

**Two things I'd rather have decided by someone with more context**, both noted on the
issue:

1. Is the client the right layer? Filtering here is defence-in-depth. If guacd should
   not be streaming sensitive values back at all, the durable fix belongs there. I
   don't know whether any protocol relies on streaming these back.
2. Name matching is a heuristic. `token` could match something that isn't secret, and a
   protocol could name a sensitive parameter something the pattern misses. A
   per-protocol flag in the protocol descriptor would be precise, at the cost of
   touching every protocol definition. Happy to do that instead if preferred.

Found while auditing a SPICE protocol implementation built on Guacamole
(GUACAMOLE-261); the issue is not SPICE-specific.

### DIFF
--- a/guacamole/src/main/frontend/src/app/client/types/ManagedClient.js
+++ b/guacamole/src/main/frontend/src/app/client/types/ManagedClient.js
@@ -63,6 +63,19 @@ angular.module('client').factory('ManagedClient', ['$rootScope', '$injector',
     var THUMBNAIL_UPDATE_FREQUENCY = 5000;
 
     /**
+     * Pattern matching connection parameter names whose values are sensitive
+     * (passwords, passphrases, private keys) and must therefore never be
+     * reflected into the UI-visible arguments model when streamed back by the
+     * server via "argv". Mutating such a parameter toward guacd is already
+     * rejected server-side; this keeps the value out of the connection-
+     * parameters panel as well (security finding L4, guacamole-server#21).
+     *
+     * @constant
+     * @type {!RegExp}
+     */
+    var SENSITIVE_ARGUMENT = /password|passphrase|private-key|client-key|secret|token/i;
+
+    /**
      * A deferred pipe stream, that has yet to be consumed, as well as all
      * axuilary information needed to pull data from the stream.
      *
@@ -645,6 +658,13 @@ angular.module('client').factory('ManagedClient', ['$rootScope', '$injector',
             // Test mutability once stream is finished, storing the current
             // value for the argument only if it is mutable
             reader.onend = function textComplete() {
+
+                // Never reflect a sensitive parameter value (password,
+                // passphrase, private key) streamed back via "argv" into the
+                // UI-visible arguments model (security finding L4).
+                if (SENSITIVE_ARGUMENT.test(name))
+                    return;
+
                 ManagedArgument.getInstance(managedClient, name, value).then(function argumentIsMutable(argument) {
                     managedClient.arguments[name] = argument;
                 }, function ignoreImmutableArguments() {});


necouchman:
> Resolves [GUACAMOLE-2305](https://issues.apache.org/jira/browse/GUACAMOLE-2305).
> 
> **Two things I'd rather have decided by someone with more context**, both noted on the issue:
> 
>     1. Is the client the right layer? Filtering here is defence-in-depth. If guacd should
>        not be streaming sensitive values back at all, the durable fix belongs there. I
>        don't know whether any protocol relies on streaming these back.
> 

If parameters are configured centrally (e.g. a connection is defined with a username and password), then I do not think there's any reason that the password should be streamed back to the client, as the client should not be able to modify this value, so providing it as part of argv should not happened.

I actually can't think of a situation where you'd want the password to be sent back - even in the case of a user providing credentials at connection time, the server rejecting those credentials, and the user being asked to send correct credentials, there's no point in filling in the original credentials.

I can't really think of a situation where you'd want the credentials - pre-configured or not - to be sent back from guacd to the client??

>     2. Name matching is a heuristic. `token` could match something that isn't secret, and a
>        protocol could name a sensitive parameter something the pattern misses. A
>        per-protocol flag in the protocol descriptor would be precise, at the cost of
>        touching every protocol definition. Happy to do that instead if preferred.
> 

Yeah, going strictly off the name seems like a bad band-aid to an underlying problem of making sure that we are marking certain credentials as sensitive. We kind of do this to some degree on the client-side, already, as password fields are marked as `password` rather than `text` fields, so it seems like maybe it's better to identify the parameters ahead of time that we expect to be sensitive?

But, then again, if we shouldn't ever be sending passwords, passphrases, etc., back from guacd to the client, maybe this doesn't matter?

ciroiriarte:
Thanks @necouchman — I think you've identified the right layering here, and I mostly agree.

**On the durable fix being server-side:** you're right that if guacd shouldn't be sending these values back, that's where the fix belongs. Worth noting what that takes today: argv registration only has `GUAC_ARGV_OPTION_ONCE` and `GUAC_ARGV_OPTION_ECHO` — there's no concept of a sensitive or non-echoed parameter. So a proper server-side fix means adding that to libguac (a sensitive/no-reflect flag, or just never streaming back a parameter that came from server-side config) and then adopting it in each protocol. That's the right fix, but it's a cross-cutting libguac + per-protocol change rather than a quick one.

**On "the client already marks password fields as `password`":** it does, and since `ManagedClient` already has the connection `forms`, it could filter on `Field.Type.PASSWORD` instead of a name regex — cleaner, as you say. The catch is coverage: `Field.Type` is `TEXT/EMAIL/USERNAME/PASSWORD/NUMERIC/…`, with no type for keys or secrets, so a private key or its passphrase renders as multiline `TEXT`, not `PASSWORD`. Filtering purely on field type would quietly leak exactly those, which the name pattern does catch today. The precise client-side option is the per-parameter marker from point 2 — a `sensitive` flag in the protocol descriptor — but that means editing every protocol's form definitions.

So there are really three layers, and I'd treat them as complementary defence-in-depth rather than either/or:
1. **Server-side no-echo** — the durable fix; biggest change (new libguac mechanism + per-protocol adoption).
2. **Client-side descriptor flag** — precise; medium change (per-protocol form edits).
3. **Client-side name regex** — this PR; immediate and cheap, though a heuristic by nature.

Since we agree guacd shouldn't be streaming these back, I've filed GUACAMOLE-2317 to track (1) as the real fix server-side. For this PR, I'll go with whatever you prefer on scope: merge (3) now as an immediate stopgap while GUACAMOLE-2317 is scoped separately, or close it in favour of GUACAMOLE-2317 if you'd rather not carry a name heuristic in the tree. Either's fine with me — I just didn't want to leave the value readable while the bigger fix is worked out.

mike-jumper:
Closing this per GUACAMOLE-2305 - there's no need to filter like this. The mechanism to avoid sending secrets is to not send them, which we already do, and there's no need to assume that no legitimate feature will ever need to send a secret to the user.

PR #1233, opened 2026-07-23
GUACAMOLE-808: Add page redirection on login

adb014:
This is quite an intrusive pull request that allows reauthentication to redirect back to the initial page so that in progress work can be directly refound.. See the discussion in the Jira for details

### DIFF
--- a/extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/AuthenticationProviderService.java
+++ b/extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/AuthenticationProviderService.java
@@ -33,6 +33,8 @@ import org.apache.guacamole.auth.openid.token.TokenValidationService;
 import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.auth.sso.NonceService;
 import org.apache.guacamole.auth.sso.SSOAuthenticationProviderService;
+import org.apache.guacamole.auth.sso.session.SSOAuthenticationSession;
+import org.apache.guacamole.auth.sso.session.SSOAuthenticationSessionManager;
 import org.apache.guacamole.auth.sso.user.SSOAuthenticatedUser;
 import org.apache.guacamole.form.Field;
 import org.apache.guacamole.form.RedirectField;
@@ -67,6 +69,12 @@ public class AuthenticationProviderService implements SSOAuthenticationProviderS
     private NonceService nonceService;
 
     /**
+     * Manager of active OpenID authentication attempts. Tracks redirection
+     */
+    @Inject
+    private SSOAuthenticationSessionManager sessionManager;
+
+    /**
      * Service for validating received ID tokens.
      */
     @Inject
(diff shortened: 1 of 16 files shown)


guacamole/src/main/java/org/apache/guacamole/GuacamoleSession.java
```diff
@@ -123,6 +129,32 @@ public void setAuthenticatedUser(AuthenticatedUser authenticatedUser) {
...
+     */
+    public String getRedirection() {
+        this.access();
+        return this.redirection;
+    }
+
+    /**
+     * Returns a redirection path for newly authenticated users, allowing
+     * redirection to where they left off when an authentication was
+     * requested.
```
necouchman, line 148: I think this is copy-pasta - this should be something like `Sets a redirection patch for newly-authenticated users...`

guacamole/src/main/java/org/apache/guacamole/GuacamoleSession.java
```diff
@@ -123,6 +129,32 @@ public void setAuthenticatedUser(AuthenticatedUser authenticatedUser) {
         this.authenticatedUser = authenticatedUser;
     }
 
+    /**
+     * Returns a redirection path for newly authenticated users, allowing
+     * redirection to where they left off when an authentication was
+     * requested.
+     *
+     * @return
+     *      A String with the redirection path
```
necouchman: ...`or null if no such path is set`?

guacamole/src/main/frontend/src/app/index/controllers/indexController.js
```diff
@@ -328,13 +328,7 @@ angular.module('index').controller('indexController', ['$scope', '$injector',
...
 
         $scope.reAuthenticating = true;
-
-        // Clear out URL state to conveniently bring user back to home screen
-        // upon relogin
-        if ($location.path() !== '/')
-            $location.url('/');
-        else
-            $route.reload();
+        $route.reload();
```
necouchman, line 331: Is this change going to adversely impact situations where SSO is not in use and where the behavior of the redirection back to the home screen actually would be expected? For example, configurations where JDBC is the only provider in use, and the re-authentication gets triggered?

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/AuthenticatedUser.java
```diff
@@ -75,4 +75,14 @@ public interface AuthenticatedUser extends Identifiable {
...
 
+    /**
+     * A function to set the originalUri value
+     */
+    public void setOriginalUri(String originalUri);
+
+    /**
+     * A function to get the originalUri value
+     */
+    public String getOriginalUri();
```
necouchman, line 92: This should be properly documented, with the `@return` JSDoc flag.

guacamole-ext/src/main/java/org/apache/guacamole/net/auth/AuthenticatedUser.java
```diff
@@ -75,4 +75,14 @@ public interface AuthenticatedUser extends Identifiable {
      */
     void invalidate();
 
+    /**
+     * A function to set the originalUri value
+     */
+    public void setOriginalUri(String originalUri);
```
necouchman, line 84: Please add the `@param` JSDoc flag for the `originalUri` parameter.

extensions/guacamole-auth-sso/modules/guacamole-auth-sso-openid/src/main/java/org/apache/guacamole/auth/openid/AuthenticationProviderService.java
```diff
@@ -121,12 +155,24 @@ public SSOAuthenticatedUser authenticateUser(Credentials credentials)
 
     @Override
     public URI getLoginURI() throws GuacamoleException {
+        return getLoginURI(null);
+    }
+
+    private URI getLoginURI(Credentials credentials) throws GuacamoleException {
```
necouchman, line 174: This method needs documentation.

necouchman (changes requested):

guacamole/src/main/frontend/src/app/index/controllers/indexController.js
adb014: Why would you expect the behavior to be different with or without SSO ?  OpenID SSO needs special treatment because of the redirection to the identity provider... Some SSO providers, like the SSL provider don't use redirection and so the base treatment is used... In all cases the idea is to come back to the requested page after authentication

guacamole/src/main/java/org/apache/guacamole/GuacamoleSession.java
adb014: Yes its a copy-paste and should be modified. Done but with different text than proposed

PR #1235, opened 2026-07-30
GUACAMOLE-2311: Profile changes are not saved when a user edits their own settings.

bbennett-ks:
When a user edits their own profile, changes to full name, email, organization, and role are silently discarded and never reach the database.

In `UserObjectTranslator.java`, self-edits take a different path than when editing another account.
```
    @Override
    public void filterExternalObject(UserContext userContext,
            User existingObject, APIUser object) throws GuacamoleException {

        // If a user is editing themselves ...
        if (existingObject != null && existingObject.getIdentifier().equals(userContext.self().getIdentifier())) {
            // ... they may only edit preference attributes
            object.setAttributes(filterAttributes(userContext.getUserPreferenceAttributes(),
                    object.getAttributes()));
        }

        else {
            // In all other cases, filter object attributes by defined schema
            object.setAttributes(filterAttributes(userContext.getUserAttributes(),
                    object.getAttributes()));
        }
    }
```

Self-edits are filtered by `UserObjectTranslator.filterExternalObject()` using `getUserPreferenceAttributes()`. For `ModeledUserContext`, that currently returns no allowed fields, so `filterAttributes()` strips all profile attributes before the update is processed. The fix is to have `ModeledUserContext `expose the standard profile fields as user-editable preferences.

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledUserContext.java
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-base/src/main/java/org/apache/guacamole/auth/jdbc/user/ModeledUserContext.java
@@ -26,6 +26,7 @@ import org.apache.guacamole.auth.jdbc.connection.ConnectionDirectory;
 import com.google.inject.Inject;
 import com.google.inject.Provider;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Date;
 import org.apache.guacamole.GuacamoleException;
 import org.apache.guacamole.auth.jdbc.base.RestrictedObject;
@@ -281,6 +282,11 @@ public class ModeledUserContext extends RestrictedObject
     @Override
     public Collection<Form> getUserAttributes() {
         return ModeledUser.ATTRIBUTES;
+    }
+
+    @Override
+    public Collection<Form> getUserPreferenceAttributes() {
+        return Collections.singletonList(ModeledUser.PROFILE);
     }
 
     @Override


necouchman:
@bbennett-ks Is there any mechanism for this access being limited? Is it tied to the "Change own password" permission, or is it just assumed that users should always be able to edit their own attributes?

bbennett-ks:
> @bbennett-ks Is there any mechanism for this access being limited? Is it tied to the "Change own password" permission, or is it just assumed that users should always be able to edit their own attributes?

It's a little odd... First, a non-admin user needs `Create new users` permission to be able to access `User settings screen`. And then you need `Change password` permission to be able to open the `User settings screen`. If not, you get `Sorry, but this user account cannot be edited`.

Maybe, one should be able to change Profile via the `Preferences` screen, like one can change the password (if `Change passwd` permission is enabled).

@necouchman Let me know if you'd like the profile data to be  `Preferences` screen unconditionally. Seems like the best design choice,..

necouchman:
> > @bbennett-ks Is there any mechanism for this access being limited? Is it tied to the "Change own password" permission, or is it just assumed that users should always be able to edit their own attributes?
> 
> It's a little odd... First, a non-admin user needs `Create new users` permission to be able to access `User settings screen`. And then you need `Change password` permission to be able to open the `User settings screen`. If not, you get `Sorry, but this user account cannot be edited`.
> 
> Maybe, one should be able to change Profile via the `Preferences` screen, like one can change the password (if `Change passwd` permission is enabled).
> 
> @necouchman Let me know if you'd like the profile data to be `Preferences` screen unconditionally. Seems like the best design choice,..

Yeah, I think there are a couple of things that need to be addressed, here:
* First, the permission currently labeled "Can change own password," maps to the `UPDATE` permission for the user's own account, which really means that the permission should more properly be labeled, "Can update own account", or something similar.
* Should the `UPDATE` permission for a user's own account trigger the ability to see the `User settings` screen, the `Preferences` screen, or both? I would say it might be better to add this to the `Preferences` section/screen rather than the `Settings` pages, and not have the `Settings` page be available for users who only have the "Can update own account" (`UPDATE`) permission.

PR #1236, opened 2026-07-31
GUACAMOLE-2243: Allow displaying who is connected.

corentin-soriano:
A permission dedicated to this use (allows to see active connections, but not history or the IP address):
image

With the permission:
image

Without (allow to see only owned active connections):
image


necouchman:
Instead of creating a new system-wide permission for this, I wonder if it makes sense to tie it more to the existing `AUDIT` permission, or implement the `AUDIT` permission as an `ObjectPermission` type, so that individual objects can have this setting applied (or not)?

corentin-soriano:
> Instead of creating a new system-wide permission for this, I wonder if it makes sense to tie it more to the existing `AUDIT` permission, or implement the `AUDIT` permission as an `ObjectPermission` type, so that individual objects can have this setting applied (or not)?

The benefit of creating an additional permission is that it does not grant access to the history. In this way, the user can only see the owner's name for an active connection.

PR #1239, opened 2026-08-21
GUACAMOLE-350: Allow SSH-Key with at least 4096 bits

adb014:
Don't like to see existing PRs open for 9 years, so here is an update of #175 addressing the comment to allow upgrade of existing installations. Please note that nvarchar in SQLServer only supports 4000 characters, so SQLServer can't support RSA-4096 keys as is.  If we can change the nvarchar to varchar SQLServer supports upto 8000 characters, so if we want this in SQLServer we'd have to change the type of 'parameter-value'

### DIFF
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/schema/001-create-schema.sql
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/schema/001-create-schema.sql
@@ -299,7 +299,7 @@ CREATE TABLE guacamole_connection_parameter (
 
   connection_id   integer       NOT NULL,
   parameter_name  varchar(128)  NOT NULL,
-  parameter_value varchar(4096) NOT NULL,
+  parameter_value varchar(8196) NOT NULL,
 
   PRIMARY KEY (connection_id,parameter_name),
 
--- a/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/schema/001-create-schema.sql
+++ b/extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/schema/001-create-schema.sql
@@ -299,7 +299,7 @@ CREATE TABLE guacamole_connection_parameter (
 
   connection_id   integer       NOT NULL,
   parameter_name  varchar(128)  NOT NULL,
-  parameter_value varchar(8196) NOT NULL,
+  parameter_value varchar(8192) NOT NULL,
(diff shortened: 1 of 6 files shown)


extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/schema/001-create-schema.sql
```diff
@@ -299,7 +299,7 @@ CREATE TABLE guacamole_connection_parameter (
 
   connection_id   integer       NOT NULL,
   parameter_name  varchar(128)  NOT NULL,
-  parameter_value varchar(4096) NOT NULL,
+  parameter_value varchar(8196) NOT NULL,
```
necouchman: Is this a typo? Should this be 8192?

necouchman (changes requested):
@adb014 Thanks for picking this one up.

Regarding SQL Server...I think it's probably fine to change from `nvarchar` to `varchar`; however, I'm not sure if that can happen transparently and keep the existing data, or if we'll need to do some sort of script to convert the existing data from one to the other?

extensions/guacamole-auth-jdbc/modules/guacamole-auth-jdbc-postgresql/schema/001-create-schema.sql
adb014: Yes, its a typo.. Damn

adb014:
> @adb014 Thanks for picking this one up.
> 
> Regarding SQL Server...I think it's probably fine to change from `nvarchar` to `varchar`; however, I'm not sure if that can happen transparently and keep the existing data, or if we'll need to do some sort of script to convert the existing data from one to the other?

The conversion of the existing data can be direct. The Unicode characters in the nvarchar will be converted to N-bytes in the varchar variable. This comes with several caveats

- Each Unicode character can be represented by several bytes.. So its possible that 4000 Unicode characters will be converted to more than 8000 chacters that will be illegal,
- As we are talking about ASCII encoded SSH keys, they should be in 7-bit ASCII characters, so unless the users has illegal characters in the SSH key the above shouldn't happen, and if they do they deserve what they get in a straight conversion.. 
 
The command to convert and then just be 

```sql
ALTER TABLE [guacamole_connection_parameter] ALTER COLUMN [parameter_value] VARCHAR(8000) NOT NULL;
``` 

I'll add this, but I have no way of testing it out, so propose to keep this based on the "main" branch to allow testing by others

PR #1248, opened 2026-09-16
GUACAMOLE-283: Add high-availability clustering support.

zivshits:
https://issues.apache.org/jira/browse/GUACAMOLE-283

This is the first slice of HA clustering for `guacamole-client`: a Redis-backed coordination module and `guacd` pooling, so multiple web application replicas can share a pool of `guacd` instances. `guacamole-server` is unmodified.

Clustering is off by default (`cluster-enabled` is `false`). With it disabled, the patched JDBC tunnel path behaves as it does today.

### What this PR does

- Adds a `guacamole-cluster` module (`ClusterStore`, Redis implementation, atomic seat script, heartbeat, Testcontainers harness).
- Resolves `guacd` membership from a static list or a Kubernetes headless Service.
- Routes new connections to the least-loaded `guacd` in the pool, joins to the instance already hosting the session, and fails closed if that route has vanished.
- Registers established tunnels in Redis so other replicas can route to them.
- Includes a two-replica Kubernetes deployment and notes from running it on a live cluster.

Verified on a two-replica EKS stack: six connections split 3/3 across two `guacd` pods; scaling `guacd` redistributes new connections; deleting a route refuses the join rather than opening a new session.

### What this PR does not do

These remain later phases of GUACAMOLE-283:

- Cluster-wide concurrency limits (the seat script exists but is not wired in).
- Cross-replica initiation of a join — the route table is in Redis, but the active-connection directory is still replica-local, so a join can only be started on the replica that owns the session.
- Cluster-wide admin active-connection view / remote kill, and share keys that work across replicas.
- Shared auth tokens and cluster-wide brute-force ban counts.

Until those land, sticky sessions on the auth token are required so a user stays on one replica.

Design, implementation notes, and the deployment manifests are under `docs/superpowers/`.


necouchman:
@zivshits Thanks for the work on this. Your pull request and commit messages will need to be tagged with the Jira issue associated with this feature. Fortunately such a Jira issue already exists - GUACAMOLE-283.

In addition to that, you have at least one merge commit that needs to be cleaned up, and your commit messages could use some cleanup - in particular, prefixes like `feat(cluster)`, `test(cluster)`, and `docs` aren't really  necessary or valuable - just make the commit messages meaningful.

zivshits:
Fixed PR name, description and commit messages

